PR #46, opened 2020-06-03
add a trouble shooting section to the installation guide

Ryanmtate:
This PR adds a `troubleshooting` section to the installation docs. This was added after encountering a few issues during installation in the hopes it will help others.


sehz:
@vijaylaxmid  Can you incorporate this into troubleshooting section?

PR #71, opened 2020-06-15
Decode max

sehz:


sehz:
New MultiplexerSocket socket is added.  This socket is capable of multiplexing kafka protocol.  From MultiplexorSocket, 2 helped sockets can be created.  First is SerialSocket which allows typical request/response.  Second is async response, where there can be multiple response in async manner.

PR #79, opened 2020-07-07
Add Local Persistent Volume option for fluvio-sys Helm chart

### LINKED ISSUE #28: use Kubernetes Local Persistent Volume instead of PVC
Since SPU performs replication,  there is no need to use PVC.

Requires changing helm templating to use different storage provider

nacardin:
Attempt to resolve #28 

This works on minikube, but I think on cloud platforms we need to have `nodeAffinity` match some label we give to nodes that have sufficient storage and are capable of running SPU.

Is there any way to define run shell commands to be run on the nodes, so that we may create a separate dir for SPU data and set its permissions? My current changes use `/tmp` because I don't know of a way to create a new dir and set permissions on it.

Do we need to write a provisioner for this? Or use something like https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/master/docs/provisioner.md

### DIFF
--- a/k8-util/helm/fluvio-sys/templates/local-pv.yaml
+++ b/k8-util/helm/fluvio-sys/templates/local-pv.yaml
@@ -5,11 +5,11 @@ metadata:
   name: fluvio-spu-pv
 spec:
   capacity:
-    storage: 1Gi
+    storage: {{ .Values.localPVStorageSize }}
   volumeMode: Filesystem
   accessModes:
   - ReadWriteOnce
-  persistentVolumeReclaimPolicy: Delete
+  persistentVolumeReclaimPolicy: {{ .Values.localPVReclaimPolicy }}
   storageClassName: fluvio-spu
   local:
     path: {{ .Values.localPVPath }}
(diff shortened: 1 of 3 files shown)


sehz:
It looks like EBS is right storage for EKS:  https://docs.aws.amazon.com/eks/latest/userguide/storage-classes.html instead of local storage.   So this means we should have 3 storage types  (cloud variable) for flexibility instead of just "local storage" type.
* minikube (k8s.io/minikube-hostpath)
* eks. (kubernetes.io/aws-ebs)
* local. (kubernetes.io/no-provisioner)

Later, we can provide additional storage class for specific cloud (GCP)

Generally, we don't want to get in how operator setup their cluster.  We assume node is available and file system is properly formatted where they wanted.  So we should make local storage file path: ```path: /tmp ``` configurable using ```values.yaml``` in the Helm template.

nacardin:
Updated.
Tested on minikube by
```
kubectl label node minikube localPVCapability=FluvioSPU
helm install fluvio-sys ./k8-util/helm/fluvio-sys --set cloud=local --set localPVPath=/tmp
```

sehz:
Nice work. thanks for contribution

Issue #86, opened 2020-07-16
Include the git commit hash in output of the version command


nacardin:
It would be useful know the exact version of the codebase when diagnosing reported issues.

Would be nice to also include relevant info:
```
$ fluvio version

Fluvio version: 0.5.1
Git Commit: 46707f594b28568afde8cdfd2a6dc0b4d2d942a6
Platform: linux/amd64
Compiler: rustc 1.45.0-nightly (1836e3b42 2020-05-06)
```

sehz:
fixed by #105

PR #105, opened 2020-07-23
Add more detailed information in `fluvio version` command

nicholastmosher:
This implements #86, adding additional information to the `fluvio version` command. The added info that gets printed includes:

* Current git hash when `cargo build` is invoked
* OS version info (as reported by `uname -a`)
* Rustc version info

The new output for the version command looks like this:

```
$ fluvio version
Fluvio version : 0.6.0
Git Commit     : d3bd95af2e2cd02997f948181681719e530ab98e
OS Details     : Linux thinkpad 5.7.8-100.fc31.x86_64 #1 SMP Thu Jul 9 15:16:52 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
Rustc Version  : 1.45.0-nightly
```

---

### Notes

* I've branched this PR starting from the commit 69ec1e6 because commit fb92e83 introduces a build error for the `cli` crate.
* By using `uname -a` for OS info I may have made it platform-dependent (unless Windows users are using cygwin). I made sure to not let that break the build by adding a default value when `uname` can't be reached, in which case the OS version would simply print `<this platform does not support uname -a>`. Ideas for a more cross-platform solution are welcome :)

### DIFF
--- a/src/cli/build.rs
+++ b/src/cli/build.rs
@@ -1,6 +1,6 @@
 use std::fs;
 use std::process::Command;
-use rustc_version::version;
+use rustc_version::version_meta;
 
 fn main() {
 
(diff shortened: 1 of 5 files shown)


src/cli/build.rs
```diff
@@ -1,6 +1,28 @@
...
+    // Fetch current git hash to print version output
+    let git_version_output = Command::new("git").args(&["rev-parse", "HEAD"])
+        .output().expect("should run 'git rev-parse HEAD' to get git hash");
+    let git_hash = String::from_utf8(git_version_output.stdout)
+        .expect("should read 'git' stdout to find hash");
+    // Assign the git hash to the compile-time GIT_HASH env variable (to use with env!())
+    println!("cargo:rustc-env=GIT_HASH={}", git_hash);
+
+    // Fetch OS information
+    let uname_output = Command::new("uname").args(&["-a"]).output().ok();
```
sehz, line 16: Can you add OS specific target here?  For example

```
#[cfg(target_os = "unix")]
```

nacardin:
You could use conditional compilation to figure out arch and OS.

https://doc.rust-lang.org/reference/conditional-compilation.html

nacardin:
The output from `uname -a` is a little verbose and hard to read. But it does seem useful to know the kernel version, perhaps you could include just the kernel version as a separate field if on a platform where uname is available.

nicholastmosher:
I added two changes in this last commit:

* I added a `cfg!(unix)` gate for printing out the `uname` output, and instead of using `uname -a` I used `uname -sro`.
* I added prettier formatting for the rustc version.

The new `fluvio version` output looks like this:

```
Fluvio version : 0.6.0
Git Commit     : 805128d24aeb1c728e74e53f318553ea2e37d032
OS Details     : Linux 5.7.8-100.fc31.x86_64 GNU/Linux
Rustc Version  : 1.45.0-nightly (47c3158 2020-06-04)
```

sehz:
ignore CI failure on local cluster,  there is separate issue for it

PR #141, opened 2020-08-05
Smok test try 2

sehz:


sehz:
Didn't succeed at running k8 smoke test but there are good changes here.

Issue #130, opened 2020-08-01
enforce clippy on CI


sehz:

sehz:
Please provide story points

PR #144, opened 2020-08-06
fix cargo clippy errors

Ryanmtate:
This PR adds a CI check for `cargo clippy` and includes suggested changes to satisfy the clippy checker.

Resolves: https://github.com/infinyon/fluvio/issues/130


Ryanmtate:
Gave the changes another review; not immediately apparent where the changes might have results in the smoke tests failing. Might end up closing this PR and attempting to put clippy in place from master... Maybe someone else wants to review the changes and see if there is somewhere that looks wrong.

sehz:
It would be easier to find causes  if you can split changes.  Need to isolate to specific crates first.

Issue #151, opened 2020-08-09
add support for tracing


sehz:
integrate with [tokio tracing](https://github.com/tokio-rs/tracing)

sehz:
please add story points

PR #154, opened 2020-08-12
Migrate logging to tracing instrumentation.

nicholastmosher:
As per #151, this PR adds support for tracing as a replacement for logging. Where possible, the dependency on the `log` crate has been removed when the `tracing` crate is added. The current exception is in crates where `flv-kf-protocol` `Decode` and `Encode` are used, as they add `log` callsites as part of their code expansion.

To get the most use out of `tracing`, use the [`#[tracing]` attribute](https://docs.rs/tracing-attributes/0.1.10/tracing_attributes/attr.instrument.html) on functions to capture relevant context.

image

In this picture log output from SC is on top and SPU is on bottom. The bolded text is part of the context captured by `#[instrument]` attributes on functions. Text inside of the bolded `{ }` represent context that is captured by the spans created by the functions. The context forms a tree-shape around the function call stack, and is very useful for tracking things such as a client connection address, SPU ID, etc. All "log" calls (e.g. "info", "debug") that are made within a span are automatically annotated with the context of that span.

Below is an example instrumentation snippet:

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -43,6 +43,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "ansi_term"
+version = "0.12.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d52a9bb7ec0cf484c551830a7ce27bd20d67eac647e1befb56b0be4ee39a55d2"
+dependencies = [
+ "winapi",
+]
+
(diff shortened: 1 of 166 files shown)


nicholastmosher:
Ah, this will require `flv-util` changes to be merged, version-bumped, and published.

Issue #147, opened 2020-08-07
Track: CLI syncing profiles from cloud


sehz:
# Purpose
To allow a user to import a list of cluster profiles from remote sources easily.

# Background:
Fluvio CLI maintains a list of cluster profiles to store address and access configuration (such as TLS, access key, etc.)  to reach fluvio clusters.  

# Requirements:
Make it easier to download cluster profiles from remote sources (such as fluvio cloud) with minimum hassle.  
Users should be able to keep downloaded cluster profiles and manage them locally (please see fluvio command ```profile``)

There are two solutions proposed:
Users need to authenticate with remote clusters manager (i.g. Fluvio cloud), then copy and paste the download link to Fluvio CLI.
Provide additional command in the CLI to authenticate against the remote manager and download the cluster profiles directly.

We have chosen to implement 2nd solutions because the first solution
Adds the extra burden for remote cluster managers to implement UI for the download link. 
The user has to perform additional steps.

# Propose solution:

New top-level keyword ```sync``` is introduced to unify sync of profiles from any sources.  Existing profiles commands are also simplified.  

```
fluvio profile <SUBCOMMAND>
FLAGS:
    -h, --help    Prints help information
SUBCOMMANDS:
    current                 Display the current context
    delete                  Delete a profile
    switch                  Switch to another profile
    sync                    Synchronize a profile from remote source
    view                    Display entire configuration
    help   
```

Then ```sync``` command should as below:
```
fluvio profile sync <SUBCOMMAND>
FLAGS:
    -h, --help    Prints help information
SUBCOMMANDS:
    local              synchronize profile from cluster running locally
    k8                 synchronize profile from cluster running Kubernetes
    cloud              synchronize profile from cluster running in Fluvio Cloud
    help
```

New ```sync cloud```  will authenticate against the remote source.   Any necessary user name and password will be prompted with essential masks.  

## Syncing clusters from Remote cluster profile:

The syncing operations follow the same principle as existing sync to ```local ``` and ```k8```.

Steps as follows:
* Retrieve cluster profiles from remote sources.
* For each cluster profiles,
   - Copy to local fluvio config file if the matching cluster doesn't exist.
   - Replace existing local config if it exists

## Authentication expiration

When authentication token has expired, the remote cluster profiles in the local fluvio config not modified..

nicholastmosher:
@sehz 

[...]

PR #157, opened 2020-08-17
Add Fluvio Cloud login to CLI

nicholastmosher:
Implements #147.

@sehz please test this out locally before merging, when I tried testing things out this morning a lot of stuff was broken but I think some of that may be because I restarted my computer and haven't started everything back up just quite right.

### New command layout

```
❯ flvd profile
Profile operations

fluvio profile <SUBCOMMAND>

FLAGS:
    -h, --help    Prints help information

SUBCOMMANDS:
    current    Display the current context
    delete     Delete the named profile
    switch     Switch to the named profile
    sync       Sync a profile from a cluster
    view       Display entire configuration
    help       Prints this message or the help of the given subcommand(s)
```

```
❯ flvd profile sync
Sync a profile from a cluster

fluvio profile sync <SUBCOMMAND>

FLAGS:
    -h, --help    Prints help information

SUBCOMMANDS:
    k8       sync profile from kubernetes cluster
    local    sync profile from local cluster
    cloud    sync profile from Fluvio Cloud
    help     Prints this message or the help of the given subcommand(s)
```

These commands map to the old CLI commands as follows:

* `profile current` -> old `profile current-profile`
* `profile delete` -> old `profile delete-profile`
* `profile switch` -> old `profile switch-profile`
* `profile sync k8` -> old `profile create-k8-profile`
* `profile sync local` -> old `profile create-local-profile`
* `profile sync cloud` -> Brand New! 🤩


src/cli/src/profile/sync/cloud/login_agent.rs
```diff
@@ -0,0 +1,255 @@
...
+        email: String,
+        password: String,
+    ) -> Result<(), CloudError> {
+        let login = LoginRequest {
+            email: email.clone(),
+            password,
+        };
+        let mut response = surf::post(format!("{}/api/v1/loginUser", &self.remote))
+            .body_json(&login)
+            .unwrap()
```
sehz, line 137: Is there reason why unwrap is here instead of mapping json error?

src/cli/src/profile/sync/cloud/login_agent.rs
```diff
@@ -0,0 +1,255 @@
...
+                Err(CloudError::AuthenticationError(email))
+            }
+        }
+    }
+
+    async fn save_credentials(&mut self, creds: Credentials) -> Result<(), CloudError> {
+        // Save credentials in agent session
+        self.session.replace(creds);
+        // Save credentials to disk
+        self.session.as_ref().unwrap().try_save(&self.path).await?;
```
sehz, line 157: This could be better if save is done first then replace it in memory:
- it is more clear
- avoid saving credential in memory if failure happens

src/cli/src/profile/sync/cloud/login_agent.rs
```diff
@@ -0,0 +1,255 @@
...
+    id: String,
+    token: String,
+}
+
+impl Credentials {
+    /// Try to load credentials from disk
+    async fn try_load<P: AsRef<Path>>(path: P) -> Result<Self, CloudError> {
+        let file_str = fs::read_to_string(path).await?;
+        let creds: Credentials =
+            toml::from_str(&*file_str).map_err(|_| CloudError::UnableToReadCredentials)?;
```
sehz, line 180: Shouldn't CloudError have original error, so it could be used for debugging later?

src/cli/src/profile/sync/cloud/login_agent.rs
```diff
@@ -0,0 +1,255 @@
...
+        Ok(creds)
+    }
+
+    /// Try to save credentials to disk
+    async fn try_save<P: AsRef<Path>>(&self, path: P) -> Result<(), IoError> {
+        fs::create_dir_all(path.as_ref().parent().unwrap()).await?;
```
sehz, line 186: avoid using unwrap

src/cli/src/profile/sync/cloud/login_agent.rs
```diff
@@ -0,0 +1,255 @@
...
+        fs::write(path, toml::to_string(self).unwrap().as_bytes()).await
```
sehz, line 187: same...

src/cli/src/profile/sync/cloud/login_agent.rs
```diff
@@ -0,0 +1,255 @@
...
+        Self::IoError(error)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use async_std::task;
+
+    #[test]
```
sehz, line 240: you can use ```#[test_async]``` macro from ```flv-future-aio``` to write async test.  see ```src/kf-socket/tests/file_fetch```

src/client-rs/Cargo.toml
```diff
@@ -29,3 +29,4 @@ flv-util = "0.2.0"
 kf-socket = { version = "2.1.0", path = "../kf-socket" }
 flv-api-sc = { version = "2.0.0", path = "../sc-api", default-features=false}
 flv-api-spu = { version = "2.0.0", path = "../spu-api"}
+async-std = "1.6.3"
```
sehz, line 32: no need for ```async-std``` since it is re-exported from ```flv-future-aio```

src/cli/src/profile/sync/cloud/login_agent.rs
nicholastmosher: So with this particular unwrap, I think it's safe to leave because we know for a fact that `Self: Serialize` and it should be impossible for a `Serialize` type to fail this. I added a comment to explain why this unwrap is safe.

src/cli/Cargo.toml
```diff
@@ -49,4 +49,9 @@ flv-types = { path ="../types", version = "0.2.0" }
 flv-spu = { path = "../spu-server", optional = true }
 flv-sc-k8 = { path = "../sc-k8", optional = true }
 rpassword = "5.0.0"
+async-std = "1.6.3"
```
sehz, line 52: unneeded since ```flv-future-aio``` re-exports async-std

nicholastmosher: Hmm, I thought I had gotten rid of that. Give me just one minute and I'll fix it

nicholastmosher: I'm running into a bit of a hiccup with this because some of the functions like `fs::write` require `path::Path`s as parameters, and while `async_std::fs` was re-exported from `flv-future-aio`, `async_std::path` was not.
It seems like I either need to keep the extra `async_std` dependency, or push an update to `flv-future-aio` with an `async_std::path` glob re-export`, or fall back to not using async io for this particular use case.

sehz: For simple and non performance fs, you don't have to make async.

PR #158, opened 2020-08-17
Refactor TlsConfig

nicholastmosher:
As I was reading through the codebase getting ready for the profiles work, I noticed that there was a lot of inconsistency regarding `TlsConfig`s, including many repeated definitions and representations. I took some time to try to organize it a little bit better. Here are some of the major points:

* Renamed all instances of `TlsConfig` in the `cli` crate to `TlsOpt` in order to avoid naming confusion
* Separated `TlsConfig` enum from `client-rs` (with variants `NoVerification`, `File`, and `Inline`) into two separate enums:
  * `TlsConfig` enum with variants `NoVerification` and `WithCerts` is the "most canonical" representation of a TlsConfig, where the `WithCerts` variant carries the inline certs.
  * `TlsConfigPaths` enum with variants `NoVerification` and `WithPaths` is one step removed from `TlsConfig`. You can convert a `TlsConfigPaths` into `TlsConfig` using the `try_into` implementation (which may fail with `IoError` if the files cannot all be read)
* In the `cli` modules where a `TlsOpt` appears, certificate path field types were changed from `String` to `PathBuf`. StructOpt will automatically parse `PathBuf` and perform path validation at parse-time. (Also, when generating shell completions, it will recommend files)
* `TlsOpt` implements `TryInto<Option<TlsConfigPaths>>`

The goal of a lot of the conversions is to make a one-way flow for dealing with the configs. So `TlsOpt` -> `TlsConfigPaths` -> `TlsConfig`, where each step along the way performs more validation.

### DIFF
--- a/src/cli/src/cluster/install/k8.rs
+++ b/src/cli/src/cluster/install/k8.rs
@@ -287,11 +287,11 @@ pub fn install_sys(opt: InstallCommand) {
 async fn set_profile(opt: &InstallCommand) -> Result<(), IoError> {
     use crate::profile::set_k8_context;
     use crate::profile::SetK8;
-    use crate::tls::TlsConfig;
+    use crate::tls::TlsOpt;
 
     let tls_config = &opt.tls;
     let tls = if tls_config.tls {
-        TlsConfig {
+        TlsOpt {
             tls: true,
             domain: tls_config.domain.clone(),
             enable_client_cert: true,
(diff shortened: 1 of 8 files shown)


nicholastmosher:
Ah, I've just realized how this interacts with the top-level `Config` object. I'm going to update this to maintain the `cert_type` flag that was used before.

sehz:
nice refactoring!
is there more updating coming?

sehz:
make sure ```make smoke-test-k8-tls``` runs ok

nicholastmosher:
I think so, I have some thoughts on how to better streamline this. I will let you know if I don't end up adding anything.

Issue #171, opened 2020-08-20
Rename crates for branding and clarity


nicholastmosher:
@sehz was talking about renaming some of the crates and getting rid of the `flv-` prefix in favor of `fluvio-`. If we're going to be renaming all of our crates, I figured we might want to have a place to discuss further renaming potential. Right now I'm only looking at crates in the `fluvio` repo itself.

Here are some possibilities I was thinking of:

- [x] `cli`
	- package: `fluvio-cli`
	- library: `fluvio_cli`
	- binary: `fluvio` (cargo install fluvio)

- [x] `client-rs` -> `client`
	- package: `fluvio`
	- library: `fluvio` (cargo add fluvio)

- [x] `cluster`
	- package: `fluvio-cluster`
	- library: `fluvio_cluster`

- [x] `event-stream-k8` -> `stream-dispatcher`
	- package: `fluvio-stream-dispatcher`
	- library: `fluvio_stream_dispatcher`

- [x] `event-stream-model` -> `stream-model`
	- package: `fluvio-stream-model`
	- library: `fluvio_stream_model`

- [x] `internal-api` -> `controlplane`
	- package: `fluvio_controlplane`
	- library: `fluvio_controlplane`

- [x] `metadata-cluster` -> `controlplane-metadata`
	- package: `fluvio-controlplane-metadata`
	- library: `fluvio_controlplane_metadata`

- [x] `sc-api` -> `sc-schema`
	- package: `fluvio-sc-schema`
	- library: `fluvio_sc_schema`

- [x] ~`sc-k8` -> `sc-cli`~ Binary target moved to `sc` crate below
	- ~package: `fluvio-sc-cli`~
	- ~binary: `fluvio-sc` (external subcommand to fluvio?)~

- [x] `sc-core` -> `sc`
	- package: `fluvio-sc`
	- library: `fluvio_sc`
	- binary: `fluvio-sc-k8`

- [x] `spu-api` -> `spu-schema`
	- package: `fluvio-spu-schema`
	- library: `fluvio_spu_schema`

- [x] `spu-server` -> `spu`
	- package: `fluvio-spu`
	- binary: `fluvio-spu`
	- ~library: `fluvio_dataplane`~ Move all shared SPU code to `spu` crate

Here are some things I was thinking when changing these names:

~* Changing `sc` -> `controlplane` and `spu` to `dataplane`~
[...]

PR #172, opened 2020-08-20
Rename package prefixes from flv- to fluvio-

nicholastmosher:
This PR implements the changes I talked about in #171. If we decide to make further name changes those can easily be added here before accepting.

# Before merging

I reset all of the changed crates' versions to `version = "0.1.0"`. I can revert that if we don't want to do that, just let me know.

# Breaking changes

Obviously, these name changes would be breaking changes. However, we can handle them just like any major version bump. Any external crates which reference these crates will be using the old `flv-<crate>` import, so when we push to a `fluvio-<crate>` package, they'll be unaffected. When we want to upgrade an external crate to use the new package names, we'll just make a major version bump to the external package and change the dependency to the `fluvio-<crate>` name.

# Major version bump

The only crates with no name changes and no version resets are `kf-service` and `kf-socket`. Only `kf-service` has any dependencies on any of the renamed fluvio crates. I see two possible ways to move forward:

1) Give `kf-service` a major version bump and update its dependencies to the newly named crates.
2) `kf-service` only uses `fluvio-types` from inside the fluvio repo, and the only thing it imports is the `print_cli_error` macro. It seems to me that we could just remove `kf-service`'s dependency on `flv-types` and be on our merry way.

I think 2) sounds like a good option, let me know your thoughts.


nicholastmosher:
This is very stale, I'll just rename everything in a new branch once we decide on new crate names per #171.

Issue #180, opened 2020-08-21
flaky CI due to spu address not available


sehz:
remediation:
if SPU > 0, wait for SPU state is ready and have address available.    if not ready, display the spu status and try until maximum number of retries.

sehz:
still issue

PR #181, opened 2020-08-23
display logs after installing k8

sehz:
to diagnose issue: #180


sehz:
Implemented following changes
- if spu > 0, installer will wait until spu are ready
- display spu status after install and before test to diagnose
- for service operator, wait 2 second before start to sync with spu store (this is workaround)

PR #183, opened 2020-08-24
Refactor ClusterConfig and ClusterClient

nicholastmosher:
This is part of my cleaning up the codebase surrounding `TlsConfig` in preparation for adding TLS support.

This PR:

* Removes the duplication between `client-rs/config::Cluster` and `client-rs/config::cluster::ClusterConfig` by removing `Cluster` and making `ClusterConfig` the default.

* Makes `ClusterClient` in charge of "connecting" itself using a `ClusterConfig, rather than forcing `ClusterConfig` to know how to do that. `ClusterConfig` is just a data structure now.

### DIFF
--- a/src/client-rs/src/client/client.rs
+++ b/src/client-rs/src/client/client.rs
@@ -138,9 +138,9 @@ impl From<String> for ClientConfig {
 }
 
 impl ClientConfig {
-    pub fn new(addr: String, connector: AllDomainConnector) -> Self {
+    pub fn new<S: Into<String>>(addr: S, connector: AllDomainConnector) -> Self {
         Self {
-            addr,
+            addr: addr.into(),
             client_id: "fluvio".to_owned(),
             connector,
         }
--- a/src/cli/src/consume/mod.rs
+++ b/src/cli/src/consume/mod.rs
@@ -21,6 +21,7 @@ mod process {
 
     use super::ConsumeLogOpt;
     use super::fetch_log_loop;
+    use fluvio::ClusterClient;
 
     /// Process Consume log cli request
     pub async fn process_consume_log<O>(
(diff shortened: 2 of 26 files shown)


src/client-rs/src/client/client.rs
```diff
@@ -138,9 +138,9 @@ impl From<String> for ClientConfig {
 }
 
 impl ClientConfig {
-    pub fn new(addr: String, connector: AllDomainConnector) -> Self {
+    pub fn new<S: Into<String>>(addr: S, connector: AllDomainConnector) -> Self {
```
sehz, line 141: minor nitpick:  you can use:
```
add: impl Into<String>
```

nicholastmosher:
In addition to adding `Into<String>`, I also did some more restructuring with TlsConfig now. I think this time should be the last time I need to touch it.

From the top down, TLS now has:

* `TlsPolicy`: Describes whether TLS is being used or not, with variants `NoTls` and `WithTls(TlsConfig)`.
* `TlsConfig`: Describes whether to read TLS keys and certs inline or from paths
* `TlsCerts`: A struct containing inline keys and certs
* `TlsPaths`: A struct containing paths to keys and certs

So instead of needing to juggle `Option<TlsConfig>`, we can just use `TlsPolicy` as a canonical top-level type.

sehz:
can you run following smoke test in your environment?
```
make smoke-test
make smoke-test-tls,
make smoke-test-k8,
make smoke-test-k8-tls

nicholastmosher:
@sehz I'm working on fixing that right now. I think I must have broken something by changing the config struct layouts but I haven't found it yet

nicholastmosher:
The smoke test failure was due to some change I made in the `TlsConfig` changes. I rolled back this branch to just include the `ClusterConfig` changes and I will solve the other problems separately.

PR #186, opened 2020-08-25
Consolidate TLS configuration options

nicholastmosher:
Previously, there was a mix of types that represented different TLS options, which made things less ergonomic and more confusing. For example:

* To not use TLS, you'd use `None::<TlsConfig>`
* To use TLS but without domain verification, you'd use `Some(TlsConfig::NoVerification)`
* To use TLS with verification, you'd use `Some(TlsConfig::WithCerts)`

I found the mix of the `Option` type into this scheme rather confusing, so I flattened it out and created some nice `From` implementations to make things more streamlined.

Now, at the top level we have `TlsPolicy`, which can be

* `Disabled`: meaning TLS is not used
* `NoVerify`: meaning that TLS is used but the certificates are not verified
* `Verify(TlsConfig)`: meaning that TLS is used and certificates are verified.

The presence of a `TlsConfig` always implies that you are working with verified TLS. The variants of `TlsConfig` describe how your certificates and keys are represented in configuration:

* `Inline(TlsCerts)`: TLS certificates and keys are held in memory as strings and serialized directly into the configuration object.
* `File(TlsPaths)`: TLS certificates and keys are referenced by paths into the filesystem. The paths are serialized into the config object.

### DIFF
--- a/src/cli/src/lib.rs
+++ b/src/cli/src/lib.rs
@@ -50,6 +50,7 @@ macro_rules! t_print_cli_err {
 
 mod target {
     use std::io::{ErrorKind, Error as IoError};
+    use std::convert::TryInto;
     use structopt::StructOpt;
 
     use fluvio::config::ConfigFile;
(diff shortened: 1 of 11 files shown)


sehz:
have you run tls smoke test?

nicholastmosher:
@sehz I just ran `make smoke-test-tls` and it passed. Running `make smoke-test-k8-tls` now. Should we eventually make these part of the CI checks?

sehz:
Yes, not sure why it wasn't part of it.

nicholastmosher:
`make smoke-test-k8-tls` has also passed 👍

sehz:
Great.  Can you create issue for updating CI with TLS?

PR #188, opened 2020-08-26
add schedule for nightly docker image publish

Ryanmtate:


k8-util/docker/fluvio/Makefile
```diff
@@ -0,0 +1,20 @@
...
+copy_binaries:
+	mkdir -p target
+	cp	${BIN_DIR}/spu-server target
+	cp	${BIN_DIR}/sc-k8-server target
+
+build:	copy_binaries
+	docker build -f Dockerfile.fluvio  -t $(NAME):$(TAG)  ./target
+
+push:	build
+	docker push $(NAME):$(TAG)
```
sehz, line 16: I think you need do perform [docker login](https://stackoverflow.com/questions/34434231/how-to-authenticate-for-docker-push) in order to publish on docker hub

Ryanmtate: commit 14c0f2e adds a step for `docker login`, however it will require the following new secrets to be added:

`DOCKER_USERNAME` and `DOCKER_EMAIL`

sehz: actually, you need to set password like here:  https://docs.docker.com/engine/reference/commandline/login/.

I created DOCKER_USERNAME and DOCKER_PASSWORD secret.

Ryanmtate: Updated and squashed commits in b53b52f

sehz:
so looks like you can manually trigger workflow:  https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch.   with this, we should be able test without waiting for cron job

Ryanmtate:
> so looks like you can manually trigger workflow: https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch. with this, we should be able test without waiting for cron job

Yes, nice find. 96cc783 adds a workflow dispatch

Ryanmtate:
> > so looks like you can manually trigger workflow: https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch. with this, we should be able test without waiting for cron job
> 
> Yes, nice find. [96cc783](https://github.com/infinyon/fluvio/commit/96cc7836510434412a572f273c7bd44a27b221d8) adds a workflow dispatch

Seems there is still some issues with the `workflow_dispatch` for private repositories. See thread: https://github.community/t/cant-trigger-workflow-manually/121740/45

Just tried renaming the file, but did not work as well.

Ryanmtate:
> > > so looks like you can manually trigger workflow: https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch. with this, we should be able test without waiting for cron job
> > 
> > 
> > Yes, nice find. [96cc783](https://github.com/infinyon/fluvio/commit/96cc7836510434412a572f273c7bd44a27b221d8) adds a workflow dispatch
> 
> Seems there is still some issues with the `workflow_dispatch` for ~private~ (just remembered this is a public repo) repositories. See thread: https://github.community/t/cant-trigger-workflow-manually/121740/45
> 
> Just tried renaming the file, but did not work as well.

cc57fe6 Seems to fix

sehz:
look at syntax for [on](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions).  For multiple events, need array syntax

Ryanmtate:
> look at syntax for [on](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions). For multiple events, need array syntax

Should have worked with the other notation as well, see: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-multiple-events-with-activity-types-or-configuration

sehz:
did publish worked?

Ryanmtate:
I still don't see the dispatch workflow button for publish; but the actions did not error on the cc57fe6. Maybe we need to merge before we have the actions?

sehz:
ok.  I will merge it then

Issue #187, opened 2020-08-25
Add smoke-test-tls and smoke-test-k8-tls to CI checks


nicholastmosher:
Right now we're running smoke tests that do not involve TLS, but even though we have TLS smoke tests written they are not part of the CI pipeline.

nicholastmosher:
Closed by #189

sehz:
can you provide story points?

PR #189, opened 2020-08-26
Add TLS smoke tests to CI

nicholastmosher:
Implements #187.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,14 +96,20 @@ jobs:
       - name: smoke test
         run: |
           FLV_CMD=true make smoke-test
-      - name: smoke test setup k8
+      - name: smoke test tls
+        run: |
+          FLV_CMD=true make smoke-test-tls
+      - name: setup smoke test k8
         run: |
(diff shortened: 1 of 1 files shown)


nicholastmosher:
This is failing with the following error and I'm not sure why

```
>> /home/runner/work/fluvio/fluvio/target/debug/fluvio produce topic1
error processing record: IoError(
    Custom {
        kind: Other,
        error: VerboseError {
            source: Custom {
                kind: Other,
                error: "failed to lookup address information: Name or service not known",
            },
            message: "could not resolve address `\":9005\"`",
        },
    },
)
```

I didn't actually change anything about the `smoke test k8` job except that it used to be named `run smoke test run k8`. I don't think that should make a difference. @sehz do you have any ideas why this is failing?

sehz:
TLS test assume certs are in place.  Probably need invoke certs set up

sehz:
Sorry.  This is not TLS test.  In that case, just rerun job.  I will take look to see why it is flaking out

nicholastmosher:
Oh weird, it just worked. Ok then!

PR #191, opened 2020-08-27
run checkout before install

Ryanmtate:


Ryanmtate:
Note: if you run `make publish_fluvio_image` as a non-admin you will receive the following error:

```json
{
  "message": "Must have admin rights to Repository.",
  "documentation_url": "https://docs.github.com/rest/reference/actions#create-a-workflow-dispatch-event"
}
```

PR #192, opened 2020-08-27
Fix/ci/docker/toolchain

Ryanmtate:
This updates the publish/nightly ci workflows to run through the smoke test before publishing.

TODO: Look into a github actions that can call into another github actions workflow step (re-using the same steps)

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -15,55 +15,11 @@ jobs:
       - uses: actions/checkout@v2
       - name: Install ${{ matrix.rust }}
         uses: actions-rs/toolchain@v1
-      - name: Start minikube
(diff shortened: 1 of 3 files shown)


.github/workflows/nightly.yml
```diff
@@ -15,6 +15,55 @@ jobs:
       - uses: actions/checkout@v2
       - name: Install ${{ matrix.rust }}
         uses: actions-rs/toolchain@v1
+      - name: Start minikube
```
sehz, line 18: minikube is not needed.  assume CI has run smoke test for all check-in.
just need to bring musl

Ryanmtate: 0817739 removes the extraneous steps; only keeps musl install step;

Issue #201, opened 2020-08-29
publish helm chart


sehz:
similar to #185, we need to publish helm chart in conjunction with docker image.

sehz:
For now, we will publish manually since automating helm chart is only needed for releases

PR #216, opened 2020-09-02
update CLI to deploy charts from helm repository

sehz:
with this change, CLI can customize chart and image location separately.


sehz:
Related to #201

PR #222, opened 2020-09-04
rename client structure

sehz:
make client struct names consistent.

### DIFF
--- a/src/client-rs/src/client/client.rs
+++ b/src/client-rs/src/client/client.rs
@@ -73,6 +73,8 @@ impl Client for RawClient {
             .await
             .map(|res_msg| res_msg.response)
     }
+
+    
 }
 
 impl RawClient {
--- a/src/client-rs/src/client/client.rs
+++ b/src/client-rs/src/client/client.rs
@@ -12,9 +12,9 @@ use flv_future_aio::net::tls::AllDomainConnector;
 
 use crate::ClientError;
 
-/// Generic client trait
+/// Frame with request and response
 #[async_trait]
-pub trait Client: Sync + Send {
+pub trait SerialFrame: Sync + Send {
     /// client config
     fn config(&self) -> &ClientConfig;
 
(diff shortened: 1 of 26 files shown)


src/client-rs/src/client/client.rs
```diff
@@ -40,21 +40,21 @@ pub trait Client: Sync + Send {
         R: Request + Send + Sync;
 }
 
-/// Client with socket connection
-pub struct RawClient {
+/// Socket with Versions
+pub struct VersionedSocket {
```
nicholastmosher, line 44: Could you add some explanation in the doc as to what you mean by "versioned" socket? What is being specified by the version? Is it the contents being sent? Is there a schema associated with the socket?

src/client-rs/src/client/client.rs
```diff
@@ -12,9 +12,9 @@ use flv_future_aio::net::tls::AllDomainConnector;
 
 use crate::ClientError;
 
-/// Generic client trait
+/// Frame with request and response
 #[async_trait]
-pub trait Client: Sync + Send {
+pub trait SerialFrame: Sync + Send {
```
nicholastmosher, line 17: What is the inspiration for this name? Does this trait describe requests? Does it describe a way for building requests?

By looking at the function names, it seems like this trait is pretty broad in what it does, it doesn't really give me a good idea as to what an implementor might look like. What is the core functionality that you're abstracting over? I don't see any instances in the codebase where a trait bound is used (`fn thing<SF: SerialFrame>(frame: SF)`). I read a good blog post a few weeks ago about [Concrete Abstraction](https://matklad.github.io/2020/08/15/concrete-abstraction.html). The idea is basically if you always know the concrete type at your call site, then it works just as well to make two types with matching function signatures, you don't actually need to abstract over them with a trait.

That's probably more review than this renaming required, but if we don't need the trait then we also don't need to name it, right? 😉

src/client-rs/src/client/cluster.rs
```diff
@@ -14,16 +14,18 @@ use super::*;
 use flv_future_aio::net::tls::AllDomainConnector;
 use std::convert::TryFrom;
 
-/// Client connection to cluster
-pub struct ClusterClient {
+/// socket to Cluster
+pub struct ClusterSocket {
```
nicholastmosher, line 18: I like this name 👍

src/client-rs/src/client/client.rs
sehz: I will add more details.  Basically, sockets knows about list of supported API versions.

src/client-rs/src/client/client.rs
sehz: It is intended to send frame using request and response manner.   But I think this doesn't need to be exposed publicly so we don't have decide name for now.

src/client-rs/src/client/cluster.rs
sehz: Yes, that's purpose of new name.  So basically, we have 2 abstraction:
- Socket (transport layer)
- Client (API).

Issue #259, opened 2020-09-10
Use `fluvio` as the new namespace for fluvio installation in kubernetes


Ryanmtate:

sehz:
Let's close this issue, more I think about this, it is not good idea to dictate namespace requirement.

sehz:
let's move this to alpha backlog and discuss how this should work

PR #262, opened 2020-09-10
rename default namespace to fluvio

Ryanmtate:
Fixes https://github.com/infinyon/fluvio/issues/259


sehz (approved):

src/cluster/src/install.rs
```diff
@@ -466,6 +467,44 @@ impl ClusterInstaller {
         }
     }
 
+    /// Create the namespace where fluvio will be installed;
+    /// `fluvio` is the default namespace for fluvio installations;
+    #[instrument(
+        skip(self),
+        fields(namespace = &*self.config.namespace),
+    )]
+    pub async fn create_namespace(&self) -> Result<(), ClusterError> {
```
sehz, line 476: so installer should not create namespace by default.  it should be explicit opt-in by installer as additional argument as convenience.

src/cluster/src/install.rs
```diff
@@ -466,6 +467,44 @@ impl ClusterInstaller {
...
+        info!("Creating namespace {}", &self.config.namespace);
+        Command::new("kubectl")
+            .args(&["create", "namespace", &self.config.namespace])
+            .inherit();
+
+        loop {
```
sehz, line 482: use K8 client apply API to create namespace

src/cluster/src/install.rs
```diff
@@ -100,7 +101,7 @@ impl ClusterInstallerBuilder {
 
     /// Sets the Kubernetes namespace to install Fluvio into.
     ///
-    /// The default namespace is "default".
+    /// The default namespace is "fluvio".
```
sehz, line 104: should not use fluvio namespace as default since we can't assume user wants to use fluvio.  it should be opt-in

src/cluster/src/install.rs
```diff
@@ -21,7 +22,7 @@ use k8_obj_metadata::InputObjectMeta;
 use crate::ClusterError;
 use crate::helm::HelmClient;
 
-const DEFAULT_NAMESPACE: &str = "default";
+const DEFAULT_NAMESPACE: &str = "fluvio";
```
sehz, line 25: revert back to using `default` since we can't assume namespace

src/cluster/src/error.rs
```diff
@@ -27,6 +30,7 @@ impl std::fmt::Display for ClusterError {
             Self::ClientError(err) => write!(f, "{}", err),
             Self::K8ConfigError(err) => write!(f, "{}", err),
             Self::K8ClientError(err) => write!(f, "{}", err),
+            Self::Utf8Error(err) => write!(f, "{}", err),
```
sehz, line 33: not needed here

sehz:
more I think about it, this is not good idea.  it breaks best practice pattern for Kubernetes Install.

sehz (changes requested):

Ryanmtate:
Closing this, we can revisit after alpha is we want to setup namespace for installation. Velero sets up the velero namespace for their services as part of install.

Issue #277, opened 2020-09-14
Create a minimal Fluvio interface definition in preparation for open-source release


nicholastmosher:
Brainstorm and document a minimal Fluvio user interface in terms of a generic domain model. Elements of the interface should not be tied to a specific technology, but rather should express the actions a user may want to take when working with Fluvio, regardless of the client they're using. The goal of this is to create a unified mental model for interacting with Fluvio which is shared among all of the client interfaces. This is the first step in each of #245, #240, and #246.

## Goals

By the end of this issue and discussion, we should have a relatively full picture of the functionality which an end user will have access to. We should expect that any existing Fluvio interface which is not described in this issue will be made private and unavailable to the user before our open-source release. We can use existing examples as inspiration for the scope of functionality that should ultimately exist in this issue's interface definition, though we may clarify, minimize, or rearrange that interface if it seems helpful.

## Examining existing functionality

I went through one of our end-to-end examples (CDC) and picked out all of the usages of the older `flv-client` crate. I'm going to list the types and methods below, and from the code I'm going to write up a plain-English description of what is going on in terms of an end-user perspective. Please let me know in the comments if I am making any assumptions or if my interpretations of any of these are incorrect.

### Producer Example

- `ScClient`
  - `.topic_metadata`: I would like to be able to discover existing topics in my cluster
  - `.create_topic`: I would like to be able to create a new topic to send events to
  - `.find_replica_for_topic_partition`: If I know my topic and partition, I would like to find out which SPU is responsible for streaming that data (SPU replica leader)
- `ScConfig`
  - `::new`: I would like to set up any settings needed to connect to a cluster
  - `.connect`: I would like to connect to a cluster using these settings
- `SpuReplicaLeader`: Acts as a reference to a particular SPU, the one that is in charge of my topic and partition
  - `.fetch_logs_once`: I would like to fetch a range of logs all at once, not as a stream
  - `.send_record`: I would like to append a message to the topic and partition managed by this SPU leader
- `ReplicaConfig`
  - `::Computed`: I would like to specify how my topic will be replicated using a partition count and replica count
- `FetchLogOption`: I would like to configure some parameters for how my fetch will behave
- `FetchOffset`: I would like to specify a log index to start or end a fetch

### Consumer Example

- `SpuReplicaLeader`
  - `.fetch_logs`: I would like to watch for new events arriving on my topic and partition

## Domain Terminology

[...]

PR #285, opened 2020-09-16
Interface experiment for Fluvio client crate

nicholastmosher:
In this PR, we're brainstorming ideas for the public API of the `fluvio` crate that we'd like to publish for our OSS developer release. This brainstorming is stemming from the analysis and discussion from issue #277.

This PR is not intended to ever be merged, but to serve as a playground for API brainstorming.


nicholastmosher:
Some questions that need to be resolved:

- In the old/current Fluvio API, messages seem to be sent directly to a SpuReplicaLeader, which I think is uniquely identified by a Topic and a Partition. However, from a user's perspective, I think it would make more sense if I didn't need to know anything about the partitions. From that standpoint, I just have a topic that I send to and receive from. Is this a feasible user model? Or is there something fundamental that I'm missing?
  - Some follow up notes. If we were to only give the user the ability to produce and consume at the Topic level, then we would need a way to reliably send messages to correct partitions. E.g. a Partition Key. Eventually, it would probably be desirable to have some sort of `FluvioEvent` trait with something like a `fn partition_key(&self) -> &str` method where the events could describe their own unique partition key. But for the time being with string messages, what if we baked the key into the function signature? So we could have something like the following:

```rust
impl Topic {
    pub async fn produce_string_message<S: Into<String>>(&self, key: &str, message: S) {}
}
```

- At what level are log offsets managed? It seems to me they would be managed at the SpuReplicaLeader level. That is, a log (event) would be uniquely identified by a (topic name, partition, offset) triple. I believe that right now, "consuming" messages is attached to the SpuReplicaLeader level, which would seem to support this idea. Would we be able to implement a topic-level consumer that would yield all of the events from the topic, regardless of their partition? It may be as simple as joining multiple partition-level consumer streams. Such a topic consumer would not be able to guarantee total ordering of events in the topic, but would be able to know that all events belonging to the same partition (having the same partition key) would arrive in order.
  - If we do decide to include a "topic-level consumer", we should probably also have a partition level consumer just for completeness. I don't really want to use the term `SpuReplicaLeader` in the public API, that seems to be very tied to our streaming implementation. I think instead, we could call it `PartitionConsumer` which is derived from a Topic using either a partition number or a partition key. Something like this:

```rust
struct Partition {}
impl Partition {
    pub fn consume_string_messages(&self) -> PartitionStringConsumer {}
    pub async fn fetch_string_messages<R: RangeBounds<Offset>>(&self, range: R) -> Result<Vec<String>, FluvioError> {}
}

struct PartitionStringConsumer {}
impl Stream for PartitionStringConsumer {
    type Item = String;
    // ...
}

impl Topic {
    pub fn partition(&self, partition: u16) -> Partition {}
    pub fn partition_by_key(&self, key: &str) -> Partition {}
}
```

nicholastmosher:
### Offset API

Current definition:

```
pub enum FetchOffset {
    Offset(i64),
    Earliest(Option<i64>),
    Latest(Option<i64>),
}
```

Questions:
- `Earliest` and `Latest` are calculated (via `calc_offset`) as relative to the `start_offset` (for Earliest) or the `last_stable_offset` (for Latest). What is the behavior for `Offset`? Is `Offset(0)` the same as `Earliest(Some(0))`?

Potential new definition

```
enum OffsetInner {
    FromBeginning(i64),
    FromEnd(i64),
}
pub struct Offset {
    offset: OffsetInner,
}
impl Offset {
    pub fn from_beginning<T: Into<Option<i64>>>(offset: T) -> Self {
        let offset = offset.into().unwrap_or(0);
        Self {
            offset: OffsetInner::FromBeginning(offset),
        }
    }
    // Similarly for `from_end`
}

// Can be constructed like this
let offset = Offset::from_beginning(None) // Offset of 0
let offset = Offset::from_beginning(10)   // Offset of 10 from beginning
let offset = Offset::from_end(None)       // Latest stable entry
let offset = Offset::from_end(10)         // 10 behind latest stable
```

nicholastmosher:
A few comments about the latest changes:

- After discussion with @sehz, I decided to change the API from being more Topic and Partition oriented to being more Producer and Consumer oriented. I am not quite done transitioning in this process, I have been investigating potential interfaces that we might be able to implement now or in the future. For example, one thought was to implement `AsyncWrite` and `AsyncRead` for `Producer` and `Consumer`, respectively. But I don't believe that we have the right machinery in place to do that properly yet, so I am trying to come up with the best interface possible for the capabilities we currently have, without interfering with the most desirable API we will be trying to achieve in the future.

- I spent a lot of time on this update working on the `Offset` API. I wanted to make sure that impossible values were impossible, which meant preventing users from creating those data structures directly and instead providing constructor functions which validate the offset (e.g. preventing negative offsets). I've also added some thorough documentation for the semantics of the different offset types.

---

After my conversation with @sehz today, I'm going to close this PR and begin taking some of the ideas here and introducing them as incremental PRs to the existing client.

PR #292, opened 2020-09-21
Add better Offset API

nicholastmosher:
This PR introduces a new API for Offsets that makes it impossible to construct illegal `Offset` values. This renames the previously-named `FetchOffset` into just `Offset`. The `Offset` type is essentially an enum with private variants, achieved by wrapping an `OffsetInner` enum inside the `Offset` struct with a private field. This makes it impossible for users to manually construct an Offset, they must use the validating constructors `Offset::absolute`, `Offset::from_beginning`, and `Offset::from_end` which only optionally produce an `Offset` when the argument is valid. This creates the invariant that if an `Offset` type exists, it is valid (e.g. there are no negative offsets).

Additionally, this PR renames `FetchLogOption` to `ConsumerConfig` since it is used both in fetching and streaming operations. I also noticed that it is a common pattern to just use `FetchLogOption::default()` in a lot of call sites, so I have separated the methods into `fetch/fetch_with_config` and `stream/stream_with_config`, where the shorter-named methods automatically use the default ConsumerConfig, and the `with_config` variants will accept a user-provided ConsumerConfig.

# TODO before merging:

- [ ] I need a good user-facing description of what the `max_bytes` configuration option does and why somebody would want to use it
- [ ] I need to know whether the user will ever need to specify a custom `Isolation` value. I have not seen any instances of this occurring, so as of now I have hidden it from the user (the default value is always used in ConsumerConfig). If we need to expose it, we should re-export it and give it documentation on how to use it.

### DIFF
--- a/src/client/src/offset.rs
+++ b/src/client/src/offset.rs
@@ -1,0 +1,267 @@
+use std::io::Error as IoError;
+use std::io::ErrorKind;
+
+use tracing::{debug, trace};
+use dataplane::ReplicaKey;
+use fluvio_spu_schema::server::fetch_offset::FetchOffsetsRequest;
+use fluvio_spu_schema::server::fetch_offset::FetchOffsetPartitionResponse;
+
+use crate::FluvioError;
+use crate::client::SerialFrame;
+
+#[derive(Debug, Clone, PartialEq)]
+pub(crate) enum OffsetInner {
+    Absolute(i64),
+    FromBeginning(i64),
+    FromEnd(i64),
+}
+
+/// Describes the location of an event stored in a Fluvio partition
+///
+/// All Fluvio events are stored as a log inside a partition. A log
+/// is just an ordered list, and an `Offset` is just a way to select
+/// an item from that list. There are several ways that an `Offset`
+/// may identify an element from a log. Suppose you sent some
+/// multiples of `11` to your partition. The various offset types
+/// would look like this:
+///
+/// ```text
+///         Partition Log: [ 00, 11, 22, 33, 44, 55, 66 ]
+///       Absolute Offset:    0,  1,  2,  3,  4,  5,  6
+///  FromBeginning Offset:    0,  1,  2,  3,  4,  5,  6
+///        FromEnd Offset:    6,  5,  4,  3,  2,  1,  0
+/// ```
+///
+/// When a new partition is created, it always starts counting new
+/// events at the `Absolute` offset of 0. An absolute offset is a
+/// unique index that represents the event's distance from the very
+/// beginning of the partition. The absolute offset of an event never
+/// changes.
+///
+/// Sometimes when a partition gets very large, Fluvio will begin
+/// deleting events from the beginning of the log in order to save
+/// space. Whenever it does this, it keeps track of the latest
+/// non-deleted event. This allows the `FromBeginning` offset to
+/// select an event which is a certain number of places in front of
+/// the deleted range. For example, let's say that the first two
+/// events from our partition were deleted. Our new offsets would
+/// look like this:
+///
+/// ```text
+///                          These events were deleted!
+///                          |
+///                          vvvvvv
+///         Partition Log: [ .., .., 22, 33, 44, 55, 66 ]
+///       Absolute Offset:    0,  1,  2,  3,  4,  5,  6
+///  FromBeginning Offset:            0,  1,  2,  3,  4
+///        FromEnd Offset:    6,  5,  4,  3,  2,  1,  0
+/// ```
+///
+/// Just like the `FromBeginning` offset may change if events are deleted,
+/// the `FromEnd` offset will change when new events are added. Let's take
+/// a look:
+///
+/// ```text
+///                                        These events were added!
+///                                                               |
+///                                                      vvvvvvvvvv
+///         Partition Log: [ .., .., 22, 33, 44, 55, 66, 77, 88, 99 ]
+///       Absolute Offset:    0,  1,  2,  3,  4,  5,  6,  7,  8,  9
+///  FromBeginning Offset:            0,  1,  2,  3,  4,  5,  6,  7
+///        FromEnd Offset:    9,  8,  7,  6,  5,  4,  3,  2,  1,  0
+/// ```
+///
+/// # Example
+///
+/// All offsets must be constructed with a positive index. Negative
+/// numbers are meaningless for offsets and therefore are not allowed.
+/// Trying to construct an offset with a negative number will yield `None`.
+///
+/// ```
+/// use fluvio::Offset;
+/// let absolute_offset = Offset::absolute(5).unwrap();
+/// let offset_from_beginning = Offset::from_beginning(100).unwrap();
+/// let offset_from_end = Offset::from_end(10).unwrap();
+///
+/// // Negative offsets will give None
+/// assert_eq!(Offset::absolute(-10), None);
+/// assert_eq!(Offset::from_beginning(-15), None);
+/// assert_eq!(Offset::from_end(-20), None);
+/// ```
+#[derive(Debug, Clone, PartialEq)]
+pub struct Offset {
+    inner: OffsetInner,
+}
+
+impl Offset {
+    /// Creates an absolute offset with the given index
+    ///
+    /// The index must not be less than zero.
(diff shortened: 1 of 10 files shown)


src/client/src/offset.rs
```diff
@@ -0,0 +1,273 @@
...
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use fluvio::Offset;
+    /// assert!(Offset::absolute(100).is_some());
+    /// assert!(Offset::absolute(0).is_some());
+    /// assert!(Offset::absolute(-10).is_none());
+    /// ```
+    pub fn absolute(index: i64) -> Option<Offset> {
```
sehz, line 109: shouldn't this be u64?,   and this return option?

src/client/src/offset.rs
```diff
@@ -0,0 +1,273 @@
...
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use fluvio::Offset;
+    /// assert!(Offset::from_beginning(10).is_some());
+    /// assert!(Offset::from_beginning(0).is_some());
+    /// assert!(Offset::from_beginning(-10).is_none());
+    /// ```
+    pub fn from_beginning(offset: i64) -> Option<Offset> {
```
sehz, line 144: why return option?

src/client/src/offset.rs
```diff
@@ -0,0 +1,273 @@
...
+    /// using the [`from_end`] function), then using `to_absolute` will calculate
+    /// the absolute offset by finding the last stably-committed event and subtracting
+    /// the relative offset from it.
+    ///
+    /// Calling `to_absolute` on an offset that is already absolute just returns
+    /// that same offset.
+    ///
+    /// Note that calculating relative offsets requires connecting to Fluvio, and
+    /// therefore it is `async` and returns a `Result`.
+    pub(crate) async fn to_absolute<F, S: Into<String>>(
```
sehz, line 216: this lead to dependencies to client.  this seems be belong to client

src/client/src/offset.rs
nicholastmosher: @sehz I thought we talked about this... This is the same representation that `FetchOffset` was using:
https://github.com/infinyon/fluvio/pull/292/files#diff-3ec5e00be2029042b2e036bb803ac947L45-L52
I thought the purpose was to maintain consistency with a potential Java client representation of `long`. I will happily make this into `u64` if you want me to though.

src/client/src/offset.rs
nicholastmosher: This returns `Option` because if a user were to call `Offset::from_beginning(-10)`, that would be invalid. Having constructors return `Option`s is a way to create the invariant of "If it exists, then it's valid". This is to prevent bad offsets from being constructed.

src/client/src/offset.rs
sehz: internal should be kept in the i64 to preserve Java compatibility.  The parameter should be u64 since offset should be positive.

src/client/src/offset.rs
nicholastmosher: I don't think it's terribly important where it goes, it's not a public function. I just thought it made sense to put a conversion function on the Offset type itself.

src/client/src/offset.rs
sehz: if you make parameter u64 then you don't need to worry about it.   I thought we are adding API so invalid argument can't be constructed

src/client/src/offset.rs
nicholastmosher: Well that leaves us with the same kind of problem, which is that there will be a portion of the range that is invalid. E.g. even if I wrote it as `fn absolute(index: u64)`, then sure they couldn't give a negative number, but they _could_ do something like `Offset::absolute(u64::MAX)` and cause a problem. I think that if we're going to store this internally as i64, then it'd communicate the allowed input values much better to ask for an `i64` argument but give the specification that it must not be less than zero. This also prevents any subtle mistakes with casting between integer types

src/client/src/offset.rs
nicholastmosher: If we made a function such as `from_beginning(offset: u64) -> Offset` (note the infallible return of an `Offset`), then that would be an inherently broken API, because not all `u64` values will produce a valid offset. If we are committed to representing this value internally as an `i64`, then we _must_ return an `Option<Offset>` in order to create a sound API. The choice now is whether to use an argument of `i64` and forbid negatives (this is the path I think we should take), or use an argument of `u64` and forbid values that overflow `i64::MAX`. I feel like the second choice feels much more arbitrary and will be more confusing for users.

src/client/src/offset.rs
sehz: If we accept i64 then it should return error if number is negative

src/client/src/offset.rs
sehz: how about if we make offset u32?,  I don't believe there is use case we jump arbitrary offsets.  u32 should be big enough

nicholastmosher: I think u32 will probably not be big enough. I think I had the same thought yesterday, I was like "a u32 should be half the size of u64, so it should have the same positive range as i64", but that's not correct. A u32 is several orders of magnitude smaller than even the positive range of an i64:

```
Max u64: 18446744073709551615
Max i64: 9223372036854775807
Max u32: 4294967295
Max i32: 2147483647
```

I think that we should just continue using `i64`

src/client/src/offset.rs
nicholastmosher: Do you mean that instead of returning an Option it should return a Result? I'd be open to doing that.

sehz: yes, invalid configuration should return Error

src/client/src/offset.rs
sehz: Do you see any use case where someone wants to do relative offset bigger than u32?  Most likely, it will be small like last 100 or 1000.

nicholastmosher: I can't immediately think of one, but I also don't think it's a good idea to prematurely restrict the things that users can do. If we change this to `from_beginning(offset: u32) -> Offset`, then there are potential valid cases that are no longer possible to express. I just feel like there's really not any harm in leaving it as `i64`

sehz: if in that case, I don't see any advantage over previous API

nicholastmosher: The previous API would not validate any of the values created in an `Offset` type. The previous definition was like this:
```rust
pub enum FetchOffset {
    Earliest(Option<i64>),
    Latest(Option<i64>),
    Offset(i64),
}
```

Since all variants of a public enum are also public, that means that I would be able to construct values such as these:
```rust
FetchOffset::Earliest(Some(-100))
FetchOffset::Latest(Some(-100))
FetchOffset::Offset(-100)
```

All of which are illegal values. This would place the responsibility of validation on every single internal piece of code that uses a `FetchOffset` value.

In contrast, the new `Offset` type can only be created via it's three constructors, `absolute`, `from_beginning`, and `from_end`. All of these functions return a `Result<Offet, FluvioError>` because there are values of `i64` which would be illegal for an offset (namely, all of the negative values of i64). The constructors will only yield an `Ok(Offset)` when the given value is valid (i.e. positive offset). Therefore, there is now an invariant on the `Offset` type which is that if an `Offset` value exists, it must be valid (i.e. non-negative). This is valuable because the internal implementation code may safely assume that all `Offset` values are valid and can eliminate repetitive checking and validating throughout the codebase. It also makes sure that a user is not given a false sense of security if they construct a `FetchOffset` that happens to be invalid but they don't find out about it until later on when they try to use it. The `Offset` API has more of a fail-fast mentality where the user will know immediately whether they created a correct value.

nicholastmosher:
After some discussion, these are the main points made:

- Internal representation stays i64 until some time in the future we change to u64
- Argument is u32 mapped onto the internal i64. It is also valid to map u32 onto u64 if we change in the future
- Later on, we could update u32 arg to u64 if internal representation is changed to u64

sehz (approved):

sehz:
doc test are failing

nicholastmosher:
Nice that the doc tests are already keeping us honest ;)

PR #293, opened 2020-09-21
update type length limit

### LINKED ISSUE #291: Failed to compile: Type-length limit reached
I'm starting to notice weird build problems involving large type lengths. Here's an example of an error I'm getting:

```
error: reached the type-length limit while instantiating `async_task::utils::abort_on_pani...::Arc<async_executor::State>]>]>`
  --> /Users/nick/.cargo/registry/src/github.com-1ecc6299db9ec823/async-task-4.0.1/src/utils.rs:24:1
   |
24 | / pub(crate) fn abort_on_panic<T>(f: impl FnOnce() -> T) -> T {
25 | |     struct Bomb;
26 | |
27 | |     impl Drop for Bomb {
...  |
36 | |     t
37 | | }
   | |_^
   |
   = note: consider adding a `#![type_length_limit="1602453"]` attribute to your crate

error: could not compile `fluvio-spu`.

error: build failed
```

That's appearing in `fluvio-spu`. When I went to apply the new type length limit, I saw we already had a type length limit and I was wondering whether this has been a problem before, and what causes it. Is there anything we can do to our code to avoid this problem from happening? I'm wondering whether we can avoid these problems rather than patching them up, I don't know if a larger type-length limit might get us into bad compile-time or other weird compiler behavior eventually.

Ryanmtate:
Fixes #291

### DIFF
--- a/src/cli/src/lib.rs
+++ b/src/cli/src/lib.rs
@@ -1,3 +1,5 @@
+#![type_length_limit = "1113484"]
+
 mod common;
 mod error;
 mod consume;
--- a/src/spu/src/lib.rs
+++ b/src/spu/src/lib.rs
@@ -1,5 +1,5 @@
 #![recursion_limit = "512"]
-#![type_length_limit = "1201870"]
+#![type_length_limit = "4006064"]
 
 mod error;
 mod start;


Cargo.lock
```diff
@@ -3376,9 +3366,9 @@ checksum = "6454029bf181f092ad1b853286f23e2c507d8e8194d01d92da4a55c274a5508c"
 
 [[package]]
 name = "vec-arena"
-version = "0.5.2"
+version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8cb18268690309760d59ee1a9b21132c126ba384f374c59a94db4bc03adeb561"
+checksum = "eafc1b9b2dfc6f5529177b62cf806484db55b32dc7c9658a118e11bbeb33061d"
```
sehz, line 3371: let's not update dependencies for now

Ryanmtate: commit 53b5e85 removes changes to dependencies

sehz (approved):

PR #304, opened 2020-09-24
fix typos on nightly image makefile

sehz:
use two different target rather than a single target for latest image CI


sehz:
fixed by #311. closing

PR #320, opened 2020-09-28
add use_serde option and make client config public

Ryanmtate:


src/client/src/client/cluster.rs
```diff
@@ -19,7 +19,7 @@ use super::*;
 /// An interface for interacting with Fluvio streaming
 pub struct Fluvio {
     socket: AllMultiplexerSocket,
-    config: ClientConfig,
+    pub config: ClientConfig,
```
nicholastmosher, line 22: Why do you need this to be public? If it's really necessary, can we mark it as `#[doc(hidden)]`?

Ryanmtate: Commit 571ab32 removes `pub` visibility. Removed the need for it in the node client.

src/client/Cargo.toml
```diff
@@ -29,7 +29,7 @@ tokio = { version = "0.2.21", features = ["macros"] }
 # Fluvio dependencies
 fluvio-future = { version = "0.1.0", features = ["tls"] }
 fluvio-types = { version = "0.1.0", path = "../types" }
-fluvio-sc-schema = { version = "0.1.0", path = "../sc-schema", default-features = false }
+fluvio-sc-schema = { version = "0.1.0", path = "../sc-schema", features = ["use_serde"] }
```
Ryanmtate, line 32: @sehz  This also appears to be how the CLI is using the features, so not sure we want to add a feature gate around this. I think most clients will need this feature.

sehz: I think serde is used by admin api.  Not need for consumer or producer

nicholastmosher: If we _do_ want to add a feature gate around the admin API, then I think we should make it specifically for that. We can make a feature called `admin` or `fluvio-admin` which depends on a `serde` as an optional feature if we want.

Ryanmtate: commit b7bef94 adds `admin` feature.

sehz (approved):

Issue #313, opened 2020-09-25
Pre-install check: Install sys chart if it is not installed


vijaylaxmid:
Install sys chart, if not already installed.

Issue #235, opened 2020-09-07
Make ClusterInstaller install `fluvio-sys` if it is not installed


nicholastmosher:
This could be a good step in improving the open-source user experience. Right now, you have to manually install both `fluvio-app` and `fluvio-sys`. We should be able to make our installer detect whether `sys` is installed and do it for the user, with options to override for specific use-cases like development.

sehz:
there was issue with installing CRD in the helm chart before.  It maybe solved.   There are 2 options:
- Integrate `sys` and `app` chart into a single chart.  And have option to install combination of them.
- Make sys chart dependencies of `app` using helm dependencies.

vijaylaxmid:
I think this is already done and can be closed. @sehz ?

sehz:
Doesn't seems it works in the master

vijaylaxmid:
I performed the following steps in the master. Seems to work fine. Do you see any difference?

```
$ fluvio cluster uninstall        
release "fluvio" uninstalled

$fluvio cluster uninstall --sys  
release "fluvio-sys" uninstalled

$fluvio cluster install --develop
NAME: fluvio-sys
LAST DEPLOYED: Thu Nov  5 00:44:44 2020
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
NAME: fluvio
LAST DEPLOYED: Thu Nov  5 00:44:49 2020
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
waiting for spu to be provisioned for: 1 seconds
1 spus provisioned 
```

sehz:
Works in the CI

PR #325, opened 2020-09-28
Streamline cluster install

nicholastmosher:
This adds better error messages to the `helm` client in such a way that it bubbles all the way up to the CLI. Before, if `helm` was not installed, if the user ran `fluvio cluster install`, they would be given the following error:

```bash
$ fluvio cluster install
error: No such file or directory (os error 2)
```

Now, it gives this error:

```bash
$ fluvio cluster install
error: Unable to find 'helm' executable
  Please make sure helm is installed and in your PATH.
  See https://helm.sh/docs/intro/install/ for more help
```

In addition, this PR makes the cluster installer attempt to fix some things if the precheck runs into errors:

- If fluvio-sys is not installed, it attempts to install it
- If the minikube context is not set right, it will set it

### DIFF
--- a/src/cli/src/cluster/install/k8.rs
+++ b/src/cli/src/cluster/install/k8.rs
@@ -43,7 +43,7 @@ pub async fn install_core(opt: InstallCommand) -> Result<(), CliError> {
         }
         // If we're in develop mode (but no explicit chart location), use hardcoded local path
         None if opt.develop => {
-            builder = builder.with_local_chart("./k8-util/helm/fluvio-app");
+            builder = builder.with_local_chart("./k8-util/helm");
         }
         _ => (),
     }
--- a/src/cluster/src/error.rs
+++ b/src/cluster/src/error.rs
@@ -5,6 +5,7 @@ use fluvio::FluvioError;
 use k8_config::{ConfigError as K8ConfigError};
 use k8_client::{ClientError as K8ClientError};
 use crate::helm::HelmError;
+use crate::check::CheckError;
 
 /// The types of errors that can occur during cluster management
 #[derive(Error, Debug)]
@@ -44,23 +45,13 @@ pub enum ClusterError {
         /// The underlying Helm client error
         source: HelmError,
     },
-    /// The installed version of helm is incompatible
-    #[error("Must have helm version {required} or later. You have {installed}")]
-    IncompatibleHelmVersion {
-        /// The currently-installed helm version
-        installed: String,
-        /// The minimum required helm version
-        required: String,
+    /// An error that occurred during pre-installation checks
+    #[error("Fluvio pre-installation check failed")]
+    PreCheckError {
+        /// The pre-check error that occurred
+        #[from]
+        source: CheckError,
     },
-    /// The fluvio-sys chart is not installed
-    #[error("The fluvio-sys chart is not installed")]
-    MissingSystemChart,
-    /// Fluvio is already correctly installed
-    #[error("The fluvio-app chart is already installed")]
-    AlreadyInstalled,
-    /// Need to update minikube context
-    #[error("The minikube context is not active or does not match your minikube ip")]
-    InvalidMinikubeContext,
     /// Timed out when waiting for SC service.
     #[error("Timed out when waiting for SC service")]
     SCServiceTimeout,
--- a/src/cluster/src/error.rs
+++ b/src/cluster/src/error.rs
@@ -55,6 +55,9 @@ pub enum ClusterError {
     /// The fluvio-sys chart is not installed
     #[error("The fluvio-sys chart is not installed")]
     MissingSystemChart,
+    /// Fluvio is already correctly installed
+    #[error("The fluvio-app chart is already installed")]
+    AlreadyInstalled,
     /// Need to update minikube context
     #[error("The minikube context is not active or does not match your minikube ip")]
     InvalidMinikubeContext,
--- a/src/cluster/src/install.rs
+++ b/src/cluster/src/install.rs
@@ -39,7 +39,7 @@ const DEFAULT_CHART_APP_NAME: &str = "fluvio/fluvio-app";
 const DEFAULT_CHART_REMOTE: &str = "https://charts.fluvio.io";
 const DEFAULT_GROUP_NAME: &str = "main";
 const DEFAULT_CLOUD_NAME: &str = "minikube";
-const DEFAULT_HELM_VERSION: &str = "3.2.0";
+const DEFAULT_HELM_VERSION: &str = "3.3.4";
 
 /// Distinguishes between a Local and Remote helm chart
 #[derive(Debug)]
--- a/src/cluster/src/install.rs
+++ b/src/cluster/src/install.rs
@@ -4,8 +4,7 @@ use std::path::PathBuf;
 use std::borrow::Cow;
 use std::process::Command;
 use std::time::Duration;
-use std::net::{IpAddr, SocketAddr};
-use std::str::FromStr;
+use std::net::SocketAddr;
 
 use tracing::{info, warn, debug, trace, instrument};
 use fluvio::{Fluvio, FluvioConfig};
@@ -25,7 +24,7 @@ use semver::Version;
 
 use crate::ClusterError;
 use crate::helm::{HelmClient, Chart, InstalledChart};
-use crate::check::get_cluster_server_host;
+use crate::check::{check_cluster_server_host, CheckError};
 
 const DEFAULT_NAMESPACE: &str = "default";
 const DEFAULT_REGISTRY: &str = "infinyon";
@@ -597,11 +596,11 @@ impl ClusterInstaller {
     ///  1. Check if compatible helm version is installed
     ///  2. Check if compatible sys charts are installed
     ///  3. Check if the K8 config hostname is not an IP address
-    fn pre_install_check(&self) -> Result<(), ClusterError> {
+    fn pre_install_check(&self) -> Result<(), CheckError> {
         // check helm version
         let helm_version = self.helm_client.get_helm_version()?;
         if Version::parse(&helm_version) < Version::parse(DEFAULT_HELM_VERSION) {
-            return Err(ClusterError::IncompatibleHelmVersion {
+            return Err(CheckError::IncompatibleHelmVersion {
                 installed: helm_version,
                 required: DEFAULT_HELM_VERSION.to_string(),
             });
@@ -612,11 +611,9 @@ impl ClusterInstaller {
             .helm_client
             .get_installed_chart_by_name(DEFAULT_CHART_SYS_REPO)?;
         if sys_charts.is_empty() {
-            return Err(ClusterError::MissingSystemChart);
+            return Err(CheckError::MissingSystemChart);
         } else if sys_charts.len() > 1 {
-            return Err(ClusterError::Other(
-                "Multiple fluvio system charts found".to_string(),
-            ));
+            return Err(CheckError::MultipleSystemCharts);
         }
 
         // check if cluster is already installed
@@ -624,7 +621,7 @@ impl ClusterInstaller {
             .helm_client
             .get_installed_chart_by_name(DEFAULT_CHART_APP_REPO)?;
         if !app_charts.is_empty() {
-            return Err(ClusterError::AlreadyInstalled);
+            return Err(CheckError::AlreadyInstalled);
         }
 
         // check k8 config hostname is not an IP address
@@ -634,24 +631,7 @@ impl ClusterInstaller {
                 // ignore server check for pod
             }
             K8Config::KubeConfig(config) => {
-                let server_host = match get_cluster_server_host(config) {
-                    Ok(server) => server,
-                    Err(e) => {
-                        return Err(ClusterError::Other(format!(
-                            "error fetching server from kube context {}",
-                            e.to_string()
-                        )))
-                    }
-                };
-                if !server_host.trim().is_empty() {
-                    if IpAddr::from_str(&server_host).is_ok() {
-                        return Err(ClusterError::InvalidMinikubeContext);
-                    };
-                } else {
-                    return Err(ClusterError::Other(
-                        "Cluster in kubectl context cannot have empty hostname".to_owned(),
-                    ));
-                }
+                check_cluster_server_host(config)?;
             }
         };
         Ok(())
@@ -659,24 +639,24 @@ impl ClusterInstaller {
 
     /// Given a pre-check error, attempt to automatically correct it
     #[instrument(skip(self, error))]
-    fn pre_install_fix(&self, error: ClusterError) -> Result<(), ClusterError> {
+    fn pre_install_fix(&self, error: CheckError) -> Result<(), ClusterError> {
 
         // Depending on what error occurred, try to fix the error.
         // If we handle the error successfully, return Ok(()) to indicate success
         // If we cannot handle this error, return it to bubble up
         match error {
-            ClusterError::MissingSystemChart if self.config.install_sys => {
+            CheckError::MissingSystemChart if self.config.install_sys => {
                 debug!("Fluvio system chart not installed. Attempting to install");
                 self._install_sys()?;
             }
-            ClusterError::InvalidMinikubeContext if self.config.update_context => {
+            CheckError::InvalidMinikubeContext if self.config.update_context => {
                 debug!("Updating to minikube context");
                 let context = MinikubeContext::try_from_system()?;
                 context.save()?;
             }
             unhandled => {
                 warn!("Pre-install was unable to autofix an error");
-                return Err(unhandled);
+                return Err(unhandled.into());
             },
         }
 
@@ -696,7 +676,7 @@ impl ClusterInstaller {
             // If all checks pass, perform the main installation
             Ok(()) => self.install_app()?,
             // If Fluvio is already installed, skip install step
-            Err(ClusterError::AlreadyInstalled) => (),
+            Err(ClusterError::PreCheckError { source: CheckError::AlreadyInstalled }) => (),
             // If there were other unhandled errors, return them
             Err(unhandled) => return Err(unhandled),
         }
--- a/src/cluster/src/install.rs
+++ b/src/cluster/src/install.rs
@@ -643,7 +643,11 @@ impl ClusterInstaller {
             // If Fluvio is already installed, skip install step
             Err(ClusterError::PreCheckError {
                 source: CheckError::AlreadyInstalled,
-            }) => (),
+            }) => {
+                debug!("Fluvio is already installed. Getting SC address");
+                let sc_address = self.wait_for_sc_service(&self.config.namespace).await?;
+                return Ok(sc_address);
+            },
             // If there were other unhandled errors, return them
             Err(unhandled) => return Err(unhandled),
         }
--- a/src/cluster/src/install.rs
+++ b/src/cluster/src/install.rs
@@ -23,7 +23,10 @@ use k8_obj_metadata::InputObjectMeta;
 
 use crate::ClusterError;
 use crate::helm::{HelmClient, Chart, InstalledChart};
-use crate::check::{check_cluster_server_host, CheckError, check_helm_version, check_system_chart, check_already_installed};
+use crate::check::{
+    check_cluster_server_host, CheckError, check_helm_version, check_system_chart,
+    check_already_installed,
+};
 
 const DEFAULT_NAMESPACE: &str = "default";
 const DEFAULT_REGISTRY: &str = "infinyon";
@@ -570,7 +573,6 @@ impl ClusterInstaller {
     /// [`with_system_chart`]: ./struct.ClusterInstaller.html#method.with_system_chart
     /// [`with_update_context`]: ./struct.ClusterInstaller.html#method.with_update_context
     fn pre_install(&self) -> Result<(), ClusterError> {
-
         // Continue fixing pre-check errors until we resolve all problems
         // or there is an error that we cannot fix
         loop {
@@ -604,7 +606,6 @@ impl ClusterInstaller {
     /// Given a pre-check error, attempt to automatically correct it
     #[instrument(skip(self, error))]
     fn pre_install_fix(&self, error: CheckError) -> Result<(), ClusterError> {
-
         // Depending on what error occurred, try to fix the error.
         // If we handle the error successfully, return Ok(()) to indicate success
         // If we cannot handle this error, return it to bubble up
@@ -621,7 +622,7 @@ impl ClusterInstaller {
             unhandled => {
                 warn!("Pre-install was unable to autofix an error");
                 return Err(unhandled.into());
-            },
+            }
         }
 
         Ok(())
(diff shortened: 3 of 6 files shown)


nicholastmosher:
This closes #235

nicholastmosher:
Hold up on merging this, I'm going to add some tracing in here to help with debugging.

src/cluster/src/install.rs
```diff
@@ -553,8 +553,21 @@ impl ClusterInstaller {
         fields(namespace = &*self.config.namespace),
     )]
     pub async fn install_fluvio(&self) -> Result<String, ClusterError> {
-        // perform pre install checks
-        self.pre_install_check()?;
+        // Preforms pre-install checks and tries to fix things if possible
```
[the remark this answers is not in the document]
nicholastmosher: The point of this is that pre-install can detect things that have exactly one fix. If we can detect a fix and we know exactly what needs to be done, we should just do it.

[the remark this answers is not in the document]
sehz: I agree but we should consolidate.  let's move action to `pre_install_check` and have option or flag to take action.

sehz (changes requested):

src/cluster/src/install.rs
```diff
@@ -457,6 +485,29 @@ pub struct ClusterInstaller {
...
+    ///
+    /// To customize the installation settings, use the [`builder()`]
+    /// function to create a [`ClusterInstallerBuilder`].
+    ///
+    /// # Example
+    ///
+    /// ```no_run
+    /// use fluvio_cluster::ClusterInstaller;
+    /// let installer = ClusterInstaller::new().unwrap();
+    /// async_std::task::block_on(installer.install_fluvio()).unwrap();
```
sehz, line 501: why use of `block_on` since this is not async fn

src/cluster/src/install.rs
```diff
@@ -465,7 +516,7 @@ impl ClusterInstaller {
     ///
     /// ```no_run
     /// use fluvio_cluster::ClusterInstaller;
-    /// let installer = ClusterInstaller::new().build().unwrap();
+    /// let installer = ClusterInstaller::builder().build().unwrap();
```
sehz, line 519: hmm. since this seems not intuitive, having build method on top of build method.

src/cluster/src/lib.rs
```diff
@@ -10,8 +10,8 @@
 //!
 //! ```no_run
 //! use fluvio_cluster::ClusterInstaller;
-//! let installer = ClusterInstaller::new().build().unwrap();
-//! fluvio_future::task::run_block_on(installer.install_fluvio()).unwrap();
+//! let installer = ClusterInstaller::new().unwrap();
+//! async_std::task::block_on(installer.install_fluvio()).unwrap();
```
sehz, line 14: unneeded use of block_on

src/cluster/src/install.rs
```diff
@@ -457,6 +485,29 @@ pub struct ClusterInstaller {
...
+    /// # Example
+    ///
+    /// ```no_run
+    /// use fluvio_cluster::ClusterInstaller;
+    /// let installer = ClusterInstaller::new().unwrap();
+    /// async_std::task::block_on(installer.install_fluvio()).unwrap();
+    /// ```
+    pub fn new() -> Result<Self, ClusterError> {
+        Ok(Self {
+            config: ClusterInstallerBuilder::new(),
```
sehz, line 505: It is weird that ClusterInstaller has dependencies to ClusterInstallBuilder.   It should be other way around?

sehz (changes requested):

sehz:
may need feed back from @Ryanmtate and @nacardin regarding new API

src/cluster/src/install.rs
nicholastmosher: `ClusterInstaller::install_fluvio` is an async function, so it needs the `block_on` to drive it

src/cluster/src/install.rs
nicholastmosher: This is probably just a problem with the example. There are now two ways to get an installer:

- With all default settings using `ClusterInstaller::new()?`
- With overridden settings using the builder

When using the builder style, you _create_ a builder using `ClusterInstaller::builder()`, and you _finalize_ a builder using `ClusterInstallerBuilder::build()`. So normally you'd have the settings in between those methods, like

```rust
let installer = ClusterInstaller::builder()
    .with_namespace("fluvio")
    .with_rust_log("debug")
    // .with_other_settings()
    .build()?;
```

I'll update the example to make more sense

src/cluster/src/lib.rs
nicholastmosher: `install_fluvio` is async

sehz: is this example or dockets?  if it is example, there is no need block on.   it should be just express as async snippet

src/cli/src/cluster/install/k8.rs
```diff
@@ -10,7 +10,7 @@ use super::*;
 pub async fn install_core(opt: InstallCommand) -> Result<(), CliError> {
     let (client, server): (TlsPolicy, TlsPolicy) = opt.tls.try_into()?;
 
-    let mut builder = ClusterInstaller::new()
+    let mut builder = ClusterInstaller::builder()
```
sehz, line 13: Typically, in the builder pattern, builder is references not the struct it is being [constructured](https://doc.rust-lang.org/1.0.0/style/ownership/builders.html).

src/cluster/src/install.rs
```diff
@@ -555,6 +624,36 @@ impl ClusterInstaller {
...
+    /// for more details:
+    ///
+    /// - [`with_system_chart`]
+    /// - [`with_update_context`]
+    ///
+    /// [`with_system_chart`]: ./struct.ClusterInstaller.html#method.with_system_chart
+    /// [`with_update_context`]: ./struct.ClusterInstaller.html#method.with_update_context
+    fn check_and_fix(&self) -> Result<(), ClusterError> {
+        // Preforms pre-install checks and tries to fix things if possible
+        match self.pre_install_check() {
```
sehz, line 640: Moving pre-check logic into installer will make Installer more complex.

sehz (changes requested):

vijaylaxmid:
I think this also fixes https://github.com/infinyon/fluvio/issues/313

src/cli/src/cluster/install/k8.rs
nicholastmosher: A few reasons I did it this way:

- The `ClusterInstallerBuilder` acts just like a config object at the end of the day, with all the options neatly laid out. I even call it `config` inside the `ClusterInstaller`
- An API user can avoid importing the `ClusterInstallerBuilder` type because all of the necessary methods are on the `ClusterInstaller` type. E.g.

```rust
use fluvio_cluster::ClusterInstaller;
let installer = ClusterInstaller::builder()
    .with_namespace("my-namespace")
    .with_rust_log("debug")
    .build()?;
installer.install_fluvio().await?;
```

- The `build()` method on the `ClusterInstallerBuilder` is the gate where things either initialize successfully or not. For example, we can do validation on whether any options can conflict with each other. Currently, this would be the point of failure if the HelmClient or K8Client could not be initialized.
- The Builder is using the consuming calls, i.e. they take `mut self`. I didn't want to have tons of consuming calls on the same object that was actually performing the installation.
- After calling `.build()`, all of the settings are locked in. That means you couldn't accidentally call `install_sys`, then change a setting, then call `install_app` with incompatible settings.

src/cluster/src/install.rs
nicholastmosher: Where do you imagine it all living? I feel like this crate is definitely the place for it. I would consider the `check` module but the `ClusterInstaller` has all of the parameters that you want the check to be compatible with, so for that reason it also kind of makes sense to put it here.

src/cluster/src/install.rs
nicholastmosher: I think sometimes it is laid out that way, but something I like about this pattern is that the `ClusterInstallerBuilder` is pretty much nothing more than a config object that is put together with builder methods where you can do validation along the way. Also, by using the `ClusterInstaller::builder() -> ClusterInstallerBuilder` method, users don't actually need to import the `ClusterInstallerBuilder` type.

src/cli/src/cluster/install/k8.rs
sehz: provisioning  helm client and K8client should be done at the time of install and should be defer to pre-check module.  Probably pre-check should can return resource as well.

src/cluster/src/install.rs
sehz: see previous comments.  Can make new struct or module, ResourceDependency.  It will be responsible for checking resources exists and return it if needed it.  That way installer can be simpler

src/cluster/src/lib.rs
nicholastmosher: This is a doc test that is marked as `no_run`. So it will do a compile-check, but not run anything. I included the `async_std` line so that even users who don't really know much about async will know that you need to do that in order to get this to run.

src/cli/src/cluster/install/k8.rs
nicholastmosher: > provisioning helm client and K8client should be done at the time of install

Why is that a requirement? This is ensuring the same order or operations

src/cluster/src/install.rs
nicholastmosher: Is that something that we need to do right now? Can we make an issue for that if we're going to start to doing such a big refactor? I don't really think that's super necessary to do immediately

sehz: This PR is major API changes. so can defer this and make smallest change now?

sehz: Since we want to start publishing crates soon, changes should be limited to additional or add minor behavior changes

nicholastmosher:
I've split out the builder changes and I'll open a new PR with it for further discussion 👍

sehz:
can you rebase from master?

src/cluster/src/error.rs
```diff
@@ -4,6 +4,7 @@ use serde::export::Formatter;
 use fluvio::FluvioError;
```
sehz, line 4: can you split out helm changes only?  this will make easier to digest changes

sehz (changes requested):

src/cluster/src/error.rs
nicholastmosher: Is there anything you'd like me to change about one or the other? If you're planning on merging them both right away can we just merge them both here? Otherwise I will spend a lot of time trying to extract the changes

src/cluster/src/install.rs
nicholastmosher: I kind of feel the opposite way actually. I feel like if there are major changes to be made they should be made before the release when they would not count as a breaking change, rather than after the release when they _would_ be a breaking change

src/cluster/src/install.rs
```diff
@@ -58,14 +59,20 @@ pub struct ClusterInstallerBuilder {
...
+    /// The location to find the fluvio sys chart
+    chart_location_sys: ChartLocation,
     /// The name of the SPU group to create
     group_name: String,
     /// The name of the Fluvio cloud
     cloud: String,
     /// Whether to save an update to the Fluvio profile
     save_profile: bool,
+    /// Whether to install fluvio-sys with fluvio-app
+    install_sys: bool,
```
sehz, line 73: chart related option seems to be should be separate out from here.    There are many different options here and this make it harder to incremental add new options later

src/cluster/src/install.rs
```diff
@@ -234,7 +251,7 @@ impl ClusterInstallerBuilder {
     /// # Example
     ///
     /// ```no_run
-    /// use fluvio_cluster::ClusterInstaller;
+    /// # use fluvio_cluster::ClusterInstaller;
     /// let installer = ClusterInstaller::new()
```
sehz, line 255: shouldn't this be refer to ClusterInstallBuilder?

src/cluster/src/error.rs
sehz: add more comments.    There are many question that need to be answered.  The helm changes are straightforward.

src/cluster/src/install.rs
sehz: major changes takes time to discuss review.   the change doesn't seem to be must have for this release

src/cluster/src/install.rs
```diff
@@ -58,14 +59,20 @@ pub struct ClusterInstallerBuilder {
     chart_name: String,
     /// A specific version of the Fluvio helm chart to install
     chart_version: String,
-    /// The location to find the helm chart
+    /// The location to find the fluvio app chart
     chart_location: ChartLocation,
+    /// The location to find the fluvio sys chart
+    chart_location_sys: ChartLocation,
```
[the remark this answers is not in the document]
nicholastmosher: The reason I added that was because when you install a local chart, you need to give a path to the exact chart file. I ran into a bug where I would try `fluvio cluster install --develop` and now that it also tries to install the -sys chart at the same time, it was attempting to install `fluvio-app` from `./k8-util/helm/fluvio-app` and it was also attempting to install `fluvio-sys` from `./k8-util/helm/fluvio-app`. So when we are using remote repositories, they will have the same value, but we want to have the flexibility to give two separate locations when using local directories.

src/cluster/src/install.rs
nicholastmosher: What specifically do you mean about chart related options? Right now the entire installation process is pretty much about configuring and installing charts in the correct way. How to install any particular chart is the responsibility of the `HelmClient`, but I think that installing the charts in such a way that works for Fluivo is exactly what the `ClusterInstaller` is meant to do.

src/cluster/src/install.rs
nicholastmosher: The way this is set up is that the user doesn't ever need to interact with the `ClusterInstallerBuilder` type. Currently, the `ClusterInstaller::new()` function actually returns a `ClusterInstallerBuilder`, which is why you see this in the documentation. I introduced #339 in order to make things a little bit nicer by making the function names more accurately represent what is happening. If we merge #339, this line would be rewritten as `let installer = ClusterInstaller::builder()...`

nicholastmosher:
I split out a PR for just the helm changes in #342

nicholastmosher:
This PR now contains just the changes for automatically fixing installation errors, building upon the new helm branch in #342

src/cluster/src/install.rs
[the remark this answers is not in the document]
sehz: can you past example of error? not able to see use case clearly

src/cluster/src/install.rs
sehz: if this flag is enabled,  `sys chart` should be installed only if it doesn't exist.

src/cluster/src/install.rs
```diff
@@ -58,14 +59,20 @@ pub struct ClusterInstallerBuilder {
...
     /// The name of the SPU group to create
     group_name: String,
     /// The name of the Fluvio cloud
     cloud: String,
     /// Whether to save an update to the Fluvio profile
     save_profile: bool,
+    /// Whether to install fluvio-sys with fluvio-app
+    install_sys: bool,
+    /// Whether to update the `kubectl` context
+    update_context: bool,
```
sehz, line 75: can we group both `install_sys` and `update_context` into separate option?  likely, there would be more option so we can be forward compatible

src/cluster/src/install.rs
```diff
@@ -88,7 +95,7 @@ impl ClusterInstallerBuilder {
     /// following:
     ///
     /// ```no_run
-    /// use fluvio_cluster::ClusterInstaller;
+    /// # use fluvio_cluster::ClusterInstaller;
     /// let installer = ClusterInstaller::new()
```
sehz, line 99: This should be refer to ClusterInstallBuilder not ClusterInstaller

src/cluster/src/install.rs
```diff
@@ -555,6 +609,36 @@ impl ClusterInstaller {
...
+    /// for more details:
+    ///
+    /// - [`with_system_chart`]
+    /// - [`with_update_context`]
+    ///
+    /// [`with_system_chart`]: ./struct.ClusterInstaller.html#method.with_system_chart
+    /// [`with_update_context`]: ./struct.ClusterInstaller.html#method.with_update_context
+    fn check_and_fix(&self) -> Result<(), ClusterError> {
+        // Preforms pre-install checks and tries to fix things if possible
+        match self.pre_install_check() {
```
sehz, line 625: let's move this under `pre_install_check` so we consolidate it `check and fix` features

src/cluster/src/install.rs
```diff
@@ -555,6 +609,36 @@ impl ClusterInstaller {
...
+            Ok(_) => (),
+            Err(ClusterError::MissingSystemChart) if self.config.install_sys => {
+                debug!("Fluvio system chart not installed. Attempting to install");
+                self._install_sys()?;
+            }
+            Err(ClusterError::InvalidMinikubeContext) if self.config.update_context => {
+                debug!("Updating to minikube context");
+                let context = MinikubeContext::try_from_system()?;
```
sehz, line 633: we should ask from confirmation from user since this is not typical installation behavior for kubernetes

sehz (changes requested):

src/cluster/src/install.rs
nicholastmosher: This is gated behind the `update_context` configuration option. If you like, we can make that option default to `false` and enable it with a CLI flag as necessary

sehz: sure

nicholastmosher:
Factored the `pre_install_check` function into a bunch of `check_X` functions and moved them to the `check` module. We also now have a dedicated `CheckError` type, which consists of errors we may see specifically when performing the pre-check. Some of the CheckError variants may be automatically fixed.

New pre-check flow:

- `install_fluvio` is invoked
- `install_fluvio` invokes `pre_install`
- `pre_install` executes a loop that does the following:
  - Runs `pre_install_check`, which returns a result.
  - If the result is Ok, return a success from pre_install
  - If the result is Err, pass that error to `pre_install_fix`
  - `pre_install_fix` will match on the error type to see if it can be automatically fixed or not.
    - If the error can be fixed, fixes it and returns Ok
    - If the error cannot be fixed automatically, returns that same error
  - If `pre_install_fix` returned Ok, re-enter the loop to check for subsequent pre-install checks and fixes
  - If `pre_install_fix` returned Err, bubble up to report to the user

src/cluster/src/install.rs
nicholastmosher: I'm not sure what you mean by grouping them into a separate option. They are two different configurations.
- `install_sys` causes the installer to check if `fluvio-sys` is installed, and to install it automatically if not
- `update_context` causes the installer to update the `kubectl` context and, if necessary, the /etc/hosts entry

src/cluster/src/install.rs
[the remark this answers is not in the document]
nicholastmosher: When I remove the `chart_location_sys` option and I run install with `--develop`, I get the following:

```
$ cargo run --bin fluvio -- cluster install --develop
>> helm version
>> helm list --filter ^fluvio-sys$ --output json
>> helm install fluvio-sys ./k8-util/helm/fluvio-app --namespace default --devel --set cloud=minikube
NAME: fluvio-sys
LAST DEPLOYED: Sat Oct  3 23:13:00 2020
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
>> helm list --filter ^fluvio-sys$ --output json
>> helm list --filter ^fluvio$ --output json
>> helm install fluvio ./k8-util/helm/fluvio-app --namespace default --devel --set image.registry=infinyon --set image.tag=f9c3749690b3a4b0882463baffc0c83f0cd32448 --set cloud=minikube --version 0.6.0-alpha.1
Exited with status code: 1
```

Note the first helm install is incorrectly using the `fluvio-app` local chart location for sys: 
```
helm install fluvio-sys ./k8-util/helm/fluvio-app --namespace default --devel --set cloud=minikube
```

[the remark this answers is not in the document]
nicholastmosher: In develop mode, should the installer use the local system chart at `k8-util/helm/fluvio-sys`? I assume so, I can update that

nicholastmosher:
Now when using the `--develop` flag, both `fluivio-app` and `fluvio-sys` will be loaded from the local chart paths correctly.

```
    Finished dev [unoptimized + debuginfo] target(s) in 0.24s
     Running `target/debug/fluvio cluster install --develop`
>> helm version
>> helm list --filter ^fluvio-sys$ --output json
>> helm install fluvio-sys ./k8-util/helm/fluvio-sys --namespace default --devel --set cloud=minikube
NAME: fluvio-sys
LAST DEPLOYED: Sat Oct  3 23:31:16 2020
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
>> helm list --filter ^fluvio-sys$ --output json
>> helm list --filter ^fluvio$ --output json
>> helm install fluvio ./k8-util/helm/fluvio-app --namespace default --devel --set image.registry=infinyon --set image.tag=f9c3749690b3a4b0882463baffc0c83f0cd32448 --set cloud=minikube --version 0.6.0-alpha.1
NAME: fluvio
LAST DEPLOYED: Sat Oct  3 23:31:17 2020
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
waiting for spu to be provisioned
1 spus provisioned
```

src/cluster/src/install.rs
```diff
@@ -708,16 +706,17 @@ impl ClusterInstaller {
                     install_settings,
                 )?;
             }
-            ChartLocation::Local(chart_location) => {
-                let chart_location = chart_location.to_string_lossy();
+            ChartLocation::Local(chart_home) => {
+                let chart_location = chart_home.join("fluvio-sys");
```
sehz, line 717: should be using `sys` chart name constant

src/cluster/src/install.rs
```diff
@@ -793,16 +792,17 @@ impl ClusterInstaller {
                 )?;
             }
             // For local, we do not use a repo but install from the chart location directly.
-            ChartLocation::Local(chart_location) => {
-                let chart_location = chart_location.to_string_lossy();
+            ChartLocation::Local(chart_home) => {
+                let chart_location = chart_home.join("fluvio-app");
```
sehz, line 803: use `app` chart constants

sehz (changes requested):

nicholastmosher:
I've removed the `chart_location_sys` option now that we're appending to a base chart_location.
Can we defer the chart factorizing to another issue? That seems more like a refactoring that is orthogonal to this PR.

sehz:
I agree.  We can postpone changes

sehz (approved):

sehz (changes requested):
need to bump up version

nicholastmosher:
You got it

sehz:
Sorry, I meant Cluster crate version

nicholastmosher:
This will also close #351

sehz (approved):

PR #333, opened 2020-09-29
Fix TLS opts not being used in CLI install

### LINKED ISSUE #332: k8 smoke test doesn't use TLS
steps to reproduce it:
`make smoke-test-k8-tls`

this result in fluvio profile:
```
~/.fluvio/config 
version = "2.0"
current_profile = "flvkube"
[profile.flvkube]
cluster = "flvkube"
[cluster.flvkube]
addr = "10.99.251.55:9003"

[cluster.flvkube.tls]
tls_policy = "disabled"
```

even thought it is using tls setup

```
installing cluster
>> /home/ubuntu/fluvio/target/debug/fluvio cluster install --spu 1 --tls --domain fluvio.local --ca-cert /home/ubuntu/fluvio/tls/certs/ca.crt --client-cert /home/ubuntu/fluvio/tls/certs/client.crt --client-key /home/ubuntu/fluvio/tls/certs/client.key --server-key /home/ubuntu/fluvio/tls/certs/server.key --server-cert /home/ubuntu/fluvio/tls/certs/server.crt --develop
```

nicholastmosher:
This should fix #332

### DIFF
--- a/src/cli/src/cluster/install/k8.rs
+++ b/src/cli/src/cluster/install/k8.rs
@@ -1,5 +1,6 @@
 use std::io::Error as IoError;
 use std::io::ErrorKind;
+use std::convert::TryInto;
 use std::process::Command;
 
 use fluvio_cluster::ClusterInstaller;
@@ -7,7 +8,7 @@ use fluvio::config::TlsPolicy;
 use super::*;
 
 pub async fn install_core(opt: InstallCommand) -> Result<(), CliError> {
-    let (client, server): (TlsPolicy, TlsPolicy) = opt.tls.into();
+    let (client, server): (TlsPolicy, TlsPolicy) = opt.tls.try_into()?;
 
     let mut builder = ClusterInstaller::new()
         .with_namespace(opt.k8_config.namespace)
--- a/src/cli/src/cluster/install/k8.rs
+++ b/src/cli/src/cluster/install/k8.rs
@@ -1,15 +1,20 @@
-use std::process::Command;
 use std::io::Error as IoError;
 use std::io::ErrorKind;
-use super::*;
+use std::process::Command;
+
 use fluvio_cluster::ClusterInstaller;
+use fluvio::config::TlsPolicy;
+use super::*;
 
 pub async fn install_core(opt: InstallCommand) -> Result<(), CliError> {
+    let (client, server): (TlsPolicy, TlsPolicy) = opt.tls.into();
+
     let mut builder = ClusterInstaller::new()
         .with_namespace(opt.k8_config.namespace)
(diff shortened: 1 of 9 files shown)


sehz:
there seems to be other TLS problem,  don't enable TLS in CI yet.

src/cli/src/cluster/install/k8.rs
```diff
@@ -1,15 +1,20 @@
...
 pub async fn install_core(opt: InstallCommand) -> Result<(), CliError> {
+    let (client, server): (TlsPolicy, TlsPolicy) = opt.tls.into();
+
     let mut builder = ClusterInstaller::new()
         .with_namespace(opt.k8_config.namespace)
         .with_group_name(opt.k8_config.group_name)
         .with_spu_replicas(opt.spu)
-        .with_save_profile(!opt.skip_profile_creation);
+        .with_save_profile(!opt.skip_profile_creation)
+        .with_tls(client, server);
```
sehz, line 17: need to unit test to ensure InstallCommand is properly translated into ClusterInstall options

src/cli/src/cluster/install/tls.rs
```diff
@@ -0,0 +1,69 @@
...
+    /// TLS: client key
+    #[structopt(long, required_if("tls", "true"), parse(from_os_str))]
+    pub client_key: Option<PathBuf>,
+
+    /// TLS: ca cert
+    #[structopt(long, required_if("tls", "true"), parse(from_os_str))]
+    pub ca_cert: Option<PathBuf>,
+}
+
+impl From<TlsOpt> for (TlsPolicy, TlsPolicy) {
```
sehz, line 37: add unit test

sehz (changes requested):

src/cli/src/cluster/install/tls.rs
nicholastmosher: *sigh* looks like StructOpt has a weird problem where `required_if` does not work for a boolean field.
I.e. all of the `required_if("tls", "true")` calls are doing nothing at all :(

src/cli/src/cluster/install/k8.rs
nicholastmosher: I'm not sure there's an easy way to do this with the current dependency structure and visibility modifiers. `ClusterInstaller` keeps it's fields private in order to keep implementation freedom for the future, and we don't want to make a two-way dependency by making the Client link to the CLI.

src/cli/src/cluster/install/tls.rs
nicholastmosher: I rolled back to custom validation

sehz (approved):

PR #336, opened 2020-09-29
add install fluvio github composite action

Ryanmtate:
This PR adds a composite github action for re-use in github workflows.

To use this github action in another workflow, you should be able to use the following:

```yaml
...
- name: Install Fluvio Local Cluster
   uses: infinyon/fluvio@latest
- name: Check Fluvio Installation
  run: |
    fluvio --help
```

We'll need to tag release once merged and publish to the github actions market place.

Not sure if we want to lock the version of the github action with the version of fluvio. I am using `latest` in the example above, and in the action.sh script it will always build from the fluvio repo master branch source.

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -1,0 +1,54 @@
+name: 'fluvio'
+description: 'Installs Fluvio CLI from source with required dependencies (e.g. Helm, Rust) and starts a Local Cluster. Used for developing fluvio.'
+inputs:
+  spus:  # number of SPUs
+    description: 'number of SPUs to create for local cluster'
+    required: false
+    default: "1"
+  cluster-type:
+    description: 'Type of cluster (i.e. local or kubernetes)'
+    required: true
+    default: 'local'
+  rust-log:
+    description: "Add rust log options (e.g. debug, warn, info)"
+    required: false
+    default: 'info'
+  version:
+    description: "Use a specific version of the fluvio CLI (e.g. latest, 0.6.0)"
+    required: false
+    default: 'latest'
+  development:
+    description: "Use a debug release version of fluvio"
+    required: false
+    default: "true"
+  helm-version:
+    description: "Use a specific helm version"
+    required: false
+    default: 'v3.3.4'
+  minikube-version:
+    description: "Use a specific minikube version"
+    required: false
+    default: 'v1.13.1'
+runs:
+  using: "composite"
+  steps: 
+    - id: set-environment-variables
+      run: |
+        echo "::set-env name=SPU_NUMBER::${{ inputs.spus }}"
+        echo "::set-env name=CLUSTER_TYPE::${{ inputs.cluster-type }}"
+        echo "::set-env name=RUST_LOG::${{ inputs.rust-log }}"
+        echo "::set-env name=VERSION::${{ inputs.version }}"
+        echo "::set-env name=DEVELOPMENT::${{ inputs.development }}"
+        echo "::set-env name=HELM_VERSION::${{ inputs.helm-version }}"
+        echo "::set-env name=MINIKUBE_VERSION::${{ inputs.minikube-version }}"
+        echo "::set-env name=OS::${{ runner.os }}"
+      shell: bash
+    - id: install-helm
+      run: ${{ github.action_path }}/dev-tools/ci-replace-helm.sh
+      shell: bash
+    - id: install-minikube
+      run: ${{ github.action_path }}/dev-tools/action-install-minikube.sh
+      shell: bash
+    - id: fluvio-cluster
+      run: ${{ github.action_path }}/dev-tools/action-install-fluvio-cluster.sh
+      shell: bash
--- a/dev-tools/action-install-fluvio-cluster.sh
+++ b/dev-tools/action-install-fluvio-cluster.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+# This script is ran by the github actions to install fluvio in
+# GitHub Action Workflows.
+echo "Installing Fluvio Local Cluster"
+if [ "$DEVELOPMENT" = "true" ]; then
+        if [ "$VERSION" = "latest" ]; then 
+                echo "Installing from latest source"
+                # Download Fluvio Repo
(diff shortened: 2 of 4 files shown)


Ryanmtate:
I'm going to close this. After reading the documentation, I don't believe we can use composite actions and use third party actions; writing a composite action may be more work than using a docker image. I am also not sure that it makes sense to try to install helm alongside fluvio in the github actions, if we need to go the docker route.

Ryanmtate:
I am re-opening; while the composite does not allow 'uses' for other actions, I'll write a shell script that will install for ubuntu.

action.sh
```diff
@@ -0,0 +1,79 @@
...
+        # Install
+        curl -Lo minikube https://storage.googleapis.com/minikube/releases/latest/minikube-${PLATFORM} \
+        && chmod +x minikube
+        sudo install minikube /usr/local/bin/
+
+        # Start Minikube with `none` driver
+        minikube start --driver=none
+
+        # Install Helm
+        curl -sSL https://get.helm.sh/helm-${HELM_VERSION}-${PLATFORM}.tar.gz | tar zx
```
sehz, line 21: can GitHub Action script can refer to another script? since we already have helm install script?

sehz:
This doesn't need to be check-into master for testing.  You can refer this actions using commit tag or branch:  https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idsteps

sehz (changes requested):

action.sh
Ryanmtate: commit 72da370 now adds a step to install helm using the ci-replace-helm.sh script

Ryanmtate:
> This doesn't need to be check-into master for testing. You can refer this actions using commit tag or branch: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idsteps


Looks like its working for the node client test 🎉 

https://github.com/infinyon/flv-client-node/runs/1191089904?check_suite_focus=true

Ryanmtate:
> > This doesn't need to be check-into master for testing. You can refer this actions using commit tag or branch: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idsteps
> 
> Looks like its working for the node client test 🎉
> 
> https://github.com/infinyon/flv-client-node/runs/1191089904?check_suite_focus=true

Spoke to soon; I think there is still some issues with the minikube installation; not ready to be merged yet.

action.yaml
```diff
@@ -0,0 +1,62 @@
...
+    default: 'info'
+  version:
+    description: "Use a specific version of the fluvio CLI (e.g. latest, 0.6.0)"
+    required: false
+    default: 'latest'
+  development:
+    description: "Use a debug release version of fluvio"
+    required: false
+    default: "true"
+  platform:
```
nacardin, line 24: I don't think there is a good way to map the OS name "ubuntu-latest", "macOS-latest", or "Linux" to platform name in the workflow. Rather than accepting "platform" as an input, is is best computed in the shell scripts used by this action. You would just need to add to `action-install-minikube.sh ` the same mapping logic in `ci-replace-helm.sh`.

action.yaml
```diff
@@ -0,0 +1,62 @@
...
+inputs:
+  spus:  # number of SPUs
+    description: 'number of SPUs to create for local cluster'
+    required: false
+    default: "1"
+  cluster-type:
+    description: 'Type of cluster (i.e. local or kubernetes)'
+    required: true
+    default: 'local'
+  log-options:
```
nacardin, line 12: It might be easier to understand by calling it `rust-log` and mentioning that it is equivalent to `RUST_LOG` in the description.

action.yaml
```diff
@@ -0,0 +1,62 @@
...
+        echo "::set-env name=OS::${{ runner.os }}"
+      shell: bash
+    - id: install-helm
+      run: ${{ github.action_path }}/dev-tools/ci-replace-helm.sh
+      shell: bash
+    - id: install-minikube
+      run: ${{ github.action_path }}/dev-tools/action-install-minikube.sh
+      shell: bash
+    - id: set-minikube-context
+      run: ${{ github.action_path }}/dev-tools/minikube-mycube.sh
```
nacardin, line 58: This shouldn't be necessary since `fluvio cluster set-minikube-context` is already executed in `dev-tools/action-install-fluvio-cluster.sh:24`.

nacardin (changes requested):

action.yaml
Ryanmtate: commit 168c946 removes.

action.yaml
Ryanmtate: commit 168c946 renames.

action.yaml
Ryanmtate: Commit 168c946 removes; I've added the case to the minikube install; but I was also thinking that maybe we should replace the case with the following for determining OS, using `$OSTYPE`:


```bash

if [[ "$OSTYPE" == "linux-gnu"* ]]; then
       # Install for Linux
elif [[ "$OSTYPE" == "darwin"* ]]; then
        # Mac OSX
elif [[ "$OSTYPE" == "cygwin" ]]; then
        # POSIX compatibility layer and Linux environment emulation for Windows
elif [[ "$OSTYPE" == "msys" ]]; then
        # Lightweight shell and GNU utilities compiled for Windows (part of MinGW)
elif [[ "$OSTYPE" == "win32" ]]; then
        # I'm not sure this can happen.
elif [[ "$OSTYPE" == "freebsd"* ]]; then
        # ...
else
        # Unknown.
fi
```

nacardin: That would be a slight improvement if `$OSTYPE` is reliable and you should also check the CPU arch. But I think this is good for now.

sehz (approved):

sehz:
should this be merged now?

Ryanmtate:
Let me run another test against it before we merge. Need to confirm it works.

Ryanmtate:
> should this be merged now?

Pending CI check, this is now ready. It is passing for https://github.com/infinyon/fluvio-client-node/pull/12

PR #339, opened 2020-09-30
Cluster install builder

nicholastmosher:
(do not merge before #325 )

I split out the builder changes I wanted to make in #325 so that the discussion wouldn't block that PR from merging. Here are the changes made here and why I think they're a good idea:

- Rename the `ClusterInstaller::new() -> ClusterInstallerBuilder` method to `ClusterInstaller::builder() -> ClusterInstallerBuilder`
  - I think this is a good idea because it better communicates what is happening. Previously, I even had to suppress a clippy warning because the `new` function was not returning a `Self` type. Now we have an appropriately-named method to start the builder process.
- Create a new implementation of `new` as `ClusterInstaller::new() -> Result<Self, FluvioError>`. This is more idiomatic because it is returning a `Self` type. This is essentially a shortcut to bypass all the builder options by using the defaults.

- ### Why not just do `ClusterInstallerBuilder::new()`?

I think there's something very clean about having one "primary" data type per module, and I think for the install module that type is the `ClusterInstaller` itself. With this proposed setup, the user would not even need to import the `ClusterInstallerBuilder` type, they can just kick off the process using `ClusterInstaller::builder()`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1002,6 +1002,7 @@ dependencies = [
 name = "fluvio-cluster"
 version = "0.1.3"
 dependencies = [
+ "async-std",
  "fluvio",
  "fluvio-controlplane-metadata",
  "fluvio-future",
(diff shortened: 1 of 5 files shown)


sehz:
Closing this for now.  Resubmit once all cluster logic has been moved out of CLI into install and come up consensus on how to configure

PR #344, opened 2020-10-01
Better error handling and reporting

nicholastmosher:
Errors are an important element of a user's experience, especially since we're at a stage where things go wrong fairly often. This PR finishes the quest to implement `std::error::Error` across all of our main error types, including `FluvioError` and `CliError`. Most of our error types are now implemented using `thiserror`, which makes it easy to nest errors such that `std::error::Error::source()` can dig down to find the root cause of many issues. Additionally, the CLI binary is now making good use of the `anyhow` crate for easy error reporting. This makes it easy to take an error with nested source errors and print them in a backtrace-like way, showing which low-level errors caused various high-level errors.

A list of important elements of this PR:

- impl `Error` for `fluvio_sc_schema::ApiError`
- impl `Error` for `fluvio_cluster::HelmError`
- impl `Error` for `fluvio_cluster::ClusterError`
- impl `Error` for `fluvio::FluvioError`
- Introduce `fluvio::config::ConfigError` as sub-error of `FluvioError`
- impl `Error` for `fluvio::config::ConfigError`
- impl `Error` for `CliError`
- Convert many CLI `Result<T, CliError>` instances to `anyhow::Result<T>`

Here are some examples of what the new error reporting looks like from the CLI:

- ### When unable to connect to minikube:

Old errors (from master right now)

```
$ fluvio topic list
error: IoError(
    Custom {
        kind: TimedOut,
        error: "time out in send and request: 3",
    },
)
```

New errors (from this branch)

```
$ fluvio topic list
Error: Fluvio socket error

Caused by:
    0: IO error: time out in send and request: 3
    1: time out in send and request: 3
```

- ### When I make `~/.fluvio/config` invlid TOML:

Old errors (from master right now)

```
$ fluvio profile view
no profile can be founded
```

New errors (from this branch)

```
$ fluvio profile view
Error: Failed to read Fluvio config file

Caused by:
    0: Fluvio config error
    1: Failed to deserialize Fluvio config
    2: expected an equals, found a newline at line 9 column 16
```


sehz:
This is great!  It will us to diagnose issue better

One thing thought, have you look at [eyre](https://crates.io/crates/eyre)?  It is replacement for anyhow.  Same author for tracing library so I am sure it has good integration.    This is not show stopper.  We could migrate to Eyre later.   Changes here are much better than status quo

sehz:
also, rebase from master

src/cli/src/cluster/install/k8.rs
```diff
@@ -1,13 +1,12 @@
...
-use std::io::ErrorKind;
 use std::convert::TryInto;
 use std::process::Command;
+use anyhow::Context;
 
 use fluvio_cluster::ClusterInstaller;
 use fluvio::config::TlsPolicy;
 use super::*;
 
-pub async fn install_core(opt: InstallCommand) -> Result<(), CliError> {
```
sehz, line 10: is there reason for removing type signature?

sehz (changes requested):

src/cli/src/cluster/install/k8.rs
nicholastmosher: So here, `CliError` was doing nothing but catching the custom `IoError` from the git_hash line. `eyre/anyhow` can catch any error type that impl's std::error::Error, so this is pretty much equivalent. In a different case, it may be useful to manually create a `CliError` instance before returning it, but there was no extra value that would have added here.

nicholastmosher:
@sehz Thanks for the suggestion about eyre, it looks great. I'm curious to see how `tracing-error` can fit into all of this, we're not using it currently and I'm not sure what exactly to do to use it but I imagine it could potentially help a lot. I'll look at that at some point in the future though.

src/cli/src/cluster/install/k8.rs
sehz: Reason for this is to statistically determine error during compile time as noted by rust-lang doc: https://doc.rust-lang.org/stable/rust-by-example/error/multiple_error_types/boxing_errors.html.   Yes it is more verbose and work but we want to preserve Rust's compile type checking much as possible

nicholastmosher: I agree with you in the context of library errors, where the structure of the errors and the internal source errors matters. That's why I didn't use anyhow/eyre in the library crates, only in the CLI crate. Note that in any of these functions returning an `eyre::Result`, we _could_ construct a `CliError` and return it, using `?`'s From-coercion to convert the type. This would be the case if we were to do something like this:

```rust
fn some_function(arg: i32) -> eyre::Result<()> {
    if arg < 0 {
        return Err(CliError::invalid_arg("Arg cannot be less than 0!").into());
    }
    Ok(())
}
```

Then the structure of the error is still preserved, i.e. the Display of the `CliError::InvalidArg` is shown in the error stack.

But overall, libraries mostly _produce_ errors, so they should be well-structured and typed, which is what `thiserror` is good at helping with. Applications mostly _consume_ errors and are interested in reporting them, which is what the `anyhow/eyre` crates are good at, and why it's fine to use `eyre::Result<T>` to capture any error. https://nick.groenen.me/posts/rust-error-handling/#libraries-versus-applications

sehz: Separation between library and application is not so clear since some of CLI could be turn into library as happened with Cluster crate.   I don't see clear justification for changing this except to save couple of lines and don't know long term consequence of losing concrete error type.  So Let's keep status quo until it is more clear.

sehz (approved):

PR #346, opened 2020-10-01
Wait for SC DNS resolution and tcp port check

nacardin:


src/cluster/src/install.rs
```diff
@@ -844,11 +863,55 @@ impl ClusterInstaller {
         Ok(None)
     }
 
+    /// Wait until the Fluvio SC public service appears in Kubernetes
+    async fn wait_for_sc_port_check(&self, addr: &str) -> Result<Option<String>, ClusterError> {
+        for i in 0..30u16 {
+            info!(attempt = i, "waiting for sc port check");
+            let sock_addr_string = format!("{}:9003", addr);
```
sehz, line 870: can't assume 9003 port, should use whatever configuration in the cluster installer

sehz (changes requested):

src/cluster/src/install.rs
```diff
@@ -844,11 +863,55 @@ impl ClusterInstaller {
...
+    async fn wait_for_sc_dns(
+        &self,
+        sock_addr_string: &str,
+    ) -> Result<Vec<SocketAddr>, ClusterError> {
+        for i in 0..30u16 {
+            info!(attempt = i, "waiting for sc dns resolution");
+            match resolve(sock_addr_string).await {
+                Ok(sock_addr) => return Ok(sock_addr),
+                Err(err) => {
+                    let sleep_ms = 1000 * 2u64.pow(i as u32);
```
nicholastmosher, line 896: Do we want to be waiting this long? It seems like this could hang forever if it never finds the SC

src/cluster/src/install.rs
```diff
@@ -844,11 +863,55 @@ impl ClusterInstaller {
...
+                    info!(
+                        attempt = i,
+                        "dns resoultion failed {}, sleeping for {} ms", err, sleep_ms
+                    );
+                    sleep(Duration::from_millis(sleep_ms)).await
+                }
+            }
+        }
+
+        Err(ClusterError::Other("unable to resolve DNS".to_owned()))
```
nicholastmosher, line 906: Should we have a special error type for this?

src/cluster/src/install.rs
```diff
@@ -871,8 +934,12 @@ impl ClusterInstaller {
                     attempt = i,
                     "Not all SPUs are ready. Waiting",
                 );
-                println!("{} of {} spu ready", ready_spu, spu);
-                sleep(Duration::from_millis(1000)).await;
+                let sleep_ms = 1000 * 2u64.pow(i as u32);
```
nicholastmosher, line 937: I don't know if exponential backoff over 30 iterations is a good idea. It doesn't really take a failure case into consideration

src/cluster/src/install.rs
sehz: or check for maximum allowable time

src/cluster/src/install.rs
sehz: Yes, maybe `SC_NOT_REACHABLE`

src/cluster/src/install.rs
nacardin: Rust doesn't fail

src/cluster/src/install.rs
nacardin: Will add error type

src/cluster/src/install.rs
nacardin: That's what it was doing before. I'll update to use config value

src/cluster/src/install.rs
nacardin: will reduce iterations

src/cluster/src/install.rs
nacardin: will reduce iterations

sehz (approved):

PR #358, opened 2020-10-03
Switch to fluvio-cloud context after login

### LINKED ISSUE #299: CLI should switch to the newly sync'd cluster profile
Currently when running `fluvio profile sync cloud`. It only create a new cluster profile in the config, but doesn't set it as the `current_profile`.

nicholastmosher:
FIxes #299


src/cli/src/profile/sync/cloud/mod.rs
```diff
@@ -88,8 +88,8 @@ fn save_cluster<O: Terminal>(
     let profile = Profile::new("fluvio-cloud".to_string());
     config.add_cluster(cluster, "fluvio-cloud".to_string());
     config.add_profile(profile, "fluvio-cloud".to_string());
+    config.set_current_profile("fluvio-cloud");
```
sehz, line 91: can we create constant for `fluvio-cloud`?

sehz (changes requested):

sehz (approved):

Issue #365, opened 2020-10-05
API simplification


sehz:
For first OSS release, we should make  following updates to make it simpler with minimum changes.

For consumer.  add `RecordIterator`, similar to iterator on hash map.  It will return tuple of metadata and records.  So example  code in the consumer can be look like this:

```
async fn consume() -> Result<(), FluvioError> {
    let consumer = fluvio::consumer(TOPIC, 0).await?;
    let mut stream = consumer.stream(Offset::beginning()).await?;

    while let Some((_,record)?) = stream.next().await {
      
          let string = String::from_utf8(record.as_byte()).unwrap();
          println!("Got record: {}", string);
          if string == "Done!" {
                  return Ok(());
          }
          
    }

    Ok(())
}

sehz:
Updated to add try operator in the while block since `next()` should return Result.   This should make it compatible with `try_next()` in the `StreamExt`

nicholastmosher:
I'm not seeing a `try_next` method in the `StreamExt` trait here: https://docs.rs/futures/0.3.6/futures/stream/trait.StreamExt.html

I think I know what method you're talking about, but does it belong to a different trait or library? Can you find it and post it here?

nicholastmosher:
Can we also start a discussion about what steps are needed in order to create a proper `Stream`-implementation? I'm not familiar enough with the inner workings of the consumer to know what the fetching process looks like.

sehz:
Sorry discard comment about about `try_next`.   It is deprecated so shouldn't have to worry about it.  Important point is that `next` should return `Option<Result<...`

For stream implementation,  challenges is map futures -> stream which means you need to hold some state in order to map to stream.  Some example: https://github.com/infinyon/k8-api/blob/master/src/k8-client/src/list_stream.rs

If this is too complex,  we can leave as futures and address stream conversion in next release

sehz:
From looking at the code, simplest implementation is to add `Stream` trait in: [AsyncResponse](https://github.com/infinyon/fluvio-socket/blob/master/crates/socket/src/multiplexing.rs)

sehz:
Or you can implement something like: [BufferedStream](https://docs.rs/futures-util/0.3.6/src/futures_util/stream/stream/buffered.rs.html#14-23) over batch stream

sehz:
Or use https://docs.rs/futures/0.3.6/futures/stream/trait.StreamExt.html#method.flat_map

nicholastmosher:
[...]

PR #372, opened 2020-10-07
Consumer api

nicholastmosher:
Implements #365 

This incorporates the changes from the `fluivo-socket` crate where `AsyncResponse` now implements `Stream`. It's important to note that `AsyncResponse` streams batches of records. The wrapper code in the consumer API is able to flat_map those batches into records so that consumers just see a nice record type. Basic consumer code will now look like this:

```rust
let mut stream = consumer.stream("my-topic", 0).await?;
    if let Some(bytes) = record.try_into_bytes() {
        let string = String::from_utf8_lossy(&bytes);
        println!("Got record: {}", string);
    }
}
```

Right now, I'm not able to get the smoke tests to work, but I'm also not sure if that's due to my changes or not. You can try out the changes by checking out the `stream` branch of `fluvio-socket` adjacently to the `fluvio` repo so the dependency patch applies. I'd like to have somewhat of a deeper review on this PR (on the fluvio-socket side) to make sure I haven't broken anything.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ members = [
 ]
 
 [patch.crates-io]
-fluvio-socket = { path = "../fluvio-socket/crates/socket" }
+fluvio-socket = { git = "https://github.com/nicholastmosher/fluvio-socket", branch = "stream" }
 
 # profile to make image sizer smaller
 # comment out for now
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,9 +19,6 @@ members = [
     "tests/runner",
 ]
 
-[patch.crates-io]
-fluvio-socket = { git = "https://github.com/nicholastmosher/fluvio-socket", branch = "stream" }
-
 # profile to make image sizer smaller
 # comment out for now
 [profile.release]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,9 @@ members = [
     "tests/runner",
 ]
 
+[patch.crates-io]
+fluvio-socket = { path = "../fluvio-socket/crates/socket" }
+
 # profile to make image sizer smaller
 # comment out for now
 [profile.release]
--- a/examples/00-echo/Cargo.toml
+++ b/examples/00-echo/Cargo.toml
@@ -8,5 +8,5 @@ edition = "2018"
 
 [dependencies]
 fluvio = { path = "../../src/client" }
-futures = "0.3.6"
+futures-lite = "1.10.1"
 async-std = "1.6.4"
--- a/examples/00-echo/Cargo.toml
+++ b/examples/00-echo/Cargo.toml
@@ -8,4 +8,5 @@ edition = "2018"
 
 [dependencies]
 fluvio = { path = "../../src/client" }
+futures = "0.3.6"
 async-std = "1.6.4"
--- a/src/cli/Cargo.toml
+++ b/src/cli/Cargo.toml
@@ -57,7 +57,7 @@ k8-config = { version = "1.3.0", features = ["context"] }
 k8-obj-core = { version = "1.1.0" }
 k8-obj-metadata = { version = "1.0.0" }
 k8-metadata-client = { version = "1.0.0" }
-fluvio = { version = "0.1.0", path = "../client" }
+fluvio = { version = "0.2.0", path = "../client" }
 fluvio-sc = { version = "0.1.0", path = "../sc", optional = true, features = ["k8"] }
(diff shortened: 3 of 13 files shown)


Cargo.toml
```diff
@@ -19,6 +19,9 @@ members = [
     "tests/runner",
 ]
 
+[patch.crates-io]
+fluvio-socket = { path = "../fluvio-socket/crates/socket" }
```
sehz, line 23: I assume this is still testing

sehz: You can make it rely on GitHub branch

examples/00-echo/Cargo.toml
```diff
@@ -8,4 +8,5 @@ edition = "2018"
 
 [dependencies]
 fluvio = { path = "../../src/client" }
+futures = "0.3.6"
```
sehz, line 11: use `futures-util` or `futures-lite`

src/client/Cargo.toml
```diff
@@ -17,6 +17,7 @@ admin = ["fluvio-sc-schema/use_serde"]
 [dependencies]
 tracing = "0.1.19"
 tracing-futures = "0.2.4"
+futures = "0.3.6"
```
sehz, line 20: use `futures-util` or `futures-lite`

src/client/src/consumer.rs
```diff
@@ -349,3 +389,18 @@ impl ConsumerConfig {
         self
     }
 }
+
```
sehz, line 392: may mark as hidden?

tests/runner/Cargo.toml
```diff
@@ -6,7 +6,8 @@ edition = "2018"
 
 [dependencies]
 bytes = "0.5.3"
-futures-util = { version = "0.3.5" }
+futures = "0.3.6"
```
sehz, line 9: use `futures-util`

sehz: revert

tests/runner/src/tests/smoke/consume.rs
```diff
@@ -75,26 +76,41 @@ async fn validate_consume_message_api(offsets: Offsets, option: &TestOption) {
...
-                            "   consumer: total records: {}, validated offset: {}",
-                            total_records, offset
-                        );
-                        assert_eq!(offset, base_offset + total_records as i64);
-                        if total_records == iteration {
-                            println!("<<consume test done for: {} >>>>", topic_name);
-                            break 'outer;
-                        }
-                    }
+        // 'outer: while let Some(Ok(event)) = stream.next().await {
```
sehz, line 79: remove comments

sehz (changes requested):

sehz:
clippy is failing.  In the CI, any failure will cause cancellation of other jobs

src/client/src/consumer.rs
nicholastmosher: Are you talking about the Record type?

src/cli/Cargo.toml
```diff
@@ -57,7 +57,7 @@ k8-config = { version = "1.3.0", features = ["context"] }
 k8-obj-core = { version = "1.1.0" }
 k8-obj-metadata = { version = "1.0.0" }
 k8-metadata-client = { version = "1.0.0" }
-fluvio = { version = "0.1.0", path = "../client" }
+fluvio = { version = "0.2.0", path = "../client" }
 fluvio-sc = { version = "0.1.0", path = "../sc", optional = true, features = ["k8"] }
 fluvio-sc-schema = { version = "0.1.0", path = "../sc-schema", features = ["use_serde"] }
 fluvio-spu = { version = "0.1.0", path = "../spu", optional = true }
```
sehz, line 63: need to bump up cli version

src/client/src/consumer.rs
sehz: Yes

src/cluster/Cargo.toml
```diff
@@ -24,7 +24,7 @@ serde_json = "1.0.57"
 thiserror = "1.0.20"
```
sehz, line 25: bump up version for cluster

sehz (changes requested):

src/client/src/consumer.rs
nicholastmosher: I think that although this Record type has `pub` in it, the `consumer` module is not pub, so any clients will not be able to import the Record name. I think this is the behavior we want.

sehz: fair enough

sehz (approved):

sehz:
rebase from master then we merge

Issue #371, opened 2020-10-07
CLI Proposal 2: Use global flags for TLS options


nicholastmosher:
While looking at the arguments to all of our subcommands, it occurred to me that we have the `tls` family of arguments in nearly every command

```
    -c, --cluster <host:port>          Address of cluster
        --domain <domain>              Required if client cert is used
        --ca-cert <ca-cert>            Path to TLS ca cert, required when client cert is enabled
        --client-cert <client-cert>    Path to TLS client certificate
        --client-key <client-key>      Path to TLS client private key
    -P, --profile <profile>
```

These are intended to allow the connection parameters to be tuned on a per-command basis, which is a good thing. However, the docs for these commands are repetitive and pollute the help screens of most of our commands. The following commands each have a copy of these arguments:

```
fluvio consume                           Reads messages from a topic/partition
fluvio produce                           Writes messages to a topic/partition
fluvio spu list                          List custom & managed SPUs
fluvio spu-group create                  Create managed SPU group
fluvio spu-group delete                  Delete managed SPU group
fluvio spu-group list                    List managed SPU groups
fluvio custom-spu register               Create custom SPU
fluvio custom-spu unregister             Delete custom SPU
fluvio custom-spu list                   List custom SPUs
fluvio topic create                      Creates a topic
fluvio topic delete                      Deletes a topic
fluvio topic describe                    Shows details of a topic
fluvio topic list                        Shows all topics
fluvio partition list                    Show all partitions
fluvio profile sync k8
fluvio profile sync local
```

StructOpt has a mechanism for creating [global args](https://docs.rs/structopt/0.3.18/structopt/#raw-methods), which would allow us to define the TLS family of arguments to live at the top-level and be implicitly accepted for any command. This would slightly change the order in which the arguments would need to be given. For example, if we wanted to run `fluvio topic list` with custom TLS arguments, the TLS args would go at the top-level right after `fluvio`, like this:

```
$ fluvio --tls \
         --domain=fluvio.io \
         --ca-cert=/tmp/ca.crt \
         --client-cert=/tmp/client.crt \
         --client-key=/tmp/client.key \
         topic list
```

I think that this change would be a good idea for a few reasons:

1) The help menus that the users see would not be polluted by the same arguments over and over
2) The user will be using profiles from their configuration file a vast majority of the time. I doubt they'll ever use these arguments anyway
[...]

PR #378, opened 2020-10-08
Cli proposal2

nicholastmosher:
Implements #371  using `ClusterTarget` as a set of global arguments rather than duplicated local arguments.


sehz:
Seems like with plug-in system, there is no really global argument.
Also, some command like install, has slightly different argument.
Deferring this

PR #392, opened 2020-10-13
Feature/alpha install script

Ryanmtate:


dev-tools/action-install-fluvio-cluster.sh
```diff
@@ -16,6 +16,12 @@ if [ "$DEVELOPMENT" = "true" ]; then
                 # Install Fluvio Src
                 cargo install --path ./src/cli
                 popd
+        elif [ "$VERSION" = "v0.6.0-alpha.3" ]; then
```
sehz, line 19: can we avoid hard coding version number? can consumer of this action use git commit or tag for dependency to this script?

dev-tools/action-install-minikube.sh
```diff
@@ -30,10 +30,26 @@ if [[ "$OSTYPE" == "linux-gnu"* ]]; then
...
+
+        # Start Minikube with `hyperkit` driver
+        # OSx does not have `none` bare metal option
+        sudo minikube start --driver=hyperkit -p minikube
+
+        # Download kubectl
+        sudo minikube kubectl -- get po -A
+
+        # Run Minikube Tunnel
+        sudo nohup minikube tunnel >/tmp/tunnel.out 2>/tmp/tunnel.out &
```
sehz, line 53: minikube tunnel is not mandatory for local cluster.  so make this optional?

dev-tools/install-fluvio.sh
```diff
@@ -0,0 +1,46 @@
+function success_msg {
+    printf "\n\n\n\e[1mFluvio Installation complete!\n\n\nRun the following command to update your PATH for the fluvio binary:\n\n\n\n"
+}
+
+function install_unix {
+    mkdir ~/.fluvio/
```
sehz, line 10: don't hard code path.  use default argument

Ryanmtate:
Opened this PR prematurely; Testing this branch using https://github.com/infinyon/fluvio-demo-apps-node/pull/11

Not ready to be merged, need to update some file locations once confirmed it is working

dev-tools/action-install-fluvio-cluster.sh
```diff
@@ -16,6 +16,12 @@ if [ "$DEVELOPMENT" = "true" ]; then
                 # Install Fluvio Src
                 cargo install --path ./src/cli
                 popd
+        elif [ "$VERSION" = "v0.6.0-alpha.3" ]; then 
+                pushd /tmp/
+                wget https://raw.githubusercontent.com/infinyon/fluvio/feature/alpha_install_script/dev-tools/install-fluvio.sh
```
Ryanmtate, line 21: will update this to `master` branch once https://github.com/infinyon/fluvio-demo-apps-node/pull/11 is working.

sehz (changes requested):

dev-tools/action-install-fluvio-cluster.sh
Ryanmtate: This version corresponds to the release version, how do you imagine using a ref for the download link?

sehz: there should be version as optional argument? if option is not specified then use latest

Ryanmtate: Commit 7041481 changes this to now search for a `v` release tag in the version, removing the hardcoded value;

```bash
...
elif [ "$VERSION" == *"v"* ]; then 
...
```

sehz (approved):

dev-tools/action-install-minikube.sh
Ryanmtate: Keeping this to satisfy load balancer error.

Ryanmtate:
This is now ready for review-- it adds a darwin support for fluvio github action and updates the install script to be compatible with the fluvio github action installer.

dev-tools/action-install-minikube.sh
```diff
@@ -24,26 +29,55 @@ if [[ "$OSTYPE" == "linux-gnu"* ]]; then
         # Update permissions for .kube and .minikube
         sudo chown -R $USER $HOME/.kube $HOME/.minikube
 
+        export IP=$(minikube ip)
```
sehz, line 32: should use `fluvio cluster set-minikube-context`

dev-tools/action-install-minikube.sh
```diff
@@ -24,26 +29,55 @@ if [[ "$OSTYPE" == "linux-gnu"* ]]; then
...
+
+        # Download kubectl
+        minikube kubectl -- get po -A
+
+        # Update permissions for .kube and .minikube
+        sudo chown -R $USER $HOME/.kube $HOME/.minikube
+
+        sudo kubectl config use-context minikube
+
+        # Run Minikube Tunnel
```
sehz, line 65: this works in the fluvio ci: 

```
- name: Install Minikube on Linux
        if: startsWith(matrix.os,'ubuntu')
        uses: manusa/actions-setup-minikube@v2.0.1
        with:
          minikube version: "v1.13.1"
          kubernetes version: "v1.19.2"
          github token: ${{ secrets.GITHUB_TOKEN }}
      - name: Install Minikube on Mac
        if: startsWith(matrix.os, 'macOS')
        run:  brew install minikube
      - name: Setup Minikube for Mac
        if: startsWith(matrix.os, 'macOS')
        run: |  
          minikube start --driver=virtualbox
      - name: Test minikube
        run: |
          minikube profile list
      - name: Build
        run: |
          cargo build
      - name: setup context
        run: |
          ./target/debug/fluvio cluster set-minikube-context
```

sehz (changes requested):

dev-tools/action-install-minikube.sh
Ryanmtate: Tried this already, this step is required, otherwise it will fail to find the cluster. 

This step is also run in the `action-install-fluvio-cluster.sh` file.

dev-tools/action-install-minikube.sh
Ryanmtate: Yes, but it did not work in the github composite action runner, so I needed to change to the the current.

without this step, calling the `fluvio cluster set-minikube context` in `action-install-fluvio-cluster.sh#L40` fails, as checked by the pre-install check called subsequently.

dev-tools/action-install-minikube.sh
sehz: so this is not show stopper but we should consolidate this into fluvio.   @vijaylaxmid @nicholastmosher let's create comprehensive ticket for `set-minikube-context` that should work anywhere

dev-tools/action-install-minikube.sh
sehz: hmm. we should create issue to resolve problem them

sehz (approved):

PR #405, opened 2020-10-19
Feature/alpha/rbac

Ryanmtate:
WIP: opening as a draft for early feedback and discussion.

For this initial version, I've opted for using a basic `role`, `object`, `action` based RBAC policy, where the role is either `root` or `default`.

## Basic Types

```rust
// Policy Mapping
pub type Policy = HashMap<Role, HashMap<Object, Vec<Action>>>;


// Authorization tuple to find enforcement policy;
pub type AuthRequest = (Role, Object, Action);
```

## Auth Trait

```rust

#[async_trait]
pub trait Auth {    
    async fn enforce(&self, request: AuthRequest, role: Role) -> Result<bool, AuthError>;
    async fn set_role(&mut self, role: Role) -> Result<(), AuthError>;
    fn get_role(&self) -> Option<&Role>;
}
```

## Basic Implementation of Auth Trait

```rust
#[async_trait]
impl Auth for BasicAuth {
    async fn set_role(&mut self, role: Role) -> Result<(), AuthError> {
        self.role = Some(role);
        Ok(())
    }

    fn get_role(&self) -> Option<&Role> {
        self.role.as_ref()
    }

    async fn enforce(&self, request: AuthRequest, role: Role) -> Result<bool, AuthError> {

        let is_role_match = request.0 == role;
        
        let is_allowed = self
            .policy
            .get(&request.0)
            .map(|role| {
                role
                    .get(&request.1)
                    .map(|actions| actions.iter().any(|action| action == &request.2))
                    .unwrap_or(false)
            })
            .unwrap_or(false);

        Ok(is_role_match && is_allowed)
    }
}
```

## Use in API SharedContext

```rust

if let Ok(authorized) = ctx.authorize((Role::Root, Object::SpuGroup, Action::Create)).await {
        if !authorized {
            trace!("authorization failed for role: {:?}", ctx.role());
            // ... handle auth error;
        }
    };
```

### DIFF
--- a/src/sc-schema/src/request.rs
+++ b/src/sc-schema/src/request.rs
@@ -4,30 +4,27 @@
 //! Maps SC Api Requests with their associated Responses.
 //!
 
-use std::convert::TryInto;
+use std::convert::{TryInto};
 use std::io::Error as IoError;
 
 use tracing::debug;
 
 use dataplane::bytes::Buf;
-
-use dataplane::api::ApiMessage;
+use dataplane::api::{ApiMessage};
 use dataplane::api::RequestHeader;
 use dataplane::api::RequestMessage;
 
 use dataplane::api::api_decode;
-use dataplane::derive::Encode;
+use dataplane::derive::{Encode};
 
 use super::versions::ApiVersionsRequest;
 use super::objects::*;
-
 use super::AdminPublicApiKey;
 
 #[derive(Debug, Encode)]
 pub enum AdminPublicRequest {
     // Mixed
     ApiVersionsRequest(RequestMessage<ApiVersionsRequest>),
-
     CreateRequest(RequestMessage<CreateRequest>),
     DeleteRequest(RequestMessage<DeleteRequest>),
     ListRequest(RequestMessage<ListRequest>),
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,8 @@
 [workspace]
 members = [
     "examples/00-echo",
+    "src/auth",
+    "src/auth-schema",
     "src/cli",
     "src/client",
     "src/cluster",
@@ -20,8 +22,8 @@ members = [
 ]
 
 [patch.crates-io]
-fluvio-socket = { git = 'https://github.com/infinyon/fluvio-socket', rev = "b717c1e"  }
-fluvio-service = { git = 'https://github.com/infinyon/fluvio-socket', rev = "b717c1e" }
+fluvio-socket = { git = 'https://github.com/infinyon/fluvio-socket', rev = "3975110" }
+fluvio-service = { git = 'https://github.com/infinyon/fluvio-socket', rev = "3975110" }
 
 
 # profile to make image sizer smaller
--- a/src/auth-schema/Cargo.toml
+++ b/src/auth-schema/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "fluvio-auth-schema"
+version = "0.1.0"
+authors = ["fluvio authors"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+log = "0.4.8"
+serde = { version = "1.0.0", features = ['derive'] }
+
+dataplane = { version = "0.1.0", path = "../dataplane-protocol", package = "fluvio-dataplane-protocol" }
+fluvio-protocol = { version = "0.2.0" }
+fluvio-service = { version = "0.2.0" }
--- a/src/auth-schema/src/lib.rs
+++ b/src/auth-schema/src/lib.rs
@@ -1,0 +1,11 @@
+mod request;
+
+pub use request::*;
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn it_works() {
+        assert_eq!(2 + 2, 4);
+    }
+}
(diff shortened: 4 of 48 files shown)


src/auth/src/abac.rs
```diff
@@ -0,0 +1,33 @@
+use super::{Auth, AuthRequest, Role, Object, Action};
```
sehz, line 1: don't need for now.

src/auth/src/auth.rs
```diff
@@ -0,0 +1,49 @@
+use crate::error::AuthorizationError;
+
+use async_trait::async_trait;
+use fluvio_service::IdentityContext;
+use fluvio_stream_model::core::Spec;
+use serde::{Serialize, Deserialize};
+
+
+#[derive(Eq, PartialEq, Hash, Serialize, Deserialize, Clone, Debug)]
+pub enum Role {
```
sehz, line 10: don't need hard wired role now

src/auth/src/basic.rs
```diff
@@ -0,0 +1,81 @@
...
+
+impl BasicAuth {
+    pub fn new() -> Self {
+        BasicAuth {
+            role: None,
+            policy: BasicAuth::policy(),
+        }
+    }
+
+    fn policy() -> Policy {
```
sehz, line 23: only root policy should be defined

src/auth/src/basic.rs
```diff
@@ -0,0 +1,81 @@
...
+        policy.insert(Role::Default, default_policy);
+
+        // Finally, return the policy
+        policy
+    }
+}
+
+#[async_trait]
+impl Auth for BasicAuth {
+    async fn enforce<I: IdentityContext + Sync>(&self, request: AuthRequest, identity: I) -> Result<bool, AuthError> {
```
sehz, line 64: add unit test

src/auth/src/error.rs
```diff
@@ -0,0 +1,10 @@
+use thiserror::Error;
+use std::io::Error as IoError;
+
+#[derive(Error, Debug)]
+pub enum AuthorizationError {
+    #[error("io error")]
+    IoError(#[from] IoError),
+    #[error("permission denied")]
+    PermissionDenied,
```
sehz, line 9: permission deny is not error.  it should be just return value from enforcer

src/auth/src/rbac.rs
```diff
@@ -0,0 +1,77 @@
+use super::{Authorization, AuthorizationRequest, Role, Object, Action, Policy};
```
sehz, line 1: duplicate of policy?

sehz (changes requested):

src/auth/src/auth.rs
Ryanmtate: why would we not need a role if we're going to write the logic to enforce the policy in the fluvio code? @nacardin I thought we all agreed that this was set?

src/auth/src/abac.rs
Ryanmtate: This file is going to be deleted. Only `policy.rs` will exist. rbac, basic, and abac are to be removed. Spoke about this yesterday on the call.

src/auth/src/basic.rs
Ryanmtate: this file is to be deleted. See comment earlier.

src/auth/src/rbac.rs
Ryanmtate: yep; to be deleted.

src/sc-schema/src/apis.rs
```diff
@@ -12,6 +12,7 @@ use dataplane::derive::Decode;
 #[derive(Encode, Decode, PartialEq, Debug, Clone, Copy)]
 #[repr(u16)]
 pub enum AdminPublicApiKey {
+    Auth = 8,
```
Ryanmtate, line 15: remove

src/sc-schema/src/objects/auth.rs
```diff
@@ -0,0 +1,22 @@
+#![allow(clippy::assign_op_pattern)]
```
Ryanmtate, line 1: this file can be removed.

src/sc-schema/src/request.rs
```diff
@@ -4,30 +4,29 @@
...
 use super::objects::*;
-
 use super::AdminPublicApiKey;
 
 #[derive(Debug, Encode)]
 pub enum AdminPublicRequest {
     // Mixed
     ApiVersionsRequest(RequestMessage<ApiVersionsRequest>),
-
+    AuthRequest(RequestMessage<AuthRequest>),
```
Ryanmtate, line 29: remove.

src/sc-schema/src/request.rs
```diff
@@ -56,6 +55,7 @@ impl ApiMessage for AdminPublicRequest {
             api_key
         );
         match api_key {
+            AdminPublicApiKey::Auth => api_decode!(Self, AuthRequest, src, header),
```
Ryanmtate, line 58: remove.

src/auth/src/auth.rs
nacardin: That's what I thought... If not where would the role definitions come from? Config?

src/sc-schema/src/apis.rs
Ryanmtate: put back

src/sc-schema/src/objects/auth.rs
Ryanmtate: put back

sehz:
replaced by #420

Issue #355, opened 2020-10-03
provide CLI plugin framework and move cloud sync out of fluvio


sehz:
We should provide way to extend CLI with extensible plug-in similar to how other platform does.
- helm
- Kubectl

sehz:
Other implementations such as : https://doc.rust-lang.org/book/ch14-05-extending-cargo.html

nicholastmosher:
Clap has a feature to enable "external subcommands", which basically let you capture a subcommand term that you were not expecting, rather than printing the help menu. For example, if we executed `fluvio foo --bar --baz=3`, then we can get back an ArgMatches object with the structured subcommand arguments in it. Then, we can check the PATH for an executable called `fluvio-foo` and forward the arguments to it.

sehz:
nice!

sehz:
If we decided to implement plug-in system.,  We may break apart CLI again rather than in monolithic mode.   Downside of monolithic CLI is that it incur increase of build time which saw in the last release.  It was taking order >10 min to build release version of CLI.  If we have plug-in system, we can still have a single CLI but have flexibility of mix and match features dynamically.   For example, we could have SC and SPU as plug-in and download them as needed.

ajhunyady:
This is also important for the Raspberry PI image. Cross compilation failed and there are so many dependencies, it would be a pain to track them all down.

sehz:
It should have at least following plug-in commands
- `install <new_plugin> -f <force>`
- `uninstall`

Plugin should be downloaded into some area we have permission to write.

nicholastmosher:
I think that having install/uninstall commands is something we would do if we had a public registry and package manager for fluvio plugins, which we don't right now. What are you imagining to happen if a user were to run `fluvio install fluvio-foo` or something? The only possible reasonable thing I can think of it doing would be to invoke cargo to install something, but that would require all users to have Rust installed, which I don't think we want.

I think we should restrict the scope of this to just searching for external subcommands. For reference, this is how Cargo handles external subcommands using clap:
- https://github.com/rust-lang/cargo/blob/master/src/bin/cargo/cli.rs#L216-L228
- https://github.com/rust-lang/cargo/blob/master/src/bin/cargo/main.rs#L142-L179

sehz:
We can also move cluster command as plug-in as well ( #335)
Other commands that can be move out
- SPU and SPG related commands

CLI should be strip down to bare bore minimum eventually.  It should just focus on:
- topic management
[...]

PR #406, opened 2020-10-19
Add external subcommand execution to fluvio-cli

nicholastmosher:
This uses the `external_subcommand` feature of clap/structopt in order to try to find and pass arguments to subcommands. For any subcommand `foo`, if `fluvio foo` is not a built-in subcommand, it will search the user's `PATH` for an executable named `fluivo-foo`. If it finds one, it will execute it, passing all subsequent arguments to it. In other words, if `fluvio-foo` exists and a user invokes `fluvio foo one two three -- four five`, then fluvio will execute `fluvio-foo one two three -- four five`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1012,6 +1012,7 @@ dependencies = [
  "tracing",
  "tracing-futures",
  "url",
+ "which",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


nicholastmosher:
This is the first step towards implementing #355

sehz (approved):

PR #408, opened 2020-10-20
add k8 image build on CI for mac

sehz:
Add testing of building k8 image on Mac to CI workflow.  This make sure we don't have regression in build system


sehz:
forgot docker isn't available on Mac.

Issue #412, opened 2020-10-20
Create minimal version index for tracking release versions


nicholastmosher:
We need a system which allows users to install and upgrade Fluvio reliably. Installing and upgrading are two tasks which need to be able to reliably check the latest release versions. We already have a base installer script, but it needs to be manually invoked with the version information of the release it should fetch.

To provide versioning information, we should create a structured index that tracks the published versions of any of our components. The way cargo and crates.io does this is by hosting an index in a git repository (https://github.com/rust-lang/crates.io-index/) which consists of one file per published item, where the file contains JSON describing the versions of that item which have been published.

In the immediate term, we will not be allowing arbitrary users to publish items, but that may be a use case we want to support in the future, so we may want to make our approach scalable and future-proof.

We also may want to track the versions of different _types_ of components. Right now, the only types of items we need to publish are binaries, but in the future we may want to manage the version information for, say, webassembly modules.

Additionally, (from the index standpoint), we may want to support various ways of representing a resource's path. For binaries, we could statically host them at a permalink which users can download from directly (e.g. using wget or curl). For other types of components, we may want to specify different types of hosting (e.g. if we publish to an external registry which hosts our items, such as mavencentral for a Java client).

nicholastmosher:
Closed by #428.

PR #428, opened 2020-10-28
Package index and self-updater

nicholastmosher:
This PR is the start to #412. Here's what is built/working so far:

- The new `fluvio-index` crate defines the data structure for the package index, as well as the http request and response handling for downloading package manifests and package files/checksums.
- The new `fluvio update` subcommand will use the package index to find the latest version of the fluvio package, named `fluvio/fluvio` (group fluvio, package fluvio). It will then download the file of the latest version, verify the checksum, and place it into `~/.fluvio/bin` and set executable permissions on it.

Here are a few things that are not done but which I think are important to include:

- The `fluvio update` subcommand does not yet handle index schema updates gracefully. The index has metadata which indicates a minimum version of the `fluvio-index` crate which is required to interact with it. The CLI should check whether it is recent enough and prompt the user to self-update before continuing with any command that depends on the index.
- The logic to install a plugin is not written yet, though it will be similar to the update code, so it's not far behind.
- There is an unhandled edge case involving platform targets. If I have `fluvio/fluvio:0.1.0` for `x86_64-apple-darwin` and I push an update `fluvio/fluvio:0.2.0` for a different target, e.g. `x86_64-unknown-linux-musl`, then the updater will fail, complaining about missing the `x86_64-apple-darwin` target for version `0.2.0`. This should be an easy fix, but needs doing.

I also have some additional thoughts regarding the design now that I have spent time in the implementation phase. They are:

- The self-update functionality does not actually require a top-level index in order to function, since the qualified name of a package (e.g. `<group>/<name>` such as `fluvio/fluvio`) maps directly to the location of that individual packages's metadata. I wonder whether we don't need the top-level index at all. One use-case where I can see it being helpful is if you want to update all of your plugins and you need to search for the package location in the registry.


src/cli/Cargo.toml
```diff
@@ -41,10 +41,15 @@ async-trait = "0.1.21"
 ctrlc = "3.1.3"
 futures-lite = { version = "1.7.0" }
 dirs = "1.0.2"
+async-h1 = "2.1.2"
+async-tls = "0.10.0"
```
sehz, line 45: use `fluvio-async-tls` which is our fork to added certificates API

sehz (changes requested):
Looks good.  just one minor changes required

src/cli/Cargo.toml
nicholastmosher: You know, I just realized that I'm not even using `async-tls`. I was originally going to, but then I copied the sendgrid client code which is using TLS via `fluvio-future`. Let me know if that works or if I need to change something. I'm going to remove this dependency at least

sehz: That works

src/cli/src/http.rs
```diff
@@ -0,0 +1,35 @@
...
+pub async fn execute(req: Request) -> Result<Response, Error> {
+    debug!("executing request: {:#?}", req);
+
+    let host = req
+        .url()
+        .host_str()
+        .ok_or_else(|| Error::from_str(StatusCode::BadRequest, "missing hostname"))?
+        .to_string();
+
+    let is_https = req.url().scheme() == "https";
```
sehz, line 14: do we need this? we  should be using https always

src/cli/src/http.rs
```diff
@@ -0,0 +1,35 @@
...
+        .url()
+        .socket_addrs(|| if is_https { Some(443) } else { Some(80) })?
+        .into_iter()
+        .next()
+        .ok_or_else(|| Error::from_str(StatusCode::BadRequest, "missing valid address"))?;
+
+    let tcp_stream = fluvio_future::net::TcpStream::connect(addr).await?;
+
+    let result = if is_https {
+        let tls_connector = fluvio_future::tls::TlsConnector::default();
```
sehz, line 26: observe feature flag: `rust_tls` and `native_tls`.  `tls` should be under `rust_tls`.  For `native_tls`, use `fluvio_future::native_tls::TlsConnector`

src/cli/src/install/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+
+    if !verify_checksum(&package_file, &package_checksum) {
+        return Err(fluvio_index::Error::ChecksumError.into());
+    }
+    debug!(hex = %package_checksum, "Verified checksum");
+    Ok(package_file)
+}
+
+fn verify_checksum<B: AsRef<[u8]>>(buffer: B, checksum: &str) -> bool {
+    let bytes = buffer.as_ref();
```
sehz, line 68: maybe use impl?
`fn verify_checksum(buffer: Impl<AsRef<[u8]>L, checksum: &str)

src/cli/src/install/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+    // Create bin_dir if it does not exist
+    let bin_dir = bin_dir.as_ref();
+    std::fs::create_dir_all(&bin_dir)?;
+
+    // Install our package to `<bin_dir>/<name>`
+    let install_path = bin_dir.join(name);
+    std::fs::write(&install_path, bytes)?;
+
+    // Mark our package as executable
+    let status = std::process::Command::new("chmod")
```
sehz, line 93: use rust std:  https://doc.rust-lang.org/std/fs/fn.set_permissions.html

sehz (changes requested):
Good jobs.  couple of minor nits

src/cli/src/install/mod.rs
nicholastmosher: I believe the code generation creates the same output for either version of the code. As a style preference, I tend to dislike using `impl` in argument position. It actually (technically) gives you less control over the type inference. When you use a generic type parameter `<B>` you have the option to pass an explicit type argument when you call the function using the turbofish, e.g. `verify_checksum::<Vec<u8>>(thing)`. That doesn't really matter here but that's why I prefer the style. That being said, if you want me to change it I will

src/cli/src/http.rs
nicholastmosher: I will start looking at this. This raises a question to me though, I know there's been a lot of work on fluvio_future lately, are there any notes on how to properly set this type of thing up? I don't really know what's changed so as I'm looking at this I'm needing to dig through source and changes and feature flags to figure it out.

nicholastmosher: I ended up just creating a function to return a TlsConnector of either implementation based on feature flag. I feel like that is the type of concern which might be better to push down into fluvio-future itself. We can have a discussion on that later perhaps

src/cli/src/install/mod.rs
nicholastmosher: Alright, rewrote it using https://doc.rust-lang.org/std/os/unix/fs/trait.PermissionsExt.html
It's weird, I had specifically gone looking for something like that and it was only now that I actually found it 💁‍♂️

src/cli/src/install/mod.rs
sehz: I tend to like `impl` because it seems to more compact.  However, this maybe more personal style so no strong position here.

src/cli/src/http.rs
sehz: Sure.  Changes were minimized to make sure existing code were not impacted.   We can add further improvements to `fluvio-future` and make it easier to consume it

sehz (approved):

sehz:
rebase.  CI is flaky now, so you may need to re-run CI couple of times to make it pass

PR #438, opened 2020-11-01
Add authorizationConfigMap option

nacardin:
Add authorizationConfigMap to helm chart and installer.
Upload SC log as artifact during CI tests.


Makefile
```diff
@@ -37,9 +36,15 @@ smoke-test-tls:	test-clean-up
 
 # test rbac with ROOT user
 smoke-test-tls-root:	test-clean-up
-	AUTH_POLICY=$(AUTH_POLICY) X509_AUTH_SCOPES=$(AUTH_SCOPE)  \
+	kubectl create configmap authorization --from-file=${AUTH_CONFIG}
```
sehz, line 39: Not appropriate to use config map here since this is local SC not K8 Sc so it doesn't know anything about pod

nacardin:
Tests pass intermittently locally

sehz (approved):

PR #461, opened 2020-11-11
Index target

nicholastmosher:
This causes `Target::from_str` to interpret `"x86_64-unknown-linux-gnu"` as `Target::X86_64UnknownLinuxMusl`, and replaces `fluvio_index::PACKAGE_TARGET` constant with `fluvio_index::package_target()` function to return a folded and structured version of the current build's target.

### DIFF
--- a/src/package-index/src/error.rs
+++ b/src/package-index/src/error.rs
@@ -21,7 +21,7 @@ pub enum Error {
     ReleaseAlreadyExists(semver::Version, Target),
     #[error("Failed to parse URL")]
     UrlParseError(#[from] url::ParseError),
-    #[error("Invalid platform {0}")]
+    #[error("Invalid target {0}")]
     InvalidPlatform(String),
     #[error(transparent)]
     HttpError(#[from] HttpError),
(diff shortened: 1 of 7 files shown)


src/package-index/src/error.rs
```diff
@@ -22,7 +22,7 @@ pub enum Error {
     #[error("Failed to parse URL")]
     UrlParseError(#[from] url::ParseError),
     #[error("Invalid platform {0}")]
-    InvalidPlatform(String),
+    InvalidTarget(String),
```
sehz, line 25: will this requires bump up to `0.2.0`?

sehz (changes requested):
question about semversion

src/package-index/src/error.rs
nicholastmosher: Hmm, possibly, yeah. I can revert that if we don't want to deal with that right now

sehz: Preferable to minimize version changes unless it is necessary

sehz (approved):

PR #466, opened 2020-11-11
reduce timeout for searching replica

sehz:
reduce time for looking up replica to 300 ms
return proper error code instead of timeout


sehz:
all test passed in the local machine

PR #467, opened 2020-11-12
if there is topic creation error return proper error

sehz:
fix where error was not returned properly for topic create


sehz:
use #468, this branch is obsolete

PR #468, opened 2020-11-12
if there is topic creation error return proper error

sehz:
fix proper error when trying to create duplicate topic


sehz:
all tests passed in local machine

PR #470, opened 2020-11-12
more changes to make it CI stable

sehz:
- force to resync topic controller every 60 seconds
- better debug message during consumer validation
- wait until test running shutdown to allow logs to catch up
- set client logging level to warn
- remove macOS from CI local cluster test


sehz:
local and k8 TLS tests passes in the local machine

PR #475, opened 2020-11-13
Package id types

nicholastmosher:
The `PackageId` type from `fluvio-package-index` was somewhat finicky in that it always carried the package version as an `Option<Version>`, because it could be written in a string like either:

```
<group>/<name>:<version>
OR
<group>/<name>
```

This made some code surrounding PackageId somewhat unweildy because we always had to unwrap the option whenever we wanted to use the Version, even in a context where we knew that we had already checked that we had the version available.

---

This PR changes that type to have two different typed variants, `PackageId<WithVersion>` and `PackageId<MaybeVersion>`. The `PackageId<MaybeVersion>` behaves essentially the same way that the old `PackageId` did. However, the new `PackageId<WithVersion>` type guarantees that it will always have a Version embedded inside it, so when you have a `PackageId<WithVersion>`, you can just use the Version without needing to match or unwrap it.

Another advantage of having distinct types is that they have different `FromStr` implementations. For example, parsing a string into a `PackageId<WithVersion>` will add the requirement that the string must contain the `:<version>` suffix. On the other hand, parsing into `PackageId<MaybeVersion>` will allow a string with or without the `:<version>`.

```rust
let id_with_version: PackageId<WithVersion> = "fluvio/fluvio:1.0.0".parse()?;
let id_without_version: PackageId<MaybeVersion> = "fluvio/fluvio".parse()?;
let id_without_version: PackageId<MaybeVersion> = "fluvio/fluvio:1.0.0".parse()?;
```

You can upgrade a `PackageId<MaybeVersion>` into a `PackageId<WithVersion>` by providing a version to use.

```rust
let id: PackageId<MaybeVersion> = "fluvio/fluvio".parse()?;
let id_with_version: PackageId<WithVersion> = id.into_versioned(Version::parse("1.0.0").unwrap());
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1130,7 +1130,7 @@ dependencies = [
 
 [[package]]
 name = "fluvio-package-index"
-version = "0.1.1"
+version = "0.2.0"
 dependencies = [
  "http-types",
  "lazy_static",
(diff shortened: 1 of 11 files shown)


sehz (approved):

sehz:
all tests passes except k8 which seems to be stuck

PR #477, opened 2020-11-16
Refactor CLI

nicholastmosher:
This branch standardizes some of the patterns that I've seen throughout the CLI codebase, which (in my opinion) makes things a bit cleaner and actually slims some things down. Some changes of note:

- All StructOpt **structs** (i.e. argument definitions) are named in the pattern `ThingOpt`
- All StructOpt **enums** (i.e. subcommand definitions) are named in the pattern `ThingCmd`
- For subcommand processing, we used to have a pattern like `ThingCmd` and a corresponding _global function_ `process_thing(opt: ThingCmd, ...)` function. Now, we have `ThingCmd` and an _associated function_ `ThingCmd::process(self, ...)`. This helps to keep related code close together, and also allows for nice patterns when branching between subcommands, such as

```
impl ThingCmd {
    fn process(self) {
        match self {
            Self::first_subcommand(cmd) => cmd.process(),
            ...
        }
    }
}
```

- We used to have a `ClusterTarget` option that was included as a "flattened" set of options on every command that required it. Now, we have a global command `RootOpt` that contains the `ClusterTarget` args, removing a bunch of duplicate code from the codebase.
- I split up a bunch of places where there were many subcommands in a single module. Now, every subcommand has its own module file, which I think makes things much clearer and easier to find in the project directory.


src/cli/src/cluster/minikube.rs
```diff
@@ -0,0 +1,22 @@
+use structopt::StructOpt;
+use k8_config::context::MinikubeContext;
+use crate::Result;
+
+#[derive(Debug, StructOpt)]
+pub struct SetMinikubeContext {
```
sehz, line 6: MinikubeContext is goner...

src/cli/src/cluster/mod.rs
```diff
@@ -1,63 +1,54 @@
...
 
-#[allow(clippy::large_enum_variant)]
 #[derive(Debug, StructOpt)]
 #[structopt(about = "Available Commands")]
-pub enum ClusterCommands {
+pub enum ClusterCmd {
     /// Install a Fluvio cluster, locally or on Minikube
     #[structopt(name = "install")]
-    Install(InstallCommand),
+    Install(Box<InstallOpt>),
```
sehz, line 20: Why Box?

src/cli/src/root_cli.rs
```diff
@@ -150,47 +168,77 @@ enum Root {
...
-            Root::External(args) => process_external_subcommand(args)?,
-        };
-        Ok(output)
-    })
-}
+impl RootCmd {
+    pub async fn process(self, root: RootOpt) -> Result<()> {
+        let out = Arc::new(PrintTerminal::new());
+        let fluvio_config = root.target.load()?;
+        let fluvio = Fluvio::connect_with_config(&fluvio_config).await?;
```
sehz, line 188: Cluster doesn't require connect to fluvio

sehz (changes requested):
Almost everything seems to be fine except dealing with global setting related to Cluster.

src/cli/src/cluster/mod.rs
nicholastmosher: Clippy gives a warning when one enum variant is much larger than the others. It appears `InstallCommand` is much larger. Clippy recommends boxing it, and structopt seems to work with it just fine.

src/cli/src/root_cli.rs
nicholastmosher: Mmm, yep. This is something that I had fixed in my `zsh` branch, because I noticed that some commands don't require connecting to Fluvio at all but they were being blocked on establishing a valid connection. For example I couldn't generate completions until connecting to a cluster. I have a fix for this, let me bring it in here.

src/cli/src/cluster/minikube.rs
nicholastmosher: Forgot to delete

sehz (approved):
Looks good

nicholastmosher:
You know, I realized the ZSH fix really boiled down to a very small change so I just included it in this PR.

nicholastmosher:
One last change, the ZSH completions are running into the same weird bug as these guys: https://github.com/volta-cli/volta/pull/772
Just need to remove all `:` from the structopt comments...

sehz (approved):

nicholastmosher:
Ok the ZSH stuff still has problems. I took it back out of this PR 😂
The rest of this should be good to go

nicholastmosher:
@sehz this CI failure is part of the preexisting problems, right? I don't think I caused any of those errors but I want to make sure.

sehz:
Hmm.  didn't see that failure before, can you run both local and k8 tests locally to sure?

nicholastmosher:
Here's my output from running `fluvio cluster check --pre-install` followed by `make smoke-test-tls`

```
❯ cargo run --bin fluvio -- cluster check --pre-install
    Finished dev [unoptimized + debuginfo] target(s) in 0.27s
     Running `target/debug/fluvio cluster check --pre-install`

Running pre-install checks....

✔️  ok: Supported kubernetes version is installed
✔️  ok: Fluvio system charts are installed
✔️  ok: Can create service
✔️  ok: Can create customresourcedefinitions
✔️  ok: Can create secret
✔️  ok: Load balancer is up

All checks passed!

❯ make smoke-test-tls
./target/debug/fluvio cluster uninstall
./target/debug/fluvio cluster uninstall --local
FLV_CMD=true ./target/debug/flv-test --spu 1 --produce-iteration 5 --tls --local --client-log warn --server-log debug
using local environment driver
>> /Users/nick/infinyon/fluvio/./target/debug/fluvio cluster uninstall --local
>> helm version
installing cluster
>> /Users/nick/infinyon/fluvio/./target/debug/fluvio cluster install --spu 1 --local --rust-log debug --tls --domain fluvio.local --ca-cert /Users/nick/infinyon/fluvio/tls/certs/ca.crt --client-cert /Users/nick/infinyon/fluvio/tls/certs/client-root.crt --client-key /Users/nick/infinyon/fluvio/tls/certs/client-root.key --server-key /Users/nick/infinyon/fluvio/tls/certs/server.key --server-cert /Users/nick/infinyon/fluvio/tls/certs/server.crt
>> sync
>> /Users/nick/infinyon/fluvio/./target/debug/fluvio run sc --tls --enable-client-cert --server-cert /Users/nick/infinyon/fluvio/tls/certs/server.crt --server-key /Users/nick/infinyon/fluvio/tls/certs/server.key --ca-cert /Users/nick/infinyon/fluvio/tls/certs/ca.crt --bind-non-tls-public 0.0.0.0:9005
>> /Users/nick/infinyon/fluvio/./target/debug/fluvio run spu --tls --enable-client-cert --server-cert /Users/nick/infinyon/fluvio/tls/certs/server.crt --server-key /Users/nick/infinyon/fluvio/tls/certs/server.key --ca-cert /Users/nick/infinyon/fluvio/tls/certs/ca.crt --bind-non-tls-public 0.0.0.0:9012 -i 5001 -p 0.0.0.0:9010 -v 0.0.0.0:9011
The application panicked (crashed).
Message:  sc : FlvSocketError { source: IoError { source: Os { code: 61, kind: ConnectionRefused, message: "Connection refused" } } }
Location: /Users/nick/infinyon/fluvio/src/cli/src/cluster/install/mod.rs:190

Backtrace omitted.
Exited with status code: 101
thread 'main' panicked at 'internal error: entered unreachable code', tests/runner/src/util/cmd.rs:90:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
make: *** [smoke-test-tls] Error 101
```

sehz:
That means some command try to create connections when there is no fluvio

sehz (approved):
Great!

sehz:
rebase

sehz:
all local test passes per @nicholastmosher

PR #478, opened 2020-11-16
various cleanup

sehz:
- add more structure logging if few places
- update to stream dispatcher to use updated timer


sehz:
K8 test works locally.

PR #482, opened 2020-11-17
Fix composite action and and add to CI

simlay:
The goal of this PR is to unify the installation and creation of helm, minikube, etc. Looks like most of it was already done, just needs to be fleshed out.


.github/workflows/ci.yml
```diff
@@ -158,42 +137,25 @@ jobs:
...
     steps:
       - uses: actions/checkout@v2
-      - name: Install Helm
-        run: dev-tools/ci-replace-helm.sh
-        env:
-          HELM_VERSION: v3.3.4
-          OS: ${{ matrix.os }}
-      - run: helm version
+      - name: Install Fluvio Local Cluster
+        uses: ./
```
simlay, line 151: This seems to be the way to use a [composite action in it's own repo](https://github.com/dmesquita/github-actions-tutorial/blob/1d533f76934cde714f780a83a0092e88b8e087cd/.github/workflows/use-action-same-repo.yml#L10).

dev-tools/action-install-fluvio-cluster.sh
```diff
@@ -1,27 +1,35 @@
...
 # GitHub Action Workflows.
+
+set -eu -o pipefail
 echo "Installing Fluvio Local Cluster"
 if [ "$DEVELOPMENT" = "true" ]; then
         case $VERSION in
                 "latest")
                         echo "Installing from latest source"
-                        curl -sSf https://raw.githubusercontent.com.infinyon/fluvio/master/dev-tools/install.sh | bash
+                        unset VERSION
```
simlay, line 11: This is bad and I will clean this up.

dev-tools/action-install-fluvio-cluster.sh
```diff
@@ -1,27 +1,38 @@
...
-                ;;
-                *)
-                        echo "Release binary is not available for $VERSION; use 'latest'"
-                ;;
-        esac
+        echo "Installing from latest source"
+        curl -sSf https://raw.githubusercontent.com/infinyon/fluvio/master/install.sh | bash
+        echo 'export PATH="$HOME/.fluvio/bin:$PATH"' >> $HOME/.bash_profile
+        . $HOME/.bash_profile
+        # case $VERSION in
```
simlay, line 12: fix this.

sehz:
can you rebase from master?

dev-tools/action-install-fluvio-cluster.sh
```diff
@@ -1,27 +1,15 @@
...
-                        echo "Installing Fluvio $VERSION"
-                        curl -sSf https://raw.githubusercontent.com.infinyon/fluvio/master/dev-tools/install.sh | bash
-                        echo 'export PATH="$HOME/.fluvio/bin:$PATH"' >> $HOME/.bash_profile
-                        . $HOME/.bash_profile
-                ;;
-                *)
-                        echo "Release binary is not available for $VERSION; use 'latest'"
-                ;;
-        esac
+        echo "Installing from latest source"
```
sehz, line 8: Can you change this to `Installing Fluvio CLI from latest source`

dev-tools/action-install-fluvio-cluster.sh
```diff
@@ -1,27 +1,15 @@
 #!/bin/bash
 # This script is ran by the github actions to install fluvio in
 # GitHub Action Workflows.
+
+set -eu -o pipefail
 echo "Installing Fluvio Local Cluster"
 if [ "$DEVELOPMENT" = "true" ]; then
```
sehz, line 7: Should get rid of development since it is not built from source code

dev-tools/action-install-fluvio-cluster.sh
```diff
@@ -1,27 +1,15 @@
...
-                ;;
-        esac
+        echo "Installing from latest source"
+        curl -sSf https://raw.githubusercontent.com/infinyon/fluvio/master/install.sh | bash
+        echo 'export PATH="$HOME/.fluvio/bin:$PATH"' >> $HOME/.bash_profile
+        . $HOME/.bash_profile
 
-        
         # Install Fluvio System Charts
         fluvio cluster start --sys
```
sehz, line 14: should be changed to: `fluvio cluster start --setup --local`

dev-tools/action-install-fluvio-cluster.sh
```diff
@@ -1,27 +1,15 @@
 #!/bin/bash
```
sehz, line 1: We should move action related files into it's own directory.  maybe `actions`?

sehz (changes requested):

dev-tools/action-install-fluvio-cluster.sh
simlay: Hmm. `actions` works. I was also thinking about doing `.gh-actions/` but I don't love that either.

sehz (approved):

sehz (changes requested):
they should be kept as is. only actions relation script should move

sehz (approved):

PR #486, opened 2020-11-18
update to use no delay to test TLS

sehz:


sehz:
setting no delay has no effect on test. closing this for now

PR #489, opened 2020-11-19
Add formatting for user errors and some initial cases

nicholastmosher:
After a lot of searching, I found that `color_eyre` has configuration options to change how things are printed out. Previously, it would add footers to our error messages that were noisy and ugly that looked like this:

```
$ fluvio cluster install
...
```

Now it will not print any backtraces, spantraces (from tracing), or ENV notices for RUST_BACKTRACE, it will only show the errors.

---

Originally, I had been thinking that the error stack was not nice to have, but I think that if we take care to word the error messages nicely it can work out really well. I discovered that eyre/color_eyre introduces a `Section` extension trait for Error and Result types https://docs.rs/color-eyre/0.5.7/color_eyre/section/trait.Section.html which allow you to annotate the error message that is printed.

The approach I ended up taking was to create an `into_report` method on `CliError` which converts it into an `eyre::Report`, the structure which contains the annotations on the errors. So the flow for creating user-facing error messages is this:

- Pattern-match on the `CliError` for the specific error case you want to annotate
- Convert that error instance to a Report using `Report::from(my_cli_error)`
- Annotate the Report object using the `Section` extension methods, such as `.suggestion("do this")`

Using this setup, we can let all errors be printed uniformly (i.e. using their std::error::Error stack printouts) with minimal noise, but add help notices and suggestions to specific cases where it is useful. An example of a printout for an annotated error looks like this:

```
❯ fluvio cluster install
Performing pre-flight checks
Error:
   0: Fluvio cluster error
   1: Failed to install Fluvio on Kubernetes
   2: Fluvio pre-installation check failed
   3: Helm client error
   4: Failed to connect to Kubernetes

Suggestion: Make sure you have run 'minikube start --driver=hyperkit'
```

And colorized, it looks like this:

image


src/cli/Cargo.toml
```diff
@@ -47,7 +47,7 @@ async-h1 = "2.1.2"
 http-types = "2.4.0"
 thiserror = "1.0.20"
 eyre = "0.6.1"
-color-eyre = "0.5.5"
+color-eyre = { version = "0.5.5", default-features = false }
 which = "4.0.2"
 sha2 = "0.9.1"
```
sehz, line 52: bump up version for cli?

src/cli/src/error.rs
```diff
@@ -65,15 +41,47 @@ pub struct HttpError {
...
     }
+
+    pub fn into_report(self) -> color_eyre::Report {
+        use color_eyre::{Report, Section};
+        use CliError::*;
+        use TheClusterError::*;
+        use TheCheckError::*;
+
+        match self {
+            e
```
sehz, line 61: typos?

sehz (changes requested):
Looks good, minor nits

sehz:
can you also run: `make smoke-test-k8-tls-policy`

src/cli/src/error.rs
nicholastmosher: Nope, that's an unfortunate side-effect of `cargo fmt` in this case because the match expression is so long. It originally looked like

```rust
match self {
    e @ ClusterError(...) => {...}
}
```

Which is a way of preserving the original value instead of picking out a destructured subpiece. Essentially `e` is equivalent to `self` in this case. So actually, thinking about it, we could probably do without the `e @`.

But it is used right below in `let report = Report::from(e)`. If we wanted to get rid of `e @`, we would replace that with `let report = Report::from(self)`

sehz: can you change from `e` to something else.  That way, it reduce confusion and increase readability

sehz (changes requested):

nicholastmosher:
Since the `fluvio-helm` crate was created after this PR was opened, there are some changes I had made to the helm client and its errors that need updating before this PR should merge. Those are over at https://github.com/infinyon/fluvio-helm/pull/3.

src/cluster/src/error.rs
```diff
@@ -8,50 +7,125 @@ use crate::helm::HelmError;
...
+    #[error("Failed to install Fluvio locally")]
+    InstallLocal(#[from] LocalInstallError),
+    /// An error occurred while trying to uninstall Fluvio
+    #[error("Failed to uninstall Fluvio")]
+    Uninstall(#[from] UninstallError),
+}
+
+/// Errors that may occur while trying to install Fluvio on Kubernetes
+#[derive(thiserror::Error, Debug)]
+pub enum K8InstallError {
```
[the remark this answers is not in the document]
nicholastmosher: What do you mean? All of the variants inside of `K8InstallError` are the same ones we had before. All that has happened here is that I've created one more layer of context in the stack of errors. Now, a `ClusterError` will communicate what type of action was being attempted when an error occurred. Those types of errors include 

- "K8 Install" errors (we were trying to install Fluvio onto Kubernetes)
- "Local Install" errors (we were trying to install Fluvio locally)
- "Uninstall" errors (we were trying to uninstall Fluvio)
  - I am even thinking about splitting this into "K8Uninstall" and "LocalUninstall" but it seemed not urgent to do so.

nicholastmosher:
A few notes about recent changes:

1) I moved some of the user-facing error conversions down into the fluvio-cluster crate.
2) I set up fluvio-cluster with its own binary target so that it may compile into a `fluvio-cluster` executable and be used as an extension.

The user-facing error messages look just about the same, although since we are converting them into pretty "Reports" early, we do happen to lose some of the higher-level error details. For example, when minikube is down, `fluvio cluster install` now prints the following:

```
❯ fluvio cluster install
Performing pre-flight checks
Error:
   0: Failed to install Fluvio on Kubernetes
   1: Fluvio pre-installation check failed
   2: Helm client error
   3: Failed to connect to Kubernetes

Suggestion: Make sure you have run 'minikube start --driver=hyperkit'
```

image

Notice that this is a slightly different output than what we had before, which was

```
❯ fluvio cluster install
Performing pre-flight checks
Error:
   0: Fluvio cluster error
   1: Failed to install Fluvio on Kubernetes
   2: Fluvio pre-installation check failed
   3: Helm client error
   4: Failed to connect to Kubernetes

Suggestion: Make sure you have run 'minikube start --driver=hyperkit'
```

image

In this specific case, it doesn't cause too much impact, notice that the only line we lost was `0: Fluvio cluster error`. In this case it's fine, since this error will eventually be printed directly by the `fluvio-cluster` executable itself, so the fact that it is a cluster error will be implied. However we should keep an eye on this for the future, if we use `fluvio-cluster` functionality as a component in another project, this kind of error reporting may cut off valuable high-level details such as what was happening before executing the `fluvio-cluster` functionality.

The above is not a blocking concern to merging this PR, just something to keep in mind going forward.

sehz (changes requested):

src/cli/Cargo.toml
```diff
@@ -46,7 +46,12 @@ k8-config = { version = "1.3.0" }
 k8-client = { version = "3.0.0", default-features = false }
 fluvio-future = { version = "0.1.8", features = ["fs", "io", "subscriber"] }
 fluvio = { version = "0.2.3", path = "../client", default-features = false }
+fluvio-sc = { version = "0.3.1", path = "../sc", optional = true, default-features = false }
```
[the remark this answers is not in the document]
nicholastmosher: Ah, that must have come from when I was rebasing and a lot of things needed to be resolved. I must have left this in unintentionally, I'll get rid of it.

src/cli/src/error.rs
```diff
@@ -1,84 +1,39 @@
...
-    },
+    IoError(#[from] IoError),
     #[error(transparent)]
-    OutputError {
-        #[from]
-        source: OutputError,
-    },
+    OutputError(#[from] OutputError),
+    #[error("Fluvio cluster error")]
+    ClusterCliError(#[from] ClusterCliError),
```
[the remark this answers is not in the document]
nicholastmosher: Removing these lines causes the project to no longer compile. This is because even though we want to split out fluivo-cluster and fluvio-extension-consumer into independent executables later, they are still directly linked as of now, and therefore we need the `From` conversions from these respective error types. I agree that this needs to eventually happen, but I didn't actually add these error variants in this PR so I believe it's unrelated and should be handled in a future issue.

src/cluster/src/check.rs
```diff
@@ -30,6 +31,51 @@ const RESOURCE_SERVICE_ACCOUNT: &str = "secret";
 /// The type of error that can occur while running preinstall checks
 #[derive(Error, Debug)]
 pub enum CheckError {
+    /// There was a problem with the helm client during pre-check
+    #[error("Helm client error")]
+    HelmError(#[from] HelmError),
+
+    /// There was a problem fetching kubernetes configuration
```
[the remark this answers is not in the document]
nicholastmosher: Kubernetes errors already have their own type, it's the `K8ConfigError` you see here. The reason it is part of the `CheckError` variants is because it may be an underlying cause of a check failure, so it gets bundled up so that it may be inspected and debugged later.

I'm not sure what the call-to-action is here, I didn't actually add this error type in this PR, I only reformatted it. If there is something we would like to do about this, I think it would be good to make a new issue.

src/cli/src/error.rs
```diff
@@ -90,15 +54,43 @@ pub struct HttpError {
...
     }
+
+    pub fn into_report(self) -> color_eyre::Report {
+        use color_eyre::{Report, Section};
+        use CliError::*;
+        use TheClusterError::*;
+        use TheCheckError::*;
+
+        match &self {
+            ClusterError(InstallK8(K8InstallError::PreCheck(HelmError(
```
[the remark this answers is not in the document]
nicholastmosher: Moved these down into the `fluvio-cluster` crate where the errors originate.

src/cluster/Cargo.toml
```diff
@@ -1,29 +1,32 @@
...
 name = "fluvio_cluster"
 path = "src/lib.rs"
 
+[[bin]]
+name = "fluvio-cluster"
+path = "src/main.rs"
+
 [features]
-default = ["native_tls"]
+default = ["cli", "native_tls"]
```
sehz, line 19: cli is only needed for binary.  you can add `required-features` to [bin](https://stackoverflow.com/questions/35711044/how-can-i-specify-binary-only-dependencies)

src/cluster/Cargo.toml
```diff
@@ -34,21 +37,22 @@ serde = "1.0.115"
...
 async-trait = "0.1.21"
+color-eyre = { version = "0.5.8", default-features = false, optional = true }
 structopt = { version = "0.3.16", default-features = false, optional = true }
 prettytable-rs = { version = "0.8.0", optional = true }
 
 # Fluvio dependencies
-fluvio-helm = { version = "0.1.0", git = "https://github.com/infinyon/fluvio-helm", branch = "main" }
 fluvio = { version = "0.2.3", path = "../client", default-features = false }
-fluvio-controlplane-metadata = { version = "0.2.0", path = "../controlplane-metadata", features = ["k8"] }
+fluvio-helm = { git = "https://github.com/infinyon/fluvio-helm", rev = "6e15fd2c789c33ba8bae51a369a0983e177c2f78" }
```
sehz, line 46: since helm changes are merged into main, can remove `rev`

src/cluster/src/cli/error.rs
```diff
@@ -0,0 +1,68 @@
...
+impl ClusterError {
+    pub fn into_report(self) -> color_eyre::Report {
+        use color_eyre::{Report, Section};
+        use fluvio_helm::HelmError as Helm;
+        use CheckError as Check;
+        use crate::K8InstallError::*;
+
+        match &self {
+            Self::InstallK8(PreCheck(Check::HelmError(Helm::FailedToConnect))) => {
+                let report = Report::from(self);
```
sehz, line 51: Isn't this already same suggestion done in Precheck? seems like this is duplicate logic

sehz: I suggest add `Suggestion` similar to `Display` of `Debug` trait.   This allow more flexibility and suggestion as well as push logic to near the error.  For example:

```
pub trait Suggestion {

       fn format(&self,  report: Report);
}
```

then you can add them to Helm error
```
impl Suggestion for HelmError {

     .....
}
```

src/cluster/src/main.rs
```diff
@@ -0,0 +1,12 @@
+use structopt::StructOpt;
+use fluvio_cluster::cli::ClusterOpt;
+use fluvio_future::task::run_block_on;
+
+fn main() -> color_eyre::Result<()> {
+    color_eyre::config::HookBuilder::blank()
+        .display_env_section(false)
+        .install()?;
+    let root: ClusterOpt = ClusterOpt::from_args();
```
sehz, line 9: Without having suggestion trait, we need to duplicate logic for suggestion.   Should create common util to process error in the common extension

sehz (changes requested):

src/cli/src/lib.rs
```diff
@@ -1,11 +1,329 @@
...
+                process_external_subcommand(args)?;
+            }
+        }
+
+        Ok(())
+    }
+}
+
+/// All top-level subcommands that require a Fluvio client
+#[derive(StructOpt, Debug)]
```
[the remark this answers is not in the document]
nicholastmosher: What do you mean? I don't believe this had ever been moved to consumer extension... If you look further down in the diff for this PR, you can see that all I did was move this block of code from `root_cli.rs` into `lib.rs`.

https://github.com/infinyon/fluvio/pull/489/files/fe5784edf0420575bc0cb9fe6f799cbbcbe82f81#diff-c5fa9bd113fd871d87ac8391cfbfa409089625f65d07de2a2de55643d3f41946L147-L190

If we want to remove this from here and consolidate it into fluvio-extension-consumer, can we do that in a future issue?

src/cluster/Cargo.toml
nicholastmosher: I could see it either way here, but I've heard that it's nice to use `rev` when using git dependencies because it will give you a more reproducible build. Using `branch = main` is kind of like using cargo where you specify `version = "*"` because any change upstream will automatically be pulled in. If somebody makes breaking changes on the main branch, this would immediately fail the build here, but pinning a `rev` version would require us to opt into that change and fix things that break. If you still want me to change it I will, that was just my reason for doing it this way.

src/cli/src/lib.rs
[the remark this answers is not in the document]
sehz: Got it.  I thought this was already moved.  Sure we can move in another PR

src/cluster/Cargo.toml
sehz: how if I publish helm repo as is?

nicholastmosher: I think that would be fine, I don't know of any other changes that would need to go there

src/cluster/src/cli/error.rs
nicholastmosher: I don't think this is duplicate logic. In `fluvio-helm`, we have this from errors.rs:

```rust
pub enum HelmError {
...
    #[error("Failed to connect to Kubernetes")]
    FailedToConnect,
...
}
```

and here we are adding a suggestion of how to possibly remedy it.

I don't really understand this `Suggestion` trait you are recommending. The current suggestions are already extension methods on the `eyre::Report` type via the `color_eyre::Section` trait. I actually think that the way things are laid out here works out pretty well. Here are the considerations that were in my head when I wrote it this way:

- Error types belonging to _library crates_ (e.g. HelmError and CheckError) should be strictly data structures with Display/Debug implementations. Some of these error data structures are nested, i.e. they hold another error inside them, which indicates that the present-level error was "caused by" the inner error. This nesting of error types creates a context that describes a relationship between the high-level errors which were caused by subsequent low-level errors.
- Different _application crates_ can define more specialized ways of rendering errors received from libraries. This is the role of the `eyre`/`color_eyre` crate and of the `Report` type from them. A Report simply wraps a library's error data type and uses the information in the Display/Debug implementations to produce a nicer, CLI-specialized rendering of those errors. It also allows you to add special "sections" to the "Report", (this is where the current .`suggestion()` lines come into play), which are all collected together and printed in a nice colorized way to the terminal (see https://docs.rs/color-eyre/0.5.8/color_eyre/section/trait.Section.html).
  - This way, different types of applications with different error-rendering needs do not need to pull in specialized code for rendering errors for a different application type. For example, a web service using `fluvio-cluster` should not need to link to the code which renders `fluvio-cluster` errors for the CLI. Instead, it should provide its own application-specific rendering, such as mapping the error into an HTTP resposne with JSON. Similarly, this HTTP-JSON error-rendering would not be linked into the CLI binary target.
  - This `into_report` function in `src/cluster/src/cli/error.rs` was where I thought there was the best boundary between the `fluvio-cluster` _library errors_ and the _application error rendering_. One thing to note is that this is a unique task and that the `into_report` function provides a single home for us to add more CLI-specific annotations to our application errors, and they won't spill out into the rest of the codebase.

Let me know if you have any problems with this design, but I don't see any glaring problems that will harm the codebase in the long run. We can always tweak and update the exact layout later if we find pain points.

src/cluster/src/main.rs
nicholastmosher: See my comment above

src/cluster/Cargo.toml
sehz: `version 0.1.0` is published

src/cluster/src/cli/error.rs
sehz: Initial thought is that  `Suggestion` trait might be more appropriate for Fluvio.   However, if `Report` type is sufficient for now, we can defer that idea later. 

Still, `into_report` should be push down to individual Error or at least to top level Error in the individual extension.  Implementation at CLI creates explicit coupling between them.   For example, if we run `Cluster` extension by itself, it will have different output.   CLI is not expected to perform any post processing.  It's job is to hand off argument to extension

nicholastmosher:
CI is failing from this clippy error

```
❯ make check-clippy check-fmt
rustup component add clippy --toolchain stable
info: component 'clippy' for target 'x86_64-apple-darwin' is up to date
cargo +stable clippy --all-targets  -- -D warnings
    Checking fluvio-extension-common v0.1.0 (/Users/nick/infinyon/fluvio/src/extension-common)
    Checking fluvio-runner-local v0.1.0 (/Users/nick/infinyon/fluvio/src/extension-runner-local)
    Checking fluvio-extension-consumer v0.1.0 (/Users/nick/infinyon/fluvio/src/extension-consumer)
    Checking fluvio-cluster v0.3.2 (/Users/nick/infinyon/fluvio/src/cluster)
    Checking fluvio-cli v0.4.1 (/Users/nick/infinyon/fluvio/src/cli)
warning: function is never used: `get_extensions`
  --> src/cli/src/install/mod.rs:29:15
   |
29 | pub(crate) fn get_extensions() -> Result<Vec<(String, PathBuf)>, CliError> {
   |               ^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: 1 warning emitted

error: function is never used: `get_extensions`
  --> src/cli/src/install/mod.rs:29:15
   |
29 | pub(crate) fn get_extensions() -> Result<Vec<(String, PathBuf)>, CliError> {
   |               ^^^^^^^^^^^^^^
   |
   = note: `-D dead-code` implied by `-D warnings`

error: could not compile `fluvio-cli`.

error: build failed
make: *** [check-clippy] Error 101
```

I believe this is a function that was newly added in another PR. @simlay do you have plans to use this function in the near future, or is it not needed?

sehz:
> I believe this is a function that was newly added in another PR. @simlay do you have plans to use this function in the near future, or is it not needed?

CI run in `master` passes as well as running locally.   Maybe rebase issue?

simlay:
> CI is failing from this clippy error
> 
> ```
> ❯ make check-clippy check-fmt
> rustup component add clippy --toolchain stable
> info: component 'clippy' for target 'x86_64-apple-darwin' is up to date
> cargo +stable clippy --all-targets  -- -D warnings
>     Checking fluvio-extension-common v0.1.0 (/Users/nick/infinyon/fluvio/src/extension-common)
>     Checking fluvio-runner-local v0.1.0 (/Users/nick/infinyon/fluvio/src/extension-runner-local)
>     Checking fluvio-extension-consumer v0.1.0 (/Users/nick/infinyon/fluvio/src/extension-consumer)
>     Checking fluvio-cluster v0.3.2 (/Users/nick/infinyon/fluvio/src/cluster)
>     Checking fluvio-cli v0.4.1 (/Users/nick/infinyon/fluvio/src/cli)
> warning: function is never used: `get_extensions`
>   --> src/cli/src/install/mod.rs:29:15
>    |
> 29 | pub(crate) fn get_extensions() -> Result<Vec<(String, PathBuf)>, CliError> {
>    |               ^^^^^^^^^^^^^^
>    |
>    = note: `#[warn(dead_code)]` on by default
> 
> warning: 1 warning emitted
> 
> error: function is never used: `get_extensions`
>   --> src/cli/src/install/mod.rs:29:15
>    |
> 29 | pub(crate) fn get_extensions() -> Result<Vec<(String, PathBuf)>, CliError> {
>    |               ^^^^^^^^^^^^^^
>    |
>    = note: `-D dead-code` implied by `-D warnings`
> 
> error: aborting due to previous error
> 
> error: could not compile `fluvio-cli`.
> 
> To learn more, run the command again with --verbose.
> warning: build failed, waiting for other jobs to finish...
> error: build failed
> make: *** [check-clippy] Error 101
> ```
> 
> I believe this is a function that was newly added in another PR. @simlay do you have plans to use this function in the near future, or is it not needed?

Hmm, maybe this feature branch didn't merge the changes quite right? It's definitely used in [current master](https://github.com/infinyon/fluvio/blob/8e5866c48b490db2ac1a1f5a6f476696f7057441/src/cli/src/root_cli.rs#L286)

src/cluster/src/cli/error.rs
nicholastmosher: > to top level Error in the individual extension

As of right now, the only suggestions that exist are in `fluivo-cluster` for the `ClusterError` type. Other `into_report` implementations will live in the respective crate for other extensions.

nicholastmosher:
Ok I'll check out what happened

nicholastmosher:
Ok, yeah it was a rebasing problem. Since I merged root_cli into lib.rs in cli/, I missed the extra changes that were added to root_cli during the latest rebase. It should all be fixed now

src/cluster/src/cli/error.rs
sehz: Yes, I see that now.   It looks good now

sehz:
There is CI failure for Mac Local test.  Try to run locally see if you can reproduce that issue

nicholastmosher:
Running locally on this branch seems to work just fine

```
❯ cargo run --bin fluvio -- cluster install --setup --local --develop
    Finished dev [unoptimized + debuginfo] target(s) in 30.56s
     Running `target/debug/fluvio cluster install --setup --local --develop`
Performing pre-flight checks
Setup successful, all the steps neccessary for cluster installation have been performed successfully
```

nicholastmosher:
`make smoke-test-tls` also passes

```
❯ make smoke-test-tls
cargo build   --bin fluvio
    Finished dev [unoptimized + debuginfo] target(s) in 0.61s
cargo build   --bin flv-test
   Compiling fluvio-system-util v0.1.0 (/Users/nick/infinyon/fluvio/src/utils)
   Compiling flv-test v1.0.1 (/Users/nick/infinyon/fluvio/tests/runner)
    Finished dev [unoptimized + debuginfo] target(s) in 8.49s
echo "clean up previous installation"
clean up previous installation
./target/debug/fluvio cluster uninstall
./target/debug/fluvio cluster uninstall --local
FLV_CMD=true ./target/debug/flv-test --spu 1 --produce-iteration 5 --tls --local --client-log warn --server-log debug --skip-checks
using local environment driver
>> /Users/nick/infinyon/fluvio/./target/debug/fluvio cluster uninstall --local
>> helm version
installing cluster
>> /Users/nick/infinyon/fluvio/./target/debug/fluvio cluster install --spu 1 --local --rust-log debug --tls --domain fluvio.local --ca-cert /Users/nick/infinyon/fluvio/tls/certs/ca.crt --client-cert /Users/nick/infinyon/fluvio/tls/certs/client-root.crt --client-key /Users/nick/infinyon/fluvio/tls/certs/client-root.key --server-key /Users/nick/infinyon/fluvio/tls/certs/server.key --server-cert /Users/nick/infinyon/fluvio/tls/certs/server.crt
Performing pre-flight checks
>> helm version
>> helm version
>> helm list --filter ^fluvio-sys$ --output json
>> sync
>> /Users/nick/infinyon/fluvio/./target/debug/fluvio cluster run sc --tls --enable-client-cert --server-cert /Users/nick/infinyon/fluvio/tls/certs/server.crt --server-key /Users/nick/infinyon/fluvio/tls/certs/server.key --ca-cert /Users/nick/infinyon/fluvio/tls/certs/ca.crt --bind-non-tls-public 0.0.0.0:9005
>> /Users/nick/infinyon/fluvio/./target/debug/fluvio cluster run spu --tls --enable-client-cert --server-cert /Users/nick/infinyon/fluvio/tls/certs/server.crt --server-key /Users/nick/infinyon/fluvio/tls/certs/server.key --ca-cert /Users/nick/infinyon/fluvio/tls/certs/ca.crt --bind-non-tls-public 0.0.0.0:9012 -i 5001 -p 0.0.0.0:9010 -v 0.0.0.0:9011
1 spus provisioned
creating test topic: <topic0>
setting rust log: warn
>> /Users/nick/infinyon/fluvio/./target/debug/fluvio topic create topic0 --replication 1
topic "topic0" created
topic: topic0, created
all 1 partitions are live
>> /Users/nick/infinyon/fluvio/./target/debug/fluvio topic list
 NAME    TYPE      PARTITIONS  REPLICAS  IGNORE-RACK  STATUS                   REASON
 topic0  computed      1          1                   resolution::provisioned
>> /Users/nick/infinyon/fluvio/./target/debug/fluvio partition list
 TOPIC   PARTITION  LEADER  REPLICAS  RESOLUTION  HW  LEO  LSR  FOLLOWER OFFSETS
 topic0  0          5001    []        Online      0   0    0    []
start testing...
found topic: topic0 offset: 0
starting fetch stream for: topic0 base offset: 0, expected new records: 5
<<consume test done for: topic0 >>>>
consume message validated!
```

sehz:
was able to reproduce the issue by `rm -rf ~/.fluvio/` and try running setup.

nicholastmosher:
This was a refactoring problem on my end. The `ClusterCmd::process` function used to accept a `ClusterTarget` as its argument and I changed it to accept a `&FluvioConfig`. However, the process of converting a ClusterTarget to a FluvioConfig may fail if it tries to load a profile from `~/.fluvio/config` and the file does not exist. I've solved this by reverting to accepting the plain `ClusterTarget`.

This has got me thinking though, I wonder if there are any improvements that could be made to the FluvioConfig API. I'm imagining something like a builder where it starts with a bunch of default settings, then various methods can override those settings from different sources. The different sources might be profiles or they may be CLI arguments, and the builder methods could "merge" newer settings into the previous or default settings. This would be worthy of an entirely new issue, I'm just thinking out loud here.

src/cluster/Cargo.toml
```diff
@@ -34,21 +38,22 @@ serde = "1.0.115"
...
```
sehz, line 47: can use published version of helm

sehz (changes requested):
Everything looks good except one small nit

sehz (approved):

PR #491, opened 2020-11-19
update to clippy for 1.48 and bump fluvio-service 0.3

sehz:


sehz:
local run of `FLV_CMD=true FLV_SOCKET_WAIT=600 make smoke-test-k8-tls-root` passes.  merging this

Issue #463, opened 2020-11-11
Merge `spu` and `custom-spu` terms


nicholastmosher:
So I'm documenting the CLI, and wanted to pitch an idea. Right now we have these commands:

```
fluvio spu list                  List custom & managed SPUs
fluivo custom-spu register       Create custom SPU
fluvio custom-spu unregister     Delete custom SPU
fluvio custom-spu list           List custom SPUs
```

I was thinking we could lay them out like this instead:

```
fluvio spu register                    Register a custom (out-of-band) SPU
fluvio spu unregister                  Unregister a custom (out-of-band) SPU
fluvio spu list [--managed|--custom]   List all SPUs (or only --managed or --custom)
```

To make things consistent, we should find all instances of SPU vs Custom SPU and consolidate them to SPU.

- [x] Rename in CLI
~Rename in Documentation~ covered by (#63)
~Rename in ControlPlane API~ future issue (#538)
~Rename in Admin API~ future issue (#538)
~Rename in SC~ future issue (#538)

sehz:
I think this makes sense.  Instead of creating another type, this allow us to consolidate.   In that case, we should change
- Doc/CLI
- ControlPlane API
- Admin API
- SC controller

implementations to conform to this

nicholastmosher:
Remaining work was moved to issues for the future.

PR #495, opened 2020-11-20
Merge custom spu and spu commands

nicholastmosher:
This one was quick so I was able to put it together during some down time.

I made the changes we talked about in #463, so the commands now look like this:

```
fluvio spu register
fluvio spu unregister
fluvio spu list                       # Lists managed and custom SPUs
fluvio spu list --managed             # Lists managed SPUs
fluvio spu list --custom              # Lists custom SPUs
fluvio spu list --managed --custom    # Lists managed and custom SPUs
```

I must say I have not tested this at all, but the code compiles, so...


sehz (approved):

sehz:
running test locally passes

PR #499, opened 2020-11-21
separate consumer extension from CLI

sehz:
- created common extension crates
- move consumer related commands into `consumer` extension
- use updated version of k8-client


sehz:
K8 failures are due to GI VM problem.  K8 smoke tests ran fine in local env

PR #502, opened 2020-11-21
remove helm into separate repo

sehz:


sehz:
CI k8-test is stuck but runs ok in local env

PR #508, opened 2020-11-23
Search $FLUVIO_DIR and $HOME/.fluvio/extensions for extension executable

### LINKED ISSUE #500: Extension:  Use CLI to search only extension folder
Currently CLI plug-in loader searches entire PATH.  Instead it should only search for extension folder.  Also, Installer should not give instruction to user about changing PATH since it is no longer required

- [x] Use  `extension` folder instead of `bin`

- [x] Only search extension folder instead of entire PATH

- [x] Update Installer to remove PATH warning

simlay:
This will close #500.

### DIFF
--- a/src/cli/src/install/mod.rs
+++ b/src/cli/src/install/mod.rs
@@ -9,10 +9,22 @@ use crate::CliError;
 pub mod update;
 pub mod plugins;
 
-fn fluvio_bin_dir() -> Result<PathBuf, CliError> {
+pub(crate) fn fluvio_extensions_dir() -> Result<PathBuf, CliError> {
+    if let Ok(dir) = std::env::var("FLUVIO_DIR") {
+        // Assume this is like `~/.fluvio
+        let path = PathBuf::from(dir).join("extensions");
+        if path.exists() {
+            return Ok(path);
+        }
+    }
+
     let home =
-        dirs::home_dir().ok_or_else(|| IoError::new(ErrorKind::NotFound, "Homedir not found"))?;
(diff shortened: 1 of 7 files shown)


src/cli/src/install/mod.rs
```diff
@@ -9,10 +9,25 @@ use crate::CliError;
...
+    if let Ok(dir) = std::env::var("FLUVIO_DIR") {
+        // Assume this is like `~/.fluvio
+        let path = PathBuf::from(dir).join("extensions");
+        if path.exists() {
+            return Ok(path);
+        }
+    }
+
     let home =
-        dirs::home_dir().ok_or_else(|| IoError::new(ErrorKind::NotFound, "Homedir not found"))?;
```
simlay, line 14: https://github.com/rust-lang/rust/issues/43321 relevant

sehz (changes requested):
Looks good.  What about removing warning about PATH in the installer?

simlay:
> What about removing warning about PATH in the installer?

Whoops. Forgot about that. Where do we want the installer to install to? `/usr/local/bin`? I actually kinda like it getting installed at `~/.fluvio/bin`.

sehz:
> > What about removing warning about PATH in the installer?
> 
> Whoops. Forgot about that. Where do we want the installer to install to? `/usr/local/bin`? I actually kinda like it getting installed at `~/.fluvio/bin`.

I see.  if it is installed at `~/.fluvio/bin`, we still need have user modify PATH

sehz (approved):

PR #510, opened 2020-11-23
fix topic controller not sleeping properly

sehz:
Instead of sleeping 60 seconds between each loop, it is only done once.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,8 +29,9 @@ install_tools_mac:
 	brew install helm
 
 build_test:
-	cargo build --${TEST_BUILD} --bin fluvio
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -29,8 +29,9 @@ install_tools_mac:
 	brew install helm
 
 build_test:
-	cargo build --${TEST_BUILD} --bin fluvio
-	cargo build --${TEST_BUILD} --bin flv-test
+	export CARGO_PROFILE=$(if $(TEST_BUILD),release,debug); \
```
nacardin, line 32: Did you mean to commit this?

PR #512, opened 2020-11-24
Initial fluvio extension metadata commands

### LINKED ISSUE #501: Extension: Extension Command Protocol
Currently CLI build top level commands by linking `StructOpt` in the extension directly using rust.  Since extension will be load dynamically and can be written any language, we need to way to build top level commands in the CLI dynamically.

For example:
```$ fluvio
All top-level subcommands that require a Fluvio client

... <common flags>

SUBCOMMANDS:
    consume      Reads messages from a topic/partition
    produce      Writes messages to a topic/partition
    spu          Manage and view Streaming Processing Units (SPUs)
    spg          Manage and view SPU Groups (SPGs)
    topic        Manage and view Topics
    partition    Manage and view Partitions
    profile      Manage Profiles, which describe linked clusters
    cluster      Install or uninstall Fluvio clusters
    run          Run a Streaming Controller (SC) or SPU
    install      Install Fluvio plugins
    update       Update the Fluvio CLI
    version      Print Fluvio version information
    help         Prints this message or the help of the given subcommand(s)
```

Commands such as consume/produce/spu/clusters need to be  derived from extensions dynamically whether they are statically linked into CLI or dynamically loaded.

- [ ] Provide Command Traits for rust extensions to generate top level help

- [ ] Update CLI to load top levels dynamically from extensions

- [ ] Migrate consumer extension to use

simlay:
This PR will eventually close #501.


src/extension-consumer/src/topic/mod.rs
```diff
@@ -64,6 +79,11 @@ impl TopicCmd {
             TopicCmd::List(list) => {
                 list.process(out, fluvio).await?;
             }
+            TopicCmd::Metadata => {
+                use crate::common::t_println;
+                t_println!(out, "{}", format!("metadata: {:#?}", self.metadata()));
```
simlay, line 84: When I run this I get an error `Connection refused (os error 111)`

src/extension-common/src/lib.rs
```diff
@@ -30,6 +30,28 @@ macro_rules! t_print_cli_err {
...
+#[derive(Debug)]
+pub struct FluvioExtensionMetaData {
+    pub command: String,
+    pub description: String,
+    pub version: String,
+}
+
+use structopt::StructOpt;
+pub trait FluvioExtension: StructOpt {
+    fn version(&self) -> String;
```
sehz, line 42: use `&str`

src/extension-common/src/lib.rs
```diff
@@ -30,6 +30,28 @@ macro_rules! t_print_cli_err {
...
+    fn description(&self) -> String;
```
sehz, line 43: use `&str`

src/extension-runner/Cargo.toml
```diff
@@ -25,4 +25,5 @@ thiserror = "1.0.20"
 # regardless of TLS, sc and spu always use rust_tls for now because we need cert API
 fluvio-future = { version = "0.1.8", features = ["subscriber"] }
 fluvio-sc = { version = "0.3.2", path = "../sc", default-features = false, features = ["k8_rustls"]}
-fluvio-spu = { version = "0.2.2", path = "../spu", default-features = false }
\ No newline at end of file
+fluvio-spu = { version = "0.2.2", path = "../spu", default-features = false }
+fluvio-extension-common = { version = "0.1.0", path = "../extension-common", features = ["target"] }
```
sehz, line 29: runner doesn't need this right now because it is not top level extension

sehz (changes requested):
good start.
need to add this to `cluster` command extension

src/extension-common/src/lib.rs
```diff
@@ -30,6 +30,28 @@ macro_rules! t_print_cli_err {
     };
 }
 
+#[derive(Debug)]
+pub struct FluvioExtensionMetaData {
```
nicholastmosher, line 34: Nit: `FluvioExtensionMetadata`

src/extension-common/src/lib.rs
nicholastmosher: I don't think there's any way to implement this using `&str` if you want to be able to use `env!()` values.

nicholastmosher (commented):
I think there may be a good way to minimize the boilerplate that individual extensions need to write. Since we're deriving the version and description from cargo anyway, we could just make a macro that creates the right structopt subcommand and metadata all at once. Something like:

```rust
fluvio_extension!(MetadataOpt);

// Which would expand to:

struct MetadataOpt {}
impl MetadataOpt {
    fn process(self) {
        // Print JSON for top-level CLI to use
        println!("{}", serde_json::to_str(&Self::metadata());
    }

    fn metadata() -> FluvioExtensionMetadata {
        FluvioExtensionMetadata {
            version: env!("CARGO_PKG_VERSION"),
            description: ...,
        }
    }
}
```

Then the only manual integration point would be dropping this `MetadataOpt` into the structopt

```rust
enum TopicCmd {
    ...
    #[structopt(hidden = true)]
    Metadata(MetadataOpt),
}
```

This would also prevent us from needing a Trait, I think this is probably the wrong use-case for a trait since there we are not using it generically anywhere. I.e., there is no instance of `<E: FluvioExtension>` anywhere and I don't imagine there ever will be (especially since the point of this exercise seems to be _avoiding_ linking the CLI modules together via code). https://matklad.github.io/2020/08/15/concrete-abstraction.html

src/extension-consumer/Cargo.toml
```diff
@@ -36,9 +36,9 @@ which = "4.0.2"
 # Fluvio dependencies
 
 flv-util = { version = "0.5.0" }
-fluvio-future = { version = "0.1.8" }
+fluvio-future = { version = "0.1.8", features = ["fs", "io"] }
```
simlay, line 39: Not sure why this broke but when I pulled master, I had to add this to fix the build.

src/extension-common/src/lib.rs
simlay: `env!` returns a `&str` but if we use the default implementation for `metadata`, it's just converting those to `String` which seems not ideal to me.

simlay: The biggest issue I have with `env!` is for the `CARGO_PKG_DESCRIPTION` environment variable, is set at the crate level and the current `extension-consumer` crate has `topic`, `partition`, `produce`, `consume` which results in them having the same `description` in the metadata.

sehz:
> I think there may be a good way to minimize the boilerplate that individual extensions need to write. Since we're deriving the version and description from cargo anyway, we could just make a macro that creates the right structopt subcommand and metadata all at once. Something like:
> 



> ```rust
> fluvio_extension!(MetadataOpt);
> 
> // Which would expand to:
> 
> struct MetadataOpt {}
> impl MetadataOpt {
>     fn process(self) {
>         // Print JSON for top-level CLI to use
>         println!("{}", serde_json::to_str(&Self::metadata());
>     }
> 
>     fn metadata() -> FluvioExtensionMetadata {
>         FluvioExtensionMetadata {
>             version: env!("CARGO_PKG_VERSION"),
>             description: ...,
>         }
>     }
> }
> ```
> 
> Then the only manual integration point would be dropping this `MetadataOpt` into the structopt
> 
> ```rust
> enum TopicCmd {
>     ...
>     #[structopt(hidden = true)]
>     Metadata(MetadataOpt),
> }
> 
> // and
> 
> impl TopicCmd {
>     pub fn process(self) {
>         match self {
>             ...
>             Self::Metadata(metadata) => metadata.process(),
>         }
>     }
> }
> ```
> 
> This would also prevent us from needing a Trait, I think this is probably the wrong use-case for a trait since there we are not using it generically anywhere. I.e., there is no instance of `<E: FluvioExtension>` anywhere and I don't imagine there ever will be (especially since the point of this exercise seems to be _avoiding_ linking the CLI modules together via code). https://matklad.github.io/2020/08/15/concrete-abstraction.html

Problem with macro is that it is sort of "magic"  and sometimes make it hard to to customize it.   Trait seems to make it more transparent and accessible.   We can have "macro" to automate them but trait allow enforcement of consistent behavior while macro does it by compiler trickery only

simlay:
> Problem with macro is that it is sort of "magic" and sometimes make it hard to to customize it. Trait seems to make it more transparent and accessible. We can have "macro" to automate them but trait allow enforcement of consistent behavior while macro does it by compiler trickery only

I kind of agree with this. Part of the reason I like a trait here is because I want to be told what I need to implement when building an extension and this helps the implementer know they did it right. There's definitely a time to when traits can become shackles as mentioned in that post.

src/extension-consumer/src/topic/mod.rs
sehz: that's weird.

src/extension-consumer/Cargo.toml
sehz: shouldn't need that. what is exact error?

sehz (changes requested):

src/extension-common/src/lib.rs
```diff
@@ -30,6 +31,13 @@ macro_rules! t_print_cli_err {
     };
 }
 
+#[derive(Debug, Serialize, Deserialize)]
```
sehz, line 34: See comments in the CLI,  we can return array of commands

sehz (changes requested):

src/extension-consumer/src/topic/mod.rs
simlay: Yeah... this is what I get:
```
simlay@tower-1:~/projects/infinyon/fluvio (extension-command-protocol)$ cargo run --bin fluvio -- topic metadata
   Compiling fluvio-extension-consumer v0.1.0 (/home/simlay/projects/infinyon/fluvio/src/extension-consumer)
   Compiling fluvio-cli v0.4.0 (/home/simlay/projects/infinyon/fluvio/src/cli)
    Finished dev [unoptimized + debuginfo] target(s) in 10.26s
     Running `target/debug/fluvio topic metadata`
Error: 
   0: Fluvio client error
   1: Fluvio socket error
   2: Connection refused (os error 111)

Location:
   src/cli/src/bin/main.rs:12

Backtrace omitted.
```

src/extension-consumer/Cargo.toml
simlay: I notice that I don't get the error when I just build the fluvio cli but when I `cargo build` in the `extension-common` subcrate, I get this:

```
simlay@tower-1:~/projects/infinyon/fluvio/src/extension-consumer (extension-command-protocol)$ cargo build
   Compiling fluvio-extension-common v0.1.0 (/home/simlay/projects/infinyon/fluvio/src/extension-common)
   Compiling fluvio-extension-consumer v0.1.0 (/home/simlay/projects/infinyon/fluvio/src/extension-consumer)
error[E0432]: unresolved import `fluvio_future::fs`
   --> src/extension-consumer/src/produce/mod.rs:118:24
    |
118 |     use fluvio_future::fs::File;
    |                        ^^ could not find `fs` in `fluvio_future`

error[E0432]: unresolved import `fluvio_future::io`
   --> src/extension-consumer/src/produce/mod.rs:119:24
    |
119 |     use fluvio_future::io::stdin;
    |                        ^^ could not find `io` in `fluvio_future`

error[E0432]: unresolved import `fluvio_future::io`
   --> src/extension-consumer/src/produce/mod.rs:120:24
    |
120 |     use fluvio_future::io::ReadExt;
    |                        ^^ could not find `io` in `fluvio_future`

error: aborting due to 3 previous errors

For more information about this error, try `rustc --explain E0432`.
error: could not compile `fluvio-extension-consumer`

```

sehz: Yup.  consumer crates need: `io` and `fs` feature flags

src/cli/src/root_cli.rs
```diff
@@ -247,6 +260,42 @@ impl VersionOpt {
...
+    pub fn process(self) -> Result<()> {
+        let metadata = self.metadata();
+
+        println!("FLUVIO metadata: {:?}", metadata);
+        Ok(())
+    }
+    pub fn metadata(self) -> Result<HashMap<String, FluvioExtensionMetadata>> {
+        // Scan for extensions, run `fluvio <extension> metadata` on them, add them to metadata
+        // hashmap, run the same on FluvioCmds
+        let mut metadata: HashMap<String, FluvioExtensionMetadata> = HashMap::new();
```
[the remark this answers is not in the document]
simlay: After playing with it for a while, I think we'd have to change the way that packages/extensions are installed and executed. Given that our packages only [download one binary (for now)](https://github.com/infinyon/fluvio/blob/b0a6688e114e23f3480d6a62a8bea4e78b01319f/src/package-index/src/http.rs#L56-L62) when they install a new package, as a future extension writer if I had to return a list of one, I'd be a little annoyed. I think that making an exception for the top level `fluvio metadata` function leads to the least amount of inconsistency.

sehz (approved):

src/extension-consumer/src/topic/mod.rs
simlay: I removed this because it's broken and I don't see the need for a statically bound command to need a `metadata` command.

PR #524, opened 2020-11-27
improve unit test time in the CI using sccache

sehz:
update unit-test-all


sehz:
Seems to improve unit test build time by 30-50%

PR #525, opened 2020-11-27
Make CLI help menu wording consistent

nicholastmosher:
These are some small wording changes on the structopt docstrings which change the way the help menu is rendered for the CLI. I changed it so that everything is using the imperative tone, e.g. "Read" rather than "Reads". This is also how the descriptions will be written for the CLI reference on the website.


sehz (approved):

sehz:
nice cleanup

PR #533, opened 2020-12-01
Add Fluvio Platform version to `fluvio version`

### LINKED ISSUE #383: Show Client and Platform version
Currently when `version` command is perform in the CLI. It shows following:
```
fluvio version
Fluvio version : 0.6.0-alpha.2
Git Commit     : 9534b85bc34866acf50a13f777ee57ebf43e35fa
OS Details     : Darwin 19.6.0 x86_64
Rustc Version  : 1.46.0 (04488af 2020-08-24)
```

We should separate in terms of client and current server version.  So it should be something like:
```
fluvio version
Fluvio Client version : 0.6.0-alpha.2
Fluvio Platform version : 0.6.0-alpha.3
Git Commit     : 9534b85bc34866acf50a13f777ee57ebf43e35fa
OS Details     : Darwin 19.6.0 x86_64
Rustc Version  : 1.46.0 (04488af 2020-08-24)
```

nicholastmosher:
Closes #383


nicholastmosher:
The errors we're seeing do seem to be an Encoding problem. Commenting out this line fixes the error
https://github.com/infinyon/fluvio/pull/533/files#diff-018b6d63a3c28d8e1cdd29f3d1864b0d21b6744c134b58944262cb6f5008f209R32

src/cli/src/lib.rs
```diff
@@ -252,13 +253,23 @@ impl FluvioCmd {
...
+    pub async fn process(self, target: ClusterTarget) -> Result<()> {
+        println!("Fluvio CLI      : {}", crate::VERSION.trim());
+
+        // Attempt to connect to a Fluvio cluster to get platform version
+        // Even if we fail to connect, we should not fail the other printouts
+        if let Ok(fluvio_config) = target.load() {
+            if let Ok(fluvio) = Fluvio::connect_with_config(&fluvio_config).await {
+                let version = fluvio.platform_version();
+                println!("Fluvio Platform : {}", version);
+            }
```
sehz, line 265: in case of error,  we should just print out error instead of swallowing

src/dataplane-protocol/src/apis.rs
```diff
@@ -4,16 +4,14 @@
...
 
-use dataplane::derive::Encode;
-use dataplane::derive::Decode;
+use crate::derive::{Decode, Encode};
 
 /// API call from client to SPU
 #[fluvio(encode_discriminant)]
 #[derive(Encode, Decode, PartialEq, Debug, Clone, Copy)]
 #[repr(u16)]
 pub enum AdminPublicApiKey {
```
sehz, line 13: This doesn't belong in the `dataplane-protocol` since this is used by consumer and producer not admin

src/sc-schema/src/lib.rs
```diff
@@ -1,4 +1,3 @@
-mod apis;
```
sehz, line 1: need to move back from data plane to here

sehz (changes requested):

src/cli/src/lib.rs
nicholastmosher: Well suppose that I am a new Fluvio user and I have never connected the CLI to a cluster yet. I try to run `fluvio version`. I would expect it not to crash or give me an error. All this is doing is optionally printing the cluster SC version info only when we can successfully connect to a cluster. Otherwise it prints just local version info.

sehz: In that case, just say `Problem connecting to Fluvio`.    For kubectl, this happens when it can't connects
```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.0", GitCommit:"e19964183377d0ec2052d1f1fa930c4d7575bd50", GitTreeState:"clean", BuildDate:"2020-08-26T14:30:33Z", GoVersion:"go1.15", Compiler:"gc", Platform:"linux/amd64"}
The connection to the server 192.168.49.2:8443 was refused - did you specify the right host or port?
``

sehz (approved):

PR #537, opened 2020-12-02
Handle nonexistent extensions dir and extension signals.

### LINKED ISSUE #532: [bug] Fluvio extensions directory not found
On master (4144c3aa), running `cargo run --bin fluvio --` exits with an error with the following message:

```
Error:
    0: Fluvio extensions directory not found
```

I recently deleted my entire `~/.fluvio/` directory, this may also happen if `~/.fluvio/` exists but `~/.fluvio/extensions` doesn't.

We should make sure that the extension-finding system either gracefully deals with no extensions directory, or creates one.

simlay:
Closes #532 also deals with the edge case where [`code`](https://doc.rust-lang.org/std/process/struct.ExitStatus.html#method.code) doesn't return a status code because it's been killed on a unix system (my version of fluvio-package segfaulted due to a `SIGSEGV`).

### DIFF
--- a/src/cli/src/install/mod.rs
+++ b/src/cli/src/install/mod.rs
@@ -13,15 +13,20 @@ pub(crate) fn fluvio_extensions_dir() -> Result<PathBuf, CliError> {
     if let Ok(dir) = std::env::var("FLUVIO_DIR") {
         // Assume this is like `~/.fluvio
         let path = PathBuf::from(dir).join("extensions");
-        if path.exists() {
-            return Ok(path);
+        if !path.exists() {
+            std::fs::create_dir(&path)?;
         }
+        return Ok(path);
     }
 
     let home =
         home::home_dir().ok_or_else(|| IoError::new(ErrorKind::NotFound, "Homedir not found"))?;
-    let path = home.join(".fluvio/extensions/");
+    let path = home.join(".fluvio");
     if path.exists() {
+        let path = path.join("extensions/");
+        if !path.exists() {
+            std::fs::create_dir(&path)?;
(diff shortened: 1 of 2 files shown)


sehz (approved):

src/cli/src/install/mod.rs
```diff
@@ -13,15 +13,20 @@ pub(crate) fn fluvio_extensions_dir() -> Result<PathBuf, CliError> {
     if let Ok(dir) = std::env::var("FLUVIO_DIR") {
         // Assume this is like `~/.fluvio
         let path = PathBuf::from(dir).join("extensions");
-        if path.exists() {
-            return Ok(path);
+        if !path.exists() {
+            std::fs::create_dir(&path)?;
```
sehz, line 17: actually,  shouldn't not create extension directory if it doesn't exists for purpose of query extensions.   It should be only done when extension is installed

sehz (changes requested):

src/cli/src/install/mod.rs
simlay: I'm fine with that. I'm just not sure what's more graceful than saying
```
Error:
    0: Fluvio extensions directory not found
```

If you're running an extension and there's no extensions directory, the user should know they've gotta reinstall or something. @nicholastmosher What do you think?

sehz: Hmm.  I think it is ok to create directory if it doesn't exists.

sehz:
rebase please

sehz (approved):

PR #530, opened 2020-12-01
Initial S3 makefile

simlay:
This will close #465.

PR #544, opened 2020-12-03
Fluvio publish in release workflow

### LINKED ISSUE #465: Integrate package publisher with GitHub Release workflow
Release workflow uses package publisher int the GitHub Actions to publish to packages S3 bucket and github releases.
- CLI

Ideally, this will be done using fluvio-packages for the S3 bucket bit.

simlay:
I forgot that releases workflows can only run from a branch in the same repo.

This is the same PR as #530 and will close #465.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,10 +22,31 @@ jobs:
         uses: davidB/rust-cargo-make@v1
         with:
           version: '0.32.9'
+
+      - name: install musl-tools
+        if: startsWith(matrix.os, 'ubuntu')
+        run: |
+          sudo apt-get update;
+          sudo apt-get install -y musl-tools build-essential;
+
+      - name: Build and publish using fluvio-packages
+        run: cargo make -l verbose --profile production publish-fluvio
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+
+      - name: Update fluvio install.sh
+        if: startsWith(matrix.os, 'ubuntu')
+        run: cargo make -l verbose --profile production s3-upload-installer
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+
       - name: Build and upload release to github
         run: cargo make -l verbose --profile production github-release-upload
         env:
           GITHUB_TOKEN: ${{ secrets.ACTION_RELEASE }}
+
 
   release_fluvio_pi:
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -139,13 +139,15 @@ check-clippy:	install-clippy
...
 
-build-all-test:	
+build-all-test:
 	cargo build --tests --all-features
 
 run-all-unit-test:
 	cargo test --all-features
 
 install_musl:
+	sudo apt-get update;
```
sehz, line 149: this need to be specifically target toward ubuntu which make it not suitable here

.github/workflows/release.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: [infinyon-ubuntu-bionic, macos-latest]
```
sehz: we should stick with GitHub runner for now

makefiles/Base.toml
```diff
@@ -8,4 +8,39 @@ CARGO_PROFILE = "release"
...
+    "build",
+    "--release",
+    "--bin", "fluvio",
+    "--target=x86_64-unknown-linux-musl",
+    "@@split(CARGO_MAKE_CARGO_BUILD_TEST_FLAGS, )"
+]
+
+[tasks.setup-musl.linux]
+script = '''
+which musl-gcc 2> /dev/null > /dev/null || {
```
sehz, line 34: this is highly specific to ubuntu or OS that has `apt` and doesn't belong here.  it should be done by whatever infrastructure provisioning tools

sehz (changes requested):

makefiles/Base.toml
simlay: Yeah, I've got mixed feelings about it. This is why the check for `musl-gcc` exists so it doesn't need to be installed. This step is also contingent on it being done with the "production" profile.

sehz: but that creates dependency to specific infrastructure.  they should move to CI configuration or provisioning tools

makefiles/S3.toml
```diff
@@ -0,0 +1,48 @@
...
+]
+
+[tasks.publish-fluvio]
+dependencies = [
+    "build-release",
+]
+script = '''
+mkdir -p ~/.fluvio/extensions
+./target/${TARGET}/release/fluvio install fluvio/fluvio-package;
+./target/${TARGET}/release/fluvio package publish ./target/${TARGET}/release/fluvio --version=${FLUVIO_VERSION} --force;
```
simlay, line 41: remove `--force.`

simlay: [Build worked](https://github.com/infinyon/fluvio/actions/runs/397617509]. The only difference between https://github.com/infinyon/fluvio/commit/0e1e509727933328b29aace642e2257420054e25 and https://github.com/infinyon/fluvio/commit/4a6d916ba29acf2e8990cba1b79732b1b7e4bf0c is the `--force` here.

sehz (approved):

PR #549, opened 2020-12-03
Improve failed checks printout for fluvio-cluster

### LINKED ISSUE #519: Improve fluvio cluster pre-check printout
Right now if any of our cluster pre-checks fail, an Error bubbles all the way up and prints on the CLI, like this:

image

We should update this so that a failed pre-check does not print an "Error" but rather prints a custom "failed pre-check" message. It should be visually consistent with the passing pre-check messages, something like:

```
❯ fluvio cluster install
Performing pre-flight checks
 X  failed: No Kubernetes installation detected
```

nicholastmosher:
Closes #519.

With this PR, check messages now render much more uniformly. Also, most of the code that prints to the CLI was moved out of the checking module itself and into the `cluster/src/cli/` set of modules, which are only compiled when `features = ["cli"]` is enabled.

Here is how some checks look:

- With a system that is fully ready-to-go, the successful check printout looks like this:

```
❯ cargo run --bin fluvio -- cluster check
Running pre-startup checks...
✅ ok: Can create service
✅ ok: Can create customresourcedefinitions
✅ ok: Can create secret
✅ ok: Load balancer is up
All checks passed!
You may proceed with cluster startup
next: run `fluvio cluster start`
```

image

- On a system where minikube tunnel is not up, checks look like this:

```
❯ cargo run --bin fluvio -- cluster check
Running pre-startup checks...
✅ ok: Can create service
✅ ok: Can create customresourcedefinitions
✅ ok: Can create secret
❌ failed: Minikube tunnel may not be running
    Run `sudo nohup  minikube tunnel  > /tmp/tunnel.out 2> /tmp/tunnel.out &`
1 check failed: Please correct it before continuing with cluster startup
```

image

- On a system where the Fluvio system chart is not installed, checks look like this:

```
❯ cargo run --bin fluvio -- cluster check
    Finished dev [unoptimized + debuginfo] target(s) in 0.22s
     Running `target/debug/fluvio cluster check`
Running pre-startup checks...
❕ warning: Missing Fluvio system charts.
  💡 note: this may be fixed automatically during startup
✅ ok: Can create service
✅ ok: Can create customresourcedefinitions
✅ ok: Can create secret
✅ ok: Load balancer is up
Some checks failed, but may be auto-corrected
You may proceed with cluster startup
next: run `fluvio cluster start`
```

image

- Starting a cluster when there are auto-fixes available (e.g. missing system chart will be installed for us)

```
❯ cargo run --bin fluvio -- cluster start --local
    Finished dev [unoptimized + debuginfo] target(s) in 0.23s
     Running `target/debug/fluvio cluster start --local`
Performing pre-flight checks
NAME: fluvio-sys
LAST DEPLOYED: Thu Dec  3 11:35:55 2020
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
1 spus provisioned
✅ ok: Fixed: Missing Fluvio system charts.
```

image

### DIFF
--- a/src/cluster/src/check.rs
+++ b/src/cluster/src/check.rs
@@ -38,7 +38,8 @@ impl From<Vec<CheckResult>> for CheckResults {
 
 impl CheckResults {
     pub fn into_statuses(self) -> CheckStatuses {
-        let statuses: Vec<_> = self.0
+        let statuses: Vec<_> = self
+            .0
             .into_iter()
             .filter_map(|it| match it {
                 Ok(status) => Some(status),
@@ -343,14 +344,14 @@ impl ClusterChecker {
 
 /// Checks that the installed helm version is compatible with the installer requirements
 pub(crate) fn check_helm_version(helm: &HelmClient, required: &str) -> CheckResult {
-    let helm_version = helm
-        .get_helm_version()
-        .map_err(CheckError::HelmError)?;
+    let helm_version = helm.get_helm_version().map_err(CheckError::HelmError)?;
     if Version::parse(&helm_version) < Version::parse(required) {
-        return Ok(CheckStatus::fail(UnrecoverableCheck::IncompatibleHelmVersion {
-            installed: helm_version,
-            required: required.to_string(),
-        }));
+        return Ok(CheckStatus::fail(
+            UnrecoverableCheck::IncompatibleHelmVersion {
+                installed: helm_version,
+                required: required.to_string(),
+            },
+        ));
     }
     Ok(CheckStatus::pass("Supported helm version is installed"))
 }
@@ -391,7 +392,9 @@ pub(crate) async fn check_load_balancer_status() -> CheckResult {
     let cluster_context = match context.config.current_context() {
         Some(context) => context,
         None => {
-            return Ok(CheckStatus::fail(UnrecoverableCheck::NoActiveKubernetesContext));
+            return Ok(CheckStatus::fail(
+                UnrecoverableCheck::NoActiveKubernetesContext,
+            ));
         }
     };
 
@@ -409,7 +412,9 @@ pub(crate) async fn check_load_balancer_status() -> CheckResult {
             // hence handle both separately
             return Ok(CheckStatus::fail(get_tunnel_error()));
         }
-        return Ok(CheckStatus::fail(UnrecoverableCheck::LoadBalancerServiceNotAvailable));
+        return Ok(CheckStatus::fail(
+            UnrecoverableCheck::LoadBalancerServiceNotAvailable,
+        ));
     }
 
     Ok(CheckStatus::pass("Load balancer is up"))
@@ -487,7 +492,9 @@ fn check_cluster_server_host() -> CheckResult {
     let cluster_context = match context.config.current_cluster() {
         Some(context) => context,
         None => {
-            return Ok(CheckStatus::fail(UnrecoverableCheck::NoActiveKubernetesContext));
+            return Ok(CheckStatus::fail(
+                UnrecoverableCheck::NoActiveKubernetesContext,
+            ));
         }
     };
 
@@ -501,7 +508,9 @@ fn check_cluster_server_host() -> CheckResult {
         .unwrap_or(false);
 
     if !host_present {
-        return Ok(CheckStatus::fail(UnrecoverableCheck::MissingKubernetesServerHost));
+        return Ok(CheckStatus::fail(
+            UnrecoverableCheck::MissingKubernetesServerHost,
+        ));
     }
 
     Ok(CheckStatus::pass("Kubernetes config is loadable"))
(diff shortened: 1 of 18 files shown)


sehz:
rebase please

src/cluster/src/start/local.rs
```diff
@@ -510,4 +530,48 @@ impl LocalClusterInstaller {
...
+        let client = Fluvio::connect().await.expect("sc ");
+        let mut admin = client.admin().await;
+
+        // wait for list of spu
+        for _ in 0..30u16 {
+            let spus = admin.list::<SpuSpec, _>(vec![]).await.expect("no spu list");
+            let live_spus = spus
+                .iter()
+                .filter(|spu| {
+                    spu.status.is_online() && !spu.spec.public_endpoint.ingress.is_empty()
```
sehz, line 557: ingress checking is not needed for local cluster

sehz (changes requested):

src/cluster/src/start/local.rs
nicholastmosher: This is code that was already in the project, I just moved it here from `src/cluster/src/cli`. What would need to change in order for this check to be correct?

sehz: Should have flag to check for ingress or not.   For local, `false`, for k8, `true`

src/cluster/src/check.rs
```diff
@@ -135,33 +150,19 @@ pub enum CheckError {
...
-    async fn perform_check(&self) -> Result<StatusCheck, CheckError>;
+    async fn perform_check(&self) -> Result<String, CheckError>;
 }
 
 pub(crate) struct LoadableConfig;
 
 #[async_trait]
 impl InstallCheck for LoadableConfig {
-    async fn perform_check(&self) -> Result<StatusCheck, CheckError> {
+    async fn perform_check(&self) -> Result<String, CheckError> {
```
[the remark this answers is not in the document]
nicholastmosher: Do you just disagree with the name here? The data contained in this new structure is essentially the same as what was in the older structure, only now it is more organized across the `Result` boundary where `Ok` represents "the good thing that happened" and `Err` represents "the bad thing that happened".

Previously, we had these possible variants for a `Result<StatusCheck, CheckError>`:

```rust
match result {
    Ok(StatusCheck::Working(String)) => {},
    Ok(StatusCheck::NotWorkingNoRemediation(CheckError)) => {},
    Ok(StatusCheck::NotWorking(CheckError, String)) => {},
    Err(CheckError) => {}
}
```

One problem I had with this is that it is unclear which variant a particular `CheckError` should belong to. If I encounter an error during a check, there are three possible places that error might go, and the semantic meaning of it changes depending on where it's located, meaning that it's very easy to accidentally signal the wrong meaning.

The way things are laid out now is like this:

```rust
match result {
    Ok(String) => { /* String is a success message */ }
    Err(CheckError::AutoRecoverable(RecoverableCheck)) => {},
    Err(CheckError::Unrecoverable(UnrecoverableCheck)) => {},
    Err(CheckError::AlreadyInstalled) => {},
}
```

I specifically divided all of the variants into separate types, `RecoverableCheck` and `UnrecoverableCheck` to communicate the types of actions which may be taken next.

This new structure maps perfectly onto the actions that we want to take in each case.

- With `Ok(message)`, we just print the successful message
- With a `RecoverableCheck`, we want to attempt to automatically fix the problem
- With an `UnrecoverableCheck`, we know we can't do anything about it and we must return
- With `AlreadyInstalled`, we simply don't have any work to do

I really like to think of this less in terms of handling "errors", but rather to think about this as just a data structure that is communicating the current state of events. In our pre-checking, we have a handful of Checks we want to make, and we want to know a summary of which checks succeeded and which ones did not. The new `CheckResults` type is just a `Vec<Result<String, CheckError>>` which holds the results of _all_ of the checks. The important thing about this is that a single `CheckResults` is a perfect summary of all of the checks that have been run. Because of this, all of the command-line printing code was able to be extracted in to the `cli` module of the crate which is only compiled with the `features = ["cli"]` flag, minimizing our library's size and stdout-side-effects. Printing the status of the checks then just becomes a simple iteration over the `CheckResults`.

If you look, the `fluvio cluster check` command never actually "bubbles up" anything contained in this `CheckResults`, it is strictly used as a data structure that is iterated over and printed. In the `fluvio cluster start` command, it is also not "bubbled up". The `CheckResults` are now returned as a piece of data inside a new return struct `StartStatus` which is returned in the `Ok()` variant of the `ClusterInstaller::install_fluvio` and `LocalClusterInstaller::install` functions. It is just a description of the summary of the checks, and it is never bubbled up as an error, just looked at as a piece of data. 

I really think that the changes here are a big improvement. I think everything is laid out in a way that makes sense and works well and can be checked off and we can move on to the next thing.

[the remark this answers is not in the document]
sehz: Before, it was clear what was the purpose of `performCheck`.   It was supposed perform validation to ensure Cluster meet all requirements before starting.  It's tempting to use Error for everything since it make easier to consolidate it but in this case, it is not semantically correctly.   For example, if I check if status of the network and network is down, that is not error because  that is what code is supposed to handle.   What could be considered as error?  For example if network driver is not installed because that's not what supposed to happen.   Without clear separation, it will be hard to maintain the code

nicholastmosher:
Update: The check module now has the following datatypes

- `type CheckResult = Result<CheckStatus, CheckError>`: Represents either a successful check or a check which encountered an error and did not complete.
- `enum CheckStatus`: Represents a successfully-performed check which either `Pass`ed or `Fail`ed.
- `enum RecoverableCheck`: A successfully-performed check which has `Fail`ed but may be auto-recoverable
- `enum UnrecoverableCheck`: A successfully-performed check which has `Fail`ed and we can do nothing about
- `enum CheckError`: The classic error type which represents a check did not complete successfully for some reason

```rust
let result: CheckResult = todo!();
match result {
    Ok(CheckStatus::Pass(message)) => {}, // `message` is a success message
    Ok(CheckStatus::Fail(RecoverableCheck(e))) => {}, // `e` describes the recoverable check
    Ok(CheckStatus::Fail(UnrecoverableCheck(e))) => {}, // `e` describes unrecoverable check
    Err(CheckError) => {}, // The check failed to perform for some reason
}
```

src/cluster/src/start/local.rs
```diff
@@ -347,7 +349,7 @@ impl LocalClusterInstaller {
             .stderr(Stdio::from(errors))
             .spawn()?;
 
-        Ok(())
+        Ok("localhost:9003".to_string())
```
sehz, line 352: why hard coding here? this should be coming from local config or profile

sehz (changes requested):
All looks good except one minor issue with config

sehz (approved):

PR #557, opened 2020-12-08
Fix composite action when not in fluvio repo

simlay:
Yesterday in standup we discussed if there was any advantage to having the composite action in it's own repo and couldn't find a good one. This fixes fluvio cluster installer when you're not in the [root of the fluvio repo](https://github.com/infinyon/fluvio-client-node/pull/36/checks?check_run_id=1517989280).


nicholastmosher:
So this is for use-cases when you want to use the Fluvio action _from_ another repo, right?

simlay:
> So this is for use-cases when you want to use the Fluvio action _from_ another repo, right?

Correct.

sehz (approved):

PR #579, opened 2020-12-18
Updated the minikube driver to docker rather than none

simlay:
This is a partial fix to infinyon/fluvio-client-node#42 CI. This will also close #570.

Issue #563, opened 2020-12-14
Refactor composite action


simlay:
Not every use case of the fluvio composite action wants to install minikube. Let's either make this a separate composite action or make this as a parameter like `no_minikube_setup`.

Needs to be working and on CI for fluvio-cloud, the node client and this repo.

Expanding the scope a little, let's add k8s support for fluvio installer.

Do pretest of of composite action in releases.

After discussion, we will move the composite action tests to a different workflow that's not required on each PR.

PR #590, opened 2020-12-21
Bump version and publish a new release with CI

### LINKED ISSUE #568: Parse VERSION to determine if a github release is a pre-release
Currently, all github releases are set to pre-releases. We should move this parse the VERSION string to see if there's a suffix regex of `-(alpha|beta).[0-9]`, this is a pre-release, if this doesn't end with that regular expression, we should publish it as a full release.

simlay:
This uses the `fluvio package bump latest` in the release.  This closes #568 as well.


makefiles/S3.toml
```diff
@@ -39,6 +39,7 @@ script = '''
 mkdir -p ~/.fluvio/extensions
 ./target/${TARGET}/release/fluvio install fluvio/fluvio-package;
 ./target/${TARGET}/release/fluvio package publish ./target/${TARGET}/release/fluvio --version=${FLUVIO_VERSION};
+./target/${TARGET}/release/fluvio package bump latest;
```
simlay, line 42: There's a small chance of a race condition between the macos release build and the ubuntu release build. Ideally, this would be done after all of them are complete. Doing this correctly in CI is annoying because you gotta re-install fluvio :/. Maybe we wait for #579 and #563?

sehz: sure

sehz (approved):

Issue #560, opened 2020-12-14
regression: fluvio cluster check


sehz:
- [x] 1: Before, cluster check is used to tell how to fix but it is no longer  displaying detail which make it difficult to fix the issue (closed by #562)
- [x] 2: Check is no longer running displaying status one bye one but it is displaying only after all checks are running

```
Running pre-startup checks...
✅ ok: Kubernetes config is loadable
✅ ok: Supported kubernetes version is installed
✅ ok: Supported helm version is installed
❕ warning: Missing Fluvio system charts.
  💡 note: this may be fixed automatically during startup
✅ ok: Can create service
✅ ok: Can create customresourcedefinitions
✅ ok: Can create secret
❕ warning: Minikube tunnel not found
  💡 note: this may be fixed automatically during startup
Some checks failed, but may be auto-corrected
```

nicholastmosher:
I can fix 1) pretty easily, when working on it I had figured that any auto-fixable thing might be better to just let the `start` command take care of it but I'll put the diagnostics back.

With regards to 2), I didn't realize that displaying the results one at a time was a requirement of the checker. The reason we were able to do that before was because we didn't have a good binary/library separation boundary. The library was just freely printing to stdout, which IMO is bad library design. I know another pattern that we can use to restore the one-by-one printout capability, but if it's not a huge deal then I'd like to push that to a future issue. Let me know what you think.

nicholastmosher:
Closed by #596

PR #596, opened 2020-12-22
Render check progress incrementally as they are run

nicholastmosher:
This fixes most of #560. When running `fluvio cluster check` or `fluvio cluster start --local`, the checks will now be rendered as they run, rather than all at once in a batch at the end. Making this work for k8 will probably be slightly trickier so I didn't include it in this PR.

This change works by turning the `ClusterChecker` into more of a builder pattern, where you can load it up with particular checks you want to run. Then, you have the option to kick off the checks with one of these methods:

- `run`: Simply runs all of the checks in a batch without trying to fix anything, returns all results at once
- `run_and_fix`: Runs all of the checks in a batch, attempts to fix them, and returns all results at once
- `run_with_progress`: Returns immediately with an `async_channel::Receiver` that yields check results as they are performed
- `run_and_fix_with_progress`: Same as `run_with_progress`, but tries to apply fixes

For the Local installer, I added a config option `render_checks` which is disabled by default, but when it is explicitly enabled, the installer will directly render the checks on stdout (despite being printed from a library. I was ok with this since it's behind a config flag).

The k8 installer needs a bit of work to be well-structured enough to easily make this setup work. I think it's something we can put off until the future.


src/cluster/Cargo.toml
```diff
@@ -20,9 +20,9 @@ required-features = ["cli"]
 cli = [
```
sehz, line 20: bump up version for crate

src/cluster/src/check/mod.rs
```diff
@@ -379,16 +398,208 @@ impl ClusterChecker {
             Box::new(CreateServiceAccountPermission),
             Box::new(LoadBalancer),
```
sehz, line 399: Load Balancer is check is not needed for local.   So we need to tie this to cluster type.

src/cluster/src/check/mod.rs
```diff
@@ -379,16 +398,208 @@ impl ClusterChecker {
...
+        self.checks.extend(checks);
+        self
+    }
 
-        // Collect results from running checks
-        let mut results = vec![];
+    /// Adds all checks required for starting a cluster on minikube.
+    ///
+    /// Note that no checks are run until [`run`] or [`run_with_progress`].
+    pub fn with_minikube_checks(mut self) -> Self {
```
sehz, line 408: not sure what meant by `minikube_check`?  We should not tie to specific Kubernetes type.  We don't want to create different configuration for `eks`, `kind`,`aks`.   Name should be something like K8Installed or something like that to make it meaning clear

src/cluster/src/check/mod.rs
```diff
@@ -379,16 +398,208 @@ impl ClusterChecker {
...
+        ];
+        self.checks.extend(checks);
+        self
+    }
+
+    /// Adds all checks required for starting a local cluster.
+    ///
+    /// Note that no checks are run until [`run`] or [`run_with_progress`].
+    pub fn with_local_checks(mut self) -> Self {
+        let checks: Vec<Box<(dyn ClusterCheck)>> = vec![
```
sehz, line 424: Seems like this should be moved to local cluster install?

src/cluster/src/check/mod.rs
```diff
@@ -379,16 +398,208 @@ impl ClusterChecker {
...
+    /// ```no_run
+    /// # use fluvio_cluster::{ClusterChecker, CheckResults};
+    /// # async fn do_run() {
+    /// let check_results: CheckResults = ClusterChecker::empty()
+    ///     .with_preflight_checks()
+    ///     .run()
+    ///     .await;
+    /// # }
+    /// ```
+    pub async fn run(&self) -> CheckResults {
```
sehz, line 451: maybe changed this to `run_and_wait` or `run_and_collect` to make it clear this is synchronous or batch processing

src/cluster/src/check/mod.rs
```diff
@@ -379,16 +398,208 @@ impl ClusterChecker {
...
+    ///     .with_preflight_checks()
+    ///     .run_with_progress();
+    /// while let Ok(check_result) = progress.recv().await {
+    ///     println!("Got check result: {:?}", check_result);
+    /// }
+    /// # }
+    /// ```
+    pub fn run_with_progress(self) -> Receiver<CheckResult> {
+        let (sender, receiver) = async_channel::bounded(100);
+        spawn(async move {
```
sehz, line 530: any reason why this need to be spawn?

src/cluster/src/check/mod.rs
```diff
@@ -379,16 +398,208 @@ impl ClusterChecker {
...
+    ///
+    /// This function will return immediately with a channel which will yield
+    /// progress updates about checks and fixes as they run.
+    ///
+    /// If you want to run checks and fixes as a single batch and receive all of
+    /// the results at once, use [`run`] instead.
+    pub fn run_and_fix_with_progress<F, R>(self, fix: F) -> Receiver<CheckResult>
+    where
+        F: Fn(RecoverableCheck) -> R + Send + Sync + 'static,
+        R: Future<Output = Result<(), UnrecoverableCheck>> + Send + Sync,
```
sehz, line 556: this seems to be duplicate of `run_and_fix`.   this should be consolidated into a single method with only difference being what to do with result in between

src/cluster/src/check/render.rs
```diff
@@ -0,0 +1,191 @@
...
+            println!("✅ {} {}", "ok:".bold().green(), success.green());
+        }
+        Fail(e @ CheckFailed::AutoRecoverable(_)) => {
+            println!("❕ {} {}", "warning:".bold().yellow(), e);
+            println!(
+                "  💡 {} this may be fixed automatically during startup",
+                "note:".bold()
+            );
+        }
+        Fail(CheckFailed::AlreadyInstalled) => {
```
sehz, line 66: Why is `AlreadyInstalled` part of CheckFailed.  Seems like `AutoRecoverable` and `AlreadyInstalled` doesn't belong together?

sehz (changes requested):

src/cluster/src/check/render.rs
nicholastmosher: This code was already like this, it's only showing up in the diff because I moved it. If we want to we can open another issue to talk about the organization of it though

src/cluster/src/check/mod.rs
nicholastmosher: The way I see it, there are two orthogonal concerns:

- Just Check
  - Batch (`run`)
  - Progress (`run_with_progress`)
- Check and Fix
  - Batch (`run_and_fix`)
  - Progress (`run_and_fix_with_progress`)

So the two concerns are a) whether to fix the checks and b) whether to report progress as they go.

I wanted to solve each of these generally so that it would work for either a local install or a k8 install.

For a), I abstracted away the idea of a fixing-function. So the caller of the checker will provide their own function that describes how to take a "failed" `RecoverableCheck` and attempt to fix it. That's the `fix: F` function argument. All of the generics basically say that the fix function may be an async function.

For b), I wanted to maintain the library/binary separation while also providing a way for the binary half (i.e. the CLI) to be receiving updates from checks being performed by code in the library half. To do this, I made it so that the `with_progress` functions immediately return with an `async_channel::Receiver<CheckResult>`. Inside the `with_progress` function, it spawns a task which is what actually performs the checks, and it sends the results through the channel as it goes. So the CLI can simply watch for new `CheckResult`s to come through the channel and print them to stdout.

src/cluster/src/check/mod.rs
nicholastmosher: Yes, this has to do with binary/library separation. `run_with_progress` now returns immediately so that the CLI does not need to wait for all checks to complete before regaining control. The checks are performed inside the spawned task, and as they complete, they are sent to the caller via the Receiver that was given.

src/cluster/src/check/mod.rs
nicholastmosher: I don't have a problem naming this `with_k8_check` or something similar. This was just distinguishing between local and k8, and some of the checks happened to be specialized to minikube so I thought it made sense. (e.g. the minikube tunnel check)

src/cluster/src/check/mod.rs
nicholastmosher: This is just a convenience method to load up local checks into the checker. It makes it nicer for the caller, e.g. the LocalClusterInstaller. Callers can then just do something like

```rust
ClusterChecker::empty()
    .with_local_checks()
    .run_and_fix(fix)
    .await
```

src/cluster/src/check/mod.rs
nicholastmosher: I believe this was already being done before this PR. Can we move this to a future issue?

src/cluster/src/check/mod.rs
sehz: ok. that's good justification

src/cluster/src/check/mod.rs
sehz: How does this communicate end of progress?

sehz:
rebase.  there are 2 issues to be address

sehz (approved):

PR #602, opened 2020-12-23
Change SPU data dir to /var/lib/fluvio/data

### LINKED ISSUE #539: [Bug]:  SPU on K8 uses /tmp/fluvio as data dir
On K8, SPU runs with `--log-base-dir /tmp/fluvio` so its not using the PV mounted at `/var/lib/fluvio/data`

**Describe the setup**
Minikube / Fluvio Cloud

**How to reproduce it**
1. Install cluster to minkube
2. Run `kubectl get pod flv-spg-main-0 -o yaml`

Container args shows `log-base-dir` value. Should be `/var/lib/fluvio/data`, not `/tmp/fluvio`.

nicholastmosher:
Closes #539.

I found two places where the data directory was being hardcoded as `/tmp`, so I replaced those with references to `SPU_LOG_BASE_DIR` from the defaults module, then I updated that default to point to `/var/lib/fluvio/data`.

I have not been able to test this out locally to check if I have broken anything else because I can't run `make minikube_image` on mac as of right now.


nicholastmosher:
Ok I was just able to confirm that the `--log-data-dir` does in fact change to `/var/lib/fluvio/data` like we want. Now I just have to figure out why CI is failing...

```
apiVersion: v1
kind: Pod
metadata:
  creationTimestamp: "2020-12-23T15:37:34Z"
  generateName: fluvio-spg-main-
  labels:
    app: spu
    controller-revision-hash: fluvio-spg-main-5977dc488c
    group: main
    statefulset.kubernetes.io/pod-name: fluvio-spg-main-0
...
  name: fluvio-spg-main-0
  namespace: default
...
spec:
  containers:
  - args:
    - /fluvio
    - run
    - spu
    - --sc-addr
    - fluvio-sc-internal.default.svc.cluster.local:9004
    - --log-base-dir
    - /var/lib/fluvio/data
    - --log-size
    - 1Gi
...
```

sehz (changes requested):
Looks good.  Can you bump up version number for effected crates?

sehz (approved):

PR #603, opened 2020-12-23
Add cargo-make task for minikube-image

nicholastmosher:
I've added some cargo-make tasks that mirror other ones that have worked well for us. I'm able to successfully build the musl images on Mac.

One thing I'm stuck on though is that `fluvio cluster start --develop` does not seem to be picking up on these new images, I'm getting `ErrImagePull` on the pods that come up. @nacardin is there anything extra that needs doing to these docker images for them to be seen?

### DIFF
--- a/makefiles/Build.toml
+++ b/makefiles/Build.toml
@@ -1,0 +1,24 @@
+extend = "Base.toml"
+
+[tasks.build-dev-musl]
+condition = { profiles = ["development"] }
+command = "cargo"
+args = ["build", "--target", "x86_64-unknown-linux-musl"]
+dependencies = ["set-musl-env", "rustup-add-target-musl"]
+
+[tasks.build-release-musl]
+condition = { profiles = ["production"] }
+command = "cargo"
+args = ["build", "--release", "--target", "x86_64-unknown-linux-musl"]
+dependencies = ["set-musl-env", "rustup-add-target-musl"]
+
+[tasks.rustup-add-target-musl]
+command = "rustup"
+args = ["target", "add", "x86_64-unknown-linux-musl"]
+
+[tasks.set-musl-env]
+condition = { platforms = ["mac"] }
+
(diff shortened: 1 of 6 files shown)


nacardin:
Should we remove duplicate functionality from the GNU makefile? Maybe we can remove `fluvio_bin_linux` and point `fluvio_image` and `minikube_image` to run `cargo make`

sehz:
Yes, we should remove duplication.  But need to make sure we don't have regression in performance and other workflows.  The linux build should be using `cargo build` as before.   Other non-unix platform can use docker based cross-compiler

nicholastmosher:
@sehz this actually is not using the `cross build` tool that uses Docker images for compilation. This is using plain-old cargo but it has all the right configurations to make the musl target build properly. I will make sure we have a way to invoke build for the particular targets we want

nicholastmosher:
@nacardin I just briefly looked through the Makefile looking for isolated pieces that I could pull out, but I realized that it's pretty tangled up right now and I don't trust myself to modify it without any regressions without a lot of careful work. I think we already have issues open for migrating to cargo-make, maybe we can leave that work for those issues in the future.

nicholastmosher:
@sehz and @nacardin is there anything you would like to see changed here before merging?

sehz:
rebase

makefiles/Build.toml
```diff
@@ -0,0 +1,24 @@
+extend = "Base.toml"
```
sehz, line 1: is this being used? if not, we should merge this later when this is integrated.  Other changes looks good

sehz (changes requested):

sehz:
rebase

sehz:
Is this still relevant?

nicholastmosher:
I think everything we needed from this has been taken care of elsewhere

PR #604, opened 2020-12-23
Composite action refactor

### LINKED ISSUE #563: Refactor composite action

simlay:
Closes #563

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -366,8 +366,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - uses: ./
-        name: Install Fluvio Local Cluster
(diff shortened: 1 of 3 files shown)


sehz:
is still relevant?

simlay:
> is still relevant?

Given that we're having issues with stability for just the local cluster in CI, this PR adds k8s clusters which makes it just too much more unstable.

Let's close for now.

PR #613, opened 2020-12-28
Fix Display impl for PackageId<MaybeVersion>

nicholastmosher:
This fixes an itty bitty bug that I found having to do with the `fluvio install` command.

When running something like `fluvio install fluvio/fluvio-cloud`, I get this output:

```
🎣 Fetching latest version for package: fluvio/fluvio-cloud:...
⏳ Downloading package with latest version: fluvio/fluvio-cloud:0.1.0...
```

This is _almost_ what we want, except there is a trailing `:` after the `fluvio/fluvio-cloud` on the first line of output

image

This is because the `Display` impl of `PackageId<MaybeVersion>` was _just barely_ wrong. So I fixed it and added a test to make sure it prints as expected. Now we have:

```
🎣 Fetching latest version for package: fluvio/fluvio-cloud...
⏳ Downloading package with latest version: fluvio/fluvio-cloud:0.1.0...
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1124,7 +1124,7 @@ dependencies = [
 
 [[package]]
 name = "fluvio-package-index"
-version = "0.2.0"
+version = "0.2.1"
 dependencies = [
  "http-types",
  "lazy_static",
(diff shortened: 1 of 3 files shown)


sehz (approved):

sehz:
rebase

PR #615, opened 2020-12-29
update version to beta2

sehz:


RELEASE.md
```diff
@@ -0,0 +1,16 @@
+# Release process
+
+bump up version `VERSION` file
+
+## Release crates
+
+Find crates that has been changed last release
+```
+cargo ws changed
```
simlay, line 9: Is this short for https://crates.io/crates/cargo-workspaces#changed?

sehz: yup

sehz: it would be handy if there are other tools that can show diff between published and what is in the workspace

simlay (approved):
:+1:

PR #599, opened 2020-12-22
Adding force dispatch argument to release CI

simlay:
Defaults to nothing, forces a new release only if ran with `--force`.

PR #616, opened 2020-12-29
Used script over command in cargo-make

simlay:
This is a fix to deal with non-forced releases as seen like this one https://github.com/infinyon/fluvio/actions/runs/449824310.

Let's wait until we're ready for another release before we merge this.

### DIFF
--- a/makefiles/Github.toml
+++ b/makefiles/Github.toml
@@ -27,24 +27,13 @@ dependencies = [
     "build-release",
     "github-release-create-pre-release",
 ]
-command = "github-release"
-args = [
-    "-v",
-    "upload",
-    "--user", "infinyon",
-    "--repo", "fluvio",
-    "--tag", "v${FLUVIO_VERSION}",
-    "--name", "fluvio-v${FLUVIO_VERSION}-${TARGET}",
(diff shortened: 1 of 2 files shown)


sehz (approved):

makefiles/Github.toml
```diff
@@ -27,24 +27,9 @@ dependencies = [
...
-    "--tag", "v${FLUVIO_VERSION}",
-    "--name", "fluvio-v${FLUVIO_VERSION}-${TARGET}",
-    "--file", "./target/${TARGET}/release/fluvio",
-    "${GITHUB_RELEASE_FORCE}",
-]
-env_scripts = [
-'''
-if [ "${FORCE_RELEASE}" = "--force" ]; then
-    set_env GITHUB_RELEASE_FORCE "--replace"
-fi
```
nicholastmosher, line 45: What did this do? Are we ok without it?

simlay: This wasn't working right anyway. I should have removed it in #600. I think I missed it because #599 was merged too early. Anyway, the `GITHUB_RELEASE_FORCE` environment variable is set in the release.yaml.

nicholastmosher (approved):

PR #621, opened 2020-12-30
Fluvio self update using rename

### LINKED ISSUE #617: CLI: self-update doesn't work
Steps produce it:
Install current release or previous release.
Try to perform self update:
```
which fluvio
/home/ubuntu/.fluvio/bin/fluvio
ubuntu@ip-172-31-20-82:~$ fluvio update
🎣 Fetching latest version for fluvio/fluvio...
⏳ Downloading Fluvio CLI with latest version: fluvio/fluvio:0.6.0-beta.2...
Error: 
   0: Text file busy (os error 26)
```

nicholastmosher:
Closes #617.

The error was caused by the fact that I was directly opening the executable's current file and trying to write to it. This fix causes the updater to write the update to a new temporary file, then `rename` that file to the path of the original executable. The original executable remains available on disk for as long as the executing process needs it, but the path becomes inhabited by the newly updated executable immediately. On unix this is actually an atomic operation, so at no point is there an inconsistent state (e.g. if some process tried to execute `fluvio` while we were halfway through writing it before).

This PR is rebased on top of #603 so that I could test the musl build on a docker ubuntu image. We can either merge #603 first or merge this one to supersede it.

### DIFF
--- a/src/cli/src/install/mod.rs
+++ b/src/cli/src/install/mod.rs
@@ -122,10 +122,7 @@ fn verify_checksum<B: AsRef<[u8]>>(buffer: B, checksum: &str) -> bool {
     &*buffer_checksum == checksum
 }
 
-pub fn install_bin<P: AsRef<Path>, B: AsRef<[u8]>>(
-    bin_path: P,
-    bytes: B,
-) -> Result<(), CliError> {
+pub fn install_bin<P: AsRef<Path>, B: AsRef<[u8]>>(bin_path: P, bytes: B) -> Result<(), CliError> {
     use std::io::Write as _;
 
     let bin_path = bin_path.as_ref();
--- a/src/cli/src/install/mod.rs
+++ b/src/cli/src/install/mod.rs
@@ -25,12 +25,6 @@ fn fluvio_base_dir() -> Result<PathBuf, CliError> {
     Err(IoError::new(ErrorKind::NotFound, "Fluvio base directory not found").into())
 }
 
-pub(crate) fn fluvio_bin_dir() -> Result<PathBuf, CliError> {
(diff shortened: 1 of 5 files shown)


src/cli/src/install/mod.rs
```diff
@@ -122,24 +116,27 @@ fn verify_checksum<B: AsRef<[u8]>>(buffer: B, checksum: &str) -> bool {
...
+    // Create directories to bin_path if they do not exist
+    if let Some(parent) = bin_path.parent() {
+        std::fs::create_dir_all(&parent)?;
+    }
+
+    // Write bin to temporary file
+    let tmp_dir = tempdir::TempDir::new("fluvio")?;
+    let tmp_path = tmp_dir.path().join("fluvio");
+    let mut tmp_file = File::create(&tmp_path)?;
+    tmp_file.write_all(bytes.as_ref())?;
```
sehz, line 133: shouldn't use https://doc.rust-lang.org/std/fs/fn.copy.html ?

sehz (changes requested):

src/cli/src/install/mod.rs
nicholastmosher: This is not the same update strategy that we started out with. Originally, we were going to have the Fluvio CLI copy itself to another file and then invoke that file. Now, instead, we are simply writing the new downloaded file to a temporary directory and then renaming it to "overwrite" the original executable. This would be equivalent to `mv old new`. Importantly, it doesn't actually write over the old (currently executing) executable (it still has the same file descriptor, but now the other file has the file _name_).

simlay (approved):
lgtm

sehz (approved):

PR #630, opened 2021-01-04
Fix self-update when /tmp is mounted on a different fs

nicholastmosher:
`std::fs::rename` relies on the source and destination filesystems being the same. This falls apart if, say `/tmp` is mounted on a different partition. This change switches to use `std::fs::copy`, which works no matter the fs setup. Since our temp file is being created inside a `tempfile` (from tempfile crate), then the temp directory and contained files will be cleaned up no matter what.

### DIFF
--- a/src/cli/src/install/mod.rs
+++ b/src/cli/src/install/mod.rs
@@ -128,7 +128,7 @@ pub fn install_bin<P: AsRef<Path>, B: AsRef<[u8]>>(bin_path: P, bytes: B) -> Res
     std::fs::create_dir_all(&parent)?;
 
     // Write bin to temporary file
-    let tmp_dir = tempdir::TempDir::new_in(parent, "fluvio")?;
+    let tmp_dir = tempdir::TempDir::new_in(parent, "fluvio-tmp")?;
     let tmp_path = tmp_dir.path().join("fluvio");
     let mut tmp_file = File::create(&tmp_path)?;
     tmp_file.write_all(bytes.as_ref())?;
--- a/src/cli/src/install/mod.rs
+++ b/src/cli/src/install/mod.rs
@@ -122,7 +122,8 @@ pub fn install_bin<P: AsRef<Path>, B: AsRef<[u8]>>(bin_path: P, bytes: B) -> Res
     let bin_path = bin_path.as_ref();
 
     // Create directories to bin_path if they do not exist
-    let parent = bin_path.parent()
+    let parent = bin_path
+        .parent()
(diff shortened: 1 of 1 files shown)


src/cli/src/install/mod.rs
```diff
@@ -122,12 +122,12 @@ pub fn install_bin<P: AsRef<Path>, B: AsRef<[u8]>>(bin_path: P, bytes: B) -> Res
     let bin_path = bin_path.as_ref();
 
     // Create directories to bin_path if they do not exist
-    if let Some(parent) = bin_path.parent() {
-        std::fs::create_dir_all(&parent)?;
-    }
+    let parent = bin_path.parent()
+        .ok_or_else(|| IoError::new(ErrorKind::NotFound, "parent directory not found"))?;
+    std::fs::create_dir_all(&parent)?;
```
nacardin, line 128: How could the `parent` not already exist?

src/cli/src/install/mod.rs
```diff
@@ -122,12 +122,12 @@ pub fn install_bin<P: AsRef<Path>, B: AsRef<[u8]>>(bin_path: P, bytes: B) -> Res
...
 
     // Write bin to temporary file
-    let tmp_dir = tempdir::TempDir::new("fluvio")?;
+    let tmp_dir = tempdir::TempDir::new_in(parent, "fluvio")?;
```
nacardin, line 131: Should the temp dir be named something that indicated that it is a temp dir for updating, such as `update_tmp`?

nacardin (commented):
Two questions

src/cli/src/install/mod.rs
nicholastmosher: I cannot think of an example, but `Path::parent` returns an `Option<&Path>`, so this is just how I went about dealing with it. I imagine this code path will never be taken

src/cli/src/install/mod.rs
nicholastmosher: Possibly? I don't really have a preference. Perhaps we can do something like `fluvio-tmp` so people know it's fluvio-related even if their binary is somewhere else (e.g. `/usr/local/lib` or something)

src/cli/src/install/mod.rs
nicholastmosher: > Returns [None] if the path terminates in a root or prefix.

nicholastmosher: So perhaps if somebody executed `fluvio` in the root directory of a container or something...

sehz (approved):

PR #632, opened 2021-01-05
add cluster installer to use cluster ip

sehz:
Add option to use `cluster_ip` instead of external LB address.   This allows installer not to wait for LB which make take while in some cloud. 
Fixed issue where k8 installer doesn't handle namespace correctly.


nacardin:
It's unnecessary to run `wait_for_sc_dns` if using cluster IP

sehz:
it's still useful to confirm that service is up and and you can connect

PR #638, opened 2021-01-06
Add 'fluvio profile rename' to rename profiles

### LINKED ISSUE #542: [CLI] Add option to rename profiles
Right now profiles are created automatically by the CLI and there is no way (other than manually editing `~/.fluvio/config`) to rename the profiles that exist. We should add an option that allows you to override the name of the profile when it is created.

Profile creation instances:

- Local profile is created
- Profile is synced directly with a cluster (k8, local)
- Profile is synced from cloud

For any existing profile, we should be able to rename it, similarly to `git remote rename`

- `fluvio profile rename <from> <to>`
  - Will rename profile `from` into a new profile called `to`

For each profile creation instance, we should be able to provide a name flag (e.g. `profile sync local --profile-name=<name>`)

When syncing with cloud, the profile name should default to the email account name, with an option to override it

- `fluvio cloud login --email=<name@email.com> --password=<password>`
  - If a profile called `name` already exists, gracefully reject and notify that the profile name exists
  - Will create a profile named `name`
- `fluvio cloud login --email=<email> --password=<password> --profile-name=foo`
  - Will create a profile named `foo`

nicholastmosher:
Closes #542 

Adds a new CLI command:

```
fluvio profile rename <from> <to>
```

which renames a profile called `from` to a new profile called `to`. If the old profile name was marked as the current profile, the "current profile" name will be updated to match the renamed profile, so this command will not change any other behavior.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -883,7 +883,7 @@ dependencies = [
 
 [[package]]
 name = "fluvio"
-version = "0.3.2"
+version = "0.3.3"
 dependencies = [
  "async-channel",
  "async-mutex",
--- a/src/cli/src/profile/mod.rs
+++ b/src/cli/src/profile/mod.rs
@@ -10,6 +10,7 @@ use structopt::StructOpt;
 mod sync;
 mod current;
 mod switch;
+mod rename;
 mod delete_profile;
 mod delete_cluster;
 mod view;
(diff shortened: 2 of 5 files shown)


sehz (changes requested):
Great.  Can you bump up version for client?

sehz (approved):

sehz:
rebase

PR #642, opened 2021-01-07
Add Examples to CI

### LINKED ISSUE #641: Add Rust client examples to CI
We have a handful of client code examples in the `examples/` directory, we should add a CI job for those to be built and run.

nicholastmosher:
Closes #641

These examples (00-produce and 01-consume) mirror the front-page code snippets on fluvio.io. This PR adds them to CI so that we are sure that the front-page code is always good


examples/00-produce/src/main.rs
```diff
@@ -0,0 +1,12 @@
+fn main() {
```
simlay, line 1: I'm a can of the `#[main]` proc macro from `async_std`.

sehz:
rebase

.github/workflows/ci.yml
```diff
@@ -251,6 +251,10 @@ jobs:
         timeout-minutes: 5
         run: |
           make RELEASE=true UNINSTALL=noclean smoke-test-tls-root
+      - name: Test examples
+        run: |
+          examples/tests/test-simple.sh
+          examples/tests/test-echo.sh
```
sehz, line 257: add timeout parameters to limit runaway tasks

nicholastmosher: Hmm I tried adding timeouts to the `test-simple.sh` script and I tried adding timeouts directly to the `echo` example... I'm trying to figure out what's going on

sehz: add `timeout-minutes` in the GH Action run which terminate task if exceeds allotted time.  See previous run

.github/workflows/ci.yml
```diff
@@ -258,6 +258,11 @@ jobs:
         timeout-minutes: 5
         run: |
           make RELEASE=true UNINSTALL=noclean smoke-test-tls-root
+      - name: Test examples
+        timeout-minutes: 10
```
sehz, line 262: can you reduce to 5 minutes? unless there is reason for bigger delay

.github/workflows/ci.yml
```diff
@@ -258,6 +258,11 @@ jobs:
...
+        run: |
+          examples/tests/test-simple.sh --release
+          examples/tests/test-echo.sh --release
```
sehz, line 265: why is there duplicate run after smoke test?

examples/02-echo/Cargo.toml
```diff
@@ -9,5 +9,5 @@ publish = false
 
 [dependencies]
 fluvio = { path = "../../src/client" }
-futures-lite = "1.10.1"
+futures = "0.3.9"
```
sehz, line 12: is there reason putting 0.3.9? should put lowest bound much as possible

examples/tests/test-echo.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -e
+
+RELEASE=$1
+
+cargo run $RELEASE --bin fluvio -- topic delete echo || true
```
sehz, line 7: I see that why `cargo run` is used for compilation.
I would prefer that to put in explicit either explicit `cargo build` or in the Makefile to clear that it is for build step

sehz (changes requested):

.github/workflows/ci.yml
nicholastmosher: One is Mac and one is Linux

examples/02-echo/Cargo.toml
nicholastmosher: Why lowest? Don't you typically choose latest version for dependencies?

examples/tests/test-echo.sh
nicholastmosher: Do you mean in the test script use `cargo build` followed by `cargo run`? Or do you mean to take the `cargo build` out of the test script all together?

examples/02-echo/Cargo.toml
sehz: No, unless you have explicit requirements for that release (bug fixes, new API,etc).   If you put higher version number, you are imposing more requirements for user of create which force them to upgrade.

examples/tests/test-echo.sh
sehz: If you look at CI steps, it is split into build and test in order take advantage of SCCACHE.   If you try to build and test in a single step, you may run into timeout issue.   We don't want to have timeout for build because it may vary but running should be bounded

.github/workflows/ci.yml
```diff
@@ -315,7 +320,17 @@ jobs:
       - name: Run smoke-test
         timeout-minutes: 5
         run: |
-          make RELEASE=true UNINSTALL=noclean smoke-test-tls-root
+          FLV_CMD=true FLV_SOCKET_WAIT=600 make RELEASE=true UNINSTALL=noclean smoke-test-tls-root
```
sehz, line 323: Can you move environment variable into ENV section at the beginning of job? Can look at other job as example:

.github/workflows/ci.yml
```diff
@@ -315,7 +320,17 @@ jobs:
...
+      - name: Build examples
+        run: |
+          cargo build --release --bin produce
```
sehz, line 326: Can we build this a single step? Every time, we add another example, we don't want to change the CI.

.github/workflows/ci.yml
```diff
@@ -258,6 +258,11 @@ jobs:
         timeout-minutes: 5
         run: |
           make RELEASE=true UNINSTALL=noclean smoke-test-tls-root
+      - name: Test examples
+        timeout-minutes: 5
+        run: |
+          examples/tests/test-simple.sh --release
```
sehz, line 264: Can you change this into a single run?

.github/workflows/ci.yml
```diff
@@ -315,7 +320,17 @@ jobs:
...
+          FLV_CMD=true FLV_SOCKET_WAIT=600 make RELEASE=true UNINSTALL=noclean smoke-test-tls-root
+      - name: Build examples
+        run: |
+          cargo build --release --bin produce
+          cargo build --release --bin consume
+          cargo build --release --bin echo
+      - name: Test examples
+        timeout-minutes: 5
+        run: |
+          examples/tests/test-simple.sh --release
```
sehz, line 332: change into a single run

sehz (changes requested):
just need minor changes  to simplify CI

sehz (approved):

PR #654, opened 2021-01-09
Added FluvioInner to Fluvio and dependent structs.

### LINKED ISSUE #645: Creating a consumer via fluvio::consumer results in PartitionNotFound
Using [`fluvio::consume`](https://github.com/infinyon/fluvio/blob/769ba7a8b3f72230abf753c91425ed315e18a58c/src/client/src/lib.rs#L158) results in `PartitionNotFound` errors for topics that already exist however this doesn't happen if you use the exact same code from within `fluvio::consume`.

This for example fails:

```rust
use fluvio::Offset;
const TOPIC: &str = "my-topic";

#[async_std::main]
async fn main() -> Result<(), Box<dyn std::error::Error>> {
    let consumer = fluvio::consumer(TOPIC, 0).await?;
    let _ = consumer.stream(Offset::beginning()).await?;

    Ok(())
}
```
However this succeeds:

```rust
use fluvio::Offset;
const TOPIC: &str = "my-topic";

#[async_std::main]
async fn main() -> Result<(), Box<dyn std::error::Error>> {

    let client = fluvio::Fluvio::connect().await?;
    let consumer = client.partition_consumer(TOPIC, 0).await?;
    let _ = consumer.stream(Offset::beginning()).await?;

    Ok(())
}
```
I've failed at finding any tools to help show when a given variable is dropped and the deconstructor is called. But the same error happens if you do the following:
```rust
let consumer = fluvio::Fluvio::connect().await?.partition_consumer(TOPIC, 0).await?;
```
I think this happens because `Fluvio::connect().await?` returns a `Fluvio` instance and `partition_consumer` uses it but returns a `PartitionConsumer`, I'm pretty sure a little mini scope is created here to the effect of:
```rust
let consumer = {
    let client = fluvio::Fluvio::connect().await?;
    client.partition_consumer(TOPIC, 0).await?
};
```
`client` is dropped at the ending of the `}`

This is all a long winded way of saying "I think that when the Fluvio client is dropped, it breaks things".

simlay:
Closes #645.

### DIFF
--- a/src/client/src/client/cluster.rs
+++ b/src/client/src/client/cluster.rs
@@ -84,7 +84,7 @@ impl Fluvio {
     fn spu_pool(&self) -> Result<Arc<SpuPool>, FluvioError> {
         self.spu_pool
             .get_or_try_init(|| -> Result<Arc<SpuPool>, FluvioError> {
-                let pool = run_block_on(SpuPool::start(self.config.clone(), &self.socket));
+                let pool = run_block_on(SpuPool::start(self.config.clone(), self.socket.clone()));
                 Ok(Arc::new(pool?))
             })
             .map(|pool| pool.clone())
--- a/src/client/src/spu/spu_pool.rs
+++ b/src/client/src/spu/spu_pool.rs
@@ -62,7 +62,7 @@ impl SpuPool {
     /// start synchronize based on pool
     pub async fn start(
         config: ClientConfig,
-        sc_socket: &AllMultiplexerSocket,
+        sc_socket: SharedAllMultiplexerSocket,
     ) -> Result<Self, FlvSocketError> {
         let metadata = MetadataStores::start(sc_socket).await?;
         debug!("starting spu pool");
--- a/src/client/src/sync/store.rs
+++ b/src/client/src/sync/store.rs
@@ -2,7 +2,7 @@ use std::sync::Arc;
 
 use tracing::debug;
 
-use fluvio_socket::AllMultiplexerSocket;
+use fluvio_socket::SharedAllMultiplexerSocket;
 use fluvio_socket::FlvSocketError;
 
 use crate::metadata::spu::SpuSpec;
@@ -17,19 +17,21 @@ pub struct MetadataStores {
     shutdown: Arc<SimpleEvent>,
     spus: StoreContext<SpuSpec>,
     partitions: StoreContext<PartitionSpec>,
+    socket: SharedAllMultiplexerSocket,
 }
 
 impl MetadataStores {
     /// start synchronization
-    pub async fn start(socket: &AllMultiplexerSocket) -> Result<Self, FlvSocketError> {
+    pub async fn start(socket: SharedAllMultiplexerSocket) -> Result<Self, FlvSocketError> {
         let store = Self {
             shutdown: SimpleEvent::shared(),
             spus: StoreContext::new(),
             partitions: StoreContext::new(),
+            socket: socket.clone(),
         };
 
-        store.start_watch_for_spu(socket).await?;
-        store.start_watch_for_partition(socket).await?;
+        store.start_watch_for_spu().await?;
+        store.start_watch_for_partition().await?;
 
         Ok(store)
     }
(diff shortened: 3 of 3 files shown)


src/client/src/client/cluster.rs
```diff
@@ -20,10 +18,23 @@ use super::*;
...
+}
+pub(crate) struct FluvioInner {
+    pub socket: SharedAllMultiplexerSocket,
+    pub config: ClientConfig,
+    pub versions: Versions,
+    pub spu_pool: Arc<SpuPool>,
+}
+impl Drop for Fluvio {
+    fn drop(&mut self) {
+        println!(
```
simlay, line 31: remove.

src/client/src/spu/spu_pool.rs
```diff
@@ -62,7 +62,7 @@ impl SpuPool {
     /// start synchronize based on pool
     pub async fn start(
         config: ClientConfig,
-        sc_socket: &AllMultiplexerSocket,
+        sc_socket: SharedAllMultiplexerSocket,
```
nicholastmosher, line 65: I believe this set of changes was only helping because it just happened to keep the Arc with the AllMultiplexerSocket alive. Since it is now held in FluvioInner, this should not be necessary

simlay: Yeah, I put this back to what it was.

src/client/src/client/cluster.rs
```diff
@@ -71,23 +72,14 @@ impl Fluvio {
         let (socket, config, versions) = inner_client.split();
         let socket = AllMultiplexerSocket::shared(socket);
 
-        let spu_pool = OnceCell::new();
-        Ok(Self {
+        let spu_pool = Arc::new(SpuPool::start(config.clone(), &socket).await?);
+        let inner = Arc::new(FluvioInner {
```
nicholastmosher, line 76: Since `Fluvio` holds `FluvioInner` in an Arc, we probably do not need `Arc<SpuPool>` inside of that as well

nicholastmosher: Other than that everything looks good

nicholastmosher: I realize that may not have been clear. Instead of `Arc<SpuPool>` in `FluvioInner` we can have just `SpuPool`

simlay: :+1:

src/client/src/client/cluster.rs
```diff
@@ -71,23 +72,14 @@ impl Fluvio {
         let (socket, config, versions) = inner_client.split();
         let socket = AllMultiplexerSocket::shared(socket);
 
-        let spu_pool = OnceCell::new();
-        Ok(Self {
+        let spu_pool = SpuPool::start(config.clone(), &socket).await?;
```
sehz, line 75: restore lazy start since we don't want need to start spu pool only when consumer/producer requested it

src/client/src/consumer.rs
```diff
@@ -48,15 +48,15 @@ use crate::spu::SpuPool;
 pub struct PartitionConsumer {
     topic: String,
     partition: i32,
-    pool: Arc<SpuPool>,
+    inner: Arc<FluvioInner>,
```
sehz, line 51: Consumer should not directly reference Fluvio which is connection to SC after looking up.  Instead we should create wrapper for SpuPool which maintain ref to Fluvio

sehz (changes requested):

src/client/src/consumer.rs
simlay: Well, part of the reason that this is needed is because the `SharedAllMultiplexerSocket` passed into `SpuPool` isn't kept. So, when the functions return, the `socket : Arc<AllMultiplexerSocket>`s go [out of scope in something like `SpuPool::start`](https://github.com/infinyon/fluvio/blob/51360927a2efc067d7015c3a186028b8300cc3d3/src/client/src/spu/spu_pool.rs#L63-L74) they drop which causes the [`self.terminate.notify` call in `MultiplexerSocket::drop()`](https://github.com/infinyon/fluvio-socket/blob/f537ef106530f74a2d0f9a772e9594dffc7f731f/crates/socket/src/multiplexing.rs#L81) which is an event to [the `MultiPlexingResponseDispatcher` to close the socket.](https://github.com/infinyon/fluvio-socket/blob/f537ef106530f74a2d0f9a772e9594dffc7f731f/crates/socket/src/multiplexing.rs#L357)

Another way to do this is to pass in owned instances of `SharedAllMultiplexerSocket`s rather than [`&AllMultiplexerSockets`](https://github.com/infinyon/fluvio/blob/51360927a2efc067d7015c3a186028b8300cc3d3/src/client/src/spu/spu_pool.rs#L65).

@nicholastmosher Thoughts?

sehz:
rebase.

sehz (approved):

Issue #37, opened 2020-06-02
SPU: implement delay flush


sehz:
Currently every write incur flush call.   Delay flush based on following criteria
* time
* number of writes or bytes transferred

Bonus: should be configurable

sehz:
Closing.  Thanks @digikata

PR #669, opened 2021-01-11
Implement flush on counted writes from #37

digikata:
Implement an option of flush policy of flushing every n writes, default is still to flush every write, unit test coverage tests other cases. Issue #37

### DIFF
--- a/src/storage/src/mut_records.rs
+++ b/src/storage/src/mut_records.rs
@@ -24,15 +24,37 @@ use crate::records::FileRecords;
 
 pub const MESSAGE_LOG_EXTENSION: &str = "log";
 
+/// MutFileRecordsFlushPolicy describes and implements a flush policy
+pub enum MutFileRecordsFlushPolicy {
+    NoFlush,
+    EveryWrite,
+    CountWrites { n_writes: u32, write_tracking: u32 },
+}
+
 /// Can append new batch to file
 pub struct MutFileRecords {
     base_offset: Offset,
     item_last_offset_delta: Size,
     f_sink: BoundedFileSink,
+    flush_policy: MutFileRecordsFlushPolicy,
+    write_count: u64,
     path: PathBuf,
 }
 
 impl Unpin for MutFileRecords {}
+
+fn get_flush_policy_from_config(option: &ConfigOption) -> MutFileRecordsFlushPolicy {
+    if option.flush_write_count == 0 {
+        MutFileRecordsFlushPolicy::NoFlush
+    } else if option.flush_write_count == 1 {
+        MutFileRecordsFlushPolicy::EveryWrite
+    } else {
+        MutFileRecordsFlushPolicy::CountWrites {
+            n_writes: option.flush_write_count,
+            write_tracking: 0,
+        }
+    }
+}
 
 impl MutFileRecords {
     pub async fn create(
@@ -48,6 +70,8 @@ impl MutFileRecords {
         Ok(MutFileRecords {
             base_offset,
             f_sink,
+            flush_policy: get_flush_policy_from_config(option),
+            write_count: 0,
             item_last_offset_delta: 0,
             path: log_path.to_owned(),
         })
@@ -68,6 +92,8 @@ impl MutFileRecords {
         Ok(MutFileRecords {
             base_offset,
             f_sink,
+            flush_policy: get_flush_policy_from_config(option),
+            write_count: 0,
             item_last_offset_delta: 0,
             path: log_path.to_owned(),
         })
@@ -98,9 +124,10 @@ impl MutFileRecords {
         if self.f_sink.can_be_appended(buffer.len() as u64) {
             debug!("writing {} bytes at: {}", buffer.len(), self.path.display());
             self.f_sink.write_all(&buffer).await?;
-            // for now, we flush for every send
-            self.f_sink.flush().await?;
-            debug!("flushed {}", self.path.display());
+            self.write_count = self.write_count.saturating_add(1);
+            if self.flush_policy.should_flush() {
+                self.f_sink.flush().await?;
+            }
             Ok(())
         } else {
             Err(StorageError::NoRoom(item))
@@ -136,6 +163,32 @@ impl FileRecords for MutFileRecords {
     }
 }
 
+impl MutFileRecordsFlushPolicy {
+    /// Evaluates the flush policy and returns true
+    /// if the policy determines the need to flush
+    fn should_flush(&mut self) -> bool {
+        use MutFileRecordsFlushPolicy::*;
+
+        match self {
+            NoFlush => false,
+
+            EveryWrite => true,
+
+            CountWrites {
+                n_writes: n_max,
+                write_tracking: wcount,
+            } => {
+                *wcount += 1;
+                if *wcount >= *n_max {
+                    *wcount = 0;
+                    return true;
+                }
+                false
+            }
+        }
+    }
+}
+
 #[cfg(test)]
 mod tests {
 
@@ -157,9 +210,10 @@ mod tests {
     use crate::ConfigOption;
 
     const TEST_FILE_NAME: &str = "00000000000000000100.log"; // for offset 100
+    const TEST_FILE_NAMEC: &str = "00000000000000000200.log"; // for offset 200
 
     #[test_async]
-    async fn test_write_records() -> Result<(), StorageError> {
+    async fn test_write_records_every() -> Result<(), StorageError> {
         let test_file = temp_dir().join(TEST_FILE_NAME);
         ensure_clean_file(&test_file);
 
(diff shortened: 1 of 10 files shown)


sehz:
can you rebase and link with issue?

src/types/src/defaults.rs
```diff
@@ -29,6 +29,7 @@ pub const SPU_LOG_SIZE: &str = "1Gi";
 pub const SPU_LOG_INDEX_MAX_BYTES: u32 = 10485760;
```
sehz, line 29: Can you move new constant to storage crates?  We are going to deprecrate this crate in future.

sehz (changes requested):
Nice work.  Thanks for contribution.
Couple of minor changes are requested

src/types/src/defaults.rs
digikata: yes this is a simple change

src/storage/src/mut_records.rs
```diff
@@ -197,4 +254,55 @@ mod tests {
...
+        assert_eq!(batch.get_header().magic, 2, "check magic");
+        assert_eq!(batch.records.len(), 2);
+        let mut records = batch.records;
+        assert_eq!(records.len(), 2);
+        let record1 = records.remove(0);
+        assert_eq!(record1.value.inner_value(), Some(vec![10, 20]));
+        let record2 = records.remove(0);
+        assert_eq!(record2.value.inner_value(), Some(vec![10, 20]));
+
+        for _ in 1..NUM_WRITES {
```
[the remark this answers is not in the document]
digikata: I'll add some discussion in comments around the tests. I did have negative tests setup, but removed it because of the variability in outcome. In practice it's very hit or miss to catch the no-flush case - the test is basically a race against the kernel layers below also flushing or fully completing their writes internally.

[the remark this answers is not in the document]
sehz: Is there way to make it reliable test for flush case?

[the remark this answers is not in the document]
digikata: That tests are added are the ones I believe should be reliable for the positive flush cases, but I think I should add more explanation along the lines of these comments regarding why there is no negative flush test.  A negative unit test could maybe be verified with some sort of mock object - but I'm not sure how much value there is in that test.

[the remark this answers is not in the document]
sehz: Sounds good

src/storage/src/mut_records.rs
```diff
@@ -98,9 +122,40 @@ impl MutFileRecords {
...
+                    debug!("EveryWrite flushed {}", self.path.display());
+                }
+
+                CountWrites {
+                    n_writes: n,
+                    count: icounts,
+                } => {
+                    let cnt = icounts + 1;
+                    if cnt >= n {
+                        self.flush_policy = CountWrites {
```
sehz, line 139: Few questions
1) Shouldn't count be updated after flushing?
2) What happens flushing fails?
3) Instead of creating new CountWrite.  Should just ask CountWrite to update.

src/storage/src/mut_records.rs
```diff
@@ -98,9 +122,40 @@ impl MutFileRecords {
...
+                            n_writes: n,
+                            count: 0,
+                        };
+                        self.f_sink.flush().await?;
+                        debug!("CountWrites flushed {}", self.path.display());
```
sehz, line 144: Add counts to debug log

src/storage/src/mut_records.rs
```diff
@@ -98,9 +122,40 @@ impl MutFileRecords {
...
+                    } else {
+                        self.flush_policy = CountWrites {
+                            n_writes: n,
+                            count: cnt,
```
sehz, line 148: Should add increment method to `CountWrite` instead of allocation

src/storage/src/mut_records.rs
```diff
@@ -197,4 +254,61 @@ mod tests {
...
+        assert_eq!(batch.records.len(), 2);
+        let mut records = batch.records;
+        assert_eq!(records.len(), 2);
+        let record1 = records.remove(0);
+        assert_eq!(record1.value.inner_value(), Some(vec![10, 20]));
+        let record2 = records.remove(0);
+        assert_eq!(record2.value.inner_value(), Some(vec![10, 20]));
+
+        for _ in 1..NUM_WRITES {
+            msg_sink.send(create_batch()).await?;
```
sehz, line 302: changed `?` to `expect`.  I guess this was code from while back.

src/storage/src/mut_records.rs
```diff
@@ -197,4 +254,61 @@ mod tests {
...
+        }
+
+        let bytes = read_bytes_from_file(&test_file)?;
```
sehz, line 305: change `?` to expect

src/storage/src/mut_records.rs
```diff
@@ -197,4 +254,61 @@ mod tests {
...
+        let nbytes = write_size * NUM_WRITES as usize;
+        assert_eq!(bytes.len(), nbytes, "should be {} bytes", nbytes);
+
+        let old_msg_sink = MutFileRecords::open(OFFSET, &options).await?;
```
sehz, line 309: change `?` to expect

sehz (changes requested):

src/storage/src/mut_records.rs
digikata: In the update, I did this inline to the count directly, as an increment method on enum would require a redundant match on the enum variant. I can still do it as a method if you think its cleaner.

src/storage/src/mut_records.rs
digikata: The counts were updated the logic before, but the new commit commit increments just the field in place instead.
If the flush fails, it should be reported upwards as a send error Result

src/storage/src/mut_records.rs
sehz: Yes, I think it would more clean  to separate out implementation

digikata: ok, updated

src/storage/src/mut_records.rs
```diff
@@ -24,16 +24,36 @@ use crate::records::FileRecords;
...
+    CountWrites { n_writes: u32 },
+}
+
 /// Can append new batch to file
 pub struct MutFileRecords {
     base_offset: Offset,
     item_last_offset_delta: Size,
     f_sink: BoundedFileSink,
+    flush_policy: MutFileRecordsFlushPolicy,
+    flush_write_count: u32, // only used for CountWrite policy
```
sehz, line 39: can you change to `write_count`?  I think this is useful for general metrics.

src/storage/src/mut_records.rs
```diff
@@ -98,9 +124,32 @@ impl MutFileRecords {
         if self.f_sink.can_be_appended(buffer.len() as u64) {
             debug!("writing {} bytes at: {}", buffer.len(), self.path.display());
             self.f_sink.write_all(&buffer).await?;
-            // for now, we flush for every send
-            self.f_sink.flush().await?;
-            debug!("flushed {}", self.path.display());
+            match self.flush_policy {
```
sehz, line 127: can you move this code to FlushPolicy? I think this would make it simpler to read and maintain.

src/storage/src/mut_records.rs
```diff
@@ -111,6 +160,16 @@ impl MutFileRecords {
     pub async fn flush(&mut self) -> Result<(), IoError> {
         self.f_sink.flush().await
     }
+
+    fn should_flush_after_increment(&mut self, n_writes: u32) -> bool {
```
sehz, line 164: move to FlushPolicy

sehz (changes requested):

src/storage/src/mut_records.rs
digikata: sure no problem

src/storage/src/mut_records.rs
digikata: sure

src/storage/src/mut_records.rs
digikata: Making this either a saturating or wrapping counter is pretty straightforward it looks like

src/storage/src/mut_records.rs
```diff
@@ -195,6 +163,32 @@ impl FileRecords for MutFileRecords {
...
+        use MutFileRecordsFlushPolicy::*;
+
+        match self {
+            NoFlush => false,
+
+            EveryWrite => true,
+
+            CountWrites {
+                n_writes: n_max,
+                write_tracking: wcount,
```
sehz, line 179: write_tracking seems to be duplicate of `write_count`.  Do we need both?

digikata: I made write_count an absolute count tied to mut_records for metrics, write tracking is tied to the flush policy but is reset to zero after a flush. I could have unified them and triggered the flush policy on a (write_count % n_writes == 0), but thought the divide would be more expensive than two counters and two increments.

sehz: Sounds good

sehz:
rebase please

sehz (approved):

PR #670, opened 2021-01-11
Add 'cancel previous runs' to CI

nicholastmosher:
I don't know if there is specific reason we did not have this here before, but it makes CI faster when you can cancel the previous and queued jobs when you push to a branch. To make this work, we need to add a `GITHUB_TOKEN` secret


sehz:
I think we had before, but there was some issue.  Let's try again

sehz (approved):

PR #687, opened 2021-01-14
Made consumer module pub

simlay:
This makes infinyon/fluvio-client-node#66 much cleaner and it will also help with our rust client users.

### DIFF
--- a/src/client/src/consumer.rs
+++ b/src/client/src/consumer.rs
@@ -391,23 +391,26 @@ impl Default for ConsumerConfig {
 }
 
 impl ConsumerConfig {
-    /// TODO describe what the max_bytes option does
+    /// Maximum number of bytes to be fetched at a time.
     pub fn with_max_bytes(mut self, max_bytes: i32) -> Self {
         self.max_bytes = max_bytes;
         self
     }
 }
 
+/// The individual record for a given stream.
 pub struct Record {
     offset: i64,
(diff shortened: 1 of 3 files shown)


src/client/src/consumer.rs
```diff
@@ -391,23 +391,27 @@ impl Default for ConsumerConfig {
 }
 
 impl ConsumerConfig {
-    /// TODO describe what the max_bytes option does
+    /// Analogous to
+    /// [KIP-74](https://cwiki.apache.org/confluence/display/KAFKA/KIP-74%3A+Add+Fetch+Response+Size+Limit+in+Bytes)
```
simlay, line 395: I found this someplace in the fluvio src as a reference but I'm not sure it's correct.

sehz, line 395: remove ref to kafka

src/client/src/lib.rs
```diff
@@ -86,7 +86,7 @@ mod error;
 mod client;
 mod admin;
 mod params;
-mod consumer;
+pub mod consumer;
```
sehz, line 89: Don't think we want to expose entire module yet.  @nicholastmosher concur?
Let's just expose what is need for this change

sehz (changes requested):

sehz (approved):

sehz:
rebase please

PR #690, opened 2021-01-15
Github actions for crates publishing

simlay:
I'm I read https://github.com/katyo/publish-crates correctly, I think this can actually work on every push of master without error and will work on all the projects in the workspace?

@sehz can you add a `CARGO_REGISTRY_TOKEN` to the settings with a token from crates.io?

### DIFF
--- a/.github/workflows/publish_crates.yml
+++ b/.github/workflows/publish_crates.yml
@@ -1,0 +1,20 @@
+name: Publish Crates
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [stable]
+
+jobs:
+  publish_crates:
+    name: Publish to crates.io
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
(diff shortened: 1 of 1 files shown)


katyo:
@simlay Course, this action never try to publish already published versions. But I still yet recommend use the push condition for specific branch or tag to avoid erroneous publishing. You can see real usage example here: [oboe-rs](https://github.com/katyo/oboe-rs/blob/a98967e15c8aaa4f285839ffba454332f6d63363/.github/workflows/rust.yml#L273-L292)

.github/workflows/publish_crates.yml
```diff
@@ -0,0 +1,20 @@
+name: Publish Crates
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [master]
```
sehz, line 6: I think it is better if we trigger release on pushing to `stable` branch so PR can trigger which can be done by anyone.

sehz:
rebase

sehz:
secret is set

sehz (approved):

PR #693, opened 2021-01-16
Implment flush policy to flush on a delay time after writes

digikata:
clusters of writes with a close enough interarrival time
will not incur extra flushes

a delay flush task will exit after associated mut_records
instance is dropped

TODO: the addition of a flush count allows a bit better visibility into how
many flushes have been flushed on some paths, but still need to increment
flush_count from delay flush task.

This is a good way to inspect a couple of different cases:
RUST_LOG=debug cargo test test_write_records_idle_delay

https://github.com/infinyon/fluvio/issues/37

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -891,7 +891,7 @@ dependencies = [
 
 [[package]]
 name = "fluvio"
-version = "0.3.4"
+version = "0.3.5"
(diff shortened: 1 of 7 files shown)


digikata:
wrong branch

PR #694, opened 2021-01-16
Implement flush policy to flush on a delay time after writes

digikata:
https://github.com/infinyon/fluvio/issues/37

With a delay flush policy set, writes will cause a flush after the set delay
time.

Clusters of writes with an interarrival time below the delay will not incur
extra flushes, a final flush is scheduled after the latest write.

The delay flush task will exit after associated mut_records instance is
dropped

TOREVIEW: the get_file method was commented out for removal because the
ownership seemed convoluted to resolve (?) even with a sharing in place
for the f_sink file. the get_file method in the FileRecords trait never
seemed to be used in the codebase

This is a good way to inspect a couple of different cases:
RUST_LOG=debug cargo test test_write_records_idle_delay

### DIFF
--- a/src/storage/src/mut_records.rs
+++ b/src/storage/src/mut_records.rs
@@ -1,12 +1,20 @@
 use std::io::Error as IoError;
 use std::path::PathBuf;
 use std::path::Path;
+use std::time::{Duration, Instant};
+use std::sync::atomic::{AtomicU32, Ordering};
+use std::sync::Arc;
+use async_mutex::Mutex;
 
 use tracing::debug;
+use tracing::warn;
 use tracing::trace;
-use futures_lite::io::AsyncWriteExt;
 
-use fluvio_future::fs::File;
+use futures_lite::io::AsyncWriteExt;
+use async_channel::Sender;
+
+use fluvio_future::timer;
+// use fluvio_future::fs::File;
 use fluvio_future::file_slice::AsyncFileSlice;
 use fluvio_future::fs::BoundedFileSink;
 use fluvio_future::fs::BoundedFileOption;
@@ -24,32 +32,40 @@ use crate::records::FileRecords;
 
 pub const MESSAGE_LOG_EXTENSION: &str = "log";
 
-/// MutFileRecordsFlushPolicy describes and implements a flush policy
-pub enum MutFileRecordsFlushPolicy {
-    NoFlush,
-    EveryWrite,
-    CountWrites { n_writes: u32, write_tracking: u32 },
-}
+// Delay flush scheduling interrival time constant
+//   not something that a user should change
+// this is a delay put in because the scheduler switch
+// to the flush task is fast enough that it might beat
+// out waiting writes which should be preferred
+const DELAY_FLUSH_SIA_MSEC: u64 = 3;
 
 /// Can append new batch to file
 pub struct MutFileRecords {
     base_offset: Offset,
     item_last_offset_delta: Size,
-    f_sink: BoundedFileSink,
-    flush_policy: MutFileRecordsFlushPolicy,
+    f_sink: Arc<Mutex<BoundedFileSink>>,
+    f_slice_root: AsyncFileSlice,
+    cached_len: u64,
+    flush_policy: FlushPolicy,
     write_count: u64,
+    flush_count: Arc<AtomicU32>,
     path: PathBuf,
+    flush_time_tx: Option<Sender<Instant>>,
 }
 
 impl Unpin for MutFileRecords {}
 
-fn get_flush_policy_from_config(option: &ConfigOption) -> MutFileRecordsFlushPolicy {
-    if option.flush_write_count == 0 {
-        MutFileRecordsFlushPolicy::NoFlush
+fn get_flush_policy_from_config(option: &ConfigOption) -> FlushPolicy {
+    if option.flush_idle_msec > 0 {
+        FlushPolicy::IdleFlush {
+            delay_millis: option.flush_idle_msec,
+        }
+    } else if option.flush_write_count == 0 {
+        FlushPolicy::NoFlush
     } else if option.flush_write_count == 1 {
-        MutFileRecordsFlushPolicy::EveryWrite
+        FlushPolicy::EveryWrite
     } else {
-        MutFileRecordsFlushPolicy::CountWrites {
+        FlushPolicy::CountWrites {
             n_writes: option.flush_write_count,
             write_tracking: 0,
         }
@@ -67,13 +83,18 @@ impl MutFileRecords {
         let log_path = generate_file_name(&option.base_dir, base_offset, MESSAGE_LOG_EXTENSION);
         debug!("creating log at: {}", log_path.display());
         let f_sink = BoundedFileSink::open_append(&log_path, sink_option).await?;
+        let f_slice_root = f_sink.slice_from(0, 0)?;
         Ok(MutFileRecords {
             base_offset,
-            f_sink,
+            f_sink: Arc::new(Mutex::new(f_sink)),
+            f_slice_root,
+            cached_len: 0,
             flush_policy: get_flush_policy_from_config(option),
             write_count: 0,
+            flush_count: Arc::new(AtomicU32::new(0)),
             item_last_offset_delta: 0,
             path: log_path.to_owned(),
+            flush_time_tx: None,
         })
     }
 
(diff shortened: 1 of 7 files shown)


digikata:
I think this would be a good point for an initial review. 

I need to analyze the composite-action failure to see how it's liked to changes here.

simlay:
As far as the composite action CI goes, it's been flakey recently. I should check that the minikube version and things match that of the other CI.

src/storage/src/mut_records.rs
```diff
@@ -104,11 +114,16 @@ impl MutFileRecords {
...
 
     pub async fn validate(&mut self) -> Result<Offset, LogValidationError> {
-        validate(self.f_sink.get_path()).await
+        let f_sink = self.f_sink.lock().await;
+        validate(f_sink.get_path()).await
     }
 
     pub fn get_pos(&self) -> Size {
-        self.f_sink.get_current_len() as Size
+        block_on(async {
```
sehz, line 122: I would rather not see block_on here.  There maybe two path I see:
1) change this to async
2) cache len

src/storage/src/mut_records.rs
```diff
@@ -145,34 +258,67 @@ impl FileRecords for MutFileRecords {
...
+    // }
 
     fn get_path(&self) -> &Path {
         &self.path
     }
 
     fn as_file_slice(&self, start: Size) -> Result<AsyncFileSlice, IoError> {
-        self.f_sink
-            .slice_from(start as u64, self.f_sink.get_current_len() - start as u64)
+        block_on(async {
```
sehz, line 282: avoid block_on

src/storage/src/mut_records.rs
```diff
@@ -145,34 +258,67 @@ impl FileRecords for MutFileRecords {
...
+            let am = self.f_sink.clone();
+            let f_sink = am.lock().await;
+            f_sink.slice_from(start as u64, f_sink.get_current_len() - start as u64)
+        })
     }
 
     fn as_file_slice_from_to(&self, start: Size, len: Size) -> Result<AsyncFileSlice, IoError> {
-        self.f_sink.slice_from(start as u64, len as u64)
+        block_on(async {
```
sehz, line 290: avoid block_on

src/storage/src/mut_records.rs
```diff
@@ -136,7 +167,89 @@ impl MutFileRecords {
...
+        f_sink.flush().await
+    }
+
+    #[allow(unused)]
+    pub fn flush_count(&self) -> u32 {
+        self.flush_count.load(Ordering::Relaxed)
+    }
+
+    fn delay_flush(&mut self, delay_millis: u32) -> Result<(), IoError> {
+        let delay_tgt = delay_millis as u64;
```
sehz, line 182: This seems like to spawn task every flush which is expensive.  And also creating channel as well.  
It probably more efficient if timer loop is created only once.

sehz (changes requested):

src/storage/src/mut_records.rs
digikata: yup can do there, I'll explore both for this

src/storage/src/mut_records.rs
digikata: The slice methods are synchronous in the FileRecords trait. I used block_on to limit the scope of changes, but if changing the trait is ok too then I can look at widening that.

src/storage/src/mut_records.rs
digikata: The delay flush task is only spawned on the first write, after that, the flush_time_tx option is Some(channel::Sender) and the task will continue to loop on the rx of that channel for the rest of the lifetime of the mut_record instance. The following calls to delay_flush skip the        "if self.flush_time_tx.is_none() {" and  update the time sent to the task - which lets the loop start a new timer delay, or reason about in-process delays.

src/storage/src/mut_records.rs
```diff
@@ -136,7 +167,89 @@ impl MutFileRecords {
...
+                    }
+                }
+                debug!("delay_flush task exited");
+            });
+        }
+
+        // update running flush
+        if let Some(flush_tx) = &self.flush_time_tx {
+            let tx = flush_tx.clone();
+            fluvio_future::task::run(async move {
```
sehz, line 246: avoid blocking call

src/storage/src/mut_records.rs
sehz: got it.  just need to ensure that spawned future will be safely terminate if record is dropped

sehz (changes requested):

src/storage/src/mut_records.rs
digikata: When the mut_records struct is dropped, the only transmit side of the channel, flush_time_tx, is closed. The delay flush task picks that up via the while(!rx.is_closed()) condition and  exits the loop, ending the task.  I've verified that happens via tracing the test_write_records_idle_delay test.

src/storage/src/mut_records.rs
digikata: good catch I'll rework this...

src/storage/src/mut_records.rs
digikata: Returning cached get_pos value now.  One thing that makes me uneasy working with the change is that the position is a u64 and the Size type is u32; this existed before any of of these changes.  But that means there might be some funny behavior showing up at 4GB log file sizes.  cargo clippy can turn on lints for casting truncations, but this might want to be a separate issue.

src/storage/src/mut_records.rs
```diff
@@ -24,32 +32,32 @@ use crate::records::FileRecords;
...
-    CountWrites { n_writes: u32, write_tracking: u32 },
-}
-
 /// Can append new batch to file
 pub struct MutFileRecords {
     base_offset: Offset,
     item_last_offset_delta: Size,
-    f_sink: BoundedFileSink,
-    flush_policy: MutFileRecordsFlushPolicy,
+    f_sink: Arc<Mutex<BoundedFileSink>>,
```
sehz, line 39: Since underlying File system is thread safe.  Is it better to create separate file from descriptor rather than tryin to use mutex?

src/storage/src/mut_records.rs
sehz: can mitigate by only allowing limit segment up to u32.  Typical size for segment is 1G

digikata: I'm thinking it might be worth putting in a front end check for overflow of 4GB (or 1GB) on write.

src/storage/src/mut_records.rs
digikata: The file descriptor isn't accessible from the boundedfilesink / or the async file type, so there's a tradeoff of changing that and maintining the difference or upstreaming the change.  

You're right about the fs side of things supporting thread safe part, but copying fs descriptor sort of gives added cleanup issues on the fd on drop of the parent BoundedFileSink. that code needs to then not close the fd and the flush task needs to know to close it.  I suspect this would be easier to address  one of the io_uring backends, and maybe defer improving this beyond a Mutex for now?

sehz: sounds goo

sehz:
rebase

src/storage/src/mut_records.rs
sehz: Yes, we should not be blocking on here since this is critical path

digikata:
no changes except the rebase here - working on removing the block_on

src/storage/src/mut_records.rs
```diff
@@ -136,7 +178,92 @@ impl MutFileRecords {
...
+                        // for wait times after the first delay runs
+                        write_time = wt;
+                    }
+                    //  A short fixed wait was added here to give a little
+                    //  breathing space to accumulate clusters of writes which
+                    //  arrive closely but not fast enough for this task loop
+                    //  between the first and following check for write times
+                    //  It prevents extra flushes and the time is still accounted for
+                    //  in the delay_dur calculation.
+                    timer::after(Duration::from_millis(3)).await;
```
sehz, line 223: can you make 3 as constant?

sehz:
So this seems to be pretty much done?  Can you transition from `draft`

digikata:
I think so - except the macos checks seem to intermittently fail.   Do you want me to squash the commits into a single one?

sehz:
Yes, that would be great.

sehz (changes requested):

sehz:
MacOS CI failure is know issue.

digikata:
Moved a parameter from literal to a constant, squashed. This change should be ready to merge after a good ci run.

src/storage/src/mut_records.rs
digikata: Yup made this a constant in the squash update

sehz (approved):
Great work!  Looks good

sehz:
`test_write_records_idle_delay` sometimes fail in CI for Mac because VM only has 2 core and performance can vary widely.  Is there way to make it more reliable?

digikata:
Most likely a change to the secondary interrival delay might change it, but
I should try to diagnose it a little better.  Do you happen to know if I
turn on tracing inside the test, if it will get logged in the CI report?

sehz:
Yes, it should be

PR #695, opened 2021-01-17
Fix crates publish ci

simlay:
We should probably have `dry-run` as part of the CI


sehz (approved):

sehz:
How about allow dispatch for dry un while actual publishing is done thru only push.  This allow to verify before we try publishing by any with write access

simlay:
Closing for now as the constraints of the publish action are a bit harsh.

PR #709, opened 2021-01-22
Cluster upgrade

nicholastmosher:
This implements `fluvio cluster upgrade`, piggybacking on much of the same command infrastructure as `fluvio cluster start`. Later I'd like to revisit this to make a clearer separation of concerns, but this works for now.


src/cluster/src/error.rs
```diff
@@ -44,6 +44,9 @@ pub enum K8InstallError {
     /// Encountered an error while performing one or more pre-checks
     #[error("Failed to perform one or more pre-checks")]
     PrecheckErrored(CheckResults),
+    /// Failed to update Fluvio cluster
+    #[error("Failed to update Fluvio")]
+    FailedUpdate,
```
sehz, line 49: Should've this include why upgrade failed?

src/cluster/src/start/k8.rs
```diff
@@ -781,6 +791,36 @@ impl ClusterInstaller {
             }
         }
 
+        // When upgrading, wait for platform version to match new version
```
sehz, line 794: This seems duplicate of sc checking? Can we consolidate to there.  Version check is useful as there too

sehz (changes requested):

src/cluster/src/start/k8.rs
nicholastmosher: Correct me if I'm wrong but it seems like the upgrade process would not need `create_managed_spu_group`? Aside from that, I can see if I can consolidate

sehz: Yes creating spus is not needed in upgrading.  It would be good if we can move post install/upgrade workflow as separate entity (with builder?) so it can be more flexible

src/cluster/src/error.rs
nicholastmosher: Right now the only reason the `FailedUpdate` error would be given is if the platform version check times out. But I can add that to the description of the error.

src/cluster/src/start/k8.rs
nicholastmosher: Yes, I have plans like that for when we refactor this crate

nicholastmosher:
@sehz can you publish `fluvio-helm:0.4.0`? That is what is causing CI to fail here

sehz:
> @sehz can you publish `fluvio-helm:0.4.0`? That is what is causing CI to fail here

done

src/cluster/src/error.rs
```diff
@@ -45,8 +45,8 @@ pub enum K8InstallError {
     #[error("Failed to perform one or more pre-checks")]
     PrecheckErrored(CheckResults),
     /// Failed to update Fluvio cluster
-    #[error("Failed to update Fluvio")]
-    FailedUpdate,
+    #[error("Failed to witness expected Fluvio cluster version: {0}")]
+    FailedVersionTimeout(String),
```
sehz, line 49: Isn't this same as SCServiceTimeout?  I would expect error related to helm upgrade failure like (existing install version, etc)

src/cluster/src/start/k8.rs
```diff
@@ -1075,6 +1049,32 @@ impl ClusterInstaller {
         }
     }
 
+    /// Wait until the platform version of the cluster matches the chart version here
+    #[instrument(skip(self))]
+    async fn wait_for_fluvio_version(&self) -> Result<(), K8InstallError> {
```
sehz, line 1054: This should leverage `wait_for_sc_service` not creating another variant

sehz (changes requested):

src/cluster/src/start/k8.rs
nicholastmosher: I was thinking about actually unifying all of the `wait_for_X` checks, but right now the project is not well fitted to do that. I am planning on revisiting this when I refactor, but for now I wanted to do the minimum to get it to work.

src/cluster/src/error.rs
nicholastmosher: I think SC Service timeout is about waiting for the endpoint to be available, and the Failed Version timeout is about waiting to see a Fluvio cluster with the specifically requested version. This is what the `upgrade` process does, because when you begin upgrading, there is already an SC up and available, but it does not indicate that the upgrade has succeeded, because you would expect the old SC to go down and to get a new one up that reports the correct version.

nicholastmosher: I can look at cleaning this up when I refactor, but this at least gets the job done

sehz: FailedVersionTimeout is confusing and misleading.  There could be multiple reason:
1:  Helm installation failure.
2.  Upgraded failed for whatever reason, SC doesn't come up?  
3.  Network failure
4.  SC come up but somehow version is different.

For 2&3, `SCServiceTimeout` can be used.

src/cluster/src/start/k8.rs
sehz: I don't think this will work since it does't take account of ns

sehz (changes requested):

src/cluster/src/error.rs
nicholastmosher: I have been playing with this for a little bit and it seems like I won't be able to actually merge `wait_for_fluvio_version` and `wait_for_sc_service` because `create_managed_spu_group` needs to run in between those calls. The flows involved here aren't quite obvious because similar code is being used for both the initial-install path and the upgrade path. I am hoping to untangle these a bit in the future.

So here's where we're at:

- Fluvio install worked just fine before I added upgrade.
- As part of upgrade, I needed to be able to look at the "platform version" reported by the cluster, because simply observing an SC and SPU standing is not enough to know that they were actually upgraded to the specified version.
- So `wait_for_fluvio_version` is really only needed for upgrade use-case.

My current thinking is that I can reword the error message to indicate that something went wrong in the upgrade process, but I don't know how to be any more specific than "the upgrade failed" if we are not going to talk about the fact that the version requests did not see the new version of the cluster appear.

sehz: `create_managed_spu_group`  is optional

sehz: new error message is fine

nicholastmosher: Alright, I've pushed a new error message which is `FailedClusterUpgrade` which prints as "Failed to upgrade cluster to version X"

src/cluster/src/start/k8.rs
```diff
@@ -1032,6 +1049,34 @@ impl ClusterInstaller {
...
```
sehz, line 1054: use `FluvioConfig` rather than rely on profile since it might not be available (there is option to skip profile save).  See: `create_managed_spu_group`

sehz (changes requested):

sehz (approved):
looks good. just rebase

PR #711, opened 2021-01-23
Updated Readme and cargo.toml to be more accurate for crates.io

simlay:
The stuff on https://crates.io/crates/fluvio says `flv-client` and doesn't have any keywords or categories. I chose the category slug of `database` because it seemed the least-not-wrong.

* [According to the cargo docs, each category should match one of the strings available](https://doc.rust-lang.org/cargo/reference/manifest.html#the-categories-field) at https://crates.io/category_slugs.

### DIFF
--- a/src/client/Cargo.toml
+++ b/src/client/Cargo.toml
@@ -4,6 +4,8 @@ version = "0.3.6"
 edition = "2018"
 license = "Apache-2.0"
 authors = ["Fluvio Contributors <team@fluvio.io>"]
+categories = ["database"]
+keywords = ["streaming", "stream", "queue"]
 repository = "https://github.com/infinyon/fluvio"
 description = "The offical Fluvio driver for Rust"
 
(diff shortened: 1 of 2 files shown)


src/client/Cargo.toml
```diff
@@ -4,6 +4,8 @@ version = "0.3.6"
 edition = "2018"
 license = "Apache-2.0"
 authors = ["Fluvio Contributors <team@fluvio.io>"]
+categories = ["database"]
+keywords = ["streaming", "stream", "queue", "kafka"]
```
sehz, line 8: remove kafka

sehz (changes requested):

src/client/Cargo.toml
simlay: Yeah, I was on the fence about that. I went with it because maybe people searching for kafka would find us. Anyway, I've removed it.

sehz (approved):

PR #714, opened 2021-01-24
test delay flush: examine macos config 

digikata:
just getting some visibility into the mac config CI


sehz:
We are moving to Mac as tier2 support which means it doesn't need to be passed as part of fluvio CI.  Mark as closing

PR #716, opened 2021-01-26
Fluvio command extension

nicholastmosher:
This closes infinyon/flv-util#10 by introducing a new `CommandExt` which does not panic. This is now located in its own crate that I put into the fluvio repo and named `fluvio-command` (name subject to bikeshedding).

This new extension trait introduces three methods:

- `inherit`: Causes child process to inherit `stdout` and `stderr`
- `display`: Produces a stringified representation of the Command
- `result`: Waits for the command to complete and returns a result representing the outcome.

I think that I've set up the `result` method in a good way to make commands flow well with regular Rust error handling. The other command-running functions (`output`, `status`, and `spawn`) tended not to compose well with typical Rust functions, you would always need to pattern match in several places to check whether it executed right, whether the exit code was good, etc. The `result` method now returns `Ok(Output)` if an only if

- The command executes successfully (e.g. we did not fail to find the named executable)
- The command _completes_, i.e. it is not terminated early
- The command returns an exit code of 0.

If any of those error cases occurs, then `result` returns the new `CommandError` type which describes the exact command that was attempted as well as which problem was encountered. (`IoError`, `Terminated`, or `ExitCode`).

The end result is that we can use command chaining nicely like this:

```rust
let _output: std::process::Output = Command::new("kubectl")
    .args(&["get", "pods"])
    .result()?;
println!("Success!");
```

And we can be confident that if we reach the "Success!" printout that everything worked as we expected. This is the behavior that we want 90% of the time when working with commands. In the case that a non-zero exit code does not necessarily indicate a total failure, we'll need to resort back to manually checking the status.


src/command/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "fluvio-command"
+version = "0.1.0"
+authors = ["Nick Mosher <nicholastmosher@gmail.com>"]
```
sehz, line 4: changes to fluvio contributors

sehz (changes requested):
one small nitpick on coding.
However, there is one major feature dropped from previous implementation which is ability to debug command execution which was control from `FLV_CMD`.

src/command/Cargo.toml
nicholastmosher: Whoops, `cargo new` got away from me...

src/command/src/lib.rs
```diff
@@ -92,6 +93,13 @@ impl CommandExt for Command {
     }
 
     fn result(&mut self) -> CommandResult {
+        let should_log = std::env::var("FLUVIO_CMD")
```
sehz, line 96: can you put this using `once_cell::Lazy` so it can be only evaluated once?

nicholastmosher: Sure thing

nicholastmosher:
This should be ready to go. @sehz if it looks good to you, can we publish this to crates.io after its merged? I'd like to be able to use it in `fluvio-helm` as well as other crates.

sehz (changes requested):
`once_cell` changes looks good.  But need to update GitHub CI to use new `FLUVIO_CMD`

sehz (approved):

Issue #705, opened 2021-01-21
Refactor `fluvio-cluster` crate


nicholastmosher:
There are a handful of things that are starting to get messy in the `fluvio-cluster` crate that could use some cleaning up. I'll list down some of the changes that I think are needed:

- [x] We need to split out sys-chart installation into its own module/builder so it can be shared
- [x] Need to refactor some functions in k8 starter to make pre-install checks print as-they-go, similar to check and local
- [x] Use derive_builder for large config objects
- [ ] Unify config options that are shared between k8 and local installers
- [ ] Stop using Lazy, prefer fields in config objects

nicholastmosher:
The bulk of items in this issue are taken care of so I am going to close it. If the remaining items become a priority in the future, we can open a new issue for them.

PR #721, opened 2021-01-27
Cluster refactor: Split out SysInstaller, simplify check-fixes

nicholastmosher:
This is a good start to some refactoring that's needed in the cluster crate. In this PR I've extracted the sys chart installation logic into its own `SysInstaller` type, which means we have less code duplication and cross-dependency between the Local and K8 installers.

Another change here is that I've added an `attempt_fix` optional function to the `ClusterCheck` trait, which allows instances of that trait to define the means by which it can attempt to fix a failed check. Previously, this fixing logic was very far removed from the check itself, so this consolidates things nicely. It also has the effect of unblocking the "incremental progress checks" for the k8 installer. Now, using the k8 installer `fluvio cluster start` will print checking outputs as they are run, the same way that `fluvio cluster start --local` already does today.

This checks a number of the boxes from #705


src/cluster/src/check/mod.rs
```diff
@@ -256,15 +260,66 @@ impl CheckSuggestion for UnrecoverableCheck {
...
+            None => {
+                return Ok(CheckStatus::fail(
+                    UnrecoverableCheck::NoActiveKubernetesContext,
+                ));
+            }
+        };
+
+        let server_url = &cluster_context.cluster.server;
+
+        // Check that the server URL has a hostname, not just an IP
```
sehz, line 309: Don't think this is needed anymore

src/cluster/src/check/mod.rs
```diff
@@ -273,8 +328,53 @@ pub(crate) struct K8Version;
...
 impl ClusterCheck for K8Version {
+    /// Check if required kubectl version is installed
     async fn perform_check(&self) -> CheckResult {
-        k8_version_check()
+        let kube_version = Command::new("kubectl")
+            .arg("version")
+            .arg("-o=json")
+            .output()
+            .map_err(CheckError::KubectlNotFoundError)?;
+        let version_text = String::from_utf8(kube_version.stdout).unwrap();
```
sehz, line 338: You don't need to convert to string in order to convert from JSON.   JSON conversion be done from byte slice

src/cluster/src/check/mod.rs
```diff
@@ -374,30 +593,31 @@ impl ClusterChecker {
...
+    /// - [`run_wait`](ClusterChecker::run_wait)
+    /// - [`run_wait_and_fix`](ClusterChecker::run_wait_and_fix)
+    /// - [`run_with_progress`](ClusterChecker::run_with_progress)
+    /// - [`run_and_fix_with_progress`](ClusterChecker::run_and_fix_with_progress)
     pub fn with_preflight_checks(mut self) -> Self {
         let checks: Vec<Box<(dyn ClusterCheck)>> = vec![
             Box::new(LoadableConfig),
             Box::new(K8Version),
             Box::new(HelmVersion),
-            Box::new(SysChart),
```
sehz, line 400: why remove sys chart check?

src/cluster/src/check/mod.rs
```diff
@@ -411,15 +631,14 @@ impl ClusterChecker {
...
-    /// - [`run_and_fix_with_progress`]
+    /// - [`run_wait`](ClusterChecker::run_wait)
+    /// - [`run_wait_and_fix`](ClusterChecker::run_wait_and_fix)
+    /// - [`run_with_progress`](ClusterChecker::run_with_progress)
+    /// - [`run_and_fix_with_progress`](ClusterChecker::run_and_fix_with_progress)
     pub fn with_k8_checks(mut self) -> Self {
         let checks: Vec<Box<(dyn ClusterCheck)>> = vec![
             Box::new(LoadableConfig),
             Box::new(HelmVersion),
-            Box::new(SysChart),
```
sehz, line 422: same question about sys chart?

src/cluster/src/cli/start/sys.rs
```diff
@@ -0,0 +1,40 @@
...
+use crate::sys::{SysConfig, SysInstaller};
+use crate::ClusterError;
+use crate::error::SysInstallError;
+
+pub fn install_sys(opt: StartOpt, upgrade: bool) -> Result<(), ClusterCliError> {
+    install_sys_impl(opt, upgrade).map_err(ClusterError::InstallSys)?;
+    Ok(())
+}
+
+fn install_sys_impl(opt: StartOpt, upgrade: bool) -> Result<(), SysInstallError> {
```
sehz, line 12: why creating another fn?

src/cluster/src/cli/start/sys.rs
```diff
@@ -0,0 +1,40 @@
...
+    let mut builder = SysConfig::builder();
+    builder.with_namespace(opt.k8_config.namespace);
```
sehz, line 14: Can collapse into a single line?

src/cluster/src/cli/start/sys.rs
```diff
@@ -0,0 +1,40 @@
...
+
+    match opt.k8_config.chart_location {
```
sehz, line 16: maybe this should be moved to `opt` rather than here

src/cluster/src/cli/start/sys.rs
sehz: Maybe whole thing should be moved into `InstallSys`

src/cluster/src/sys.rs
```diff
@@ -0,0 +1,223 @@
...
+    /// let builder = SysConfig::builder();
+    /// ```
+    pub fn builder() -> SysConfigBuilder {
+        SysConfigBuilder::default()
+    }
+}
+
+impl SysConfigBuilder {
+    /// Validates all builder options and constructs a `SysConfig`
+    pub fn build(&self) -> Result<SysConfig, SysInstallError> {
```
sehz, line 63: Shouldn't this just return `SysConfig` rater than Result?  Why would this fail?

sehz (changes requested):
Looks good with few minor points.

src/cluster/src/check/mod.rs
nicholastmosher: I only removed it from this helper method because it now requires a `SysConfig` object with the namespace and other info required to install the sys chart, and we do not have that information available here. If you look at where this is actually called (from k8 installer and local installer), the SysChartCheck is now manually added to the checker using a config that is derived from the installer's config itself.

https://github.com/infinyon/fluvio/pull/721/files#diff-acfbf9dfd3f6cc9ca6cba75c1b476d4144794784b4c93ef7c8e8d716c5a65591R333

This means that the SysConfig is now derived from the K8Install config or the LocalInstall config, so it will inherit certain options rather than them being hardcoded (e.g. the namespace).

src/cluster/src/check/mod.rs
nicholastmosher: See answer above

src/cluster/src/cli/start/sys.rs
nicholastmosher: This was done just to convert all of the error types at once. Otherwise, we would need map_err all over the place which is messy

src/cluster/src/cli/start/sys.rs
nicholastmosher: Ooh, with the current code layout that was not possible, but I just thought of something that will make this work

src/cluster/src/sys.rs
nicholastmosher: This is intentional, the reason I put a Result there is because even though there might not be a reason for this to fail today, in the future we may need to add configuration options that require validation or which need to be collected from the environment, which may be fallible operations. It doesn't hurt us today to put a Result here if we don't need it, but it may hurt us in the future (by introducing a breaking change) if we end up needing a Result but don't have one.

src/cluster/src/cli/start/sys.rs
nicholastmosher: Hmm, maybe, but right now this does what we need it to do. I don't see an urgent need to move it, and it's not immediately obvious to me how things would look by trying to rewrite this.

sehz: sounds good

nicholastmosher:
It'd be nice to merge and publish https://github.com/infinyon/fluvio-helm/pull/11 before this, so that we can bump our dependency to fluvio-helm:0.4.1 and use the latest version. The big benefit there is that the new fluvio-helm updates do not forward all of the noise from the helm command

sehz (approved):
Looks good

PR #723, opened 2021-01-28
Cluster sys version

nicholastmosher:
Add `chart_version` to `SysConfig`. Also puts back `build.rs` in all of the crates that need VERSION file in order to prevent problems with UNDEFINED version on certain configurations.

Bumps fluvio-cluster to `0.7.0` because the refactoring PR that was merged earlier had breaking changes.
Bumps fluvio platform VERSION to `0.7.0-alpha.1` due to the protocol changes a little while ago


.gitignore
```diff
@@ -5,3 +5,4 @@ target/
 target-docker
 .vscode/tasks.json
 .idea/
+VERSION
```
sehz, line 8: This make it not check-in the file at the top.

src/cluster/build.rs
```diff
@@ -0,0 +1,4 @@
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    std::fs::copy("../../VERSION", "./src/VERSION")?;
```
sehz, line 2: this may fail in the `cargo publish`

src/sc/build.rs
```diff
@@ -0,0 +1,4 @@
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    std::fs::copy("../../VERSION", "./src/VERSION")?;
```
sehz, line 2: allow failure

sehz (changes requested):

.gitignore
nicholastmosher: git is smart enough to continue tracking changes to files that match .gitignore rules if those files are already committed in the repository

src/cluster/build.rs
nicholastmosher: Hmm I thought this was the solution we used originally and that it worked? Or am I mistaken? I have never used `cargo publish` so I don't know how picky it is and I can't really test it out

sehz: This will fail because VERSION doesn't exists

nicholastmosher: Hmm can we make a custom publish flow that just does `cargo check && cargo publish ARGS`? I think that running cargo check should trigger the VERSION copy and place the file in the crate directory, so it should get zipped up by cargo publish. What do you think of that idea?

nicholastmosher:
There were actually some underlying bugs where CLI options were not being propagated into the installer configuration. I have fixed those issues

nicholastmosher:
Once we figure out a solution for VERSION, I can add back the version bump to 0.7.0-alpha.1 and we can merge

src/cluster/build.rs
nicholastmosher: Actually, since I have made the `copy` infallible, I think that we can make a manual publish flow work because you can just run the `cargo check` or `cargo build` locally yourself before running `cargo publish` and it should work just fine

nicholastmosher:
Newest changes:

- `chart_version` is now a required field for both `SysConfig` and `ClusterConfig`
- `fluvio-cli` passes VERSION to `fluvio-cluster
- Fixed broken doc tests
- Added tests for required config options

src/cli/build.rs
```diff
@@ -2,6 +2,9 @@ use std::process::Command;
 use rustc_version::version_meta;
 
 fn main() {
+    // Copy VERSION file. Do not fail e.g. when built via `cargo publish`
+    let _ = std::fs::copy("../../VERSION", "./src/VERSION");
```
sehz, line 6: maybe we put it back old way of referring VERSION?  That way, we don't need to maintain `build.rs`

src/client/src/consumer.rs
```diff
@@ -37,7 +37,7 @@ use crate::spu::SpuPool;
 ///
```
sehz, line 37: Can you move this as separate PR since this is not related to this purpose of this PR

src/cluster/src/main_disabled.rs
```diff
@@ -0,0 +1,12 @@
+// use structopt::StructOpt;
+// use fluvio_cluster::cli::ClusterOpt;
+// use fluvio_future::task::run_block_on;
+
+// fn main() -> color_eyre::Result<()> {
+//     color_eyre::config::HookBuilder::blank()
```
sehz, line 6: remove if not used

src/cluster/src/start/k8.rs
```diff
@@ -49,103 +51,27 @@ static MAX_SC_NETWORK_LOOP: Lazy<u16> = Lazy::new(|| {
...
     ///
     /// The default namespace is "default".
     ///
     /// # Example
     ///
-    /// ```no_run
-    /// # use fluvio_cluster::ClusterInstaller;
-    /// let installer = ClusterInstaller::new()
+    /// ```
+    /// # use fluvio_cluster::{ClusterConfig, ClusterConfigBuilder, ClusterError};
```
sehz, line 65: Suggest simplifying doc.   Instead of duplicating examples.  Just have a single example.   Then each setting will just detail what it does.

src/cluster/src/start/k8.rs
```diff
@@ -49,103 +51,27 @@ static MAX_SC_NETWORK_LOOP: Lazy<u16> = Lazy::new(|| {
...
-    pub fn build(self) -> Result<ClusterInstaller, ClusterError> {
-        Ok(ClusterInstaller {
-            config: self,
-            kube_client: K8Client::default().map_err(K8InstallError::K8ClientError)?,
-            helm_client: HelmClient::new().map_err(K8InstallError::HelmError)?,
-        })
-    }
-
+#[derive(Builder, Debug)]
+#[builder(build_fn(skip), setter(prefix = "with"))]
```
sehz, line 56: I would to remove `with` prefix.  It seems to have more verbose.

src/cluster/src/start/k8.rs
```diff
@@ -560,94 +632,47 @@ impl ClusterInstallerBuilder {
...
+/// # }
 /// ```
 ///
 /// [Helm Charts]: https://helm.sh/
 /// [Minikube]: https://kubernetes.io/docs/tasks/tools/install-minikube/
 #[derive(Debug)]
 pub struct ClusterInstaller {
     /// Configuration options for this installation
-    config: ClusterInstallerBuilder,
+    config: ClusterConfig,
```
sehz, line 651: Do we need another layer of abstraction?  I would prefer we don't make changes to how config works until there is need or better abstraction.

sehz (changes requested):

src/cluster/src/start/k8.rs
nicholastmosher: The reason for needing this change is that previously, we had no config options that did not have a default value. Now, it should be _invalid_ to produce a `ClusterConfig` without first providing a `chart_version`. However, we do not want to put `chart_version: Option<String>` directly into our Config object because then our business logic would need  to check that it is present at every usage point.

Now, we distinguish between the Builder, which is where we collect all of our options, and the Config, which is where we are guaranteed to have all of the options we need. To get from Builder to Config you use `.build()`, which will fail if not all of the options are provided.

src/cluster/src/start/k8.rs
nicholastmosher: This gives us the same API we already had. Our builder functions already used names such as `with_chart_version`. This is just telling the `derive_builder` crate how to take a field name e.g. `chart_version` and produce a method for us, `with_chart_version`

src/cluster/src/start/k8.rs
nicholastmosher: The reason for having individual docs is so that the doc tests will fail if we change the interface. I actually want to open another issue because it turns out our CI has not actually been running doc tests and our docs were out of date.

src/cli/build.rs
nicholastmosher: The CLI already has a build.rs that we need to maintain. I think that this is the simplest way to handle our VERSION needs, I really (_really_) disliked having feature flags that set VERSION="UNDEFINED".

sehz: Can you add `rerun-if-changed` this to `build.rs` ?  https://stackoverflow.com/questions/49077147/how-can-i-force-build-rs-to-run-again-without-cleaning-my-whole-project

src/cluster/src/start/k8.rs
sehz: Still like to have a just a single example

src/cluster/Cargo.toml
```diff
@@ -11,20 +11,19 @@ description = "Tools for installing and managing Fluvio clusters"
 name = "fluvio_cluster"
 path = "src/lib.rs"
 
-[[bin]]
-name = "fluvio-cluster"
-path = "src/main.rs"
-required-features = ["cli"]
+#[[bin]]
```
simlay, line 14: delete?

sehz (approved):

PR #729, opened 2021-01-30
Fix inverted install/upgrade for sys

nicholastmosher:
Clearly this should have been opposite what it was


sehz (approved):

sehz (changes requested):
bump up VERSION to `alpha.2`

sehz:
bump up VERSION to `0.7.0-alpha.2`

sehz (approved):

src/cluster/src/cli/start/sys.rs
```diff
@@ -38,11 +38,11 @@ fn install_sys_impl(
     let installer = SysInstaller::from_config(config)?;
 
     if upgrade {
```
sehz, line 40: Seems like we need another mode?
In one mode, we know exactly whether we want install/upgrade. For example, new installation.
In other case, we don't really care.  If sys chart exists, upgrade is done otherwise new sys chart is installed.

sehz:
hope everything works

PR #731, opened 2021-01-31
Add helm-publish-sys to Makefile and release CI

nicholastmosher:
We have been experiencing build breakages because the fluvio-cluster installer always assumes that for any version of `fluvio-app`, there will be an equivalent version of `fluvio-sys`, but our release.yml currently does not always push them together. This adds a make task and a CI step to always push fluvio-sys with fluvio-app at the same version number.


sehz (approved):
Great!

PR #732, opened 2021-02-01
Add missing chart_version to SysConfig in check

nicholastmosher:
This should fix the remaining CI breakage. The problem was that `SysConfig` was being used in the `cli/check` module but had not been updated since `chart_version` was made into a required field, causing the "missing chart_version" error we were seeing before.

Do we want to force publish over `0.7.0-alpha.2` for this or bump to `0.7.0-alpha.3`?

### DIFF
--- a/src/cluster/src/cli/check.rs
+++ b/src/cluster/src/cli/check.rs
@@ -12,8 +12,7 @@ impl CheckOpt {
     pub async fn process(self, default_chart_version: &str) -> Result<(), ClusterCliError> {
         use colored::*;
         println!("{}", "Running pre-startup checks...".bold());
-        let sys_config: SysConfig = SysConfig::builder()
-            .chart_version(default_chart_version)
+        let sys_config: SysConfig = SysConfig::builder(default_chart_version)
             .build()
             .map_err(ClusterError::InstallSys)?;
         let mut progress = ClusterChecker::empty()
--- a/src/cluster/src/cli/check.rs
+++ b/src/cluster/src/cli/check.rs
@@ -9,10 +9,11 @@ use crate::check::SysChartCheck;
 pub struct CheckOpt {}
 
 impl CheckOpt {
-    pub async fn process(self) -> Result<(), ClusterCliError> {
+    pub async fn process(self, default_chart_version: &str) -> Result<(), ClusterCliError> {
         use colored::*;
         println!("{}", "Running pre-startup checks...".bold());
         let sys_config: SysConfig = SysConfig::builder()
+            .chart_version(default_chart_version)
             .build()
             .map_err(ClusterError::InstallSys)?;
         let mut progress = ClusterChecker::empty()
--- a/src/cluster/src/cli/mod.rs
+++ b/src/cluster/src/cli/mod.rs
@@ -87,7 +87,7 @@ impl ClusterCmd {
                 uninstall.process().await?;
             }
             Self::Check(check) => {
-                check.process().await?;
+                check.process(default_chart_version).await?;
             }
             Self::Releases(releases) => {
                 releases.process().await?;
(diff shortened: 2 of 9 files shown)


src/cluster/src/cli/check.rs
```diff
@@ -9,10 +9,11 @@ use crate::check::SysChartCheck;
 pub struct CheckOpt {}
 
 impl CheckOpt {
-    pub async fn process(self) -> Result<(), ClusterCliError> {
+    pub async fn process(self, default_chart_version: &str) -> Result<(), ClusterCliError> {
         use colored::*;
         println!("{}", "Running pre-startup checks...".bold());
         let sys_config: SysConfig = SysConfig::builder()
```
sehz, line 15: Chart version should be mandatory so this error should caught in compile time

sehz (changes requested):

src/cluster/src/cli/check.rs
nicholastmosher: I have mixed feelings about this. I think that this might make sense for this specific instance of applying the builder pattern, but that the pattern of mandating required arguments be available before even constructing the builder would be an anti-pattern in  general. Imagine a scenario where we have a config with 10 fields and all of them are required. We wouldn't want to have a 10-argument constructor, and we also wouldn't want to have a 10-field "required fields" struct to pass into the builder - that defeats the purpose of the builder.

In my eyes, the purpose for having a builder is to have an intermediate data type which allows us to collect fields incrementally, and potentially from different sources. The builder pattern is not meant to only be used when we have default fields for everything. Imagine a scenario where we have multiple sources where we may be reading configuration options from. I think it is not unreasonable to imagine a case where we load options from a file, from environment variables, and from the CLI all in the same application. The builder is the place where we allow all of these sources to collect these values together, then verify that all of the required values have been collected successfully. A nice pattern with builders is to actually implement helper methods for loading configuration from the various sources, such as:

```rust
impl ConfigBuilder {
    fn load_file(&mut self, path: &Path) -> &mut Self {}
    fn load_env(&mut self) -> &mut Self {}
    fn load_cli(&mut self, opts: SomeOpts) -> &mut Self {}
}
```

Used as

```rust
let config = Config::builder()
    .load_file(&config_path)
    .load_env()
    .load_cli(opts)
    // .hardcoded(value)
    .build()?;
```

One thing to note about this kind of pattern is that no single source might actually need to provide _all_ of the required arguments, but as long as the builder can collect all of the required values by the time `.build()` is called, then everything is fine.

---

Also, I just remembered another thing. `derive_builder` provides `impl Default for Builder`, meaning that it is always possible to construct your `ConfigBuilder` in such a way that it is still empty. That is, even if we put the "required options" into our own `.builder()` constructor method, users would still be able to construct an instance of the builder that has empty required fields. This essentially means that this "compile time check" can be subverted, essentially undermining its usefulness. Before I remembered this point I was ready to just go ahead and put the required fields into the constructor (even despite my previous points), but I think that this is a sufficient argument against doing so. 

Let me know what you think

sehz: In this case, safety/correctness override other concerns.  If we can't disable `derive_builder`, then we should not use `derive_builder`.   

The other alternative is not to specify `chart_version` in the Builder but require in the `build`.   So in this case, use `build(chart_version)`

nicholastmosher: I went ahead and added the chart_version as a parameter to the `builder` constructor. I did that for K8, Local, ans Sys installers. All of our examples use this construction pattern, so if somebody decides to subvert it with Default I think we can call that an advanced use-case and let them deal with handling the Result from `.build()` on their own.

sehz (approved):

PR #733, opened 2021-02-01
Fix broken doc tests in consumer

nicholastmosher:
There have been some API changes that broke some doc tests. For some reason it was not picked up by CI, I will open an issue to make sure doctests are run in CI. This fixes broken doc tests from consumer. This can be verified by running `cargo test --doc`.

### DIFF
--- a/src/client/src/consumer.rs
+++ b/src/client/src/consumer.rs
@@ -37,7 +37,7 @@ use crate::spu::SpuPool;
 ///
 /// ```no_run
 /// # use fluvio::{Fluvio, Offset, ConsumerConfig, FluvioError};
-/// # async fn do_create_consumer(fluvio: &Fluvio) -> Result<(), FluvioError> {
+/// # async fn example(fluvio: &Fluvio) -> Result<(), FluvioError> {
 /// let consumer = fluvio.partition_consumer("my-topic", 0).await?;
 /// let records = consumer.fetch(Offset::beginning()).await?;
 /// # Ok(())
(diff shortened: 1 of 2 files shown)


src/client/src/consumer.rs
```diff
@@ -76,15 +76,12 @@ impl PartitionConsumer {
...
+    /// # async fn example(consumer: &PartitionConsumer) -> Result<(), FluvioError> {
     /// let response = consumer.fetch(Offset::beginning()).await?;
     /// for batch in response.records.batches {
     ///     for record in batch.records {
-    ///         if let Some(record) = record.value.inner_value() {
-    ///             let string = String::from_utf8(record)
-    ///                 .expect("record should be a string");
-    ///             println!("Got record: {}", string);
-    ///         }
+    ///         let string = String::from_utf8_lossy(record.value.as_ref());
```
sehz, line 83: there is new `to_string_lossy` in the record

nicholastmosher: Hmm that's right... I wish I had seen that before we added it, I think it was not a good idea to add `to_string_lossy` to our record type. Is this a blocker in your eyes or could we merge it like this anyways?

sehz: @simlay opinion?

simlay: I'm fine with it as is and we can remove to_string_lossy

sehz (approved):

sehz:
that's fine. I will merge it.

PR #734, opened 2021-02-01
Protocol compatibility for Fluvio client

nicholastmosher:
This change adds a `MINIMUM_PLATFORM_VERSION` constant to the client crate and causes it to check the platform version of the cluster it is communicating with. If the cluster's platform version is below the minimum, then the client will produce a new `FluvioError::MinimumPlatformVersion` error and refuse to connect.

I believe that we may want to have a bigger design discussion about this. I think it's possible that the solution here is rather brittle, since a future protocol change may change the offsets of data and the position of `platform_version` in the message may not be reliable, making this change useless. We probably want to think about creating a stable encoding of the _protocol_ version in such a way that is very future compatible and which can be checked by clients and servers alike.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -961,7 +961,7 @@ dependencies = [
 
 [[package]]
 name = "fluvio"
-version = "0.4.0"
+version = "0.4.1"
 dependencies = [
  "async-channel",
  "async-mutex",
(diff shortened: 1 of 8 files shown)


sehz (changes requested):
looks good.  can you bump up crate version?

sehz (approved):

PR #738, opened 2021-02-02
[Merged by Bors] - Verify topic before producer waits on stdin

### LINKED ISSUE #697: [Bug]: CLI is silent when producing to non-existing topic
When topic `mytopic` does not exist.
```
❯ fluvio produce mytopic
```
does not give any feedback until actually producing a record.

```
❯ fluvio version
Fluvio CLI      : 0.6.1
Fluvio Platform : 0.6.1
Git Commit      : 9f48ca7c14f39fda357932dff59ab9038ee55531
OS Details      : Linux 5.8.0-1017-aws x86_64 GNU/Linux
Rustc Version   : 1.49.0 (e1884a8 2020-12-29)
```

tjtelan:
Add pub async `topic_exists` to TopicProducer
Resolves #697

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Release Notes
 
+## Platform Version 0.8.4 - 2020-05-29
+* Don't hang when check for non exist topic. ([#697](https://github.com/infinyon/fluvio/pull/697))
+  
 ## Platform Version 0.8.3 - 2020-05-25
 * Added builder for fluvio_storage::config::ConfigOption. ([#1076](https://github.com/infinyon/fluvio/pull/1076))
 * Use batch record sending in CLI producer ([#915](https://github.com/infinyon/fluvio/issues/915))
--- a/src/client/src/fluvio.rs
+++ b/src/client/src/fluvio.rs
@@ -124,7 +124,7 @@ impl Fluvio {
         debug!(topic = &*topic, "Creating producer");
 
         let spu_pool = self.spu_pool().await?;
-        if !spu_pool.topic_exists(&topic).await {
+        if !spu_pool.topic_exists(&topic).await? {
             return Err(FluvioError::TopicNotFound(topic));
         }
 
--- a/src/client/src/fluvio.rs
+++ b/src/client/src/fluvio.rs
@@ -123,11 +123,12 @@ impl Fluvio {
         let topic = topic.into();
         debug!(topic = &*topic, "Creating producer");
 
-        if !self.spu_pool().await?.topic_exists(&topic).await {
+        let spu_pool = self.spu_pool().await?;
+        if !spu_pool.topic_exists(&topic).await {
             return Err(FluvioError::TopicNotFound(topic));
         }
 
-        Ok(TopicProducer::new(topic, self.spu_pool().await?))
+        Ok(TopicProducer::new(topic, spu_pool))
     }
 
     /// Creates a new `PartitionConsumer` for the given topic and partition
--- a/src/client/src/fluvio.rs
+++ b/src/client/src/fluvio.rs
@@ -122,6 +122,11 @@ impl Fluvio {
     ) -> Result<TopicProducer, FluvioError> {
         let topic = topic.into();
         debug!(topic = &*topic, "Creating producer");
+
+        if !self.spu_pool().await?.topic_exists(&topic).await {
+            return Err(FluvioError::TopicNotFound(topic));
+        }
+
         Ok(TopicProducer::new(topic, self.spu_pool().await?))
     }
 
--- a/src/client/src/spu.rs
+++ b/src/client/src/spu.rs
@@ -181,9 +181,9 @@ impl SpuPool {
         let replica = ReplicaKey::new(topic, 0);
         self.metadata
             .partitions()
-            .lookup_by_key(&replica)
+            .store()
+            .contains_key(&replica)
             .await
-            .is_ok()
     }
 
     pub fn shutdown(&mut self) {
--- a/src/client/src/spu.rs
+++ b/src/client/src/spu.rs
@@ -181,9 +181,9 @@ impl SpuPool {
         let replica = ReplicaKey::new(topic, 0);
         self.metadata
             .partitions()
-            .store()
-            .contains_key(&replica)
+            .lookup_by_key(&replica)
             .await
+            .is_ok()
     }
 
     pub fn shutdown(&mut self) {
--- a/src/client/src/spu.rs
+++ b/src/client/src/spu.rs
@@ -177,6 +177,15 @@ impl SpuPool {
         Ok(stream)
     }
 
+    pub async fn topic_exists<S: Into<String>>(&self, topic: S) -> bool {
+        let replica = ReplicaKey::new(topic, 0);
+        self.metadata
+            .partitions()
+            .store()
+            .contains_key(&replica)
+            .await
+    }
+
     pub fn shutdown(&mut self) {
         self.metadata.shutdown();
     }
--- a/src/client/src/spu.rs
+++ b/src/client/src/spu.rs
@@ -177,13 +177,11 @@ impl SpuPool {
         Ok(stream)
     }
 
-    pub async fn topic_exists<S: Into<String>>(&self, topic: S) -> bool {
+    pub async fn topic_exists<S: Into<String>>(&self, topic: S) -> Result<bool,FluvioError> {
         let replica = ReplicaKey::new(topic, 0);
-        self.metadata
+        Ok(self.metadata
             .partitions()
-            .store()
-            .contains_key(&replica)
-            .await
+            .lookup_by_key(&replica).await?.is_some())
     }
 
     pub fn shutdown(&mut self) {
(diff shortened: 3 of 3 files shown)


src/extension-consumer/src/produce/mod.rs
```diff
@@ -70,6 +70,15 @@ impl ProduceLogOpt {
         let (cfg, file_records) = self.validate()?;
         let mut producer = fluvio.topic_producer(&cfg.topic).await?;
 
+        // Check if topic exists before reading files or stdin (#697)
+        if !&producer.topic_exists().await {
```
nicholastmosher, line 74: nit: remove the `&` here

nicholastmosher (commented):
Hi @tjtelan, thanks for the PR! I was able to confirm that this works on my machine. To anyone else who wants to test this out, it does take about a minute for the error message to pop up so be patient. The error output looks like this:

```
Error:
   0: Consumer Error
   1: Topic hello not found
```

nicholastmosher:
If you wouldn't mind fixing the one nit and then rebasing, that would be great 🙏

src/extension-consumer/src/produce/mod.rs
```diff
@@ -70,6 +70,15 @@ impl ProduceLogOpt {
         let (cfg, file_records) = self.validate()?;
         let mut producer = fluvio.topic_producer(&cfg.topic).await?;
 
+        // Check if topic exists before reading files or stdin (#697)
```
sehz, line 73: Issue with this approach is that it moves burden of error checking into application.   Instead,  it should return `PartitionNotFound` or `TopicNotFound` from the client.

sehz (changes requested):

src/extension-consumer/src/produce/mod.rs
tjtelan: I am unsure if I understand the objection. However, I did just change the error from `IoError::ErrorKind` to `FluvioError::TopicNotFound` to make the errors fit in with the standard.

src/client/src/producer.rs
```diff
@@ -56,6 +56,12 @@ impl TopicProducer {
 
         send_record_raw(spu_client, &replica, record).await
     }
+
+    /// Returns true if a simple connection to `self.topic/0` is successful
+    pub async fn topic_exists(&self) -> bool {
+        let replica = ReplicaKey::new(&self.topic, 0);
+        self.pool.create_serial_socket(&replica).await.is_ok()
```
sehz, line 63: can you move this to `SpuPool` ?  In that case, can you use `lookup_by_key` and not have to create socket

src/extension-consumer/src/produce/mod.rs
sehz: Issue is that we want to make work for any client not just CLI.  @simlay thoughts?

sehz (changes requested):

src/extension-consumer/src/produce/mod.rs
simlay: Ah, yeah. [`topic_producer`](https://github.com/infinyon/fluvio/blob/dcdfa58286c3472e24a295a1c4c645e2dc8d02c6/src/client/src/client/cluster.rs#L111-L118) should throw this error if the topic doesn't exist.

tjtelan:
I made the changes, but I wanted to point out that the check is slow in my testing. Much slower than how it was before because  `lookup_by_key` has to time out `MAX_WAIT_TIME` before returning.

I took a moment to compare exec time for 3 different scenarios between master and this branch.

On master
```
# Existing topic
$ time echo hello | target/release/fluvio produce hello
Ok!

real    0m0.017s
user    0m0.015s
sys     0m0.008s

---

# Interactive mode, nonexisting topic
$ time target/release/fluvio produce nontopic
[Not printed, I pressed enter here immediately. Otherwise never returns]
Error: 
   0: Consumer Error
   1: Fluvio client error
   2: timed out searching metadata Partition failed due to timeout: 60000 ms

real    1m4.098s
user    0m0.000s
sys     0m0.024s

---

# Piped stdin, nonexisting topic
$ time echo hello | target/release/fluvio produce nontopic
Error: 
   0: Consumer Error
   1: Fluvio client error
   2: Partition not found: nontopic-0

real    0m0.017s
user    0m0.008s
sys     0m0.018s
```

This branch

```
# Existing topic
$ time echo hello | target/release/fluvio produce hello
Ok!

real    0m0.005s
user    0m0.005s
sys     0m0.002s

---

# Interactive mode, nonexisting topic
$ time target/release/fluvio produce nontopic
Error: 
   0: Consumer Error
   1: Fluvio client error
   2: Topic not found: nontopic

real    1m0.033s
user    0m0.042s
sys     0m0.013s

---

# Piped stdin, nonexisting topic
$ time echo hello | target/release/fluvio produce nontopic
Error: 
   0: Consumer Error
   1: Fluvio client error
   2: Topic not found: nontopic

real    1m0.036s
user    0m0.041s
sys     0m0.019s
```

src/client/src/spu/spu_pool.rs
```diff
@@ -159,6 +159,16 @@ impl SpuPool {
         Ok(stream)
     }
 
+    /// Returns true if key `topic/0` is found
+    pub async fn topic_exists<S: Into<String>>(&self, topic: S) -> bool {
+        let replica = ReplicaKey::new(topic, 0);
+        self.metadata
+            .partitions()
+            .lookup_by_key(&replica)
```
sehz, line 167: You are right.  `lookup_by_key` potentially waits for while.  

You can do this to remove wait:

```
        self.metadata
            .partitions()
            .store()
            .contains_key(&replica)
            .await
``

Comment (tjtelan) [reply src/client/src/spu/spu_pool.rs]: Cool. That performs a lot better now.

```
$ time target/release/fluvio produce nontopic
Error: 
   0: Consumer Error
   1: Fluvio client error
   2: Topic not found: nontopic

real    0m0.015s
user    0m0.007s
sys     0m0.014s
```

Comment (sehz) [reply src/client/src/spu/spu_pool.rs]: Nice!

Rebase and we can merge this

**Review by sehz — APPROVED** [review]
Thanks for contribution

**Comment by sehz** [comment]
Examples are failing.  @nicholastmosher idea?

**Comment by nicholastmosher** [comment]
Hey @tjtelan, I took a look at what's going on and it seems to be a regression caused by your changes. Here's what I'm seeing:

When running `examples/tests/test-simple.sh`, the following happens:

- `cargo run --bin fluvio -- topic create simple`
- `cargo run --bin produce` (runs the `examples/produce` binary)
  - It prints `Produce error: TopicNotFound("simple")`, which is incorrect because the `simple` topic does exist

I think it is likely that when you check in `metadata.partitions().store()`, you are observing a cache which has not been sync'd with the cluster.

@sehz let me know if this sounds like a reasonable idea, but I think the best way to implement this would be to use `admin.list` in order to list the current topics, then make sure the topic we want exists. I haven't checked to see if this compiles, but @tjtelan I think it would look something like this:

```rust
let responses = admin.list::<TopicSpec>(vec![my_topic_name]).await?;
let topic_exists = !responses.is_empty();
```

sehz:
> Hey @tjtelan, I took a look at what's going on and it seems to be a regression caused by your changes. Here's what I'm seeing:
> 
> When running `examples/tests/test-simple.sh`, the following happens:
> 
> * `cargo run --bin fluvio -- topic create simple`
> * `cargo run --bin produce` (runs the `examples/produce` binary)
>   
>   * It prints `Produce error: TopicNotFound("simple")`, which is incorrect because the `simple` topic does exist
> 
> I think it is likely that when you check in `metadata.partitions().store()`, you are observing a cache which has not been sync'd with the cluster.
> 
> @sehz let me know if this sounds like a reasonable idea, but I think the best way to implement this would be to use `admin.list` in order to list the current topics, then make sure the topic we want exists. I haven't checked to see if this compiles, but @tjtelan I think it would look something like this:
> 
> ```rust
> let admin = fluvio.admin().await;
> let responses = admin.list::<TopicSpec>(vec![my_topic_name]).await?;
> let topic_exists = !responses.is_empty();
> ```

That is inefficient because it incurs round trip topic and even that may not have data.   Looks like may need new API to expose cache only when it has data

nicholastmosher:
@sehz can we move forward an implement this the inefficient way for now and improve it later? What should the status of this PR be?

sehz:
Need to perform fencing on cache to ensure cache are populated before doing lookup

sehz:
@nicholastmosher should update PR

sehz:
`lookup_and_wait` is API to uses.  Here is example:

```
impl StoreContext<SpuSpec> {
        pub async fn look_up_by_id(
            &self,
            id: i32,
        ) -> Result<CacheMetadataStoreObject<SpuSpec>, FluvioError> {
            self.lookup_and_wait(|g| {
                for spu in g.values() {
                    if spu.spec.id == id {
                        return Some(spu.inner().clone());
                    }
                }
                None
            })
            .await?
            .ok_or(FluvioError::SPUNotFound(id))
        }
    }
```

nicholastmosher:
Since we have an issue to track this, I am going to remove the PR from the engineering board to avoid clutter

sehz:
update the PR with correct API that waits for metadata before checking

PR #739, opened 2021-02-02
Create basedir during `fluvio install` if missing

### LINKED ISSUE #718: [Bug]: `fluvio install` fail if `~/.fluvio/` directory does not exist
**What happened**

The `fluvio install` command reports an error if the `~/.fluvio/` directory does not exist

```
cargo run --bin fluvio -- install fluvio/fluvio-cloud
    Finished dev [unoptimized + debuginfo] target(s) in 0.19s
     Running `target/debug/fluvio install fluvio/fluvio-cloud`
🎣 Fetching latest version for package: fluvio/fluvio-cloud...
⏳ Downloading package with latest version: fluvio/fluvio-cloud:0.1.3...
Error:
   0: Fluvio base directory not found
```

This should not happen via the regular curl | bash flow but we should fix it nonetheless.

**Expected behavior**
The `fluvio install` command should create the `~/.fluvio/` directory if it does not exist.

1) `rm -rf ~/.fluvio`
2) `cargo run --bin fluvio -- install fluvio/fluvio-cloud`

tjtelan:
Removed Err return
Resolves #718

### DIFF
--- a/src/cli/src/install/mod.rs
+++ b/src/cli/src/install/mod.rs
@@ -13,20 +13,21 @@ fn fluvio_base_dir() -> Result<PathBuf, CliError> {
     if let Ok(dir) = std::env::var("FLUVIO_DIR") {
         // Assume this is like `~/.fluvio
         let path = PathBuf::from(dir);
-        return Ok(path);
+        return fluvio_base_dir_create(path);
     }
     let home =
         home::home_dir().ok_or_else(|| IoError::new(ErrorKind::NotFound, "Homedir not found"))?;
     let path = home.join(".fluvio");
 
-    match path.exists() {
-        true => Ok(path),
-        false => {
-            // Create the base dir if it doesn't exist yet (#718)
-            std::fs::create_dir(&path)?;
-            Ok(path)
-        },
+    fluvio_base_dir_create(path)
+}
+
+fn fluvio_base_dir_create(path: PathBuf) -> Result<PathBuf, CliError> {
+    if !path.exists() {
+        // Create the base dir if it doesn't exist yet (#718)
+        std::fs::create_dir_all(&path)?
     }
+    Ok(path)
 }
 
 pub(crate) fn fluvio_extensions_dir() -> Result<PathBuf, CliError> {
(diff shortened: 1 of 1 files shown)


src/cli/src/install/mod.rs
```diff
@@ -19,10 +19,14 @@ fn fluvio_base_dir() -> Result<PathBuf, CliError> {
         home::home_dir().ok_or_else(|| IoError::new(ErrorKind::NotFound, "Homedir not found"))?;
     let path = home.join(".fluvio");
 
-    if path.exists() {
-        return Ok(path);
+    match path.exists() {
+        true => Ok(path),
+        false => {
+            // Create the base dir if it doesn't exist yet (#718)
+            std::fs::create_dir(&path)?;
```
nicholastmosher, line 26: Instead of `create_dir`, use `create_dir_all`, which is more like `mkdir -p` (it will create all parent directories needed)

src/cli/src/install/mod.rs
```diff
@@ -19,10 +19,14 @@ fn fluvio_base_dir() -> Result<PathBuf, CliError> {
         home::home_dir().ok_or_else(|| IoError::new(ErrorKind::NotFound, "Homedir not found"))?;
     let path = home.join(".fluvio");
 
-    if path.exists() {
-        return Ok(path);
+    match path.exists() {
```
nicholastmosher, line 22: We probably don't need a match here, just

```rust
if !path.exists() {
    // ...
}
Ok(path)
```

src/cli/src/install/mod.rs
```diff
@@ -19,10 +19,14 @@ fn fluvio_base_dir() -> Result<PathBuf, CliError> {
...
+        true => Ok(path),
+        false => {
+            // Create the base dir if it doesn't exist yet (#718)
+            std::fs::create_dir(&path)?;
+            Ok(path)
+        },
```
nicholastmosher, line 28: This will work if `FLUVIO_DIR` is not set, but if it is set (above), then it will return early before creating the directory.

nicholastmosher (changes requested):
Thanks again for the PR @tjtelan! I left a few comments about potential edge cases, let me know if you need any more clarification :)

nicholastmosher (approved):
Looks good to me

sehz:
Great. thanks for contribution

PR #740, opened 2021-02-02
Add sha256 hash to `fluvio version`

### LINKED ISSUE #680: Add sha256 hash to `fluvio version`
We should make the Fluvio CLI hash itself and paste that hash into the `fluvio version`. This will give us an indisputably unique identifier of the executable which is being run. The git hash is good, but slightly misleading, since it does not account for working directory changes that have not been committed.

tjtelan:
Minor alignment fix for existing columns
Resolves #680


src/cli/src/lib.rs
```diff
@@ -243,7 +246,13 @@ struct VersionOpt {}
...
 impl VersionOpt {
     pub async fn process(self, target: ClusterTarget) -> Result<()> {
-        println!("Fluvio CLI      : {}", crate::VERSION.trim());
+        println!("Fluvio CLI        : {}", crate::VERSION.trim());
+
+        // Read CLI and compute its sha256
+        let fluvio_bin = File::open(std::env::current_exe()?.into_os_string())?;
+        let reader = BufReader::new(fluvio_bin);
+        let fluvio_bin_sha256 = Sha256::digest(reader.buffer());
+        println!("Fluvio CLI SHA256 : {:x}", fluvio_bin_sha256);
```
nicholastmosher, line 255: I tried this out and the hash this produced did not match the output of hashing it manually via

```
$ shasum -a256 < target/debug/fluvio
```

I believe the problem here is that `reader.buffer()` does not read the full contents of the file, just what the buffered reader has buffered in memory.

Try using `let bytes = std::fs::read(std::env::current_exe()?)?` for reading the file, it's a handy helper that will read the full contents

nicholastmosher (changes requested):
Thanks for the PR @tjtelan! I found one issue which I commented below. Can you check that the hash given in `version` matches a manual hash check? Depending on your system you can use `shasum -a256` or `sha256sum`

src/cli/src/lib.rs
tjtelan: Hm. My manual testing initially had the hash output match the output of `shasum`, but I'm seeing that they are different as well. That's disappointing, and I agree that the buffered reader doesn't have the full file contents.

The reason I was using `BufReader` instead of `std::fs:read` is due in part to a very lengthy pause to the output of `fluvio version`. The output *is correct* using the `std::fs::read` though.

The first line will print, then almost 30 seconds will pass before the remaining version output prints.


```
$ time fluvio version
Fluvio CLI        : 0.7.0-alpha.2
Fluvio CLI SHA256 : e34467b0685a9c611e917c06ee5352a3f7bf83ba863ea4c2dcaa25d5c4b5c007
Fluvio Platform   : 0.7.0-alpha.2
Git Commit        : 7cd3e8df148e4c9a7a042a510e1081f73f1e6155
OS Details        : Linux 5.10.7-3-MANJARO x86_64 GNU/Linux
Rustc Version     : 1.48.0 (7eac88a 2020-11-16)

real    0m26.951s
user    0m26.803s
sys     0m0.147s
```

Just for comparison, here is the timed output for `shasum`
```
$ time shasum -a256 < target/debug/fluvio
e34467b0685a9c611e917c06ee5352a3f7bf83ba863ea4c2dcaa25d5c4b5c007  -

real    0m1.199s
user    0m1.167s
sys     0m0.031s
```

nicholastmosher: I saw this issue in `fluvio cloud version` when we implemented this there. If you build the binary in release mode, the time to self-hash becomes imperceptible, both because it is higher performance and because the binary is smaller

tjtelan: The release mode binary has much better performance. And the output is correct too.

```
$ time target/release/fluvio version
Fluvio CLI        : 0.7.0-alpha.2
Fluvio CLI SHA256 : 3faa035a15e766029300f1ea3748ebda1bc54aff1f90664574245c377544d72b
Fluvio Platform   : 0.7.0-alpha.2
Git Commit        : a7ecb18a03892ec97d0c754e2a1683ced4afad33
OS Details        : Linux 5.10.7-3-MANJARO x86_64 GNU/Linux
Rustc Version     : 1.48.0 (7eac88a 2020-11-16)

real    0m0.188s
user    0m0.146s
sys     0m0.049s
```

Vs  `shasum`
```
$ time shasum -a256 < target/release/fluvio
3faa035a15e766029300f1ea3748ebda1bc54aff1f90664574245c377544d72b  -

real    0m0.175s
user    0m0.165s
sys     0m0.010s
```

nicholastmosher (approved):
LGTM

PR #749, opened 2021-02-04
Protocol socket service

nicholastmosher:
This merges this `fluvio-protocol` and `fluvio-socket` repositories into this one, preserving git history. There are some failing doc tests, but everything else seems to build just fine.


src/fluvio-socket/.github/workflows/ci.yaml
```diff
@@ -0,0 +1,65 @@
+name: CI
```
simlay, line 1: Do we wanna remove this?

nicholastmosher: Yes, there are some things in here that are duplicate infrastructure, but I want to make sure that anything that was unique in fluvio-socket or fluvio-protocol gets merged into the fluvio infra before deleting it

sehz (approved):

PR #751, opened 2021-02-05
Updated dependencies on socket,service and protocol to use path

simlay:
I think this is a follow up from #749.


nicholastmosher (approved):
Everything builds and tests pass on my machine, good job!

PR #706, opened 2021-01-21
Remove optional in DefaultAsyncRecord.

simlay:
~This is dependent on https://github.com/infinyon/fluvio-protocol/pull/4 but I've written the refactor using a work around.~

Depends on https://github.com/infinyon/fluvio-protocol/pull/5

We should also probably rename `try_into_bytes` to `bytes`. 

Record updates:
* Added `AsRef` implementation `as_ref()` returns `&[u8]`
* `to_utf8_lossy` (returns a string)
* `to_bytes` takes ownership and returns a `Vec<u8>`
* `impl TryFrom<Record> for String`

PR #752, opened 2021-02-05
Use encode_varint to encode vector length in DefaultAsyncBuffer and RecordSets.

simlay:
Closes #706.


sehz:
rebase

src/dataplane-protocol/src/record.rs
```diff
@@ -453,9 +431,28 @@ mod test {
         Ok(())
     }
 
-    /// test decoding of records when one of the batch was truncated
     #[test]
-    fn test_decode_batch_truncation() {
```
simlay, line 458: I deleted the truncation stuff because it seems that it's a bad use case to take the bytes and modify them directy. Thoughts @sehz?

sehz: LOL.  That is to  simulate where SPU truncates output bytes in order to comply with max bytes per response.  This is critical test to demonstrate receivers can recover much of batches even thought some of the batches are incomplete.

sehz (changes requested):

simlay:
> LOL. That is to simulate where SPU truncates output bytes in order to comply with max bytes per response. This is critical test to demonstrate receivers can recover much of batches even thought some of the batches are incomplete.

Well then that clears up the importance of it.

src/dataplane-protocol/src/record.rs
simlay: Part of the reason I removed it was because I couldn't figure out how to change the truncation bytes. Given that the SPU truncates it, that'll also have to change (probably why the build is currently failing) so should we keep the encoder changes to `RecordSet` or not?

sehz: We should make it changes only to `DefaultAsyncBuffer`

sehz (changes requested):
crate version need to be updated as well min platform version

simlay:
> crate version need to be updated as well min platform version

I'm pretty sure I bumped all the right versions. It's kind of a breaking change to all things depending on the `dataplane-protocol` crate which means there's a breaking change to just about every crate in the project so that's nice.

-
```diff
@@ -0,0 +1,22 @@
+[package]
```
sehz, line 1: what is this file?

sehz (changes requested):

-
simlay: haha woops. I dunno how `-` got written.

sehz (approved):

PR #754, opened 2021-02-06
make k8 CI run more reliable

sehz:
increase FLV_CLUSTER_MAX_SC_NETWORK_LOOP, VERSION check count to higher to make it K8 run reliable


sehz:
closing in favor for #755

Issue #580, opened 2020-12-19
Simplify error message in the CLI


sehz:
For example, when trying to create duplicate topic;
```
$ fluvio topic create test
Error: 
   0: Consumer Error
   1: Fluvio client error
   2: Fluvio SC schema error
   3: Received error code: TopicAlreadyExists (Some("topic \'test\' already defined"))
```

This is too much information.  We should review and come up with more format

---

This seems to be one of several cases where a somewhat "expected" outcome of running a command should not be rendered as an error. We shouldn't change our general-purpose error-handling mechanism, but rather for the purpose of the `fluvio topic create test` command, we should watch for this specific error and manually render it in a well-defined way.

We should keep an eye out for other commands that exhibit a similar behavior where they are rendering "expected" results as errors, and come up with a uniform way of rendering them nicely.

PR #762, opened 2021-02-08
Change error types back to tuples

nicholastmosher:
A little while ago I made the mistake of turning a bunch of error types into struct-variants. This makes it very difficult to pattern match on errors when we need to inspect them. I would like to change them back to being tuple-variants.

Updating the error enums is a breaking change, so I will open this PR now, and we can either merge it right away or we can wait until we are already ready to do a version bump on a few crates and we can include this in those changes.

The main motivation behind doing this is that it will make implementing things like https://github.com/infinyon/fluvio/issues/580 much easier

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1327,7 +1327,7 @@ dependencies = [
 
 [[package]]
 name = "fluvio-extension-consumer"
-version = "0.2.0"
+version = "0.3.0"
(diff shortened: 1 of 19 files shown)


sehz:
Rebase please

sehz (approved):

PR #763, opened 2021-02-09
Use varint for default async buffer

sehz:


simlay (approved):
🚀

sehz:
k8 works with right version. closing

PR #766, opened 2021-02-09
Cli error printout

### LINKED ISSUE #580: Simplify error message in the CLI

nicholastmosher:
Closes #580 

This adds graceful error handling to `fluvio topic create` (if a named topic already exists) and to `fluvio topic delete` (if the named topic does _not_ already exist).

Topic create used to look like this:

```
❯ cargo run --bin fluvio -- topic create foobar
Error:
   0: Consumer Error
   1: Fluvio client error
   2: Fluvio SC schema error
   3: Received error code: TopicAlreadyExists (Some("topic \'foobar\' already defined"))

```

Topic create now looks like:

```
❯ cargo run --bin fluvio -- topic create foobar
topic "foobar" created

❯ cargo run --bin fluvio -- topic create foobar
Topic already exists
```

Topic delete now looks like this:

```
❯ cargo run --bin fluvio -- topic delete foobar
Topic not found
```

One thing I added here is a new error variant: `CliError::ExitWithCode`. It is intended to signal that an error occurred _but was already handled_, and that after execution has bubbled all the way back up to main, it should exit with a particular error code. This is in order to avoid planting `std::process::exit(code)` throughout the codebase.

### DIFF
--- a/src/cli/src/lib.rs
+++ b/src/cli/src/lib.rs
@@ -19,7 +19,7 @@ use install::update::UpdateOpt;
 use install::plugins::InstallOpt;
 pub use error::{Result, CliError};
 
-use fluvio::{Fluvio, FluvioError};
+use fluvio::Fluvio;
 use fluvio_extension_common::FluvioExtensionMetadata;
 use fluvio_extension_consumer::consume::ConsumeLogOpt;
 use fluvio_extension_consumer::produce::ProduceLogOpt;
@@ -33,9 +33,6 @@ use common::target::ClusterTarget;
 use common::output::Terminal;
 use common::PrintTerminal;
 use fluvio::config::ConfigFile;
-use fluvio_sc_schema::ApiError;
-use fluvio_sc_schema::errors::ErrorCode;
-use fluvio_extension_consumer::ConsumerError;
 
 pub const VERSION: &str = include_str!("VERSION");
 static_assertions::const_assert!(!VERSION.is_empty());
@@ -232,22 +229,7 @@ impl FluvioCmd {
                 produce.process(&fluvio).await?;
             }
             Self::Topic(topic) => {
-                let result = topic.process(out, &fluvio).await;
-                match &result {
-                    Err(ConsumerError::ClientError(FluvioError::ApiError(api))) => match api {
-                        ApiError::Code(ErrorCode::TopicAlreadyExists, _) => {
-                            println!("Topic already exists");
-                            return Err(CliError::ExitWithCode(1));
-                        }
-                        ApiError::Code(ErrorCode::TopicNotFound, _) => {
-                            println!("Topic not found");
-                            return Err(CliError::ExitWithCode(1));
-                        }
-                        _ => result?,
-                    },
-                    Err(_) => result?,
-                    Ok(_) => (),
-                }
+                topic.process(out, &fluvio).await?;
             }
             Self::Partition(partition) => {
                 partition.process(out, &fluvio).await?;
--- a/src/cli/src/lib.rs
+++ b/src/cli/src/lib.rs
@@ -19,7 +19,7 @@ use install::update::UpdateOpt;
 use install::plugins::InstallOpt;
 pub use error::{Result, CliError};
 
-use fluvio::Fluvio;
+use fluvio::{Fluvio, FluvioError};
 use fluvio_extension_common::FluvioExtensionMetadata;
 use fluvio_extension_consumer::consume::ConsumeLogOpt;
 use fluvio_extension_consumer::produce::ProduceLogOpt;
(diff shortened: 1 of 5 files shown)


sehz:
need to be rebased

src/cli/src/lib.rs
```diff
@@ -229,7 +232,22 @@ impl FluvioCmd {
                 produce.process(&fluvio).await?;
             }
             Self::Topic(topic) => {
-                topic.process(out, &fluvio).await?;
+                let result = topic.process(out, &fluvio).await;
+                match &result {
+                    Err(ConsumerError::ClientError(FluvioError::ApiError(api))) => match api {
```
sehz, line 237: Why specific custom handling here instead of just printing out error?

sehz (changes requested):

src/cli/src/lib.rs
nicholastmosher: That was the point of this whole PR. The behavior we want for `fluvio topic create` is:

- If the topic does not already exist, create it and print `topic "X" created`
- If the topic already exists, do nothing, but print `Topic already exists` and return exit code 1

previously, what would happen is that the `TopicAlreadyExists` error would bubble all the way up and get printed as a color_eyre error with the red error stack. Now, that specific error type is handled in our specified way.

nicholastmosher: (similarly for `fluvio topic delete`)

sehz: Why not update `eyere` format?

nicholastmosher: I think that the eyre format is good for what it is intended for. I see it as these are two different cases:

1) When trying to create a topic that already exists, this is a "domain error", something that is well-understood and all we need to do is inform the user what happened.
2) When something unexpected goes wrong, this is a "system error", something that we can consider a bug. When this happens, we want to be presented with as much information as possible so that we can understand where the problem originated. This is where eyre's error stack comes in handy: we can get a summary of where the error originated and a general idea of what went wrong.

sehz: This seems too complex for what we need right now.  Let's just display error and reserve stack based rendering for advanced or debug mode.

nicholastmosher:
@sehz I gave this another shot by moving the error display code to `error.rs` so it is no longer handled right in the app logic. Let me know if this looks better to you.

sehz:
What about other errors?  They all should return -1.  Not sure why only topic need to be handle special.  
```
#[derive(Error, Debug)]
pub enum FluvioError {
    #[error("Topic not found: {0}")]
    TopicNotFound(String),
    #[error("Partition not found: {0}-{1}")]
    PartitionNotFound(String, i32),
    #[error("Spu not found: {0}")]
    SPUNotFound(i32),
    #[error(transparent)]
    IoError(#[from] IoError),
    #[error("Fluvio socket error")]
    FlvSocketError(#[from] FlvSocketError),
    #[error("Fluvio SC schema error")]
    ApiError(#[from] ApiError),
    #[error("Fluvio config error")]
    ConfigError(#[from] ConfigError),
    #[error("Attempted to create negative offset: {0}")]
    NegativeOffset(i64),
    #[error("Cluster (with platform version {cluster_version}) is older than the minimum required version {client_minimum_version}")]
    MinimumPlatformVersion {
        cluster_version: Version,
        client_minimum_version: Version,
    },
    #[error("Unknown error: {0}")]
    Other(String),
}
```

nicholastmosher:
The other errors get bubbled up untouched, meaning that `color_eyre` will print them as the typical error stack and return exit code 1, just like we've already been doing.

To your point about "just displaying the error", the reason I'm having a hard time thinking about that is because it seems like that would still not give us a good output either way. Would we print the Display of the most top-level error? Because then we would end up with a message such as "Fluvio Client error", which tells us almost nothing about what happened. But if we print the Display of the lowest-level error, then we could end up with a message such as "file not found (OS error X)", which also does not help the user to know what has happened. To me, the only way to communicate what has actually happened is to either include the context of the entire stack of errors, or to convert that stack of errors into a single statement that describes the situation, which is what I'm trying to do in this PR.

sehz:
There doesn't need to be just one way to display error.  In this case, we should optimize for simple case.  Can add flag or env variable to display detail stack as before like `BACKTRACE` flag for rust binary.   

I also think we need to separate error into separate categories to make it easier to handle in the future.

sehz (approved):

PR #773, opened 2021-02-11
Fixed fluvio cluster delete when sys chart not installed

nicholastmosher:
I finally figured out why `fluvio cluster delete` was behaving weird but only sometimes. If `fluvio-sys` was installed, `fluvio cluster delete` works just fine. BUT, if `fluvio-sys` was not installed, then we would get a Kubernetes 404 during the uninstall step because we were attempting to interact with CRD objects that weren't there.

This is a quick fix that skips certain uninstall steps if fluvio-sys is not installed.

### DIFF
--- a/src/cluster/src/delete.rs
+++ b/src/cluster/src/delete.rs
@@ -3,12 +3,13 @@ use std::process::Command;
 use std::fs::remove_dir_all;
 
 use tracing::{info, warn, debug, instrument};
-use fluvio_future::timer::sleep;
 use k8_client::ClientError as K8ClientError;
 use k8_client::{load_and_share, SharedK8Client};
 use k8_client::http::status::StatusCode;
 use k8_types::Spec;
 use k8_types::core::pod::PodSpec;
+use fluvio_future::timer::sleep;
+use fluvio_command::CommandExt;
 
 use crate::helm::HelmClient;
 use crate::{DEFAULT_CHART_APP_REPO, DEFAULT_NAMESPACE, DEFAULT_CHART_SYS_REPO};
@@ -141,7 +142,7 @@ impl ClusterUninstaller {
         let client = load_and_share().map_err(UninstallError::K8ClientError)?;
 
         self.wait_for_delete::<PodSpec>(client, "fluvio-sc").await?;
-        self.cleanup().await?;
+        self.cleanup().await;
 
         Ok(())
     }
@@ -169,7 +170,7 @@ impl ClusterUninstaller {
             )
             .map_err(UninstallError::HelmError)?;
         debug!("fluvio sys chart has been uninstalled");
-        self.cleanup().await?;
+        self.cleanup().await;
 
         Ok(())
     }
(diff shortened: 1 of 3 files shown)


sehz:
why would this be problem?  If cluster is installed, we assume sys chart is installed otherwise how would cluster be installed.  This also break other workflows where we don't want to uninstall sys chart.

One option is to add explicit flag to uninstall `sys` chart

nicholastmosher:
The case that I ran into was during testing. In the test flow, it first calls `fluvio cluster delete`, _then_ it goes to install the cluster to perform the test. If you are starting from a fresh system, this will prevent the test from running because it fails at the delete step.

nicholastmosher:
Since there seems to be some question on this PR I have moved the clippy fixes to another PR

sehz:
relevant?

nicholastmosher:
Yes, actually. `fluvio cluster delete` still exhibits this behavior. This PR is ready to go

src/cluster/src/delete.rs
```diff
@@ -248,13 +248,32 @@ impl ClusterUninstaller {
...
+            .iter()
+            .nth(0)
+            .is_some();
+
+        Ok(sys_chart_installed)
+    }
+
     /// in order to remove partitions, finalizers need to be cleared
+    #[instrument(skip(self))]
     async fn remove_partitions(&self, namespace: &str) -> Result<(), UninstallError> {
```
sehz, line 266: this won't work for object objects as well.  Instead,  change `remove_objects` to ignore errors

sehz (changes requested):

src/cluster/src/delete.rs
```diff
@@ -206,28 +207,33 @@ impl ClusterUninstaller {
...
     /// Remove objects of specified type, namespace
-    fn remove_objects(
+    fn remove_custom_objects(
         &self,
         object_type: &str,
         namespace: &str,
         selector: Option<&str>,
     ) -> Result<(), UninstallError> {
+        // If fluvio-sys is not installed, there are no custom objects to remove
+        if !self.is_sys_installed()? {
```
sehz, line 233: let's make it more less dependent on `sys-charts`.  Just try to remove it and ignore errors.

sehz: sys installed check is repeated for every call, either only done once or just don't check

src/cluster/src/delete.rs
```diff
@@ -243,18 +249,36 @@ impl ClusterUninstaller {
             info!("deleting all {} in: {}", object_type, namespace);
             cmd.arg("--all");
```
sehz, line 250: So I suggest check if if succeed or not, if fail, don't return error.   In fact, signature could be as:
```
fn remove_custom_objects(
         &self,
         object_type: &str,
         namespace: &str,
         selector: Option<&str>,
     ) -> Result<bool, UninstallError>
```

If it failed to delete, return false

sehz (changes requested):

src/cluster/src/delete.rs
nicholastmosher: I forgot to remove this when I made deletes non-error. Should be good now

src/cluster/src/delete.rs
nicholastmosher: I actually just made everything ignore the results of deleting, which I think does what we need

nicholastmosher:
@sehz I think this is ready to go

sehz (approved):

PR #781, opened 2021-02-12
Use public consumer API for CLI consumer

### LINKED ISSUE #397: CLI: refactor consume to use documented API
Currently, CLI uses undocumented API to perform consumer fetch and display.   We need to convert this to use published API

nicholastmosher:
This closes #397 

- Simplify CLI consumer
- Reduce dependency of CLI consumer on the internal details of API consumer
- Enhance API consumer with any required information

### DIFF
--- a/src/client/src/consumer.rs
+++ b/src/client/src/consumer.rs
@@ -11,12 +11,14 @@ use futures_util::stream::{StreamExt, once, iter};
 use fluvio_spu_schema::server::stream_fetch::{DefaultStreamFetchRequest, DefaultStreamFetchResponse};
 use dataplane::Isolation;
 use dataplane::ReplicaKey;
+use dataplane::ErrorCode;
 use dataplane::fetch::DefaultFetchRequest;
 use dataplane::fetch::FetchPartition;
 use dataplane::fetch::FetchableTopic;
 use dataplane::fetch::FetchablePartitionResponse;
 use dataplane::record::RecordSet;
 use dataplane::record::DefaultRecord;
+use dataplane::batch::DefaultBatch;
 use fluvio_types::event::offsets::OffsetPublisher;
 
 use crate::FluvioError;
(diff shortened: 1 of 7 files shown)


src/client/src/consumer.rs
```diff
@@ -392,7 +405,13 @@ impl ConsumerConfig {
 
 /// The individual record for a given stream.
 pub struct Record {
```
sehz, line 407: these are already in the batch header why creating duplicate information?

sehz:
Instead of trying to use a single monolithic api for everything,  create different iterator for different use case.

For example, in the `HashMap`,  there is `values` and `keys`.    Likewise, we can have `batches` which will iterate batches or  `record` which will may have tuple of headers, and records.

src/client/src/consumer.rs
nicholastmosher: I removed the `partition_index` and `log_start_offset` as they ended up not being needed by the CLI consumer

nicholastmosher:
In the newest update, there is now `stream_with_config` and `stream_batches_with_config` as the basic streaming-consumer functions.

src/client/src/consumer.rs
```diff
@@ -390,9 +437,53 @@ impl ConsumerConfig {
     }
 }
 
+pub struct RecordBatch {
```
sehz, line 440: Instead of creating another duplicated struct, can you add to Batch in the data plane

sehz (changes requested):

src/client/src/consumer.rs
nicholastmosher: Got it. I added new methods `.records()` and `.into_records()` that return iterators over the `DefaultRecord`s inside of the batch.

sehz:
can you rebase on this PR? https://github.com/infinyon/fluvio/pull/769

nicholastmosher:
After #769 is merged I will rebase this one more time so that the diff is clear

sehz:
Please rebase from master since #769 is merged. thanks

sehz (approved):
LGTM.

PR #795, opened 2021-02-18
Encapsulate host and port in Endpoint struct

nacardin:


sehz:
Still relevant?

sehz:
Still needed?

Issue #676, opened 2021-01-11
CLI Design: `fluvio produce`


nicholastmosher:
It feels like there is some good opportunity to improve the `fluvio produce` CLI options. Right now there are several ways to do one thing, and the arguments that are available are not really composable. I want to gather some requirements about what capabilities the `fluivo produce` command should have, and what information it should make available to the user.

## Requirements

Please check the boxes below to indicate your thoughts on each requirement:

1) I should be able to provide input from `stdin` or from named files
- This would allow me to pipe each line of some command output as a record
- E.g. `rg "hello" | fluvio produce my-topic [some args]`
  - Each discovered line containing "hello" would be produced as a record
2) When providing input from named files, I should be able to use `xargs`
- E.g. when taking filenames to produce from, all trailing arguments should be accepted
- E.g. `cat record-names.txt | grep ".toml" | xargs fluvio produce my-topic [some args]`
  - Expands to something like `fluvio produce my-topic -f one.toml two.toml three.toml four.toml`
3) I should be able to take the specified input and produce it line-by-line or as a single record
4) I should be able to choose whether to observe confirmation (via e.g. stdout) after each record that is produced
- Today, we observe confirmation with an `Ok!` printout after each record or file is produced
- When observing confirmation, it should be structured in a script-friendly way
  - E.g. when producing one line at a time, confirmation output should be each line produced
  - E.g. when producing one file at a time, confirmation output should be the filename of each file produced
5) I should be able to view human-friendly output without interfering with script-friendly output
- Script-friendly output should go to stdout
- Human-friendly output should go to stderr

Consumer concerns:

1) I should be able to gracefully deal with non-text files, i.e. binary
- Does it make sense to attempt to render non-text content to a terminal?
   - aj: we don't have a good understanding of this use case. Maybe we should remove until users ask. @sehz?
- Should we provide a mechanism to predictably place binary content into files?
  - aj: same as above, I suggest we defer under we understand the use case. @sehz?
  - E.g. instead of attempting to print a `file.jpg` to the terminal, simply place it into a `file-000001.jpg` file.
  - Maybe we can define an argument that takes a file-template-string, e.g `--template=image-{{OFFSET}}.jpg`

@infinyon/dev are there any more requirements you'd like to see here? I will begin writing a proposed interface that hopefully will meet these requirements

---

## Proposed Interface

Here, I'll list out the specific arguments that I think would fulfill each use case above, and demonstrate some example usages and expected output.

```
fluvio-produce

[...]

PR #797, opened 2021-02-18
Simplified producer CLI

nicholastmosher:
This PR implements the core ideas of #676


sehz:
this is marginally better than existing feature.  I would rather see more generalized way to break down lines so can handle more variety of uses cases (i.g. CSV).

https://docs.rs/regex/0.2.11/regex/struct.Split.html
https://doc.rust-lang.org/std/primitive.str.html#method.split

nicholastmosher:
Superseded by upcoming key/value producer

PR #802, opened 2021-02-18
Ignore pkill failure in CI

nacardin:


simlay (approved):
Seems okay. I wish CI would tell us which command exited with nonzero.

sehz:
only way is to split the command

PR #814, opened 2021-02-23
Remove rustc version printout from 'fluvio version'

### LINKED ISSUE #784: CLI: Remove Rustc version from `fluvio version`
It doesn't seem like having the Rustc version that the CLI was compiled with gives us much value, so we can remove it.

nicholastmosher:
Closes #784


simlay (approved):
zoom zoom lgtm

PR #815, opened 2021-02-23
Merge Master into Stable as part of the 0.7 Release

simlay:
I'll wait until we're sure we're done with the release to merge this but we're nearly there.


sehz (approved):
This should be either merged or rebased not squashed.

Issue #566, opened 2020-12-15
Test runner infrastructure


sehz:
There are many tests that we would like to run where we want to have a `fluvio` client which is connected to a fresh fluvio cluster. We would like a test harness that can automatically provision this fresh cluster as easily as possible so that it is easy to run reproducible tests.

Implement procedure macro to mark test running test runner:

For example: 

```
#[fluvio_test(min_spu=2,topic=”test”)]
async fn smoke_test(config: TestConfig,client: Client) -> Result<(),...>
{
      let produce =  client.partition_producer(“topic”).await;
      …..

} 

#[cfg(type=”local”]
#[test]
Fn  smoke_test_runner()  {
  
       async fn smoke_test(config: TestConfig,client: Client) -> Result<(),...>
       }


       Let test_setup = new TestRunner()
       smoke_test(...)
      test_setup.cleanup()
  
}

#[fluvio_test(cluster=k8,min_spu=2,topic=”test”)]
async fn k8_test1(config: TestConfig,client: Client,cluster: &Cluster) -> Result<(),...>
{
      let produce =  client.partition_producer(“topic”).await;
      …..

} 

#[fluvio_test(cluster=local,min_spu=2,topic=”test”)]
async fn election_test(config: TestConfig,client: Client,cluster: &Cluster) -> Result<(),...>
{
      let produce =  client.partition_producer(“topic”).await;
      …..
     clutter.shutdown_spu(5001).await.expect(“shutdown”);
     ….

} 
```

They can be run using test runner:
```
fluvio_test_runner test smoke_test
fluvio_test_runner test --spu=2,
```

* Test runner run any test matches name pattern:  `smoke_test`, `smoke` (smoke_test,smoke_test1,test). similar to `cargo test`.  (Per requirements review, only a single test needs to be selected)  
* And tests must satisfy criteria matched in the test option (Optional minimum test criteria is declared via proc-macro attributes). 
    * e.g., `-k8` that test must able able to run on k8.
    * e.g., if test specifies, `min_spu=3`,  but test option specifics 1, then it will not able to criteria.  It should then print out how many tests it matches and failed to matches (name).
* Test runner can pass any of the options in the Test Runner as command line option [to be used for setting up a test environment, and to pass to a test]
* ~~Test runner can run group of tests using regex test runner name.~~ (Removed from scope, per requirements review)
* It is okay to only search for tests in `fluvio-integration-tests` crate.  The crate must only have tests.  
[...]

Issue #808, opened 2021-02-20
Generate latest and nightly release


sehz:
To make it easier to consume fluvio, we should perform following development releases
- nightly (every night)
- latest (every commit)

Difference between regular and development release is that these release won't be added to Github releases.  The release should:
- Generate Fluvio binary
- Generate Helm Chart
- Generate Docker Image

tjtelan:
It appears that most of this is complete.
* Latest docker images are built and pushed to Docker Hub (per commit)
* Nightly docker images are build and pushed to Docker Hub
* Helm charts are published (per commit)

The last piece missing are Fluvio binary per commit. Where should Fluvio development binaries be stored?

PR #820, opened 2021-02-24
Reorganize integration test setup/cleanup

### LINKED ISSUE #759: Create cluster integration test harness
We need more tests in fluvio, and one of the best ways to encourage more tests is to make them easier to write. Currently it is very difficult to write integration tests because it requires setting up and tearing down the environment very precisely every time. If we create a test harness to do this for us, it should make this process easier.

I believe the test harness should be responsible for the following:

- Create a new namespace for each individual test to be run (note this will require running on Kubernetes)
- Install a Fluvio cluster in each test's namespace
- Populate a `config: FluvioConfig` with the cluster connection details for each test's cluster
- Instantiate a `fluvio: Fluvio` client via `Fluvio::connect_with_config(&config)` which will connect to the test's specific cluster
- Pass the `fluvio` client into the body of the test

We can use the `fluvio-cluster` crate to implement this setup, and we can write a procedural macro to hide these details from the test author. Let me illustrate what the non-proc-macro version might look like:

```rust
#[fluvio_future::test_async]
async fn test_create_topic() -> Result<(), ()> {
    let namespace: String = generate_namespace(); // May be UUID or pseudo-random unique english name
    let cluster_config: ClusterConfig = fluvio_cluster::ClusterConfig::builder()
        .namespace(&namespace)
        // .add_any_other_configurations()
        .build()?;
    let installer = ClusterInstaller::from_config(cluster_config)?;
    let status = installer.install_fluvio().await?;
    let fluvio_config = FluvioConfig::new(status.address);
    let fluvio = Fluvio::connect_with_config(&fluvio_config).await?;

    // Enter test body with `fluvio`
}
```

With the help of a procedural macro, we could clean that up to look like this instead:

```rust
#[fluvio_cluster::test]
async fn test_create_topic(fluvio: Fluvio) {
    // Test body
}
```

We could add any of the cluster builder options as arguments to the proc macro, such as:

```rust
#[fluvio_cluster::test(
    namespace = "...",
    image_tag = "...",
    chart_location = "...",
    chart_version = "...",
)]
```

@infinyon/dev I'd like to get your input as to whether any of the ideas here should be changed, added to, or whether you can see any potential problems with it. I have a few unresolved questions already:

- Should this harness live in `fluvio_cluster`? This could create a circular dependency if we added tests to the `fluvio` crate
- Would this harness allow tests to be run in parallel? At first glance it seems like that could work, but I'm not 100% sure
- (bikeshed) are there any other preferred names than `#[fluvio_cluster::test]`?
- Should we add a test timeout? (this could actually be pushed down into the cluster installer itself)

---

Update: This issue is too big to be its own action item, so I'm breaking it down into bite-sized pieces:

- [ ] Clean up test runner to use programmatic fluvio-cluster rather than using `fluvio cluster start` CLI (https://github.com/infinyon/fluvio/issues/774)
- [ ] Parameterize tests to run iteratively or using random data in order to increase confidence (https://github.com/infinyon/fluvio/issues/777)
- [ ] Create proc macros to minimize boilerplate from testing helper methods

tjtelan:
Refactor the `flv-test` test runner and integration tests. Temporarily removes dynamic `TestRunner` in favor of a static test selection.

Test cluster setup now provides a `Fluvio` client to pass into user tests.

Proc macro has 2 optional arguments:
* topic
* min_spu

```rust
#[fluvio_test(min_spu=2, topic="topic")]
async fn run(client: Arc<Fluvio>, option: TestOption) {...}
```

Above is a test that requires 2 spu and during setup, it's topic will be named `topic`.

---
2 tests are available. (`--develop` mode is more consistent. Selecting a release version is not stable. It is out of scope to address in this PR.)

```shell
make build_tests
./target/debug/flv-test smoke --develop
./target/debug/flv-test concurrent --develop
```

Related #566 
Closes #759


tests/integration/src/tests/concurrent.rs
```diff
@@ -0,0 +1,101 @@
...
+
+const TOPIC: &str = "test-bug";
+const PARTITION: i32 = 0;
+
+type Record = Vec<u8>;
+
+#[fluvio_test_macro::fluvio_test(topic = "concurrent")]
+fn concurrent_test() {
+    // Hardcoding for now, until ToTrait is implemented for TestOption
+    let test_option = TestOption::default();
```
nicholastmosher, line 19: Note that the only thing that test_option is used for in this particular test case is to manually set up the topic. If the macro has already covered the topic setup, then TestOption is not needed here.

tests/integration/src/tests/smoke.rs
```diff
@@ -0,0 +1,34 @@
...
+        .arg(&format!("{}{}", test_option.topic_name, 0)) // Fix this assumption in the smoke test
+        .arg("--replication")
+        .arg(&test_option.replication().to_string());
+    if let Some(log) = &test_option.client_log {
+        command.env("RUST_LOG", log);
+    }
+
+    let _output = command
+        .result()
+        .expect("fluvio topic create should succeed");
```
nicholastmosher, line 24: (I know you're probably still working on this so if you already have plans for this just ignore me)

Once you have the topic-creation-via-macro all worked out, you can get rid of this shelling-out-to-create-topic code.

tests/integration/src/tests/smoke.rs
```diff
@@ -0,0 +1,34 @@
+#[allow(unused_imports)]
+use fluvio_command::CommandExt;
+
+#[fluvio_test_macro::fluvio_test(topic = "smoke")]
+fn smoke_test() {
+    // Magic variable
+    println!("TestOption: {:?}", &test_option);
```
nicholastmosher, line 7: Are you planning on getting rid of the magic variable? I don't think it should be included in the final thing

tests/integration/src/fluvio-test-options/src/lib.rs
```diff
@@ -0,0 +1,262 @@
+use structopt::StructOpt;
```
nicholastmosher, line 1: This entire crate seems like it is revolving around TestOption, when TestOption is really just test parameters for the one single smoke test. I think that the test macro should not need to interact with TestOption at all, we should push TestOption down into the single test case. I believe all of the cluster-related options in TestOption can be represented directly via the `fluvio_cluster::ClusterConfig`.

tests/integration/src/fluvio-test-macro/src/lib.rs
```diff
@@ -0,0 +1,114 @@
...
+            use serde::Deserialize;
+
+            fluvio_future::subscriber::init_logger();
+
+            // Magic variable
+            let test_option : TestOption = serde_json::from_str(#test_opt_str).unwrap();
+
+            run_block_on(async move {
+                fn cleanup_cluster() {
+                    run_block_on(async move {
```
nicholastmosher, line 50: Just make this `async fn cleanup_cluster() {}` and remove the run_block_on

tests/integration/src/fluvio-test-macro/src/lib.rs
```diff
@@ -0,0 +1,114 @@
...
+                    .skip_checks(true)
+                    .save_profile(true)
+                    .build()
+                    .unwrap();
+                let installer = ClusterInstaller::from_config(cluster_config).unwrap();
+                let status = installer.install_fluvio().await.expect("");
+
+                // Magic variables
+                let fluvio_config = FluvioConfig::new(status.address());
+                let fluvio = Fluvio::connect_with_config(&fluvio_config).await.unwrap();
```
nicholastmosher, line 94: Since `cleanup_cluster` is its own function, it may make sense to put this in a function like `setup_cluster`. Also note that this code is sitting directly next to the test body, so all of these values would become accessible via "magic values" in the body, which I don't think is great

tests/integration/src/tests/concurrent.rs
tjtelan: Good to know. Though, it would be convenient to have the option to inject the number of iterations. Especially if we want to scale or longevity test on a regular frequency.

tests/integration/src/tests/smoke.rs
tjtelan: Yes. I have that println!() there as on-going debugging / a reminder to get rid of the variable. Providing the test writer with the test context for their own test case setup will not be provided through magic variables. (Or at least, it won't have the appearance of coming from nowhere).

Sehyo gave me an example that I'll be trying out. Using a function signature like:

```rust
#[fluvio_test_macro::fluvio_test(topic = "smoke")]
async fn smoke_test(test_opt: IntegrationTestOptionX) { ... }
```

Where `test_opt` can be the name of the var that I end up creating with the macro. The macro args and initialized cluster info available in a more obvious way.

tests/integration/src/fluvio-test-options/src/lib.rs
tjtelan: When I abandoned running our tests w/ `cargo test`, I ended up not needing to make up complex ways to pass a test option. 

Tests are now passed a `Fluvio` client and a `TestOption`.

(`TestOption`s are still specialized to the smoke tests at the moment)

tests/runner/src/main.rs
```diff
@@ -1,41 +1,50 @@
...
+        // Connect to cluster in profile
+        Fluvio::connect().await.unwrap()
+    } else {
+        let mut test_cluster = fluvio_test_util::setup::TestCluster::new(option.clone());
+        test_cluster.start().await.unwrap()
+    };
+
+    // TODO: Build this with Test Runner
+    if let Some(test_name) = &option.test_name {
+        match test_name.as_str() {
```
sehz, line 29: this doesn't scale, we can't hard code tests every time we add test

tests/runner/src/main.rs
```diff
@@ -1,41 +1,50 @@
...
+            "smoke" => flv_test::tests::smoke::run(fluvio_client, option.clone()).await,
+            "concurrent" => flv_test::tests::concurrent::run(fluvio_client, option.clone()).await,
+            "many_producers" => {
+                flv_test::tests::many_producers::run(fluvio_client, option.clone()).await
+            }
+            _ => panic!("Test not found"),
+        };
+    }
+
+    //// Create a default topic
```
sehz, line 39: remove these lines

tests/runner/src/tests/concurrent.rs
```diff
@@ -0,0 +1,104 @@
...
+
+pub async fn test_concurrent_consume_produce(option: &TestOption) {
+    println!("Testing concurrent consumer and producer");
+    setup_produce_then_consume(option);
+    let (sender, receiver) = std::sync::mpsc::channel();
+    spawn(consumer_stream(receiver));
+    producer(sender).await;
+}
+
+fn setup_produce_then_consume(option: &TestOption) {
```
sehz, line 35: topic should be provisioned

sehz (changes requested):

Makefile
```diff
@@ -46,18 +46,18 @@ endif
...
 
 smoke-test-tls:	test-clean-up build_test
-	$(TEST_BIN) --spu ${DEFAULT_SPU} --produce-iteration ${DEFAULT_ITERATION} --tls --local ${TEST_LOG} ${SKIP_CHECK}
+	$(TEST_BIN) smoke --spu ${DEFAULT_SPU} --produce-iteration ${DEFAULT_ITERATION} --tls --cluster-type local ${TEST_LOG} ${SKIP_CHECK}
 
 smoke-test-tls-policy:	test-clean-up build_test
 	AUTH_POLICY=$(SC_AUTH_CONFIG)/policy.json X509_AUTH_SCOPES=$(SC_AUTH_CONFIG)/scopes.json  \
 	FLV_SPU_DELAY=$(SPU_DELAY) \
-	$(TEST_BIN) --spu ${DEFAULT_SPU} --produce-iteration ${DEFAULT_ITERATION} --tls --local ${TEST_LOG} ${SKIP_CHECK}
+	$(TEST_BIN) smoke --spu ${DEFAULT_SPU} --produce-iteration ${DEFAULT_ITERATION} --tls --cluster-type local ${TEST_LOG} ${SKIP_CHECK} --skip-cluster-delete
```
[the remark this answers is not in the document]
tjtelan: This is a workaround for the existing test. The `flv-test` now tries to clean up after running a test, but `smoke-test-tls-policy` was being used to set up the actual test target `smoke-test-k8-tls-root`.

Makefile
```diff
@@ -46,18 +46,18 @@ endif
 #
 
 smoke-test:	test-clean-up	build_test
-	$(TEST_BIN) --spu ${DEFAULT_SPU} --produce-iteration ${DEFAULT_ITERATION} --local ${TEST_LOG} ${SKIP_CHECK} ${EXTRA_ARG}
+	$(TEST_BIN) smoke --spu ${DEFAULT_SPU} --produce-iteration ${DEFAULT_ITERATION} --cluster-type local ${TEST_LOG} ${SKIP_CHECK} ${EXTRA_ARG}
```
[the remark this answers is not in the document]
tjtelan: Instead of `--local` it is `--cluster-type local`. And for k8 is is `--cluster-type k8`. I thought offering the options in this manner was clearer to use.

Makefile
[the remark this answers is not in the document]
sehz: but why only for `smoke-test-tls-policy` not other.  I expect to apply consistently

tests/runner/src/main.rs
tjtelan: Understood. I was landing short so the change would not be as large and risky for this PR because we use the smoke tests.

We still only support smoke tests running successfully, and nothing more. But we have an "improved but still static" interface to add new tests. A more scaled approach will come in the next PR if that is ok with you?

Makefile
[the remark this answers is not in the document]
sehz: it's more typing and we don't have other cluster type so it's not offer advantage.

Makefile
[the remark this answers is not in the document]
tjtelan: That's a really good question. `smoke-test-k8-tls-root` is not a consistent test. It is a negative test that only exists in the makefile, and not in the rest of the test code. My priority was to decouple the runner and keep the tests the same.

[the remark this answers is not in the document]
sehz: I was referring to: `smoke-test-tls-policy` which should have same argument as other scenario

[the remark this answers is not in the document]
tjtelan: It should, but the CI only refers to `smoke-test-k8-tls-root`. `smoke-test-tls-policy` is currently being used as a setup step.

The `test-clean-up` step skips cleanup when `UNINSTALL` is set. (Which is how it is used in the `master` branch). It is effectively the same as `--skip-cluster-delete`. This took time some to figure out.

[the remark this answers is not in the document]
sehz: k

Makefile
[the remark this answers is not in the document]
tjtelan: Changed back to `--local`

tests/runner/src/fluvio-test-framework/src/lib.rs
```diff
@@ -0,0 +1,30 @@
+use proc_macro::TokenStream;
+use proc_macro2::Span;
+
+use syn::{ItemFn, parse_macro_input, Ident};
+use quote::quote;
+
+// This macro will collect minimum Fluvio cluster requirements for runner
```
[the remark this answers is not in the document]
tjtelan: Expanded code provided

sehz:
Can you squash commit?

tests/runner/Cargo.toml
```diff
@@ -1,16 +1,17 @@
...
 name = "flv-test"
 version = "1.0.1"
-authors = ["fluvio.io"]
+authors = ["Fluvio Contributors <team@fluvio.io>"]
 edition = "2018"
+autotests = false
 
 [dependencies]
 log = "0.4.8"
-bytes = "0.5.3"
```
sehz, line 9: don't update this yet since other parts of fluvio is not upgrade to this version

tests/runner/Cargo.toml
```diff
@@ -1,16 +1,17 @@
...
+bytes = "1.0.1"
 futures-lite = "1.11.0"
 structopt = "0.3.5"
 async-trait = "0.1.21"
-tokio = { version = "0.2.21", features = ["macros"] }
+tokio = { version = "1.2.0", features = ["full"] }
```
sehz, line 14: ditto

tests/runner/src/fluvio-integration-derive/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    let out_fn_iden = Ident::new(&test_name.to_string(), Span::call_site());
+
+    let output_fn = quote! {
+
+        pub async fn #out_fn_iden(client: Fluvio, option: TestOption) {
+            use fluvio_test_util::test_meta::FluvioTestRequirements;
+            let test_reqs : FluvioTestRequirements = serde_json::from_str(#fn_test_reqs_str).expect("Could not deserialize test reqs");
+            if test_reqs.min_spu > option.spu {
+                panic!("Test requires {} spu", test_reqs.min_spu)
```
sehz, line 48: if min SPU is not satisfied then test should be skipped not panic

tests/runner/src/fluvio-integration-derive/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+///    let test_reqs : FluvioTestRequirements = [...]
+///    if test_reqs.min_spu > option.spu {
+///        panic!("Test requires {} spu", test_reqs.min_spu)
+///    }
+///     println!("Test")
+/// }
+#[proc_macro_attribute]
+pub fn fluvio_test(args: TokenStream, input: TokenStream) -> TokenStream {
+    let fn_test_reqs: FluvioTestRequirements = parse_macro_input!(args as AttributeArgs).into();
+    let fn_test_reqs_str =
```
sehz, line 31: Idea is that test attribute should be parsed.  With json parsing, this doesn't propagated to parser and we have no feedback

tests/runner/src/fluvio-integration-derive/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+                panic!("Test requires {} spu", test_reqs.min_spu)
+            }
+
+            #test_body
+        }
+
+    };
+
+    output_fn.into()
+}
```
sehz, line 57: Need unit test for procedure macro.  See `node-bindgen`

tests/runner/src/tests/concurrent.rs
```diff
@@ -0,0 +1,81 @@
...
+use fluvio_test_util::test_meta::TestOption;
+
+const TOPIC: &str = "test-bug";
+const PARTITION: i32 = 0;
+
+type Record = Vec<u8>;
+
+use fluvio_integration_derive::fluvio_test;
+
+#[fluvio_test]
```
sehz, line 16: In this case, topic requirement should be specified in procedure macro and test runner should set that up.

tests/runner/src/tests/concurrent.rs
```diff
@@ -0,0 +1,81 @@
...
+
+pub async fn test_concurrent_consume_produce(_option: &TestOption) {
+    println!("Testing concurrent consumer and producer");
+    let (sender, receiver) = std::sync::mpsc::channel();
+    spawn(consumer_stream(receiver));
+    producer(sender).await;
+}
+
+async fn consumer_stream(digests: Receiver<String>) {
+    let fluvio = Fluvio::connect().await.unwrap();
```
sehz, line 30: should  use client that is passed

tests/runner/src/tests/smoke.rs
```diff
@@ -0,0 +1,12 @@
+#[allow(unused_imports)]
+use fluvio_command::CommandExt;
+
+use fluvio_integration_derive::fluvio_test;
+use fluvio::Fluvio;
+use fluvio_test_util::test_meta::TestOption;
+
+#[fluvio_test(min_spu = 1)]
```
sehz, line 8: since there will be always be a single SPU, this is redundant

sehz (changes requested):

tests/runner/src/fluvio-integration-derive/src/lib.rs
sehz: See:  https://github.com/infinyon/node-bindgen/blob/master/nj-derive/src/ast/attribute.rs.

tests/runner/src/tests/concurrent.rs
sehz: Each test should be able to specify what topic they needed.  Before test is invoked, all specifies should be provisioned, so tests are ensure that topics are ready to be produce and consume

tests/runner/src/fluvio-integration-derive/src/lib.rs
sehz: this looks like syntax problem.  it should been:
```
let output_fn = quote! {

       pub async fn #out_fn_iden(client: Fluvio, option: TestOption) {
                use fluvio_test_util::test_meta::TestRequirements;
        }
};
```

tjtelan: It's a syntax problem of from That is the output from cargo. The line numbers jump from line 55 to line 80.

The only difference is the method for passing the `TestRequirements` struct into `quote!{}`

```
let test_reqs : TestRequirements = serde_json::from_str(#fn_test_reqs_str).expect("Could not deserialize test reqs");
```
vs
```
let test_reqs : TestRequirements = #fn_test_reqs
```

sehz: Define what should be in the generated code then reverse map to template.

sehz:
test runner failed:
```
ues: smoke, concurrent, many_producers]
ubuntu@ip-172-31-20-82:~/fluvio$ flvt concurrent
deleting cluster
installing cluster
panic panicked at 'Failed to install k8 cluster: InstallK8(HelmChartNotFound("fluvio/fluvio-app:0.7.1-alpha.0"))', tests/runner/src/utils/setup/environment/k8.rs:54:14
ubuntu@ip-172-31-20-82:~/fluvio$ flvt smoke
deleting cluster
installing cluster
panic panicked at 'Failed to install k8 cluster: InstallK8(HelmChartNotFound("fluvio/fluvio-app:0.7.1-alpha.0"))', tests/runner/src/utils/setup/environment/k8.rs:54:14
ubuntu@ip-172-31-20-82:~/fluvio$ 
```

tjtelan:
@sehz 

I know this works in CI because it uses `--develop`, but I am not familiar with how to select the correct version otherwise. I will look into this.

tjtelan:
@sehz 
This is the same behavior that exists in `master`.  What version am I supposed to reference for Release-mode testing, and how will I find that version at build time?

My only experience with this working has been with the `--develop` flag in CI. Is this dependent on #808?

```
$ ./target/debug/flv-test
using k8 environment driver
installing cluster
panic panicked at 'called `Result::unwrap()` on an `Err` value: InstallK8(HelmChartNotFound("fluvio/fluvio-app:0.7.1-alpha.0"))', tests/runner/src/environment/k8.rs:50:42
```

---
This might be a different problem, because it doesn't work for me w/ `--develop` on master locally either. Only works in CI.

```
$ ./target/debug/flv-test --develop
using k8 environment driver
installing cluster
[...]
```

The pod won't start
```
NAME                        READY   STATUS             RESTARTS   AGE
fluvio-sc-7fd4d9cb4-r9wzs   0/1     ImagePullBackOff   0          69s
```

Because it is trying to pull this image, which doesn't exist in Docker Hub

```
image: infinyon/fluvio:b7b3d2c16880ea21f1c67246ce9765368c991ebe
```

tjtelan:
There was a topic naming convention bug in the smoke test that was getting hit when `--produce-iteration` value == 1. It should be fixed now

sehz:
please update https://github.com/infinyon/fluvio/blob/master/tests/README.md

sehz:
this doesn't work as per doc:
```
flvt --skip-cluster-start --produce-iteration 200
```

sehz:
can you make create short cuts for `--skip-cluster-delete` `--skip-cluster-start`.

tjtelan:
> can you make create short cuts for `--skip-cluster-delete` `--skip-cluster-start`.

Sure, they are really long to type! 

I've renamed `--skip-cluster-start` back to it's original flag `--disable-install` (with the short flag `-d`)

And `--skip-cluster-delete` to `--keep-cluster`, with the short flag `-k`.

The README is updated using the long flags, but you should see the short flags in the `--help` text.

```
    -d, --disable-install    don't attempt to delete cluster or start a new cluster before test topic creation will be
                             skipped
    -k, --keep-cluster       don't delete cluster after test
```

sehz (approved):
LGTM.  Will merge after 0.7.1 release

PR #828, opened 2021-03-02
Update producer and consumer to allow sending and receiving Key/Value records

nicholastmosher:
Here's a list of changes in this PR:

- Added `Record::key(&self) -> &[u8]` and `Record::value(&self) -> &[u8]` in the consumer
- Removed `impl AsRef<[u8]> for Record` in consumer because we now want to explicitly distinguish between keys and values
- Added `TopicProducer::send_key_value_record(&self, key: K, value: V, partition: i32)` to producer
- Changed the representation of record keys in the protocol from `key: B` to `key: Option<B>` (where B is a byte buffer)
- Added a test that shows that decoding a record with an old encoding _just so happens_ to work as expected
  - Old records are always represented as "empty keys" in the form of zero-length buffers
  - Zero-length buffers are encoded as a length whose value is zero `0x00`.
  - The new record encoding represents "absent keys" in the form of `Option::None`
  - `Option::None` tag is encoded as `0x00` (as opposed to `Option::Some`, whose tag is `0x01`)
  - Therefore, decoding _old records with empty keys_ into _new records with absent keys_ works out-of-the-box
  - However, we must never attempt to decode a new record with any key using the older decoder.

Some potentially-open-questions:
- Are we happy with the new function names? These names are all potentially up for debate:
  - `Record::key`
  - `Record::value`
  - ~`TopicProducer::send_key_value_record`~
  - `TopicProducer::send`


ajhunyady:
@nicholastmosher, is there a sample CLI on how to use this feature? We need to prepare a simple app or blog to introduce it.

nicholastmosher:
@ajhunyady this PR does not yet add a CLI interface for using key/value records. It occurred to me that K/V on the CLI may take some significant additional design consideration. We can either hold this PR until the CLI is ready, or we can merge this and work on the CLI interface as a separate PR. The CLI should not require any additional changes to the Rust client, so it would be fine to merge this first and CLI second if we wanted to.

The primary unanswered question about the CLI is: How should a user distinguish between the contents of the keys and the values in their data? The reason this is tricky is because, until now, the CLI has largely not cared about the contents of messages - all records were sent as simple binary buffers, with the single exception to this being the ability to send records "line-by-line", interpreting newlines as a delimiter between records in a single file.

With Key/Value records, we now need a way for users to indicate what data should be treated as a Key and what data should be treated as a Value. I am thinking of opening an issue for tracking potential ideas about how to do this. Here are some initial thoughts of mine on this:

- One easy method would be to specify one file whose contents will be a key and one file whose contents will be a value:
  - E.g. `fluvio produce topic --key=key.txt --value=value.txt`
  - In its simplest form, this command invocation would send only a single record

- What if we wanted to specify multiple key/value records at the same time?
  - We might say that one file holds one-key-per-line and another file holds one-value-per-line
  - E.g. `fluvio produce topic --lines --key=key.txt --value=value.txt`
  - This relies on the assumption that they key/value data is textual, so would not work for binary data
  - This would rely on keys and values being written on corresponding lines in the respective files
  - Could we define some sort of globbing for key-files versus value-files? How would we associate the right keys with the right values?
    - `fluvio produce topic --keys="key-file-*.txt" --values="value-file-*.txt"`

- What if we wanted to co-locate key and value data in a single file?
  - We now need a way to encode keys and values separately in a single file
  - For textual data, something simple like CSV might be sufficient: keys and values are comma-separated, unique records are line-separated
  - We would need another solution for binary data

- Do we want to start adding support for specific types of encoded data?
  - We could add support for JSON/YAML/other object files
  - E.g. the entire file is treated as a record's value, and a keypath is given to "pick out" the data to use for the key
  - Perhaps `fluvio produce topic -f data.json --key-path=".name"`
  - We would need to choose some type of json-path DSL

There are probably more considerations as well. I will add these notes and anything else I can think of to an issue just for the CLI, but needless to say it is something that still needs more thinking-through.

ajhunyady:
Hmmm... this looks very complicated to me and it shouldn't be.

sehz:
For CLI, here is one way to do:
```
$ fluvio produce --key-seperator="="
name=John Doe 
country=US
```

sehz:
We should simplify API much as possible.  
In most cases, K/V will be typical:
```
producer.send("name","doe")
```

For case, where only values is send:
```
producer.send(None,"doe")
```

ajhunyady:
I suggest keeping the .send() for values and have an alternative API for send KV. That way the most common use case `send` will remain unchanged.

sehz:
K/V will be more typical use case rather than other way around.

ajhunyady:
In that case, send it is...

ajhunyady:
> For CLI, here is one way to do:
> 
> ```
> $ fluvio produce --key-seperator="="
> name=John Doe 
> country=US
> ```

There should be some indication this is a key/value record.

sehz:
`--kv` to indicate kv value.

```
$ fluvio produce --kv --separator "="
name=John Doe 
country=US
```

nicholastmosher:
I have added basic CLI support for Key/Value to this PR. Here is the basic usage pattern:

```
fluvio-produce 0.5.1
Write messages to a topic/partition

By default, this reads input from stdin until EOF, and sends the contents as one
record. Alternatively, input files may be given with '-f' and input may be sent
line-by-line with '-l'. In any case, the producer continues reading until it
encounters EOF.

USAGE:
    fluvio produce [FLAGS] [OPTIONS] <topic>

FLAGS:
    -l, --lines    Send each line of input as its own record (using '\n')
    -q, --quiet    Do not print progress output when sending records
    -h, --help     Prints help information

OPTIONS:
    -p, --partition <integer>
            The ID of the Partition to produce to [default: 0]

        --key-separator <key-separator>
            Sends key/value records split on the first instance of the
            separator. Implies --lines
    -f, --files <files>...
            Paths to files to produce to the topic. If absent, producer will
            read stdin

ARGS:
    <topic>    The name of the Topic to produce to
```

And here are some examples of producing and consuming records with keys and values.

All examples are using a topic named `kv-example`.

### Producing from a File

```
$ cat data.txt
alice=Alice in Wonderland
bob=Bob the Builder
eve=Eve the Listener

$ cargo run --bin fluvio -- produce kv-example --key-separator "=" -f data.txt
data.txt:0 alice=Alice in Wonderland
data.txt:1 bob=Bob the Builder
data.txt:2 eve=Eve the Listener
Ok!
```

### Producing from stdin

```
$ cargo run --bin fluvio -- produce kv-example --key-separator "="
bunny=The Easter Bunny
stdin:0 bunny=The Easter Bunny
zeus=God of Lightning
stdin:1 zeus=God of Lightning
^D
Ok!
```

### Consuming

```
$ cargo run --bin fluvio -- consume kv-example -B --key-separator="::"
alice::Alice in Wonderland
bob::Bob the Builder
eve::Eve the Listener
bunny::The Easter Bunny
zeus::God of Lightning
```

src/client/src/consumer.rs
```diff
@@ -579,22 +580,13 @@ impl Record {
...
 
-    pub fn to_string_lossy(&self) -> String {
-        String::from_utf8_lossy(self.as_ref()).to_string()
+    /// Returns the contents of this Record's key, if it exists
+    pub fn key(&self) -> Option<&[u8]> {
+        self.record.key.as_ref().map(|it| it.as_ref())
     }
-}
 
-impl AsRef<[u8]> for Record {
```
simlay, line 587: This is a breaking change so update the client `Cargo.toml` to `0.6.0` as the node client uses `as_ref`.

examples/00-produce/src/main.rs
```diff
@@ -25,7 +25,10 @@ async fn main() {
 async fn produce() -> Result<(), fluvio::FluvioError> {
     let producer = fluvio::producer("simple").await?;
     let message = "Hello, Fluvio!";
-    producer.send_record(&message, 0).await?;
+    producer
+        .send_key_value_record("Producer key", &message, 0)
```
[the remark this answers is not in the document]
nicholastmosher: Good point, I will add an instance of each 👍

src/client/src/consumer.rs
nicholastmosher: Yes, you're right. I would prefer to go ahead and make this change and bump the version to 0.7.0 unless anybody objects

examples/01-produce-key-value/src/main.rs
```diff
@@ -0,0 +1,40 @@
...
+//!
+//! After running this example, you can see the messages that have
+//! been sent to the topic using the following command:
+//!
+//! ```text
+//! $ fluvio consume key-value -B -d
+//! [Hello] Fluvio
+//! ```
+
+#[async_std::main]
```
sehz, line 24: Can we collapse this into produce example? (with different binary). Otherwise, we are starting to create too many example crates.
Or alternatively, combine with original produce example

examples/02-consume/src/main.rs
```diff
@@ -47,7 +47,7 @@ async fn consume() -> Result<(), fluvio::FluvioError> {
     let mut stream = consumer.stream(fluvio::Offset::beginning()).await?;
 
     while let Some(Ok(record)) = stream.next().await {
-        let bytes = record.as_ref();
+        let bytes = record.value();
         let string = String::from_utf8_lossy(&bytes);
```
sehz, line 51: Why this need ref since value returns ref

src/client/src/producer.rs
```diff
@@ -25,6 +25,45 @@ impl TopicProducer {
...
+        let key = key.as_ref();
+        let value = value.as_ref();
+        debug!(
+            key_size = key.len(),
+            value_size = value.len(),
+            "sending records:"
+        );
+
+        let replica = ReplicaKey::new(&self.topic, 0);
+        let spu_client = self.pool.create_serial_socket(&replica).await?;
```
sehz, line 59: This is not optimized.  But looks this was previous implementation.  let's create issue to optimize this (not creating unnecessary virtual connection every time)

src/dataplane-protocol/src/record.rs
```diff
@@ -341,42 +341,6 @@ where
     }
 }
 
-impl<B> From<String> for Record<B>
```
sehz, line 344: Why remove this?

src/dataplane-protocol/src/record.rs
```diff
@@ -480,7 +444,11 @@ mod test {
         use crate::record::DefaultRecord;
 
         fn create_batch() -> DefaultBatch {
-            let record: DefaultRecord = vec![0x74, 0x65, 0x73, 0x74].into();
+            let value = vec![0x74, 0x65, 0x73, 0x74];
+            let record = DefaultRecord {
+                value: DefaultAsyncBuffer::new(value),
```
sehz, line 449: Should not use `DefaultAsyncBuffer`.   It should be something like:
`DefaultRecord::new(key, value)`

src/dataplane-protocol/src/record.rs
```diff
@@ -511,6 +479,66 @@ mod test {
         let decoded_batches = RecordSet::decode_from(&mut Cursor::new(out), 0).expect("decoding");
         assert_eq!(decoded_batches.batches.len(), 2);
     }
+
+    #[test]
+    fn test_key_value_encoding() {
+        let key = "KKKKKKKKKK".to_string();
+        let value = "VVVVVVVVVV".to_string();
+        let record = DefaultRecord {
```
sehz, line 487: See comments previously

sehz (changes requested):

examples/02-consume/src/main.rs
nicholastmosher: I think the general pattern I see in rust is the following naming convention:
- `value(&self)` the simple name takes self by immutable reference
- `value_mut(&mut self)` the `_mut` name takes self by mutable reference
- `into_value(self)` the `into_` name consumes self by ownership

so I think `value` is the right choice here

src/dataplane-protocol/src/record.rs
nicholastmosher: Previously, we were using `Record` almost as if it were a newtype wrapped around `value`, a single piece of data in the Record. Now that we are using both key and value, it does not make sense to me that we have all of these conversions that are heavily biased towards `value` and ignore `key`. Plus this was largely unused anyways.

examples/02-consume/src/main.rs
sehz: Actually, it was not about name.   
```
String::from_utf8_lossy(&bytes);
```
Since value() is ref, shouldn't be:
```
String::from_utf8_lossy(bytes);
```

src/client/src/producer.rs
nicholastmosher: https://github.com/infinyon/fluvio/issues/840

src/dataplane-protocol/src/record.rs
sehz: It still useful and this breaks contract with previous version unnecessary.

examples/01-produce-key-value/src/main.rs
nicholastmosher: The main reason I did not combine the two is because the regular `produce` example is actually used in a test. So there wasn't really a great way to keep the examples small and clean while also satisfying the test and putting both together.

sehz: fair enough.

src/dataplane-protocol/src/record.rs
nicholastmosher: I went ahead and put this back 👍

examples/02-consume/src/main.rs
nicholastmosher: Oh I see, you're right, I will fix that

nicholastmosher: Fixed

src/client/src/consumer.rs
```diff
@@ -579,22 +580,13 @@ impl Record {
...
-    fn as_ref(&self) -> &[u8] {
+    /// Returns the contents of this Record's value
+    pub fn value(&self) -> &[u8] {
         self.record.value.as_ref()
     }
 }
-
```
sehz, line 592: revert back this as well

src/dataplane-protocol/src/record.rs
```diff
@@ -329,51 +329,51 @@ where
...
+        DefaultRecord {
+            key: Some(DefaultAsyncBuffer::new(key.into())),
+            value: DefaultAsyncBuffer::new(value.into()),
             ..Default::default()
         }
     }
 }
 
-impl<B> From<Vec<u8>> for Record<B>
+impl<B> From<String> for Record<B>
```
sehz, line 356: ? it was from Vec<u8>

src/dataplane-protocol/tests/file_fetch.rs
```diff
@@ -115,8 +120,10 @@ async fn setup_client(addr: &str) -> Result<(), FlvSocketError> {
     assert_eq!(batches.len(), 1);
     let records = &batches[0].records();
     assert_eq!(records.len(), 2);
-    assert_eq!(records[0].value.to_string(), "record 0");
-    assert_eq!(records[1].value.to_string(), "record 1");
+    assert_eq!(records[0].value.to_string(), "value 0");
+    assert_eq!(records[0].key.as_ref().unwrap().to_string(), "key 0");
```
sehz, line 124: `records[0].key().to_string()`

src/dataplane-protocol/tests/file_fetch.rs
```diff
@@ -115,8 +120,10 @@ async fn setup_client(addr: &str) -> Result<(), FlvSocketError> {
...
+    assert_eq!(records[1].value.to_string(), "value 1");
+    assert_eq!(records[1].key.as_ref().unwrap().to_string(), "key 1");
```
sehz, line 126: records[1].key().unwrap().to_string()

sehz (changes requested):

sehz:
@simlay Can you review this as well?

src/dataplane-protocol/src/record.rs
nicholastmosher: Oh there were a handful of them I think. Do you want them all back?

src/dataplane-protocol/tests/file_fetch.rs
nicholastmosher: There is no `key()` method for `dataplane::record::Record`. Do you want me to add one?

sehz: Yes.  we should add them

src/dataplane-protocol/src/record.rs
sehz: Yes, don't need to break old client unnecessary

src/client/src/consumer.rs
```diff
@@ -579,22 +580,19 @@ impl Record {
         self.offset
     }
 
-    pub fn to_string_lossy(&self) -> String {
-        String::from_utf8_lossy(self.as_ref()).to_string()
+    /// Returns the contents of this Record's key, if it exists
+    pub fn key(&self) -> Option<&[u8]> {
+        self.record.key.as_ref().map(|it| it.as_ref())
```
sehz, line 585: Actually, key and value should move to `DefaultRecord`.  
Can use `Deref` to delegate methods

nicholastmosher: Hm, I think using Deref for anything except smart-pointers is [considered an anti-pattern](https://github.com/rust-unofficial/patterns/blob/master/anti_patterns/deref.md), but I can make this method just forward to the `key()` method of `DefaultRecord` 👍

sehz: Regarding `Deref`.  I disagree.  It's being used frequently and I see no problem here.
Alternative, then just implement `inner()` which will return DefaultRecord.  Caller just need to do extra hop

Creating delegation method is really anti-pattern in my opinion.

src/client/src/producer.rs
```diff
@@ -25,6 +25,45 @@ impl TopicProducer {
...
+    /// The partition that the record will be sent to is derived from the Key.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use fluvio::{TopicProducer, FluvioError};
+    /// # async fn example(producer: &TopicProducer) -> Result<(), FluvioError> {
+    /// producer.send("Key", "Value").await?;
+    /// # Ok(())
+    /// # }
```
simlay, line 39: What's up with this docstring?

src/client/src/lib.rs
```diff
@@ -102,15 +103,19 @@ pub use crate::admin::FluvioAdmin;
 pub use crate::client::Fluvio;
 
 /// The minimum VERSION of the Fluvio Platform that this client is compatible with.
-const MINIMUM_PLATFORM_VERSION: &str = "0.7.0-alpha.5";
+const MINIMUM_PLATFORM_VERSION: &str = "0.7.1-alpha.0";
 
-/// Creates a producer that sends events to the named topic
+/// Creates a producer that sends records to the named topic
```
simlay, line 108: nit: "sends records or keys/values to the named topic" or something.

simlay (changes requested):
Sehyo managed to cover most of my concerns about things. I had some minor nits.

Also what's the file `src/dataplane-protocol/tests/test_old_record_empty_key.bin`?

src/client/src/producer.rs
nicholastmosher: So this code block will be compiled as Rust code to ensure we don't break the code in the example. In the rust doc-string code-blocks, any line starting with a `#` will not be displayed on the docs page. In this example, the only thing the user will see on docs.rs is 

```
producer.send("Key", "Value").await?;
```

they will not see the rest of this code block, which is nice to reduce visual boilerplate and get right to the meat of the example.

nicholastmosher:
@simlay that `.bin` file is what I used to verify that records in the new schema can be successfully decoded from old records that were encoded with the old schema. See https://github.com/infinyon/fluvio/pull/828/files#diff-e50536c96490b842b8f91fab3fca42353b0beca28ebe99ecd1eb80bcfc53c4b6R568-R606

Basically, I used the old encoder to encode a specific record into bytes, and saved those bytes into `test_old_record_empty_key.bin`. Then I wrote a test which attempts to decode a new-schema-record from those same bytes, and assert that the data landed where it was supposed to in the new record.

src/client/src/lib.rs
nicholastmosher: When I think "record" I consider that to include both the key and the value. This seems like the shortest way to describe it, saying "records or key/values" seems clumsy, and this is the one-line doc description.

simlay: :+1:

src/client/src/producer.rs
simlay: Cool.

simlay (approved):
I think I'm good with this.

sehz (approved):

sehz:
Can you squash commits and produce clean commit logs?

PR #829, opened 2021-03-03
Python client mvp

### LINKED ISSUE #494: Python Client Library MVP
I wanted to write down some of the research I've done on the python-rust-setup. Similar to the node-client library, the python client library should use fluvio rust crate. To do this, I think we should use some combination of:
* https://github.com/dgrunwald/rust-cpython
* https://github.com/PyO3/setuptools-rust
* https://github.com/PyO3/maturin
* https://github.com/mozilla/uniffi-rs

I think the first version we should have the rust toolchain be a dependency of this library but once we've got things more stable, we should have the the setup.py download the fluvio static library and link to it.

Examples:
* https://github.com/etesync/etebase-py - uses [flapigen](https://crates.io/crates/flapigen)

For the MVP, this task is to:
* Depend on having the rust toolchain to build the rust side
* Only have consume and produce

simlay:
Closes #494.


simlay:
Closing in favor of https://github.com/infinyon/fluvio-client-python/pull/1

Issue #700, opened 2021-01-20
CLI: `fluvio consume` should give user feedback


nacardin:
From @ajhunyady:

We had quite a bit of discussion about this, we all fell into the trap thinking that there is a bug in the system because we forgot to pass -B.
By default the consumer is waiting for new messages, which seems reasonable but not all that intuitive. For new users it seems like something went wrong and the system hangs.

The main question is who is the user for this command? Is it always users or there could also be scripts?
I would assume, we expect users, in which case we should provide some level of feedback. Maybe read the HOL and say something to the extent "2345 records stored" then "Waiting for new records... ".

nacardin:
I agree, when consume is called by a user, it should indicate offset and that it is waiting for new records. Creating separate issue to discuss consume cli.

nicholastmosher:
In the CLI we have a notion of "interactive mode", which happens when the user is sitting and watching the command on their terminal, rather than piping the output somewhere. We can detect which case is happening.

I think a good solution for this would be:

- In "interactive mode", print a notice on stderr to the effect of "Waiting for new records to arrive. To see all records, use -B"
- Don't print anything in non-interactive mode, so that the messages don't end up in somebody's records file

PR #830, opened 2021-03-04
CLI Key/Value interface

nicholastmosher:
Adds a key/value interface for the producer and consumer CLI. Should not be merged before #828.

The first draft of the documentation for these CLI changes is here: https://github.com/infinyon/fluvio-website/pull/58


src/extension-consumer/src/consume/mod.rs
```diff
@@ -74,6 +92,160 @@ impl ConsumeLogOpt {
...
+        config: ConsumerConfig,
+    ) -> Result<(), ConsumerError> {
+        let mut stream = consumer.stream_with_config(offset, config).await?;
+
+        while let Some(result) = stream.next().await {
+            let result: Result<Record, _> = result;
+            let record = match result {
+                Ok(record) => record,
+                Err(FluvioError::ApiError(ApiError::Code(code, _))) => {
+                    println!(
```
sehz, line 159: this should be output using `eprintln`.
topic and partition information is redundant since it is already specified by the user and repeat on every record.

src/extension-consumer/src/consume/mod.rs
```diff
@@ -74,6 +92,160 @@ impl ConsumeLogOpt {
...
+                    continue;
+                }
+                Err(other) => return Err(other.into()),
+            };
+
+            self.print_record(record.key(), record.value())?;
+        }
+
+        debug!("fetch loop exited");
+        println!("Consumer stream has closed");
```
sehz, line 174: use `eprintln`

src/extension-consumer/src/consume/mod.rs
```diff
@@ -74,6 +92,160 @@ impl ConsumeLogOpt {
...
+        Ok(())
+    }
+
+    /// Process fetch topic response based on output type
+    pub fn print_record(&self, key: Option<&[u8]>, value: &[u8]) -> Result<(), OutputError> {
```
sehz, line 179: this should be pass as record? which should be more robust on more changes

src/extension-consumer/src/produce/mod.rs
```diff
@@ -1,149 +1,269 @@
...
+/// If '--key-separator' or '--json-path' are used, records are sent as
+/// key/value pairs, and the keys are used to determine which partition
+/// the records are sent to.
 #[derive(Debug, StructOpt)]
 pub struct ProduceLogOpt {
     /// The name of the Topic to produce to
     #[structopt(value_name = "topic")]
     pub topic: String,
 
-    /// The ID of the Partition to produce to
```
sehz, line 35: why these options are removed?

src/extension-consumer/src/produce/mod.rs
```diff
@@ -1,149 +1,269 @@
...
+    /// Print progress output when sending records
+    #[structopt(short, long)]
+    pub verbose: bool,
+
+    /// Sends key/value records split on the first instance of the separator.
+    #[structopt(long, validator = validate_key_separator)]
+    pub key_separator: Option<String>,
+
+    /// Sends key/value JSON records where the key is selected using this JSON path.
+    #[structopt(long, conflicts_with("key-separator"))]
```
sehz, line 46: This is out of scope for key/value.  I would prefer this as separate PR to discuss how to consume/produce json

src/extension-consumer/src/produce/mod.rs
```diff
@@ -1,149 +1,269 @@
...
+    pub json_path: Option<String>,
+
+    /// Paths to files to produce to the topic. If absent, producer will read stdin.
+    #[structopt(short, long)]
+    pub files: Vec<PathBuf>,
```
sehz, line 51: I don't think we should handle multiple files which is opinionated process (for example, what is ordering of file input).  Also, we should support short format `-f`

src/extension-consumer/src/produce/mod.rs
```diff
@@ -1,149 +1,269 @@
...
-        short = "r",
-        long = "record-file",
-        value_name = "filename",
-        parse(from_os_str),
-        conflicts_with = "record-per-line"
-    )]
-    record_file: Vec<PathBuf>,
+    /// Send each line of input as its own record (using '\n')
+    #[structopt(hidden = true, short, long)]
+    pub lines: bool,
```
sehz, line 35: if we are going to support lines, let's rename this as `split` and support splitting arbitrary

src/extension-consumer/src/produce/mod.rs
```diff
@@ -1,149 +1,269 @@
...
-            Some(FileRecord::Files(self.record_file))
-        } else {
-            None
-        };
+    /// Sends records to a Topic based on the file configuration given
+    ///
+    /// This will either send the lines of a single file as individual records,
+    /// or it will send the entirety of a list of files as records, where each
+    /// whole file is one record.
+    async fn produce_from_files(&self, producer: &mut TopicProducer) -> Result<()> {
```
sehz, line 90: ditto. see comments about files

sehz (changes requested):

src/extension-consumer/src/produce/mod.rs
```diff
@@ -1,149 +1,269 @@
...
-            produce_from_files(&mut producer, cfg, records).await?;
+        // --key-separator implies --lines
+        if self.key_separator.is_some() {
+            self.lines = true;
+        }
+
+        if self.files.is_empty() {
+            self.lines = true;
+            if atty::is(atty::Stream::Stdin) {
+                eprintln!("Reading one record per line from stdin:");
```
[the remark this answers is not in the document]
nicholastmosher: What about this is redundant?

src/extension-consumer/src/consume/mod.rs
```diff
@@ -34,6 +48,10 @@ pub struct ConsumeLogOpt {
     #[structopt(short = "d", long)]
     pub disable_continuous: bool,
 
+    /// Print records in "[key] value" format, with "[null]" for no key
+    #[structopt(short, long)]
+    pub key_value: bool,
```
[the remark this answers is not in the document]
sehz: actually, this flag is confusing.  it sounds like it is about consuming.  it should be more like: `--print-kv-header`.  Also printing key should be default so in this case, it should be `-disable-print-kv-header`

src/extension-consumer/src/produce/mod.rs
nicholastmosher: Partition is now auto-assigned, and "continuous" was not being used.

src/extension-consumer/src/produce/mod.rs
```diff
@@ -1,149 +1,269 @@
...
-            partition: self.partition,
-            continuous: self.continuous,
-        };
+        Ok(())
+    }
 
-        Ok((produce_log_cfg, file_records))
+    /// Sends each line from a file as a unique record to the producer's topic.
+    async fn produce_file_lines(&self, producer: &mut TopicProducer, path: &Path) -> Result<()> {
+        let file = File::open(path)?;
```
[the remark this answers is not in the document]
nicholastmosher: Hm I am reading some of your other comments and I think I see what you're getting at. I will see if I can condense some of this. I left it separate previously because I had originally wanted to print status messages that were unique to the mode being used (e.g. print file with line number, etc.). But I will take a look at this now.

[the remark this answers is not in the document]
sehz: file is input (it implements async `read` and `write`), there is no need to differentiate from stdin or any other input.  So we should have a single pipeline for producer just that source is different

src/extension-consumer/src/produce/mod.rs
sehz: Yes, make sense for `partition` but why remove `disable_continuous`?

src/extension-consumer/src/consume/mod.rs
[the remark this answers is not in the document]
nicholastmosher: Originally I had made the consumer print keys by default, but @ajhunyady said he preferred otherwise. I don't have a preference one way or the other. Let me know what you guys decide.

src/extension-consumer/src/produce/mod.rs
nicholastmosher: `disable_continuous` is a consumer flag, `-d`, I did not remove that. What I removed here was the `--continuous` flag for producer, which had no code usages, it was dead code.

src/extension-consumer/src/produce/mod.rs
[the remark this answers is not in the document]
nicholastmosher: Ok, I consolidated the implementations, things are a bit cleaner now. I also remembered why I split it the way that I did: it was because there were slightly different strategies for printing acknowledgement to the user based on where the input was coming from. For example, reading from stdin would cause the producer to print `Ok!` after each line was typed (to notify user), but reading from files would only cause producer to print `Ok!` once at the end of the file. I think I have figured out a good middle ground to have both.

src/extension-consumer/src/consume/mod.rs
nicholastmosher: There is a specific reason I could not pass `fluvio::consumer::Record`, which is because the `fetch` interface from the consumer is still using the internal datatypes, not the `consumer::Record` data type. If we consolidated this, it would help to clean this up a little bit, but until then the common denominator is the bytes for key/value.

src/extension-consumer/src/produce/mod.rs
```diff
@@ -1,149 +1,269 @@
...
+                .await?;
+            return Ok(());
+        }
+
+        Err(ConsumerError::Other(
+            "Failed to send key-value record".to_string(),
+        ))
+    }
+
+    async fn produce_key_value_via_separator(
```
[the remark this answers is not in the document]
nicholastmosher: I have generalized all of the producer input reader logic that I can. This function in particular is the actual business logic to split keys and values of particular records, so I am going to leave it as-is 👍

src/extension-consumer/src/consume/mod.rs
nicholastmosher: This is directly pulled from old implementation https://github.com/infinyon/fluvio/pull/830/files#diff-fba8aea514e4ee4d3a739e2ccb439a3f43a63f35ab42df364a786a8253b11c95L103-L110

I can change this to eprintln and just print error code sentence 👍

src/extension-consumer/src/produce/mod.rs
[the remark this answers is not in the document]
nicholastmosher: The reason I put this in here was kind of for the same reason we opened this issue https://github.com/infinyon/fluvio/issues/700
Sometimes it is not obvious to the user that the program is waiting for input on stdin, so it is easy to think that something went wrong or it is frozen, when it is really just waiting on you. Plus, it is a nice confirmation for the user that the arguments they gave were the ones they actually intended to give.

We can change up what we print here, but I feel like we should have at least _something_ to visually indicate that the CLI is waiting for stdin.

src/extension-consumer/src/consume/mod.rs
sehz: reasonable

src/extension-consumer/src/produce/mod.rs
sehz: k

src/extension-consumer/src/produce/mod.rs
sehz: Need to re-think about support multiple files whether it's warranted or not.  Let's remove until we got come up with better design

sehz (changes requested):

src/extension-consumer/src/produce/mod.rs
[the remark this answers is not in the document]
sehz: Looks good

src/extension-consumer/src/produce/mod.rs
nicholastmosher: I actually just saw something that makes me unsure about doing this. I noticed that `BufReader::lines` actually will split on both `\n` and `\r\n`, meaning that there is not a single "split sequence", but two. I don't think there is an easy way to replicate that functionality via a "split" field in structopt.

I think this is something we may want to postpone until we have a better idea about what we want. I think, however, that we should leave this code as-is, because:

- If you look, this `--lines` field is marked as "hidden", meaning it does not show up in the help menu
- We do not document `--lines` so people will not be depending on it
- We can consider it an experiment as to whether it is useful:
  - If people ask for line-by-line functionality, we can tell them to try out the hidden `--lines` feature
  - If people ask for split-by functionality, we can ask them for their use-cases and come up with good requirements.

sehz: That's fair comment.  Here's I think we should do.   Let's make lines as default since we need standard way to separate one record from another.   This way, we don't need  `--lines` option. We can bring back separator in the future.

nicholastmosher: Right now, "line-by-line" is the default behavior for stdin, and "whole file is one record" is the default behavior for files. How does that sound?

sehz: Binary files is no different from string just with different separator.  I think if we need to deal with binary file, we can provide file type option similar to consumer

nicholastmosher:
@sehz I think all of the changes have been added. I'm ready to merge this if you are 👍

sehz (approved):
Great work!

PR #852, opened 2021-03-11
Re-enable VERSION in GH action

nicholastmosher:
This is required in order for other repos using our `action.yaml` to specify a pre-release version for testing against. E.g. fluvio-client-node.

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -13,10 +13,11 @@ inputs:
     description: "Add rust log options (e.g. debug, warn, info)"
     required: false
     default: 'info'
-  #version:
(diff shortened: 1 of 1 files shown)


simlay (approved):
This should be fine. I commented this out originally because the `VERSION` variable was getting squashed in the old `action-install-fluvio-cluster.sh` but I think it'll work now The [`VERSION` environment variable I think is still on here](https://github.com/infinyon/fluvio/blob/master/actions/action-install-fluvio-cluster.sh#L9).

nicholastmosher:
Yep, I just double-checked that the install.sh script respects the VERSION. Blank or undefined VERSION gives stable, `latest` gives prerelease, and anything else is treated like a literal version

PR #858, opened 2021-03-13
Made fluvio-cluster an optional dependency

simlay:
Not sure how I feel about this, was good practice to re-learn cargo features (I've been getting rusty on it).

Probably need to update release CI to include `fluvio-cluster` as a parameter.


nicholastmosher (approved):
Did we want to eventually separate out `fluvio-cluster` as a plugin? This seems like a good first step 👍

nicholastmosher:
> Probably need to update release CI to include fluvio-cluster as a parameter.

What are the cases in which we want to build with or without fluvio-cluster? It seems to me that until we make it a standalone plugin, the sane default would be to keep it built-in by default. So what is the opt-out case?

sehz:
is still needed?  https://github.com/infinyon/fluvio/pull/928

simlay:
> is still needed? #928

No. This isn't needed anymore.

PR #865, opened 2021-03-15
Make `fluvio update` update plugins as well as CLI

### LINKED ISSUE #672: CLI: `fluvio update` should also update all extensions
We should leverage our extension metadata mechanism to allow the Fluvio CLI to check all of the versions of the external extensions and check whether they need upgrading, then upgrade them.

nicholastmosher:
Closes #672 

This updates the `fluvio update` command to behave as follows:

- `fluvio update` with no parameters will update the CLI and any plugins that support being updated* to the latest stable version
- `fluvio update <pluginA> <pluginB>` will update any of the named plugins to the latest stable version
- `fluvio update` with a `--develop` flag with update the targets to the latest prerelease (rather than stable)

# *Plugin Update Mechanism

In order to be compatible with auto-updating, each plugin must now provide a new field of metadata called `package`, which identifies the plugin on packages.fluvio.io. For example, the `fluvio-cloud` plugin would provide a package field containing `fluvio/fluvio-cloud`. This is delivered via the already-existing metadata mechanism, where any plugin executed with a `metadata` argument is expected to return a `FluvioExtensionMetadata`, e.g.

```
$ fluvio-cloud metadata
{
  "title": "Fluvio Cloud CLI",
  "package": "fluvio/fluvio-cloud",
  "description": "A tool for interacting with Fluvio Cloud",
  "version": "0.1.4"
}
```

In the JSON above, the `package` field is new, and allows the plugin to tell the CLI where to search for its updates. Older plugins that do not provide the `package` field will not be auto-updated with `fluvio update` (with no arguments), but updates will still work fine if the plugin to update is named, e.g. `fluvio update fluvio-cloud`.

Example:

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1486,6 +1486,7 @@ version = "0.3.0"
 dependencies = [
  "async-trait",
  "fluvio",
+ "fluvio-package-index",
  "futures-lite",
  "prettytable-rs",
  "semver 0.11.0",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1593,7 +1593,7 @@ dependencies = [
 
 [[package]]
 name = "fluvio-package-index"
-version = "0.2.2"
+version = "0.3.0"
 dependencies = [
  "http-types",
  "lazy_static",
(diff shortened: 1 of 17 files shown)


simlay (approved):
I think I'm okay with this.

Issue #850, opened 2021-03-10
Release checklist


simlay:
Part of this task is to add a `RELEASES.md` someplace with a list of things that we need to check.

- [x] Review/run sample code on Home Page - Node @nicholastmosher 
- [x] Review/run sample code on Home Page - Rust @simlay
- [x]  ~Review/run sample code on Home Page - Python @simlay~ (not finished yet)
- [x] Review/run the Getting Started docs on fluvio.io/docs @nicholastmosher 
- [x] Rebase master into stable @simlay
- [x] Update `CHANGELOG.md` notes @simlay 
- [x] Publish rust crates
- [x] Run Release CI and mark as a stable release.
- [ ] Announce the release on Discord and twitter

sehz:
Add running examples codes

PR #867, opened 2021-03-15
Update changelog 0.7.1 and release document.

simlay:
Subtask of #850.

Does [`MINIMUM_PLATFORM_VERSION` in the rust client](https://github.com/infinyon/fluvio/blob/05eaf3aa1a6dd874792f492af8f99f73f703aaa0/src/client/src/lib.rs#L106) need updating to `0.7.1`? I think not because it should be semver compatible.


nicholastmosher:
Any client using `0.7.1` should not accept a connection to a cluster that is older than `0.7.1-alpha.0`. There have not been any additional breaking changes since `0.7.1-alpha.0`, so MINIMUM_CLIENT_VERSION should not need to be updated 👍

sehz (approved):

PR #871, opened 2021-03-16
Re-export fluvio config from cluster

nacardin:
Bump fluvio-cluster to update fluvio dep to 0.6.0


src/cli/src/install/update.rs
```diff
@@ -58,7 +58,7 @@ impl UpdateOpt {
             return Ok(());
         }
 
-        let s = (updates.len() != 1).then(|| "s").unwrap_or("");
+        let s = if updates.len() != 1 { "s" } else { "" };
```
nicholastmosher, line 61: I am curious as to how this ended up here? Was clippy upset at something or did you change this for another reason?

nicholastmosher (approved):
Looks good to me 👍

src/cli/src/install/update.rs
nacardin: I got 
```
error[E0658]: use of unstable library feature 'bool_to_option'
  --> src/cli/src/install/update.rs:61:38
   |
61 |         let s = (updates.len() != 1).then(|| "s").unwrap_or("");
   |                                      ^^^^
   |
   = note: see issue #64260 <https://github.com/rust-lang/rust/issues/64260> for more information
```

PR #851, opened 2021-03-10
Add doc test job to CI

tjtelan:
For #735.
This fails for me locally, but maybe OS specific error.

PR #875, opened 2021-03-17
Fix doc tests

### LINKED ISSUE #873: Update doc tests to pass CI
This is a prerequisite for #735. The tests should pass before they are included in CI on master.

New doc tests added in PR #851 are failing.

* Fail @ `fluvio-command` on the Ubuntu host
* Fail @ `fluvio-protocol-derive` on the MacOS host

It seems reasonable that the doc tests minimally pass on Linux, but I don't know what expectations we want to set for additionally passing the MacOS host.

nicholastmosher:
Closes #873 

- Fixes the `fluvio-protocol-derive` doc tests, including adding `fluvio-protocol` to dev-dependencies
- Fixed the `fluvio-command` doc tests. This one was tricky/interesting. I had written a doc test to demonstrate that the error returned by `Command::result` would carry the stderr output that the command returned. As a test case, I used `ls some-file-that-doesn't-exist`, but apparently there is a difference in the behavior of `ls` on different platforms. On Mac, it gives an exit code of 1, and on Linux it appeared not to. I wrote a new example of a failing command with stderr output.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1288,7 +1288,7 @@ dependencies = [
  "event-listener",
  "fluvio-dataplane-protocol",
  "fluvio-future 0.2.0",
- "fluvio-protocol",
+ "fluvio-protocol 0.4.0",
  "fluvio-sc-schema",
  "fluvio-socket",
  "fluvio-spu-schema",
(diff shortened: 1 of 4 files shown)


tjtelan (approved):
Passes in CI - https://github.com/infinyon/fluvio/runs/2141830304

nicholastmosher:
@tjtelan if I'm not mistaken, you've pulled these commits into your `run-doctests-ci` branch, is that right? If you've got what you need and you are planning on merging them with another PR, please feel free to close this one 👍

nicholastmosher:
Commits were pulled into https://github.com/infinyon/fluvio/pull/851

PR #885, opened 2021-03-18
Pass custom test vars from CLI into integration tests

### LINKED ISSUE #882: Test runner: Allow passing custom test-specific vars to tests
The tests all require a fluvio client handle, and a struct called `TestOption`, which is a collection of environment setup info, and a test case information used by the smoke test.

I still don't know how far we need to go down the key/value rabbit hole because we only have a few tests. So we're just going to stop with passing a `HashMap<String, String>` from the CLI to a test.

For now, the test itself will become responsible for validating all of the custom inputs, and converting from `String` to whatever it needs, and choosing default values.

---

There are a few things to be done.
* I am going to rename `TestOption` to something more specific: `TestCase`, and provide structure for environment setup while decoupling the struct from the smoke test.
* Modify the CLI to accept some basic key/value inputs that we can pass to test through this new `TestCase` struct.
* Update the smoke test in CI so it functions the same.
* Update test runner README to document the smoke test's extra vars, and provide guidance for how to make a new test.

tjtelan:
Resolves #882 

* Renames `TestOption` to `TestCase`.
* Separate environment information from test case into `EnvironmentSetup`
* Adding custom `key=val` support in CLI with `--var` or `-v` flag. Stringy-types must be parsed and validated by the test, at the moment.
* Decouple `TestCase` from the smoke test by removing smoke test specific fields and migrating the smoke test to use `HashMap<String, String>`
* Overcommunicate with smoke test. Custom variables aren't type checked. So print out the values of variables before starting test.
* Update CI + README
* Added CLI unit tests


tests/runner/src/main.rs
```diff
@@ -7,15 +11,39 @@ async fn main() {
...
+
+    // Cluster cleanup
+    cluster_cleanup(option).await;
+}
+
+async fn cluster_cleanup(option: fluvio_test_util::test_meta::TestCase) {
+    if option.skip_cluster_delete() {
+        println!("skipping cluster delete");
+    } else {
+        let mut setup = fluvio_test_util::setup::TestCluster::new(option.clone());
```
sehz, line 41: not fan of prefixing, can you just import at top?

tests/runner/src/main.rs
```diff
@@ -7,15 +11,39 @@ async fn main() {
...
+        setup.remove_cluster().await;
+    }
+}
+
+async fn cluster_setup(option: &fluvio_test_util::test_meta::TestCase) -> Arc<fluvio::Fluvio> {
```
sehz, line 46: ditto. remove prefix

tests/runner/src/main.rs
```diff
@@ -34,25 +62,132 @@ async fn main() {
...
+            "flv-test",
+            "smoke",
+            "--var",
+            "producer.iteration=9000",
+            "--var",
+            "producer.record_size=1000",
+        ]);
+        let test_case: TestCase = args.into();
+
+        let mut expected_hashmap: HashMap<String, String> = HashMap::new();
```
sehz, line 108: don't think you need to add type

tests/runner/src/tests/smoke.rs
```diff
@@ -3,11 +3,53 @@ use fluvio_command::CommandExt;
...
+// Disable producer
+// Producer count
+
 #[fluvio_test]
-pub async fn run(client: Arc<Fluvio>, option: TestOption) {
+pub async fn run(client: Arc<Fluvio>, option: TestCase) {
+    // Communicate what our expected smoke test values are.
+    // We will need a more robust pattern for all tests to validate inputs
+    // This is unwieldy duplication. But this is a small step forward
+    println!("\n###\nSmoke test vars: (Override with `-v key=val`)");
```
sehz, line 25: Can we just centralize logging for Option in once place instead of putting into test?

tests/runner/src/utils/smoke/consume.rs
```diff
@@ -24,19 +24,21 @@ fn consume_wait_timeout() -> u64 {
 }
 
 /// verify consumers
-pub async fn validate_consume_message(client: Arc<Fluvio>, option: &TestOption, offsets: Offsets) {
-    if option.use_cli() {
+pub async fn validate_consume_message(client: Arc<Fluvio>, option: &TestCase, offsets: Offsets) {
+    let use_cli = option.vars.get("use_cli").unwrap_or(&"false".to_string()) == "true";
```
sehz, line 28: since we are going extract arguments repeatedly, can we create type utility?  For example, here's what I would like to see:
`let use_cli = options.vars::<bool>("use_cli").expect("use_cli")`

tests/runner/src/utils/smoke/consume.rs
```diff
@@ -77,19 +79,27 @@ async fn get_consumer(client: &Fluvio, topic: &str) -> PartitionConsumer {
...
+    let replication = option.environment.replication;
+
+    let producer_iteration = usize::from_str_radix(
+        option
+            .vars
+            .get("producer.iteration")
+            .unwrap_or(&"1".to_string()),
+        10,
+    )
+    .expect("TODO: Fix this") as u16;
```
sehz, line 95: ditto: `let produce_iteration = options.vars::<u16>("produce_iteration").expect("iteration");`

tests/runner/src/utils/smoke/message.rs
```diff
@@ -11,14 +11,26 @@ fn generate_pre_fix(topic: &str, offset: i64) -> String {
 /// generate test data based on iteration and option
 ///
 #[allow(clippy::all)]
-pub fn generate_message(offset: i64, topic: &str, option: &TestOption) -> Vec<u8> {
-    let mut bytes = Vec::with_capacity(option.produce.record_size);
+pub fn generate_message(offset: i64, topic: &str, option: &TestCase) -> Vec<u8> {
+    //let mut bytes = Vec::with_capacity(option.produce.record_size);
+
+    //
+    let produce_record_size = usize::from_str_radix(
```
sehz, line 18: ditto. see previous comments about getting arg

sehz (changes requested):
LGTM with few minor nits:

tests/runner/src/tests/smoke.rs
tjtelan: I think so, yes. The purpose of this logging was more about over-communicating because I found the lack of input validation for custom vars difficult to use. I don't think we're done iterating with how it will work if we want to add more tests with their own variables, but I didn't have time to make it perfect without feedback from everyone else.

tests/runner/src/tests/smoke.rs
```diff
@@ -19,39 +19,15 @@ use std::sync::Arc;
 
 #[fluvio_test]
 pub async fn run(client: Arc<Fluvio>, option: TestCase) {
```
sehz, line 21: Idea will be 
```
async fn run(client: Arc<Fluvio>, option: TestCase<SmokeTestOption>) {
....
}
```
or alternatively:
```
async fn run(client: Arc<Fluvio>, option: TestCase) {
let smoke_option = option.into();
}
```



`SmokeTestOption` can be any struct that complies with `StructOpt`.   With that, can restore previous way to define SmokeTestOption:
```
pub struct SmokeTestOptions {
    /// number of produce iteration for a single producer
    #[structopt(short, long, default_value = "1")]
    pub produce_iteration: u16,

    // record size
    #[structopt(long, default_value = "100")]
    pub record_size: usize,

    // number of parallel producer
    #[structopt(long, default_value = "1")]
    pub producer_count: u16,
}
```

Can can set option similar like `cargo test`, where you can specify any extra arguments
```
$ flv-test smoke --local  --  --producer.iteration=10
```
and can even show help:
```
```
$ flv-test smoke --local  --help

```
```

tjtelan: I have the latter partially implemented in tests/runner/src/utils/smoke/mod.rs.

The only part that is missing is the `StructOpt` piece, which will be a little more complicated than we initially planned. I'll try to give it a go to get rid of the `HashMap<String, String>`, but I believe this is a good and usable improvement for custom vars.

sehz: Doesn't seems to be complicated:  https://docs.rs/structopt/0.3.21/structopt/trait.StructOpt.html#method.from_iter.

Nice thing about keeping `StructOpt` is that we can provide defaults, env var, etc.

tests/runner/src/utils/smoke/consume.rs
```diff
@@ -12,7 +12,7 @@ use fluvio_system_util::bin::get_fluvio;
 use fluvio::{Fluvio, Offset, PartitionConsumer};
```
sehz, line 12: All smoke test related files should be toghether since they will be maintained as unit.  Why moved to separate `util` directory?

sehz (changes requested):
Much better.  I just have a minor nit about file locations.  Once that has been addressed, we can merge this

Makefile
```diff
@@ -46,18 +46,18 @@ endif
 #
 
 smoke-test:	test-clean-up	build_test
-	$(TEST_BIN) smoke --spu ${DEFAULT_SPU} --produce-iteration ${DEFAULT_ITERATION} --local ${TEST_LOG} ${SKIP_CHECK} ${EXTRA_ARG}
+	$(TEST_BIN) --spu ${DEFAULT_SPU} --local ${TEST_LOG} ${SKIP_CHECK} ${EXTRA_ARG} -- smoke --producer-iteration=${DEFAULT_ITERATION}
```
sehz, line 49: Hmm.   This is little bit awkward.  I think we should still keep test name as argument before so: it should be like:
```
$(TEST_BIN) smoke  --spu ${DEFAULT_SPU} --local ${TEST_LOG} ${SKIP_CHECK} ${EXTRA_ARG} -- producer-iteration=${DEFAULT_ITERATION}
```

You can still inject test name by manipulating command

tests/runner/src/main.rs
```diff
@@ -1,21 +1,87 @@
...
+fn main() {
+    run_block_on(async {
+        let option = CliArgs::from_args();
+        //println!("{:?}", option);
+
+        let test_name;
+
+        // We want to get a TestOption compatible struct back
+        let test_opt: Box<dyn TestOption> = if let TestCli::CliCmd(cmd) = option.test_cmd {
+            test_name = cmd[0].clone();
```
sehz, line 24: Can still keep test name as structured argument.  Just add test name to `cmd`

sehz (changes requested):
See comments.  Just need to change argument position

sehz (approved):
Great.  Looks good now

sehz:
Yuck.  Mac VM is timing out

PR #886, opened 2021-03-19
Update CI to publish each commit to master

nicholastmosher:
This PR adds new `cargo make` tasks for publishing Fluvio to the `latest` release channel, as well as a brand new workflow file that will trigger these tasks, `publish_latest_fluvio`. Here's a breakdown of the new tasks:

- `cargo make publish-fluvio-latest` will publish the current release build of Fluvio to `latest`
  - FLUVIO_VESRION: `$(cat VERSION)+$(git rev-parse HEAD)`, e.g. `0.7.2-alpha.1+abcdef`
- `cargo make publish-fluvio` will publish the current release build of Fluvio to VERSION.
- `cargo make bump-fluvio-latest` will bump the `latest` version of Fluvio to the current VERSION+git hash
  - Only if Fluvio has been published on all targets for that VERSION, e.g. in prior steps of the CI job.
- `cargo make bump-fluvio` will use FLUVIO_VERSION to bump the appropriate tag
  - e.g. if FLUVIO_VERSION is `x.y.z` it will bump `stable`, but if it is `x.y.z-alpha.w+abcdef`, it will bump `latest`.
  - This will also always bump `latest`

I have done some limited testing locally, but in order to test this in earnest we will actually need to merge this PR, because that's the only way the new workflow will show up in Github.

### DIFF
--- a/.github/workflows/publish_latest_fluvio.yml
+++ b/.github/workflows/publish_latest_fluvio.yml
@@ -8,15 +8,10 @@ on:
       force:
         required: false
         description: 'Force push this release'
-
-env:
-  # Prod keys
-  # AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
-  # AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-
-  # Test keys
-  AWS_ACCESS_KEY_ID: ${{ secrets.PACKAGE_TEST_AWS_ACCESS_KEY_ID }}
-  AWS_SECRET_ACCESS_KEY: ${{ secrets.PACKAGE_TEST_AWS_SECRET_ACCESS_KEY }}
+      test:
+        required: false
+        description: 'Whether to run a test release'
+        default: ''
 
 jobs:
   publish_fluvio_cli:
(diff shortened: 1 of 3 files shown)


.github/workflows/publish_latest_fluvio.yml
```diff
@@ -0,0 +1,55 @@
...
+        required: false
+        description: 'Force push this release'
+
+env:
+  # Prod keys
+  # AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+  # AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+
+  # Test keys
+  AWS_ACCESS_KEY_ID: ${{ secrets.PACKAGE_TEST_AWS_ACCESS_KEY_ID }}
```
sehz, line 18: You can define 4 env: 2 for prod and 2 for test.  Then dynamically import env using `if`

.github/workflows/release.yml
```diff
@@ -14,9 +14,18 @@ on:
         description: 'Whether to update the install.sh script'
         required: false
 
+env:
```
sehz, line 17: ditto

sehz (changes requested):
LGTM with minot nits about not hard coding secrets

.github/workflows/publish_latest_fluvio.yml
sehz: Please see this: https://github.community/t/accessing-secrets-by-index-using-an-environment-variable/17966/4

sehz (approved):
LGTM

PR #888, opened 2021-03-20
SPU controller uses SVC ingress annotation

nacardin:


VERSION
```diff
@@ -1 +1 @@
-0.7.2-alpha.0
\ No newline at end of file
+0.7.2-alpha.1
```
sehz, line 1: don't need to bump up since we didn't publish `alpha.0`

sehz (approved):

VERSION
nacardin: fixed

sehz (approved):

PR #890, opened 2021-03-22
Move consumer code to CLI crate

nicholastmosher:
A little while ago we moved the code for the subcommands `consume`, `produce`, `topic`, and `partition` into a new crate called `fluvio-extension-consumer`. This doesn't really add any benefit for us, and it has had the effect of splitting up code that solves common problems. I think it will make things neater and easier to maintain to move those commands back into the CLI.


src/cli/src/consume/mod.rs
```diff
@@ -10,20 +10,19 @@ use std::convert::TryFrom;
...
+use fluvio_future::io::StreamExt;
 
 mod record_format;
 
 use fluvio::{Fluvio, PartitionConsumer, Offset, ConsumerConfig, FluvioError};
 use fluvio_sc_schema::ApiError;
 use fluvio::consumer::Record;
 
-use crate::ConsumerError;
+use crate::{Result, CliError};
```
sehz, line 21: I think it is better to keep Error specific to Features instead of using CLI Error

src/cli/Cargo.toml
```diff
@@ -39,20 +40,24 @@ prettytable-rs = "0.8.0"
 static_assertions = "1.1.0"
 tempdir = "0.3.7"
 sysinfo = "0.16.1"
+atty = "0.2.14"
```
sehz, line 43: I am ok to moving consumer extension back into CLI crate due to complexity of maintaining multiple crates.  However, we should isolate consumer related functionality under a feature flag.

src/cli/src/error.rs
```diff
@@ -73,19 +70,17 @@ impl CliError {
...
-                    ApiError::Code(ErrorCode::TopicAlreadyExists, _) => {
-                        println!("Topic already exists");
-                        Ok(())
-                    }
-                    ApiError::Code(ErrorCode::TopicNotFound, _) => {
-                        println!("Topic not found");
-                        Ok(())
-                    }
-                    _ => Err(self),
+            Self::ClientError(FluvioError::ApiError(api)) => match api {
```
sehz, line 73: Ditto.  Let's keep ConsumerError

sehz (changes requested):

src/cli/src/error.rs
nicholastmosher: I wonder if we can come up with a better name for it. Before we split the crates apart, there was no special type for ConsumerError, all of those variants simply lived in CliError, so this was returning to that previous pattern. If we're going to keep a new error type, it might make sense to come up with something that fits better.

Honestly, one thing I would really like to do is to only use `eyre::Result` in the entire CLI crate. That would make it so we would not need to keep track of the specific error types regardless of the libraries we use. This would essentially be equivalent to our current error-handling setup but without the hassle of maintaining and updating the CliError type. Let me know what you think.

src/cli/Cargo.toml
nicholastmosher: Sure. I'm trying to think of a better name than "consumer" to describe this handful of features. These include consume/produce/topic/partition, so "consumer" seems too specific. Something like "cluster client"? That might be to vague, and clash with the idea of fluvio-cluster/fluvio(client). Let me know if you have any naming ideas or whether we should just stick with "consumer"

src/cli/src/error.rs
sehz: Is there reason to depend on `eye:Result`?

src/cli/Cargo.toml
nicholastmosher: Yeah... it seems almost like we have a "data" or "record" client (what we call "consumer" today), and then the "admin" client.

src/cli/src/error.rs
```diff
@@ -40,6 +39,28 @@ pub enum CliError {
     Other(String),
 }
 
+#[derive(thiserror::Error, Debug)]
```
sehz, line 42: Let's move to consumer module to make it isolated

src/cli/src/error.rs
```diff
@@ -40,6 +39,28 @@ pub enum CliError {
...
+    ClientError(#[from] FluvioError),
+    #[error("Invalid argument: {0}")]
+    InvalidArg(String),
+    #[error("Error finding executable")]
+    WhichError(#[from] which::Error),
+    #[error("Unknown error: {0}")]
+    Other(String),
+}
+
+impl ConsumerError {
```
sehz, line 58: Use From Trait

src/cli/src/produce/mod.rs
```diff
@@ -7,7 +7,7 @@ use tracing::debug;
 use fluvio::{Fluvio, TopicProducer};
 use fluvio_types::print_cli_ok;
 use crate::common::FluvioExtensionMetadata;
-use crate::{Result, ConsumerError};
+use crate::error::ConsumerError;
```
sehz, line 10: let's move consumer related under `consumer` module to keep isolated and make it easier to use feature flag

sehz (changes requested):

src/cli/src/error.rs
nicholastmosher: From what? CliError already implements From<ConsumerError>

sehz:
```
pub fn invalid_arg<M: Into<String>>(reason: M) -> Self {
         Self::InvalidArg(reason.into())
     }
```

nicholastmosher: Do you mean to implement `From<String> for ConsumerError` instead? I try to avoid implementing `From<String>` for errors because it can be easy to accidentally throw a `?` after a String value and turn an actual value into an error by mistake. I don't really see a big problem with having this function here, plus that's already how it was written, I didn't add this, I just moved it.

sehz: fair enough

nicholastmosher:
Is there anything else to do for this?

sehz:
feature flag?

src/cli/Cargo.toml
```diff
@@ -39,20 +44,24 @@ prettytable-rs = "0.8.0"
...
+ctrlc = "3.1.3"
```
sehz, line 48: these dependencies should be under feature flag

src/cli/Cargo.toml
```diff
@@ -39,20 +44,24 @@ prettytable-rs = "0.8.0"
...
+hostname-validator = "1.0.0"
+content_inspector = "0.2.4"
 
 # Fluvio dependencies
 
 k8-config = { version = "1.3.0" }
 k8-client = { version = "5.0.0" }
-fluvio-future = { version = "0.2.0", features = ["fs", "io", "subscriber"] }
 fluvio = { version = "0.6.0", path = "../client", default-features = false }
+fluvio-types = { version = "0.2.0" , path = "../types" }
```
sehz, line 57: ditto

src/cli/Cargo.toml
```diff
@@ -39,20 +44,24 @@ prettytable-rs = "0.8.0"
...
+fluvio-future = { version = "0.2.0", features = ["fs", "io", "subscriber"] }
 fluvio-cluster = { version = "0.7.0", path = "../cluster", default-features = false, features = ["cli"] }
 fluvio-command = { version = "0.2.0" }
 fluvio-package-index = { version = "0.3.0", path = "../package-index" }
 fluvio-extension-common = { version = "0.3.0", path = "../extension-common", features = ["target"]}
-fluvio-extension-consumer = { version = "0.3.0", path = "../extension-consumer" }
 fluvio-controlplane-metadata = { version = "0.7.0", path = "../controlplane-metadata", features = ["use_serde", "k8"] }
-fluvio-sc-schema = { version = "0.7.0", path = "../sc-schema" }
+fluvio-sc-schema = { version = "0.7.0", path = "../sc-schema", features = ["use_serde"] }
```
sehz, line 64: ditto

src/cli/src/error.rs
```diff
@@ -1,12 +1,13 @@
...
-use fluvio_extension_common::output::OutputError;
-use fluvio_extension_consumer::ConsumerError;
 use fluvio_cluster::cli::ClusterCliError;
-use crate::common::target::TargetError;
 use fluvio_sc_schema::ApiError;
 use fluvio_sc_schema::errors::ErrorCode;
+use fluvio_extension_common::output::OutputError;
+
+use crate::common::target::TargetError;
+use crate::consumer::error::ConsumerError;
```
sehz, line 10: under features flag

src/cli/src/lib.rs
```diff
@@ -15,6 +15,16 @@ mod profile;
 mod version;
 mod metadata;
 
+#[cfg(feature = "consumer")]
```
sehz, line 18: would prefer to use cfg_if to make it tidy

src/cli/src/lib.rs
```diff
@@ -173,14 +180,15 @@ impl RootCmd {
 // Please don't change it unless you want to update the top-level help menu "about".
 /// Fluvio command-line interface
 #[derive(StructOpt, Debug)]
```
sehz, line 175: move to consumer module

sehz (changes requested):

src/cli/src/lib.rs
nicholastmosher: Moving this stuff into a module cleaned it up

src/cli/src/error.rs
nicholastmosher: Hmmmmm..... you know I think it is a really very bad idea to `cfg` variants in a public enum. It causes this weird schism where the enum would be considered two different types under the two different feature flags... I think we should avoid that at all costs.

sehz: Is there reason why this need explicit dependency to ConsumerConfig?

sehz (approved):
LGTM.  ship it

PR #892, opened 2021-03-23
Create publish_latest_fluvio workflow for publishing `latest` on every master commit.

### LINKED ISSUE #808: Generate latest and nightly release

nicholastmosher:
Closes #808 

I have finally gotten this workflow working. It has three jobs:

- Build fluvio in release mode for x86_64-apple-darwin
- Build fluvio in release mode for x86_64-unknown-linux-musl
- Bump up the `latest` tag on fluvio-packages with the version of the build, including git commit


makefiles/Base.toml
```diff
@@ -7,26 +7,61 @@ CARGO_PROFILE = "release"
...
     "build",
     "--bin", "fluvio",
-    "--release",
-    "--target=x86_64-apple-darwin",
-    "@@split(CARGO_MAKE_CARGO_BUILD_TEST_FLAGS, )"
+    "--target=x86_64-unknown-linux-musl",
 ]
 
-[tasks.build-release.linux]
+[tasks.build-prod-musl.linux]
```
sehz, line 57: is there reason to renaming and changing this task? Since this is used by other action, they need to be tested as well

makefiles/Base.toml
```diff
@@ -35,12 +70,8 @@ which musl-gcc 2> /dev/null > /dev/null || {
     echo "Install musl-tools on your system!";
     exit 1
 }
-rustup target add "$CARGO_MAKE_RELEASE_FLOW_TARGET"
```
sehz, line 38: ditto

makefiles/S3.toml
```diff
@@ -33,15 +33,22 @@ args = [
 
 [tasks.publish-fluvio]
 dependencies = [
-    "build-release",
+    "build-prod",
     "install-fluvio-package",
     "publish-fluvio-env",
 ]
-script = "${HOME}/.fluvio/bin/fluvio package --test publish --version=${FLUVIO_VERSION} ${FORCE_RELEASE} ./target/${TARGET}/release/fluvio"
+env = { FLUVIO_VERSION = { script = ["cat VERSION"] } }
```
sehz, line 40: Why changing this?
Injecting env make this less reusable

makefiles/S3.toml
```diff
@@ -58,8 +65,20 @@ script = "${HOME}/.fluvio/bin/fluvio install --develop fluvio/fluvio-package"
 
 [tasks.bump-fluvio]
 dependencies = ["install-fluvio-package", "bump-fluvio-latest"]
-script = """${HOME}/.fluvio/bin/fluvio package --test bump dynamic ${FLUVIO_VERSION} """
+script = """
+${HOME}/.fluvio/extensions/fluvio-package \
+    ${FLUVIO_PUBLISH_TEST} \
```
sehz, line 70: why it is using `FLUVIO_PUBLISH_TEST` ?

sehz (changes requested):
need to be careful about changing build script since this will impact existing release flow.
I will let @simlay approve for script in general.

This kind of convince me that `cargo make` is not suitable for build scripts and we should move back to Makefile

makefiles/Base.toml
```diff
@@ -7,26 +7,61 @@ CARGO_PROFILE = "release"
...
+#
+
+[tasks.build-dev-mac.mac]
+command = "cargo"
+args = ["build", "--bin", "fluvio", "--target", "x86_64-apple-darwin"]
+
+[tasks.build-prod-mac.mac]
+command = "cargo"
+args = ["build", "--bin", "fluvio", "--release", "--target=x86_64-apple-darwin"]
+
```
simlay, line 36: Split up the `--target` and `x86_64-apple-darwin` to be the same to the `build-dev-mac`

makefiles/Base.toml
nicholastmosher: Yes, there is a specific reason, which is that `build-release` is one of the magic "default tasks" of `cargo make`

- https://github.com/sagiegurari/cargo-make/blob/master/src/lib/descriptor/makefiles/rust.toml#L380

So when we add definitions to a task by the same name, we are not creating an original task, but rather we are editing the existing task, which is error prone and causes weird implicit effects in the task. I have found it much more effective to always create unique task names, hence `build-prod-musl`

makefiles/Base.toml
nicholastmosher: Here, we were the only ones assigning CARGO_MAKE_RELEASE_FLOW_TARGET, and it was with a constant value, so I simply substituted the constant into the  ENV variable.

makefiles/S3.toml
nicholastmosher: Publishing Fluvio always requires knowing the contents of the `VERSION` file and this is an elegant way to do it. This is the "default" `publish-fluvio` task, which uses the VERSION directly, there is also another task, `publish-fluvio-latest` which simply overrides FLUVIO_VERSION to include both the VERSION + the git commit.

See https://github.com/infinyon/fluvio/pull/892/files/038e3333c72f9fee0c46ebbd9cf02e91731ea6c0#diff-2dcdc917058b2578ec902f5e1691c7c4c6ce3da1e59036f159ee6230548c4adfR57

makefiles/S3.toml
nicholastmosher: `FLUVIO_PUBLISH_TEST` has two possible values: empty string `""` or `--test`. When it it `--test`, this will attempt to run the  bump command against `/test/` in the package registry instead of `/v1/`, so we can verify the effect of the command.

makefiles/Base.toml
```diff
@@ -7,26 +7,61 @@ CARGO_PROFILE = "release"
...
+extend = "build-prod-mac"
+[tasks.build-prod.linux]
+extend = "build-prod-musl"
+
+
+################################################################################
+# Building for x86_64-apple-darwin can only be done on Mac
+#
+
+[tasks.build-dev-mac.mac]
```
simlay, line 29: I'm a bit confused by the task named `build-dev-mac` and `build-prod-mac`, what's wrong with just `build-dev.mac` and `build-prod.mac` as well as their linux counter parts? After fighting with the cross compilation tooling, I think it's easier and more maintainable to just compile it for the host itself.

simlay (changes requested):

makefiles/S3.toml
simlay: Does this work? I tried to do this sort of thing with the github-release command. I think the "" case might actually pass in the commandline parameter of "" which wasn't acceptable for `github-release`. It might be okay with fluvio-publish.

makefiles/Base.toml
nicholastmosher: Sure. I think of this as two "levels" of tasks. At the lower level, you have tasks that always produce the same thing, and are not influenced by the environment they are run in. These would be things like

```
build-dev-mac
build-prod-mac
build-dev-musl
build-prod-musl
```

By looking at these, the goal is to be able to know _exactly_ what the output artifact is. For example, if I want to build for `x86_64-apple-darwin` in release mode, I'd use `build-prod-mac`, no matter what host I am using.

Then at a higher level, there are tasks that _do_ adapt to the environment in which they are run. The examples of these are

```
build-dev
build-prod
```

The implicit information here is that the build artifact coming out will have a target that matches the host that you are building on. So rather than saying "build a debug image for target X", you are saying "build a debug image for my host". This is essentially the same pattern that cargo uses, if you give it no `--target`, it will build for your host. If you give it a `--target`, it will build exactly that target, regardless of the host.

---

Previously, we _only_ had "high-level" tasks like `build-dev` and `build-prod`. I found this to be a problem in many cases where I wanted to quickly make a build for `musl` from my mac machine, and there was no way to express that with the tasks we had.

makefiles/Base.toml
sehz: which weird effect?

makefiles/S3.toml
sehz: fair enough.

makefiles/S3.toml
nicholastmosher: Yes, this works if you put it in a `script` block. If you wrote it as a task with `command` and `args` it would not work, you're right.

makefiles/Base.toml
sehz: I think we should not put configuration specific tasks.   This leads to combinational explosion.  For example, if we add another target: `raspberry`.   We should only have 1 single build task which can take parameters as needed

nicholastmosher: I guess the moral of the story is that the low-level tasks are very useful for local development, and the high-level tasks make it easy to make CI jobs that do the right thing per platform.

makefiles/Base.toml
nicholastmosher: Just the fact that you would _think_ you are looking at the full task definition when in reality there are other things that get implicitly mixed in.

makefiles/S3.toml
```diff
@@ -33,15 +33,22 @@ args = [
...
 ]
-script = "${HOME}/.fluvio/bin/fluvio package --test publish --version=${FLUVIO_VERSION} ${FORCE_RELEASE} ./target/${TARGET}/release/fluvio"
+env = { FLUVIO_VERSION = { script = ["cat VERSION"] } }
+script = """
+${HOME}/.fluvio/extensions/fluvio-package \
+    ${FLUVIO_PUBLISH_TEST} \
+    publish \
+    --version=${FLUVIO_VERSION} \
+    ${FORCE_RELEASE} \
+    ./target/${TARGET}/release/fluvio
```
sehz, line 47: Why not `cargo run` instead of referring target directory?

nicholastmosher: What do you mean? This command is naming the executable file that should be published, it is not invoking it.

sehz: got it

makefiles/S3.toml
```diff
@@ -58,8 +65,20 @@ script = "${HOME}/.fluvio/bin/fluvio install --develop fluvio/fluvio-package"
...
-script = """${HOME}/.fluvio/bin/fluvio package --test bump dynamic ${FLUVIO_VERSION} """
+script = """
+${HOME}/.fluvio/extensions/fluvio-package \
+    ${FLUVIO_PUBLISH_TEST} \
+    bump \
+    dynamic \
+    ${FLUVIO_VERSION}
+"""
 
 [tasks.bump-fluvio-latest]
```
sehz, line 76: Can we consolidate into a single `bump` task  instead of duplicating scripts?

nicholastmosher: If you look closely, these are actually two separate commands. One is running

- `fluvio package bump latest VERSION+git`

and the other is running

- `fluvio package bump dynamic VERSION`

It would take a whole lot more conditionals and scripting in order to consolidate them into one task, this is much simpler.

sehz: fair enough

nicholastmosher:
A few notes of changes I made while rebasing (to @simlay specifically):

- I noticed there was a new task for bumping just the stable Fluvio version, and that it gets run when a regex in the CI script matches (or rather, does not match)
  - I removed that task because the `bump-fluvio` task (which runs `fluvio package bump dynamic`) will already detect the type of release suitable for the given VERSION. i.e. it will automatically bump stable if the given version looks like `x.y.z`, or prerelease if it is `x.y.z-alpha.w`. So I got rid of the regex in CI and the `bump-fluvio-stable` task in the makefiles.
  - So now, that step just uses `cargo make bump-fluvio`
- I kept the regex in the github release step, that one looks good 👍

nicholastmosher:
Here is a passing workflow for these changes: https://github.com/infinyon/fluvio/actions/runs/683220750

sehz (approved):
We should revisit build script arch.  But for now, good enough

VERSION
```diff
@@ -1 +1 @@
-0.7.2
\ No newline at end of file
+0.7.3-alpha.1
```
simlay, line 1: Can't have a newline at the end of the file I think.

sehz (approved):

PR #893, opened 2021-03-23
Update SerialFrame to take &self only

nicholastmosher:
This is a little bit of upkeep in the client code. Prior to this, `SerialFrame`'s `send_receive` method took `&mut self`, but it was only necessary in one particular implementation of `SerialFrame`: VersionedSocket. In This PR, I changed `SerialFrame::send_receive` to take `&self`, and wrapped the `VersionedSocket` in a `Mutex` where the interior mutability was required.

This will make things nicer in some upcoming changes where we want to be able to do things like cache the socket connections for a Producer, which would be a headache and a half to do with the current API.

### DIFF
--- a/src/client/src/client/client.rs
+++ b/src/client/src/client/client.rs
@@ -13,6 +13,7 @@ use fluvio_socket::{AllFlvSocket, SharedAllMultiplexerSocket};
 use fluvio_future::native_tls::AllDomainConnector;
 
 use crate::FluvioError;
+use async_mutex::Mutex;
 
 /// Frame with request and response
 #[async_trait]
@@ -37,7 +38,7 @@ pub(crate) trait SerialFrame: Sync + Send + Display {
     }
 
     /// send and receive
-    async fn send_receive<R>(&mut self, request: R) -> Result<R::Response, FlvSocketError>
+    async fn send_receive<R>(&self, request: R) -> Result<R::Response, FlvSocketError>
     where
         R: Request + Send + Sync;
 }
@@ -45,7 +46,7 @@ pub(crate) trait SerialFrame: Sync + Send + Display {
 /// This sockets knows about support versions
 /// Version information are automatically  insert into request
 pub struct VersionedSocket {
-    socket: AllFlvSocket,
+    socket: Mutex<AllFlvSocket>,
     config: ClientConfig,
     versions: Versions,
 }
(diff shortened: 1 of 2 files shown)


nicholastmosher:
@sehz any feedback on this PR or does it look good?

src/client/src/client/client.rs
```diff
@@ -37,15 +38,15 @@ pub(crate) trait SerialFrame: Sync + Send + Display {
     }
 
     /// send and receive
-    async fn send_receive<R>(&mut self, request: R) -> Result<R::Response, FlvSocketError>
+    async fn send_receive<R>(&self, request: R) -> Result<R::Response, FlvSocketError>
```
sehz, line 41: What's use case for this?

Also, only send (Sink) need to be mutable.   There is already `InnerExclusiveFlvSink` for that.

src/client/src/client/client.rs
```diff
@@ -37,15 +38,15 @@ pub(crate) trait SerialFrame: Sync + Send + Display {
...
     where
         R: Request + Send + Sync;
 }
 
 /// This sockets knows about support versions
 /// Version information are automatically  insert into request
 pub struct VersionedSocket {
-    socket: AllFlvSocket,
+    socket: Mutex<AllFlvSocket>,
```
sehz, line 49: I have prototype branch: https://github.com/sehz/fluvio/tree/duplicate_socket which try to use thread safe TcpStream.  In that case, you can duplicate sockets without problem.

sehz (changes requested):

src/client/src/client/client.rs
nicholastmosher: The original thing I was thinking about wanting to do was cache different sockets to different replica leaders. For example, in the producer, we use `spu_pool.create_serial_socket` on every record send. I was thinking it would be good to have like a `HashMap<ReplicaKey, VersionedSerialSocket>` to cache the sockets so we wouldn't need to open them over and over. But doing that is complicated when the sockets require `&mut` access to do anything.

To be clear, I'm not convinced that caching these sockets will _actually_ get us any benefit, but I noticed that changing this `&mut` to a `&` made the API cleaner all around. Currently the only way to achieve this is to wrap the inner socket in a Mutex because the sockets we are using are not thread-safe. If we were to switch to thread-safe sockets in the future, this API change would still benefit from it.

sehz: SerialFrame is designed for used in a single task  not multi-threaded.  Whole purpose of `MultiplexorSocket` to allow multiple concurrent operation to a single socket.  It doesn't make sense to share a same serial SerialFrame in different thread.  So we should defer until if we actually hit the use case

nicholastmosher:
I'm going to go ahead and close this, we can look back on it in the future if we end up needing something like this

PR #894, opened 2021-03-23
0.7.2 release

### LINKED ISSUE #568: Parse VERSION to determine if a github release is a pre-release

simlay:
Releases 0.7.2 and closes #568.


VERSION
```diff
@@ -1 +1 @@
-0.7.2-alpha.0
\ No newline at end of file
+0.7.2-alpha.1
```
sehz, line 1: I assume this is for testing release workflow?

sehz (changes requested):

VERSION
simlay: Yup. I'm doing a pre-release test now.

sehz (approved):

makefiles/S3.toml
```diff
@@ -63,3 +63,7 @@ script = """${HOME}/.fluvio/bin/fluvio package --test bump dynamic ${FLUVIO_VERS
 [tasks.bump-fluvio-latest]
 dependencies = ["install-fluvio-package"]
 script = """${HOME}/.fluvio/bin/fluvio package --test bump latest "$(cat VERSION)+$(git rev-parse HEAD)" """
+
+[tasks.bump-fluvio-stable]
+dependencies = ["install-fluvio-package"]
+script = """${HOME}/.fluvio/bin/fluvio package --bump stable $(cat VERSION)"""
```
nicholastmosher, line 69: This should be `fluvio package bump stable $(cat VERSION)`, not `--bump`

simlay: Ah yes, good catch.

.github/workflows/release.yml
```diff
@@ -49,13 +49,13 @@ jobs:
...
         run: |
-          echo "::set-env name=AWS_ACCESS_KEY_ID::${{ secrets.PACKAGE_TEST_AWS_ACCESS_KEY_ID }}"
-          echo "::set-env name=AWS_SECRET_ACCESS_KEY::${{ secrets.PACKAGE_TEST_AWS_SECRET_ACCESS_KEY }}"
+          echo "AWS_ACCESS_KEY_ID=${{ secrets.PACKAGE_TEST_AWS_ACCESS_KEY_ID }}" >> $GITHUB_ENV
+          echo "AWS_SECRET_ACCESS_KEY=${{ secrets.PACKAGE_TEST_AWS_SECRET_ACCESS_KEY }}" >> $GITHUB_ENV
       - name: Set prod env
         if: ${{ github.event.inputs.test == '' }}
         run: |
-          echo "::set-env name=AWS_ACCESS_KEY_ID::${{ secrets.AWS_ACCESS_KEY_ID }}"
-          echo "::set-env name=AWS_SECRET_ACCESS_KEY::${{ secrets.AWS_SECRET_ACCESS_KEY }}"
```
nicholastmosher, line 58: Dang, sorry that was my bad. I thought the ::set-env stuff was something I'd only experimented with in my test-release branch, I didn't know that had found its way into master :(

simlay: No worries.

PR #896, opened 2021-03-25
Simple batch produce

nicholastmosher:
Here is the first shot at implementing the batch producer. I have included an example and everything is working. The only thing that is still up in the air is trying to see whether we can create a more flexible API.

Currently, `send_all` can take any argument that can turn into an iterator over `(Option<Key>, Value)` pairs. This is one example of how calling it looks:

```rust
let batch: Vec<_> = (0..10)
    .map(|i| (Some(i.to_string()), format!("This is record {}", i)))
    .collect();
producer.send_all(batch).await?;
```

I am currently experimenting with other API designs. Here are other caller patterns that we would like to support if at all possible:

Iterator over `(Key, Value)` (without Option)

```rust
let batch: Vec<_> = (0..10)
    .map(|i| (i.to_string(), format!("This is record {}", i))) // Note, no 'Some' here
    .collect();
producer.send_all(batch).await?;
```

Iterator over `Value` (without key)

```rust
let batch: Vec<_> = (0..10)
    .map(|i| format!("This is record {}", i)) // These records would have no key
    .collect();
producer.send_all(batch).await?;
```

---

My current attempt at implementing this is having trouble. What I'm attempting to do is create a trait called `IntoBatch` and write various blanket-implementations for different kinds of iterators. For example:

```rust
trait IntoBatch {
    type Key;
    type Value;
    fn into_batch(self) -> Vec<(Option<Self::Key>, Self::Value)>;
}

impl<I, K, V> IntoBatch for I
where
    I: IntoIterator<Item = (K, V)>,
{
    type Key = K;
    type Value = V;
    fn into_batch(self) -> Vec<(Option<Self::Key>, Self::Value)> {
        self.into_iter()
            .map(|(key, value)| (Some(key), value))
            .collect()
    }
}

impl<I, V> IntoBatch for I
where
    I: IntoIterator<Item = V>,
{
    type Key = ();
    type Value = V;
    fn into_batch(self) -> Vec<(Option<Self::Key>, Self::Value)> {
        self.into_iter().map(|(_k, value)| (None, value)).collect()
    }
}
```

However, this is causing overlapping implementations:

```
error[E0119]: conflicting implementations of trait `producer::IntoBatch`:
   --> src/client/src/producer.rs:222:1
    |
209 | / impl<I, K, V> IntoBatch for I
210 | | where
211 | |     I: IntoIterator<Item = (K, V)>,
212 | | {
...   |
219 | |     }
220 | | }
    | |_- first implementation here
221 |
222 | / impl<I, V> IntoBatch for I
223 | | where
224 | |     I: IntoIterator<Item = V>,
225 | | {
...   |
230 | |     }
231 | | }
    | |_^ conflicting implementation
```

If this is going to work, I need a better way to constrain the generics on these implementations so that they do not overlap. However, I'm not actually convinced that this can be done for the cases we have laid out here. I will continue brainstorming, but we may need to be willing to accept the less flexible API for now.


nicholastmosher:
After a good bit of experimentation, I have not found a way to support all three calling patterns. The one that I have included in this PR is the most general one, and it is still pretty easy to use. @sehz I would like to move forward with merging this as-is, and if we want to prioritize new calling patterns in the future we can make a new issue for it.

src/client/src/producer.rs
```diff
@@ -61,7 +61,36 @@ impl TopicProducer {
...
+        let records: Vec<_> = records.into_iter().collect();
+        let records: Vec<_> = records
+            .iter()
+            .map(|(key, value): &(Option<K>, V)| {
+                (key.as_ref().map(|it| it.as_ref()), value.as_ref())
+            })
+            .collect();
+        send_records_raw(spu_client, &replica, &records).await?;
+        Ok(())
     }
```
sehz, line 95: `send` should be re-written using `send_all`

src/client/src/producer.rs
```diff
@@ -61,7 +61,36 @@ impl TopicProducer {
...
+        let records = &[(Some(key), value)];
+        send_records_raw(spu_client, &replica, records).await?;
+        Ok(())
+    }
+
+    #[instrument(
+        skip(self, records),
+        fields(topic = %self.topic),
+    )]
+    pub async fn send_all<K, V, I>(&self, records: I) -> Result<(), FluvioError>
```
sehz, line 73: since we are converting to `Vector` anyway, we should take `Vector` rather than `Iterator`.  This may solve API issue as well.

sehz (changes requested):

src/client/src/producer.rs
nicholastmosher: I think that using `IntoIterator` should be just fine... Using `Vec` would not solve any of the open API rough-edges, it would actually make them worse. By using `IntoIterator` we can change the internal representation without breaking the API, but if we bake Vec into the type signature, any change would become a breaking change.

nicholastmosher: Since we've recently decided to use `Bytes` as the internal representation of records, I think that using `Into<Vec<u8>>` as our parameters should actually be a good API here 👍

src/dataplane-protocol/src/record.rs
```diff
@@ -155,26 +155,24 @@ impl Encoder for DefaultAsyncBuffer {
...
+        len.decode_varint(src)?;
+        let len = len as usize;
+
+        // Take `len` bytes from `src` and put them into a new BytesMut buffer
+        let slice = src.take(len);
+        let mut bytes_mut = bytes::BytesMut::with_capacity(len);
+        bytes_mut.put(slice);
+
+        // Replace the inner Bytes buffer of this DefaultAsyncBuffer
+        self.0 = bytes_mut.freeze();
```
nicholastmosher, line 174: @sehz this is a section where I would like some careful review when you get a chance

src/client/src/producer.rs
```diff
@@ -93,16 +110,18 @@ impl TopicProducer {
...
 }
 
 /// Sends record to a target server (Kf, SPU, or SC)
-async fn send_record_raw<F: SerialFrame>(
+async fn send_records_raw<F: SerialFrame>(
     mut leader: F,
     replica: &ReplicaKey,
-    key: Option<&[u8]>,
-    value: &[u8],
+    records: &[(Option<Bytes>, Bytes)],
```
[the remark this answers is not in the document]
sehz: This could be Vec<DefaultRecord>, that way, can eliminate unnecessary conversion

sehz (changes requested):

sehz:
rebase as well

src/dataplane-protocol/src/record.rs
```diff
@@ -155,26 +155,24 @@ impl Encoder for DefaultAsyncBuffer {
...
-                ErrorKind::Other,
-                "Can't decode buffer while cloning".to_string(),
-            ))
-        }
+
+        let mut len: i64 = 0;
+        len.decode_varint(src)?;
+        let len = len as usize;
+
+        // Take `len` bytes from `src` and put them into a new BytesMut buffer
```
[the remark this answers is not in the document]
nicholastmosher: What specifically do you mean? I am already using `take` here...

src/client/src/producer.rs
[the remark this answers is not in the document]
nicholastmosher: Which unnecessary conversion are you referring to? The reason to use a slice rather than an array is that you can pass a slice of any size, whereas the array must be a fixed size. I also don't think a Vec is necessary here, unless you are talking about using Vec instead of Bytes? But if that's what you're asking I'm not sure what the advantage to that is either.

src/dataplane-protocol/src/record.rs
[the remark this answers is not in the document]
sehz: Yes, you are right.  I thought this can be directly convert into `Bytes`.

src/client/src/producer.rs
```diff
@@ -116,22 +135,26 @@ async fn send_record_raw<F: SerialFrame>(
...
-        ..Default::default()
-    };
-    if let Some(key) = key {
-        record_msg.key = Some(key.into());
-    }
+    debug!("Putting together batch with {} records", records.len());
+
     let mut batch = DefaultBatch::default();
-    batch.add_record(record_msg);
+    for (key, value) in records {
```
sehz, line 141: can you `map` iterator rather than `for..in`

src/client/src/producer.rs
```diff
@@ -93,16 +110,18 @@ impl TopicProducer {
 
         debug!("connect to replica leader at: {}", spu_client);
 
-        send_record_raw(spu_client, &replica, None, record).await
+        let record = Bytes::from(Vec::from(buffer.as_ref()));
+        let records = &[(None, record)];
```
sehz, line 114: can combine into a single line

src/client/src/producer.rs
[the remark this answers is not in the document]
sehz: This is only used internally.  `send_all` uses Vec.  So you are reconstructing Vec twice rather than just constructing them once.

sehz (changes requested):

src/client/src/producer.rs
nicholastmosher: I just took a look and tried to do this, but it turned out to be trickier than it looks. Usually we can just use iterators to map and collect things, but we cannot collect directly into a DefaultBatch and we also cannot directly assign the internal records Vec, we need to use `add_record` for each individual record. So right now the for loop seems like the best option. Perhaps in the future we could add something to the API of DefaultBatch to make this nicer.

src/client/src/producer.rs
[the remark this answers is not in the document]
nicholastmosher: If you look at how the `records: &[(Option<Bytes>, Bytes)]` slice gets used inside of `send_records_raw`, it is simply being iterated over and the inner `Bytes` objects are being loaded into the request. I don't think there would be any advantage to passing the entire owned Vec through this call, there are no extra allocations being made.

In fact, if you look at the other usage of `send_records_raw` in the `send_record` method, it is actually directly constructing a slice on the stack, and does not require the use of Vec at all. If we were to change `send_records_raw` to require a Vec, then `send_record` would gain an extra unnecessary allocation per record sent.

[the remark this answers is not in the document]
nicholastmosher: Updated based on our conversation earlier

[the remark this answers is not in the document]
sehz: Looks good

src/dataplane-protocol/src/batch.rs
```diff
@@ -111,7 +111,25 @@ where
     }
 }
 
-impl Batch<DefaultBatchRecords> {
+impl DefaultBatch {
+    // TODO write test
```
sehz, line 115: Can close PR once this test is completed

nicholastmosher: Added test showing batch behavior is the same as individual-adding behavior

PR #897, opened 2021-03-25
Fix clippy lints

nicholastmosher:
Rust 1.51 introduced new lints to fix

### DIFF
--- a/src/stream-dispatcher/src/dispatcher/k8_dispatcher.rs
+++ b/src/stream-dispatcher/src/dispatcher/k8_dispatcher.rs
@@ -171,7 +171,7 @@ where
                         },
                         Err(err) => {
                             error!("WS channel error: {}", err);
-                            panic!(-1);
+                            panic!("WS channel error: {}", err);
                         }
                     }
                 }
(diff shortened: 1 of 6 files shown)


src/stream-dispatcher/src/dispatcher/k8_dispatcher.rs
```diff
@@ -171,7 +171,7 @@ where
                         },
                         Err(err) => {
                             error!("WS channel error: {}", err);
```
sehz, line 173: since we have panic, we don't need error?

sehz (changes requested):

src/stream-dispatcher/src/dispatcher/k8_dispatcher.rs
nicholastmosher: You know, I dunno. It was like that when I got here. Do you want me to remove that? I mean it doesn't hurt at the very least.

Actually, thinking about it more, if we at some point add more tracing modules such as opentelemetry, this would report the error to a remote system before the panic happens. So I think it's worth keeping. 🤷‍♂️

sehz: Thats good point

PR #898, opened 2021-03-25
Combine clippy and fmt jobs

nicholastmosher:
I think by putting these as two separate steps in one job rather than two jobs, we can save build time by not needing to do two separate full builds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,7 +46,7 @@ jobs:
   #      - name: Check Versions
   #        run: cargo cvm -x
 
-  check_clippy:
+  check_clippy_fmt:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


sehz:
What's before and after this change in terms of performance?

nicholastmosher:
I think the total CI job time is about the same. However, I think that GH allocates like X jobs per repo that are allowed to concurrently run, so if this reduces the number of jobs needed in a workflow I think that is a win.

sehz:
let's get actual number

nicholastmosher:
On the last job merged to master, CI took 42 minutes

image

This branch took 40m to run CI

image

Here you can see that in general, the CI job usually takes somewhere from 38m to 42m, so this seems pretty on-par

image

sehz:
Downside of this change:
* Increase overall duration since 2 jobs are executed in sequentially.
* With existing workflow, can see either of failure immediately and have better visibility.

Unless this show dramatic improvement in performance, don't think we should change.

PR #900, opened 2021-03-25
Update helm chart publish version with git commit

nicholastmosher:
I need this merged so that it starts publishing helm charts with the same version scheme as the Fluvio CLI. Please let me know if there is anything that needs to be changed :)

### DIFF
--- a/.github/workflows/publish_latest_chart.yml
+++ b/.github/workflows/publish_latest_chart.yml
@@ -21,10 +21,6 @@ jobs:
       - name: Helm Add Repo
         run: helm repo add chartmuseum https://gitops:${{ secrets.HELM_PASSWORD }}@charts.fluvio.io
       - name: Push Sys Chart
-        run: |
-          helm push k8-util/helm/fluvio-sys --version="$(cat VERSION)" --force chartmuseum
-          helm push k8-util/helm/fluvio-sys --version="$(cat VERSION)+$(git rev-parse HEAD)" --force chartmuseum
+        run: helm push k8-util/helm/fluvio-sys --version="$(cat VERSION)+$(git rev-parse HEAD)" chartmuseum
       - name: Push App Chart
-        run: |
-          helm push k8-util/helm/fluvio-app --version="$(cat VERSION)" --force chartmuseum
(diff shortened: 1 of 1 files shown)


.github/workflows/publish_latest_chart.yml
```diff
@@ -21,6 +21,10 @@ jobs:
       - name: Helm Add Repo
         run: helm repo add chartmuseum https://gitops:${{ secrets.HELM_PASSWORD }}@charts.fluvio.io
       - name: Push Sys Chart
-        run: helm push k8-util/helm/fluvio-sys --version="$(cat VERSION)-latest" --force chartmuseum
+        run: |
+          helm push k8-util/helm/fluvio-sys --version="$(cat VERSION)" --force chartmuseum
+          helm push k8-util/helm/fluvio-sys --version="$(cat VERSION)+$(git rev-parse HEAD)" --force chartmuseum
```
nacardin, line 26: We shouldn't need `--force` for this, as we are not overriding.

.github/workflows/publish_latest_chart.yml
```diff
@@ -21,6 +21,10 @@ jobs:
       - name: Helm Add Repo
         run: helm repo add chartmuseum https://gitops:${{ secrets.HELM_PASSWORD }}@charts.fluvio.io
       - name: Push Sys Chart
-        run: helm push k8-util/helm/fluvio-sys --version="$(cat VERSION)-latest" --force chartmuseum
+        run: |
+          helm push k8-util/helm/fluvio-sys --version="$(cat VERSION)" --force chartmuseum
```
nacardin, line 25: Do we need to publish release versions before releasing, is there a use case? Could this conflict with the release workflow?

nacardin (changes requested):

.github/workflows/publish_latest_chart.yml
nicholastmosher: Oh I see, I didn't see the release workflow for helm. I'll get rid of the duplicate one with just VERSION

.github/workflows/publish_latest_chart.yml
nicholastmosher: Good point

nacardin (approved):

PR #901, opened 2021-03-25
implement WASM based smart stream filter phase 1

sehz:
Smart Stream Filter/Transform(SSF) MVP.

With SSF, SPU can filter out records before sending them to a client.  SPU filter processing has the advantage of reducing time and bandwidth compared with filtering out in the client.  The filter can be written in any WASM-compatible language.   Each filter will be invoked for each batch.   Only records that have passed filter logic will be sent back.  

For client, filter can be only attached to consumer stream thru config as below:
```
let mut config = ConsumerConfig::default();
let buffer = std::fs::read(filter_path)?;
consume_config = consume_config.with_wasm_filter(buffer);
```
The client is responsible for loading WASM binaries.

When a consumer connects to SPU, WASM binaries are sent along as part of the request.  The SSF module is only valid for a single consumer streaming session.  For each new consumer streaming request, it must be re-loaded.

For each batch, a copy is sent to SSF.   The SSF is responsible for decoding and re-encoding back to SPU in the WASM context.   Since SSF is a WASM module, it only has access to data send by SPU and has no access to SPU.  Think of as micro VM.

This implementation uses non-thread-safe execution, which means it is possible to block SPU execution.


nicholastmosher:
Are there any specific instructions for building this? I'm running just `cargo build` in the project root and coming up with errors like this:

```
  = note: Undefined symbols for architecture x86_64:
            "_copy_records", referenced from:
                _filter in fluvio_filter_test.3uox9od8246ezzju.rcgu.o
          ld: symbol(s) not found for architecture x86_64
          clang: error: linker command failed with exit code 1 (use -v to see invocation)


error: could not compile `fluvio-filter-simple`

error: build failed
```

sehz:
Shouldn't be.  Let me check

PR #902, opened 2021-03-25
Bump version of fluvio-auth to fix publishing of fluvio-sc

simlay:
Some of the changes from https://github.com/infinyon/fluvio/pull/856 need version bumps. If I read that PR right I think `fluvio-controlplane` may also need a version bump but I'm not sure.

### DIFF
--- a/src/auth/Cargo.toml
+++ b/src/auth/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio-auth"
-version = "0.6.0"
+version = "0.5.0"
 authors = ["Fluvio Contributors <team@fluvio.io>"]
 edition = "2018"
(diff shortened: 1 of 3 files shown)


src/auth/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio-auth"
-version = "0.4.0"
+version = "0.6.0"
```
sehz, line 3: shouldn't this be 5.0?

sehz (changes requested):

src/auth/Cargo.toml
simlay: oh, whoops.

sehz (approved):

PR #904, opened 2021-03-27
Add test reporting and timeouts

### LINKED ISSUE #878: Test runner: Add basic test reporting
The test runner doesn't report any information about pass/fail of the test run. We want to change that so it has behavior that is similar to `cargo test`. (But we're only dealing with single tests at the moment).

We want to collect:
* Basic pass/fail status of a test
* Panic information from failures
* Duration of a test

And we want to print that information at the end of cleanup

This has 2 steps:
* Collection of report
* impl Display of report

### LINKED ISSUE #880: Test runner: Add test timeouts w/ graceful fail
For our high frequency tests, we want to set upper limits to their execution time.

Github actions is not the correct place to handle, because if the test exceeds this timeout, we want to clean up and print any information that may help narrow down where to look.

We want to add this to the fluvio_test macro, but additionally I'm wondering if there should be default timeout behavior? Like a 45/60 min upper limit by default. Typically all the CI services have one anyway, so this has more consequences to local usage.

I want to start with timeouts having an assumed unit. Probably **seconds**, since that's the largest unit offered by `std::time::Duration`.

The timer should start after any setup that occurs outside of the test itself. If the test run is longer than the timer duration, we want to assume the test has failed, begin cleanup and report the timeout.

### LINKED ISSUE #881: Test runner: Make topic creation idempotent
Topics are created prior to running tests by shelling out and using the fluvio cli. When we attempt to create a topic which already exists (like we might do with a cluster that is already deployed), we panic.

Rather than skip topic creation when we are using `--disable-install`, we should just not fail if the topic exists.

tjtelan:
Added new macro keys: name, test_name, timeout, cluster_type

Generated test code moved into `FluvioTest` so rust-analyzer can be more helpful

We catch the panic from thrown by user tests, and clean up the cluster (unless opted out)

`TestReport` is returned by successful tests, and includes the duration the user's test code was run (no setup time included)

Topic creation is done via API and is also idempotent (so we don't error out while using `--disable-install`)

Resolves #878
Resolves #880
Resolves #881


sehz:
Looks good:
```
deleting cluster
installing cluster
Performing pre-flight checks
Creating the topic: test
topic "test" created
found topic: test offset: 0
<<consume test done for: test >>>>
consume message validated!, records: 1000

Test Results:
Pass    : true
Duration: 10.754302676s

deleting cluster
```

sehz (approved):
LGTM

PR #908, opened 2021-03-29
Explicitly use latest Rust toolchain

### LINKED ISSUE #906: Mac CI latest failing
https://github.com/infinyon/fluvio/actions/runs/691862229

nicholastmosher:
Fixes #906 

I believe the reason the CI was failing is because an upstream crate started using const generics, and I'm not sure if we have explicitly declared using the latest stable toolchain.


nicholastmosher:
Re-opening on an upstream branch

PR #910, opened 2021-03-30
Bump fluvio to 0.6.1-alpha.1

nicholastmosher:


src/cluster/Cargo.toml
```diff
@@ -42,7 +42,7 @@ proclist = "0.9.2"
 remoteprocess = "0.4.2"
 
 # Fluvio dependencies
-fluvio = { version = "0.6.0", path = "../client", default-features = false }
+fluvio = { version = "0.6.1-alpha.1", path = "../client", default-features = false }
```
sehz, line 45: Is there reason need newer client api?

sehz (changes requested):

src/extension-common/Cargo.toml
```diff
@@ -27,5 +27,5 @@ futures-lite = { version = "1.7.0" }
 thiserror = "1.0.20"
 semver = { version = "0.11.0", features = ["serde"] }
 
-fluvio = { version = "0.6.0", path = "../client",  optional = true }
+fluvio = { version = "0.6.1-alpha.1", path = "../client",  optional = true }
```
sehz, line 30: Ditto

src/cluster/Cargo.toml
nicholastmosher: Since we also specify the `path` dependency as well, it requires you to use an exact matching version as what is in the repo.

src/extension-common/Cargo.toml
nicholastmosher: Using version and path together requires exact version as in repo

src/cluster/Cargo.toml
sehz: Yuck.  I guess we can revert back once we published regular version

sehz (approved):

PR #912, opened 2021-03-31
Update fluvio dependency on dataplane-protocol

nicholastmosher:
I went to publish `fluvio:0.6.1-alpha.1` this morning and it failed, making me realize I forgot to bump the dependency on `fluvio-dataplane-protocol` since I made changes there.

Also @sehz could you either publish `fluvio-dataplane-protocol:0.4.1` or invite me as an owner to the crate so I can do it?


nicholastmosher:
I'll use a git patch where I need it for now and we can publish these before next release

PR #918, opened 2021-04-02
Updates for 0.7.3 release

simlay:
I'm not missing anything, right?


nicholastmosher:
I think you may need to bump crate dependencies before you are allowed to publish, can you try doing a dry-run publish of the `fluvio` crate?

simlay:
> I think you may need to bump crate dependencies before you are allowed to publish, can you try doing a dry-run publish of the `fluvio` crate?

I did part of it. fluvio-sc-schema and fluvio-dataplane-protocol had patch version bumps that need publishing. I ran a `cargo check` on the package directory without those published to see if it would compile... the only error that's worrisome is:

```
error[E0277]: the trait bound `Arc<Vec<u8>>: From<bytes::Bytes>` is not satisfied
   --> src/producer.rs:141:23
    |
141 |             let key = key.map(DefaultAsyncBuffer::new);
    |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `From<bytes::Bytes>` is not implemented for `Arc<Vec<u8>>`
```

nicholastmosher:
Oh I think you need to bump the version of `dataplane-protocol` in the `fluvio` dependencies

simlay:
> Oh I think you need to bump the version of `dataplane-protocol` in the `fluvio` dependencies

Yeah, some of the stuff in #905 were breaking changes. I wish it were easier to know if I needed to bump everything that depends on `dataplane-protocol`.

simlay:
I feel like this needs more version bumps but I'm not sure where.

PR #927, opened 2021-04-05
Remove unused modules. Renamed some modules

nicholastmosher:
This is the second half of the earlier client PR that I split up. All I did here was move a few modules to make some things clearer:

- Changed `src/client/cluster.rs` to `src/fluvio.rs` since it is home to the main `Fluvio` type
- Changed `src/client/client.rs` to `src/sockets.rs` since it deals with `VersionedSocket` and `VersionedSerialSocket`
- Removed `src/client/mod.rs` since it was empty at this point
- Changed `src/spu/spu_pool.rs` to `spu.rs`
- Removed `src/spu/mod.rs` since it was empty at this point

I think this makes it much more clear what each module is for, and it reduces the number of empty `mod.rs` files that don't really do anything.


sehz (approved):
LGTM

sehz:
rebase

PR #928, opened 2021-04-05
[Merged by Bors] - Decouple cluster components from Fluvio cluster

### LINKED ISSUE #866: Separate Fluvio cluster components into an independent plugin
We are beginning to run into problems building the CLI and the cluster components (SC and SPU) together. The latest problem is trying to build for the Raspberry Pi targets. If we split these into multiple binaries, we can tackle these problems separately. Since we have the infrastructure for custom plugins on packages.fluvio.io, it would work nicely to distribute the cluster components as an independent binary plugin.

The main question to be resolved here is what we should call the plugin, since that will have an impact on how it is invoked from the CLI going forward. I think an easy candidate would be `fluvio-run`, so we could run the SC and SPU as `fluvio run sc` and `fluvio run spu`. If anybody has any objections or other preferences for this, please comment below.

nicholastmosher:
Closes #866 

- Removes `fluvio-runner-local` as a dependency from `fluvio-cluster`
- Renames `fluvio-runner-local` to `fluvio-run` and makes it a standalone executable
- Adds `fluvio-run` as an independent plugin on packages.fluvio.io
- Updates install.sh to include `fluvio-run` as a default plugin during installation


sehz:
Can you try to squash commit to get good commit log?

nicholastmosher:
I can update the PR description so that it becomes a nice commit message when Bors squashes it 👍

src/runner/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
-name = "fluvio-runner-local"
-version = "0.4.0"
+name = "fluvio-run"
+version = "0.1.0-alpha.1"
```
sehz: let's remove `alpha` since we are not publishing and causing problem

src/sc/build.rs
```diff
@@ -0,0 +1,5 @@
+fn main() {
+    // Copy VERSION file. Do not fail e.g. when built via `cargo publish`
```
sehz, line 2: since we don't need to publish, we can directly refer to VERSION

src/cluster/Cargo.toml
```diff
@@ -16,7 +16,6 @@ cli = [
     "structopt",
     "color-eyre",
     "prettytable-rs",
-    "fluvio-runner-local/platform",
     "fluvio-extension-common/target",
 ]
```
sehz, line 21: does this need version bump?

sehz (changes requested):
great work.  few minor comments

src/runner/Cargo.toml
nicholastmosher: I think it would be good for us to use the Platform version for the `fluvio-run` version. I set it to 0.7.4 since that is the version this will be released in, let me know if you prefer something different.

src/cluster/Cargo.toml
nicholastmosher: I don't think so, because the API of the cluster crate has stayed the same

src/cluster/src/delete.rs
```diff
@@ -191,6 +191,7 @@ impl ClusterUninstaller {
         Command::new("pkill")
             .arg("-f")
             .arg("fluvio cluster run")
```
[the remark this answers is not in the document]
nicholastmosher: I have confirmed that including both "fluvio cluster run" and "fluvio-run" as arguments allows the `fluvio cluster delete` command to work with both the new local cluster and the old local cluster setups 👍

src/command/src/lib.rs
```diff
@@ -113,7 +143,17 @@ impl CommandExt for Command {
...
+        debug!("Command> {:?}", self.display());
+        self
+    }
+
+    fn print(&mut self) -> &mut Self {
+        println!("Command> {:?}", self.display());
+        self
+    }
+
+    fn display(&self) -> String {
```
sehz, line 156: this changes API.  So need to bump up to 0.3?

src/runner/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
-name = "fluvio-runner-local"
-version = "0.4.0"
+name = "fluvio-run"
+version = "0.7.4"
```
sehz, line 3: Tying this to fluvio version means, we have to update them every time.  We should just change this to basic version `0.1.0`

sehz (changes requested):

src/runner/Cargo.toml
nicholastmosher: Hmm, gotcha. In that case I wonder if we should not even touch the Cargo.toml version for this crate at all, just use VERSION in our deploy scripts for pushing this to packages.fluvio.io. Because when users install via `install.sh`, we need a way to make sure they get a matching `fluvio` and `fluvio-run`.

src/command/src/lib.rs
nicholastmosher: It doesn't edit any existing API, just adds new. So I think bump to `0.2.1` should be good

src/runner/Cargo.toml
nicholastmosher: Set to `0.1.0` in Cargo.toml 👍

sehz: Yup.  This is where need to override for publishing

sehz:
Running smoke test didn't delete `run` components:
```
make smoke-test
cargo build   --bin fluvio 
    Finished dev [unoptimized + debuginfo] target(s) in 0.19s
cargo build   --bin fluvio-run 
    Finished dev [unoptimized + debuginfo] target(s) in 0.20s
cargo build   --bin flv-test 
    Finished dev [unoptimized + debuginfo] target(s) in 0.20s
echo "clean up previous installation"
clean up previous installation
./target/debug/fluvio cluster delete
./target/debug/fluvio cluster delete --local
FLUVIO_CMD=true ./target/debug/flv-test smoke --spu 1 --local --client-log warn --server-log fluvio=debug --skip-checks  -- --producer-iteration=1000
deleting cluster
installing cluster
Performing pre-flight checks
Creating the topic: test
topic "test" created
found topic: test offset: 0
<<consume test done for: test >>>>
consume message validated!, records: 1000
Test completed
deleting cluster
+--------------+---------------+
| Test Results |               |
+--------------+---------------+
| Pass?        | true          |
+--------------+---------------+
| Duration     | 12.267184004s |
+--------------+---------------+

ubuntu@ip-172-31-20-82:~/fluvio$ ps -ef | grep fluvio
ubuntu   3645556       1  0 16:45 pts/11   00:00:00 /home/ubuntu/fluvio/target/debug/fluvio run sc
ubuntu   3645566 3645556 88 16:45 pts/11   00:02:45 /home/ubuntu/fluvio/target/debug/fluvio-run sc
ubuntu   3645583       1  0 16:45 pts/11   00:00:00 /home/ubuntu/fluvio/target/debug/fluvio run spu -i 5001 -p 0.0.0.0:9010 -v 0.0.0.0:9011 --log-base-dir /tmp/fluvio
ubuntu   3645593 3645583  6 16:45 pts/11   00:00:11 /home/ubuntu/fluvio/target/debug/fluvio-run spu -i 5001 -p 0.0.0.0:9010 -v 0.0.0.0:9011 --log-base-dir /tmp/fluvio
ubuntu   3652910 3000105  0 16:48 pts/11   00:00:00 grep --color=auto fluvio
```

sehz:
Also,  can you update `DEVELOPER.md`?

sehz:
Delete still doesn't work in Linux:
```
flvd cluster delete --local
ubuntu@ip-172-31-20-82:~/fluvio$ ps -ef | grep fluvio
ubuntu   3645556       1  0 16:45 pts/11   00:00:00 /home/ubuntu/fluvio/target/debug/fluvio run sc
ubuntu   3645566 3645556  1 16:45 pts/11   00:02:52 /home/ubuntu/fluvio/target/debug/fluvio-run sc
ubuntu   3645583       1  0 16:45 pts/11   00:00:00 /home/ubuntu/fluvio/target/debug/fluvio run spu -i 5001 -p 0.0.0.0:9010 -v 0.0.0.0:9011 --log-base-dir /tmp/fluvio
ubuntu   3645593 3645583  0 16:45 pts/11   00:00:36 /home/ubuntu/fluvio/target/debug/fluvio-run spu -i 5001 -p 0.0.0.0:9010 -v 0.0.0.0:9011 --log-base-dir /tmp/fluvio
ubuntu   3737568 3000105  0 20:09 pts/11   00:00:00 grep --color=auto fluvio
```

sehz:
Also, there are two process spawn for each SC and SPU.  Can we collapse into a single process for each?

For example, `running sc` spawn two process.
```
ubuntu   3738481       1  0 20:10 pts/11   00:00:00 /home/ubuntu/fluvio/target/debug/fluvio run sc
ubuntu   3738493 3738481  2 20:10 pts/11   00:00:00 /home/ubuntu/fluvio/target/debug/fluvio-run sc
```

nicholastmosher:
I noticed that as well, the reason that there are two processes is because the actual invocation for cluster components is now

- `fluvio run spu` -> `fluvio-run spu`, and
- `fluvio run sc` -> `fluvio-run sc`.

This has to do with how we use `fluvio` to call out to other subcommands, and right now that simply runs the subcommand inside itself rather than spawning as an independent process. I think this is better for now because otherwise, running any kind of external fluvio subcommand would appear to just quit the prompt early, even though the subcommand is still running. I think this would really hurt plugins like `fluvio-cloud`.

So I think  it's fine how it is right now. I'll make a quick change that will hopefully fix the pkill on linux.

sehz:
Hooray, cluster delete finally work!

sehz (approved):

nicholastmosher:
Whoops, I broke local test, one sec

PR #931, opened 2021-04-05
Removed publish crates workflow

simlay:
Remove unused and broken publish workflow.


sehz:
Still need to be done right?

simlay:
> Still need to be done right?

Yeah, I mean this would be awesome if we could use the publish workflow. Its just too militant on the versions needing to match exactly.

nicholastmosher:
I ended up removing `publish_crates` sometime while reworking the CI/CD workflows, closing

PR #937, opened 2021-04-06
Fix sccache and rewrite CI jobs as a matrix

nicholastmosher:
Alright, I managed to make some substantial breakthroughs in the last half of today. The entire CI workflow is now consistently using sccache and running in about 9-10 minutes end-to-end. I also managed to cut down on the workflow file by quite a lot by writing a single job that uses a matrix to test our different scenarios.

```yaml
  tests:
    name: ${{ matrix.make.name }} (${{ matrix.os }})
    runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        os: [ubuntu-latest, macos-latest]
        rust: [stable]
        make:
          - name: Clippy
            task: "check-clippy"
          - name: Unit tests
            task: "build-all-test run-all-unit-test"
          - name: Doc tests
            task: "run-all-doc-test"
        include:
          - os: ubuntu-latest
            sccache-path: /home/runner/.cache/sccache
          - os: macos-latest
            sccache-path: /Users/runner/Library/Caches/Mozilla.sccache
        exclude:
          - os: macos-latest
            rust: stable
            make:
              name: Clippy
```

This creates a set of jobs like the following:

image

(I will be re-enabling the cluster tests)

Basically, clippy, unit tests, and doc tests all want the same job setup: sccache, rust toolchain, etc. so they all share a job definition, and we put any of the varying details into the matrix. The matrix also contains the name of a Make task to run, which is how the one job definition can serve as multiple jobs. From here on out, if we want any major new test setup, it should be contained within a single new Make task that we want to run within the same job environment.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,14 @@ on:
   pull_request:
     branches: [master]
   workflow_dispatch:
+    inputs:
+      verbose:
+        description: "Set --verbose to get verbose build output"
+        required: false
+        default: ''
+
+env:
+  VERBOSE: ${{ github.events.input.verbose }}
 
 jobs:
   cancel_previous_runs:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,72 +15,9 @@ jobs:
       - uses: styfle/cancel-workflow-action@0.4.1
         with:
           access_token: ${{ secrets.GITHUB_TOKEN }}
-  # check_helm_versions:
-  #   runs-on: ubuntu-latest
-  #   steps:
-  #     - uses: actions/checkout@v1
-  #     - name: Install YQ CLI Tools
-  #       run: |
-  #         sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys CC86BB64
-  #         sudo add-apt-repository ppa:rmescandon/yq
-  #         sudo apt update
-  #         sudo apt install yq -y
-  #     - name: Cancel Workflow Action
-  #       uses: styfle/cancel-workflow-action@0.4.1
-  #       with:
-  #         access_token: ${{ secrets.GITHUB_TOKEN }}
-  #     - name: check versions
-  #       run: make -s check_version
 
-  #  version_check:
-  #    runs-on: ubuntu-latest
-  #    steps:
-  #      - uses: actions/checkout@v2
-  #        with:
-  #          fetch-depth: 0
-  #      - name: Install cargo-cvm
-  #        uses: actions-rs/install@v0.1
-  #        with:
-  #          crate: cargo-cvm
-  #          version: latest
-  #      - name: Check Versions
-  #        run: cargo cvm -x
-
-  check_clippy:
-    runs-on: ${{ matrix.os }}
-    strategy:
-      matrix:
-        os: [ubuntu-latest]
-        rust: [stable]
-    steps:
-      - uses: actions/checkout@v2
-      - name: Install ${{ matrix.rust }}
-        uses: actions-rs/toolchain@v1
-        with:
-          toolchain: ${{ matrix.rust }}
-          override: true
-      - name: check clippy
-        run: make check-clippy RUSTV=${{ matrix.rust }}
-
-  check_fmt:
-    name: check cargo fmt
-    runs-on: ${{ matrix.os }}
-    strategy:
-      matrix:
-        os: [ubuntu-latest]
-        rust: [stable]
-    steps:
-      - uses: actions/checkout@v2
-      - name: Install ${{ matrix.rust }}
-        uses: actions-rs/toolchain@v1
-        with:
-          toolchain: ${{ matrix.rust }}
-          override: true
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -170,180 +80,139 @@ jobs:
...
-          name: fluvio-local-sc-logs
-          path: /usr/local/var/log/fluvio/flv_sc.log
-      - name: Upload Spu logs
-        timeout-minutes: 5
-        if: failure()
-        uses: actions/upload-artifact@v2
-        with:
-          name: fluvio-local-spu-logs
-          path: /usr/local/var/log/fluvio/spu_log_5001.log
+#  mac_local_cluster_test:
```
sehz, line 119: let's remove commented out code

sehz (changes requested):

sehz:
What's improvement from previous changes?  I am looking at stat and not seeing any improvement

sehz:
But this PR seems to clean up job definitions

.github/workflows/ci.yml
nicholastmosher: Sure... the only reason I hadn't removed this was because it had specific setup that I didn't re-capture anywhere else. Actually, I may be able to incorporate it into the other local cluster test now that I know how to use the matrix functionality better. But I'll remove it for now, if we want to get it back we can dig in git.

nicholastmosher:
Yeah so the changes do not _look_ significant but I did make some changes to the sccache configuration that makes it hit the cache even while using GH cache. Let me try to remember all of the little changes in here:

- Consolidate jobs into one definition
- Fix sccache setup so it uses GH cache and has majority cache-hits. I think this improved some build times:
  - Unit tests on linux dropped from 12m to about 9m
  - Unit tests on mac dropped from I think about 16m to maybe 12ish?
- I updated the clippy make task to first build with `cargo check` and then use `cargo clippy` because clippy does not leverage sccache but check does, but then clippy will use the build output from check.

.github/workflows/ci.yml
```diff
@@ -15,151 +15,61 @@ jobs:
...
+      - name: Check rustfmt
+        run: make check-fmt
 
-  doc_test:
-    name: Doc Test
+  tests:
+    name: ${{ matrix.make.name }} (${{ matrix.os }})
     runs-on: ${{ matrix.os }}
     strategy:
-      fail-fast: false
```
sehz, line 142: why remove this?

nicholastmosher: I was trying a whole bunch of things and moving stuff around, there probably was not an exact reason. I can put it back

sehz:
From previous run: https://github.com/infinyon/fluvio/runs/2292688450?check_suite_focus=true,
 
Clippy build time: 11m 50 s -> ~7M
Fmt: 7s -> 7s
Unit Test Linux build: 12m 6s -> 10m 21s
Unit Test Mac:  15m -> 12m

.github/workflows/ci.yml
```diff
@@ -170,180 +81,41 @@ jobs:
...
         uses: actions/cache@v2
-        continue-on-error: true
-        with:
-          path: ~/.cargo/registry/cache
-          key: test-${{ runner.os }}-${{ matrix.conf }}-cargo-registry-${{ hashFiles('**/Cargo.lock') }}
-          restore-keys: |
-            test-${{ runner.os }}-${{ matrix.conf }}-cargo-registry-
-      - name: Cache sccache output
-        uses: actions/cache@v2
-        continue-on-error: true
```
sehz, line 234: Can you remove verbose?
```
 Running `sccache rustc --crate-name cfg_if --edition=2018 /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/cfg-if-1.0.0/src/lib.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts --crate-type lib --emit=dep-info,metadata,link -C embed-bitcode=no -C debuginfo=1 -C metadata=7be9226c66ab2048 -C extra-filename=-7be9226c66ab2048 --out-dir /Users/runner/work/fluvio/fluvio/target/debug/deps -L dependency=/Users/runner/work/fluvio/fluvio/target/debug/deps --cap-lints allow`
```

sehz (changes requested):

.github/workflows/ci.yml
sehz: Problem is that it is hard to look at unit test log because SCCACHE log was filling up output.  How about option to set verbose flag when invoked by `workflow_dispatch`

nicholastmosher: I updated the Makefile to append the value of $(VERBOSE) to build tasks and I set the workflow dispatch up with an input for a "verbose" string that you can fill with `--verbose` in order to make it run in  verbose mode. So now by default things will run without verbose but you can use the workflow dispatch input to set it

sehz (approved):
LGTM.  Thanks for refactoring to make it cleaner structure

sehz:
Unit test failure are covered by: #947

PR #940, opened 2021-04-08
[Merged by Bors] - Update 'latest' workflows to enable cross-repo testing

nicholastmosher:
This changes the workflow to publish charts with versions that are unique rather than using build metadata.

Instead of `VERSION+COMMIT` we now have `VERSION-COMMIT`. This will cause all of the chart versions _not_ to be considered equal, as was the problem before. I have not yet re-enabled the workflow on-push, but I will after we have confirmed that everything works end-to-end.

Update: I've also now added a build-time environment variabel, `FLUVIO_VERSION_SUFFIX`, which should be set to contain the git hash of the checked-out branch. When it is present, the VERSION that gets baked into the cluster and CLI will be something like `0.7.4-alpha.0-abcdef`, where `abcdef` is the FLUVIO_VERSION_SUFFIX` that was passed. This variable will only be set in the "fluvio_latest" workflows, in regular builds and releases it will be absent.


sehz (approved):

sehz:
@nacardin Ok with this change?

sehz (approved):

nicholastmosher:
Hmm, seems like it was a docker registry problem?

```
rustup target add x86_64-unknown-linux-musl
info: component 'rust-std' for target 'x86_64-unknown-linux-musl' is up to date
cargo build --release   \
    Finished release [optimized] target(s) in 0.20s
	--bin fluvio-run --target x86_64-unknown-linux-musl
echo "Building Fluvio musl image with version: 0.7.4-alpha.0"
Building Fluvio musl image with version: 0.7.4-alpha.0
export CARGO_PROFILE=release; \
export MINIKUBE_DOCKER_ENV=true; \
export DOCKER_TAG=f2a2259fd16f57708127d30575740abf1530ced8; \
k8-util/docker/build.sh
Sending build context to Docker daemon   30.7MB

Step 1/2 : FROM alpine:3.12
unauthorized: authentication required
make: *** [fluvio_image] Error 1
Makefile:224: recipe for target 'fluvio_image' failed
```

Let's try this again

bors r+

Issue #849, opened 2021-03-10
SPU K8 healthcheck / liveness probe


nacardin:

nacardin:
Will not merge

PR #943, opened 2021-04-08
SPU liveness check

nacardin:
Resolves https://github.com/infinyon/fluvio/issues/849

### DIFF
--- a/src/client/src/error.rs
+++ b/src/client/src/error.rs
@@ -7,12 +7,14 @@ use crate::config::ConfigError;
 use semver::Version;
 
 /// Possible errors that may arise when using Fluvio
+#[allow(clippy::upper_case_acronyms)]
 #[derive(Error, Debug)]
 pub enum FluvioError {
     #[error("Topic not found: {0}")]
     TopicNotFound(String),
     #[error("Partition not found: {0}-{1}")]
     PartitionNotFound(String, i32),
+    #[allow(clippy::upper_case_acronyms)]
     #[error("Spu not found: {0}")]
     SPUNotFound(i32),
     #[error(transparent)]
--- a/src/client/src/error.rs
+++ b/src/client/src/error.rs
@@ -7,14 +7,12 @@ use crate::config::ConfigError;
 use semver::Version;
 
 /// Possible errors that may arise when using Fluvio
-#[allow(clippy::upper_case_acronyms)]
 #[derive(Error, Debug)]
 pub enum FluvioError {
     #[error("Topic not found: {0}")]
     TopicNotFound(String),
     #[error("Partition not found: {0}-{1}")]
     PartitionNotFound(String, i32),
-    #[allow(clippy::upper_case_acronyms)]
     #[error("Spu not found: {0}")]
     SPUNotFound(i32),
     #[error(transparent)]
--- a/src/extension-runner-local/src/bin/main.rs
+++ b/src/extension-runner-local/src/bin/main.rs
@@ -1,6 +1,6 @@
 use structopt::StructOpt;
 use fluvio_future::task::run_block_on;
-use fluvio_runner_local::{RunCmd, RunnerError};
+use fluvio_runner_local::{SpuLivenessCheckCmd, RunCmd, RunnerError};
 
 fn main() {
     fluvio_future::subscriber::init_tracer(None);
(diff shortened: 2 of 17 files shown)


src/client/src/error.rs
```diff
@@ -7,12 +7,14 @@ use crate::config::ConfigError;
 use semver::Version;
 
 /// Possible errors that may arise when using Fluvio
+#[allow(clippy::upper_case_acronyms)]
```
nicholastmosher, line 10: We added `-A clippy::upper_case_acronyms` to the `make check-clippy` task, so I think you shouldn't need to have these to pass CI now

src/extension-runner-local/src/bin/main.rs
```diff
@@ -14,8 +14,11 @@ fn main() {
 #[derive(Debug, StructOpt)]
 pub enum RunnerCmd {
     /// Run a Streaming Controller (SC) or SPU
-    #[structopt(name = "run")]
+    #[structopt()]
     Run(RunCmd),
+    /// Run a new Streaming Processing Unit (SPU)
+    #[structopt()]
```
nicholastmosher, line 20: Is this supposed to have a short name?

nicholastmosher:
I see there's a new CLI command here, what is the intended usage? The command is on the SPU cluster component, so is it supposed to be a different process that also runs inside the cluster? Is this code the client or the server to probe requests?

src/client/src/error.rs
sehz: I agree.

src/sc/src/k8/objects/spg_group.rs
```diff
@@ -217,16 +220,29 @@ mod k8_convert {
...
+            args.push(format!("0.0.0.0:{}", spu_non_tls_public));
         }
 
         // add RUST LOG, if passed
         if let Ok(rust_log) = std::env::var("RUST_LOG") {
             env.push(Env::key_value("RUST_LOG", &rust_log));
         }
 
-        // env.append(&mut spu_template.env.clone());
+        let liveness_probe = Probe {
```
sehz, line 234: Don't think liveness probes should run in the SC

src/spu/src/liveness_check.rs
```diff
@@ -0,0 +1,24 @@
...
+    net::{SocketAddr, TcpStream},
+};
+use std::io::prelude::*;
+
+use tracing::debug;
+
+use fluvio_protocol::{Encoder, api::RequestMessage};
+use dataplane::versions::ApiVersionsRequest;
+
+pub fn probe(endpoint: SocketAddr) -> Result<()> {
```
sehz, line 13: Is checking SPU or SC?

src/client/src/error.rs
nacardin: Removed

sehz (changes requested):

src/extension-runner-local/src/bin/main.rs
nacardin: No

src/spu/src/liveness_check.rs
nacardin: SPU

src/sc/src/k8/objects/spg_group.rs
nacardin: Here it is being added to the SPG StatefulSet config

nacardin:
> I see there's a new CLI command here, what is the intended usage? The command is on the SPU cluster component, so is it supposed to be a different process that also runs inside the cluster? Is this code the client or the server to probe requests?

K8 executes this as a separate process on an interval in order to check for liveness. This is run inside the same pod as the SPU. This is a client which checks that the SPU is responsive.

sehz:
bors +r

src/sc/src/k8/objects/spg_group.rs
nicholastmosher: This just allows the SC to configure the K8 probe right? So K8 will be running the probe but the SC is the thing that sets it up in the  cluster? If I am interpreting the flow right

nacardin: @nicholastmosher Correct

src/spu/src/liveness_check.rs
sehz: As I understand, probe should exist outside application not inside

nicholastmosher: I think what he's doing is adding the client and server side of this probe in the same executable. So the SPU server running in the pod will respond to probe requests, but the K8 cluster also needs to know how to send requests that the server will respond to. So this subcommand of the SPU executable will act as a client for that probe request. Is that right @nacardin?

nacardin: @nicholastmosher Correct

src/extension-runner-local/src/bin/main.rs
nicholastmosher: Gotcha. Ok, then this attribute isn't really necessary :P but it also doesn't hurt really

nicholastmosher:
How can I observe this at work? Is there a way to watch the probe going?

nacardin:
> How can I observe this at work? Is there a way to watch the probe going?

Just `fluvio cluster start --develop`

PR #948, opened 2021-04-09
[Merged by Bors] - Add nodeSelector option for SC and SPU

nacardin:
This lets the pods target K8 nodes by label so that SC and SPU can be deployed on specialized nodes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2564,14 +2564,13 @@ dependencies = [
 
 [[package]]
 name = "k8-types"
-version = "0.2.1"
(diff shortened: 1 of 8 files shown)


sehz (approved):

nicholastmosher:
(I'm so sorry, the bors PR errored out and bors.toml has not landed. Until then @sehz can merge as admin)

nicholastmosher:
I'm merging an update to `bors.toml` that will let it skip the macos jobs, this should work once that lands.

PR #952, opened 2021-04-12
Make Bors ignore macos-latest jobs because of brew failing

nicholastmosher:
This will allow us to `bors r+` on PRs even if these jobs are failing.

I'm introducing this because today, the macos runners seem to be unable to use brew correctly

image


sehz (approved):

nicholastmosher:
Found a proper fix and implemented in #953

PR #960, opened 2021-04-14
[Merged by Bors] - Make publisher flow build fluvio and fluvio-run

nicholastmosher:
Prior to this, the publisher flow would only build `--bin fluvio`, now it will build `--bin fluvio-run` as well.

### DIFF
--- a/makefiles/Base.toml
+++ b/makefiles/Base.toml
@@ -27,12 +27,16 @@ extend = "build-prod-musl"
 #
 
 [tasks.build-dev-mac.mac]
-command = "cargo"
-args = ["build", "--bin", "fluvio", "--target", "x86_64-apple-darwin"]
+script = """
+cargo build --bin fluvio --target=x86_64-apple-darwin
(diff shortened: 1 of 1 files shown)


makefiles/Base.toml
```diff
@@ -47,22 +51,17 @@ script = "echo Cannot build for Mac from Linux && false"
...
 [tasks.build-dev-musl.linux]
 dependencies = ["setup-musl"]
-command = "cargo"
-args = [
-    "build",
-    "--bin", "fluvio",
-    "--target=x86_64-unknown-linux-musl",
-]
+script = """
+cargo build --bin fluvio --target=x86_64-unknown-linux-musl
```
sehz, line 55: This is ok for now.  But this is where we could collapse into a single script/rule rather than duplicating logic.  @tjtelan Please add this to where we can clean up.

sehz: We will be adding more targets pretty soon

sehz (approved):

Issue #962, opened 2021-04-15
Support upgrade for local cluster with persistent data


tjtelan:
The process for upgrading a local cluster is to stop the current cluster and start a new cluster at a new version.

Data is lost when the initial cluster is stopped. We want to provide the capability of running local clusters that can keep data while upgrading from one version to another.

- [x] #4136
- [x] #4137

digikata:
Some background which to focus this issue a bit more:

https://github.com/infinyon/fluvio/pull/3999 - added some good protection from accidentally upgrading local cluster versions  (e.g. a `fluvio cluster start` - local start with one version of fluvio,  `fluvio cluster shutdown`, a fluvio version upgrade occurs, `fluvio cluster resume` with a new version)

Adding a code path to run local data upgrade logic when a version difference is detected on `fluvio cluster resume` would allow some upgrade logic to be specified. 

Disallowing many configuration changes on `resume` will simplify this process. e.g. if the initial cluster was started with 2 spus, a resume should only be initialized with the former spu count, and no path to change the option on resume should be given.

Since most metadata updates are additions of fields, upgrade logic for this kind of change will mostly be reading all older data and ensuring deserializing works with a new default. 

More complex upgrades should be accommodated later on a case by case basis - but probably triggered on different specific version boundaries. (either semver based boundaries, or specified versions in the upgrade code).

digikata:
The most basic use case to support is to allow version upgrades between known good versions -e.g. 0.11.10 -> 0.11.11

PR #923, opened 2021-04-04
[Merged by Bors] - Add experimental benchmark support

tjtelan:
Built off of PR #895

* Add benchmarking support utilizing the [bencher](https://crates.io/crates/bencher) crate.
* Add new benchmark supported test `producer_stress`, a single-core producer
* Update README

Resolves #879
Resolves #883 
Resolves #951 
Closes #566 

### Without benchmarking
```
$ cargo run --bin flv-test -- producer_stress -d -k -- --producers 3 --iteration 100
    Finished dev [unoptimized + debuginfo] target(s) in 0.16s
     Running `target/debug/flv-test producer_stress -d -k -- --producers 3 --iteration 100`
Start running fluvio test runner
skipping cluster start
Creating the topic: test
topic "test" already exists

Starting single-process producer stress
Producers              : 3
# messages per producer: 100
skipping cluster delete

+--------------+--------------+
| Test Results |              |
+--------------+--------------+
| Pass?        | true         |
+--------------+--------------+
| Duration     | 289.408879ms |
+--------------+--------------+
```

### With benchmarking

```
$ cargo run --bin flv-test -- producer_stress -d -k --benchmark -- --producers 3 --iteration 100
    Finished dev [unoptimized + debuginfo] target(s) in 0.17s
     Running `target/debug/flv-test producer_stress -d -k --benchmark -- --producers 3 --iteration 100`
Start running fluvio test runner
skipping cluster start
Starting benchmark test (Usual test output may be silenced)
+-------------------------------+------------------------------------------+
| Perf Test Summary             | Measurement                              |
+-------------------------------+------------------------------------------+
| Total time elapsed            | 1.402356271s                             |
+-------------------------------+------------------------------------------+
| Total time elapsed (ns)       | 1402356271                               |
+-------------------------------+------------------------------------------+
| Time per iteration            | 280.471254ms                             |
+-------------------------------+------------------------------------------+
| Time per iteration (ns)       | 280471254                                |
+-------------------------------+------------------------------------------+
| # of iteration                | 5                                        |
+-------------------------------+------------------------------------------+
| Sum (ns)                      | 14306013620.45                           |
+-------------------------------+------------------------------------------+
| Min (ns)                      | 281061233.6                              |
+-------------------------------+------------------------------------------+
| Max (ns)                      | 290169616.55                             |
+-------------------------------+------------------------------------------+
| Mean (ns)                     | 286120272.40900004                       |
+-------------------------------+------------------------------------------+
[...]

PR #964, opened 2021-04-15
[Merged by Bors] - Add Upgrade test to CI

### LINKED ISSUE #951: Add upgrade testing workflow in CI
The basic upgrade workflow:
- Install latest released version of Fluvio and start a cluster
- Build Fluvio from master branch
- Upgrade running cluster w/ master branch (--develop) version

We will want to make this workflow more robust in the future (e.g. adding data into cluster and verifying after upgrade) but this basic workflow is sufficient as we're doing it manually today.

tjtelan:
Split off of #923 

Adds kubernetes based upgrade testing (on custom runner due to #962)

The workflow is as follows:

- Install stable version Fluvio, and start a cluster
- Build minikube image from master branch (which we assume is always same or newer version of stable. Not downgrade)
- Upgrade cluster
- Run smoke test k8

Resolves #951

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -266,3 +266,91 @@ jobs:
         with:
           name: fluvio-k8-logs
           path: /tmp/flv_sc.log
+
+  k8_cluster_upgrade:
+    name: Kubernetes cluster upgrade test
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [infinyon-ubuntu-bionic]
+        rust: [stable]
+    env:
+      FLUVIO_CMD: true
+      FLV_SOCKET_WAIT: 600
+      FLV_CLUSTER_MAX_SC_VERSION_LOOP: 120
+      FLV_CLUSTER_MAX_SC_NETWORK_LOOP: 60
+      FLV_TEST_CONSUMER_WAIT: 300000
+    steps:
+      - uses: actions/checkout@v2
+      - run: helm version
+      - name: Install Rust ${{ matrix.rust }}
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: ${{ matrix.rust }}
+          profile: minimal
+          override: true
+      - name: Setup Minikube for Linux
+        if: startsWith(matrix.os, 'infinyon-ubuntu')
+        run: |
+          pkill -f "minikube tunnel" || true
+          minikube delete
+          minikube start --driver=docker --kubernetes-version 1.19.6
+          nohup  minikube tunnel --alsologtostderr > /tmp/tunnel.out 2> /tmp/tunnel.out &
+      - name: Test minikube
+        run: |
+          minikube profile list
+          minikube status
+      - name: Install stable Fluvio CLI and start cluster
+        run: |
+          curl -fsS https://packages.fluvio.io/v1/install.sh | bash
+          ~/.fluvio/bin/fluvio cluster start
+          ~/.fluvio/bin/fluvio version
+      # Build, upgrade and test current version
+      - name: Build
+        run: |
+          make RELEASE=release TARGET=x86_64-unknown-linux-musl build_test
+      - name: Setup installation pre-requisites
+        run: |
+          make RELEASE=true TARGET=x86_64-unknown-linux-musl  k8-setup
+      - name: Make image
+        run: make RELEASE=true minikube_image
+      - name: Upgrade cluster
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -266,3 +266,90 @@ jobs:
...
+    env:
+      FLUVIO_CMD: true
+      FLV_SOCKET_WAIT: 600
+      FLV_CLUSTER_MAX_SC_VERSION_LOOP: 120
+      FLV_CLUSTER_MAX_SC_NETWORK_LOOP: 60
+      FLV_TEST_CONSUMER_WAIT: 300000
+    steps:
+      - uses: actions/checkout@v2
+      - run: helm version
+      - name: Install ${{ matrix.rust }}
```
nicholastmosher, line 285: I like to put `Install Rust ${{ matrix.rust }}` so that the action printout says "Install Rust stable" rather than "Install stable"

nicholastmosher:
Which commits belong just to this PR? I think a lot of the diff here is shared with your benchmarking PR

.github/workflows/ci.yml
```diff
@@ -266,3 +266,90 @@ jobs:
...
+          toolchain: ${{ matrix.rust }}
+          profile: minimal
+          override: true
+      - name: Setup Minikube for Linux
+        if: startsWith(matrix.os, 'infinyon-ubuntu')
+        run: |
+          pkill -f "minikube tunnel" || true
+          minikube delete
+          minikube start --driver=docker --kubernetes-version 1.19.6
+          nohup  minikube tunnel --alsologtostderr > /tmp/tunnel.out 2> /tmp/tunnel.out &
```
sehz, line 297: there is existing script in the `k8-util/minikube.sh`.  Can you create issue to use common that same script to set up everywhere?

sehz, line 297: There is script to install minikube at: `k8-util/reset-minikube.sh`.  Can you create separate issue to use that in all Github action that uses custom-runner?

Cargo.lock
```diff
@@ -392,6 +392,12 @@ version = "0.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "904dfeac50f3cdaba28fc6f57fdcddb75f49ed61346676a78c4ffe55877802fd"
```
sehz, line 394: don't think there should be any changes in here

tests/README.md
```diff
@@ -63,8 +63,19 @@ Test runner can be a running in two ways:
 >
 > If you want the cluster to stick around, then you should start the cluster locally, and pass `--disable-install` and `--keep-cluster` flags.
 
-
```
sehz, line 66: this doesn't seem belong in this PR

tests/runner/Cargo.toml
```diff
@@ -18,6 +18,8 @@ serde = "1.0"
 serde_json = "1.0"
 inventory = "0.1"
 tokio = { version = "1.4", features = ["macros"] }
+bencher = "0.1"
```
sehz, line 21: ditto

tests/runner/src/fluvio-integration-derive/src/lib.rs
```diff
@@ -40,40 +40,44 @@ pub fn fluvio_test(args: TokenStream, input: TokenStream) -> TokenStream {
             Err(_err) => panic!("Parse failed"),
```
sehz, line 40: remove this to benchmarking PR

sehz (changes requested):

.github/workflows/ci.yml
nicholastmosher: Could we add that to the `actions/` folder that is used across workflows? It seems like a good place for it

nicholastmosher:
@sehz this PR is a superset of the benchmarking PR, so a lot of changes from that PR will "appear" in here until the benchmarking PR is merged, then they will disappear from this PR's diff when TJ rebases

.github/workflows/ci.yml
tjtelan: https://github.com/infinyon/fluvio/issues/967

tjtelan:
> @sehz this PR is a superset of the benchmarking PR, so a lot of changes from that PR will "appear" in here until the benchmarking PR is merged, then they will disappear from this PR's diff when TJ rebases

I just removed the benchmarking stuff from this PR because it isn't actually dependent on those changes.

nicholastmosher (approved):
This looks great! Nice job :D

sehz (approved):
LGMT.  Thanks

PR #965, opened 2021-04-16
[Merged by Bors] - feat: Add Partitioner trait with SiphashRoundRobinPartitioner impl

### LINKED ISSUE #838: Implement initial Key Partitioning strategy for key/value records
We need to create strategies for deriving the partition a record should be sent to from the key of that record. The fundamental constraint is:

- Records with the same key must always be sent to the same partition

We would like to add a simple key-hashing, round-robin partitioning scheme as the first implementation of this feature. Some things to keep in mind while implementing:

- For first-round implementation, we should choose the quickest and easiest strategy. One hashing algorithm and round-robin will be sufficient
- We should ideally not make any API changes that we will need to propagate to language clients for this first round.
- We should also make sure that we leave room for ourselves to come back and add configuration for things later, e.g. choosing a different hashing algorithm or key partitioning strategy.

Some research notes:

- Kafka uses murmur2 hash for partitioning. However, murmur2 is not cryptographically secure and can lead to hash collisions, presenting availability vulnerabilities (e.g. records from untrusted sources could choose keys to overload one partition)
  - https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/producer/internals/DefaultPartitioner.java#L71
  - https://en.wikipedia.org/wiki/MurmurHash
- Most people nowadays are using siphash, which cryptographically secure but high-performance, used in many modern implementations of hash tables (e.g. Rust HashMap).
  - https://en.wikipedia.org/wiki/SipHash

---

Scope: implement key hashing with siphash-2-4

nicholastmosher:
Closes #838 

- Introduces `Partitioner` trait to enable new implementations in the future
- Adds `SiphashRoundRobinPartitioner` as a first implementor of `Partitioner`
  - Uses Siphash-2-4 on records that have keys
  - Uses round-robin to assign partitions to records with no keys
- Adds a `StoreContext<TopicSpec>` to `MetdataStores` for syncing topic metadata
- Implements watch requests on SPU for Topic and SPG metadata

### DIFF
--- a/examples/00-produce/src/main.rs
+++ b/examples/00-produce/src/main.rs
@@ -32,9 +32,9 @@ async fn main() {
 async fn produce() -> Result<(), fluvio::FluvioError> {
     let producer = fluvio::producer("simple").await?;
 
-    let record = "Hello, Fluvio!";
-    producer.send_record(&record, 0).await?;
-    println!("{}", &record);
+    let value = "Hello, Fluvio!";
+    producer.send_keyless(value).await?;
+    println!("{}", value);
 
     Ok(())
 }
--- a/examples/00-produce/src/main.rs
+++ b/examples/00-produce/src/main.rs
@@ -33,7 +33,7 @@ async fn produce() -> Result<(), fluvio::FluvioError> {
     let producer = fluvio::producer("simple").await?;
 
     let value = "Hello, Fluvio!";
-    producer.send_keyless(value).await?;
+    producer.send_record(value, 0).await?;
     println!("{}", value);
 
     Ok(())
--- a/examples/01-produce-key-value/src/main.rs
+++ b/examples/01-produce-key-value/src/main.rs
@@ -33,7 +33,7 @@ async fn main() {
 }
 
 async fn produce_key_value() -> Result<(), fluvio::FluvioError> {
-    let producer = fluvio::producer("twop").await?;
+    let producer = fluvio::producer("key-value").await?;
 
     let key = "Hello";
     let value = "Fluvio";
--- a/examples/01-produce-key-value/src/main.rs
+++ b/examples/01-produce-key-value/src/main.rs
@@ -23,17 +23,22 @@
 
 #[async_std::main]
 async fn main() {
+    let _ = tracing_subscriber::fmt()
+        .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
+        .try_init();
+
     if let Err(e) = produce_key_value().await {
         println!("Produce error: {:?}", e);
     }
 }
 
 async fn produce_key_value() -> Result<(), fluvio::FluvioError> {
-    let producer = fluvio::producer("key-value").await?;
+    let producer = fluvio::producer("twop").await?;
 
     let key = "Hello";
     let value = "Fluvio";
 
+    println!("About to send");
     producer.send(key, value).await?;
     println!("[{}] {}", key, value);
     Ok(())
--- a/examples/03-echo/src/main.rs
+++ b/examples/03-echo/src/main.rs
@@ -144,7 +144,7 @@ async fn produce() -> Result<(), FluvioError> {
             .send(format!("Key {}", i), format!("Value {}", i))
             .await?;
     }
-    producer.send_keyless("Done!").await?;
+    producer.send_record("Done!", 0).await?;
 
     Ok(())
 }
--- a/examples/03-echo/src/main.rs
+++ b/examples/03-echo/src/main.rs
@@ -138,13 +138,13 @@ async fn main() {
 async fn produce() -> Result<(), FluvioError> {
     let producer = fluvio::producer(TOPIC).await?;
 
-    for i in 0..10 {
+    for i in 0..10u32 {
         println!("Sending record {}", i);
         producer
             .send(format!("Key {}", i), format!("Value {}", i))
             .await?;
     }
-    producer.send_record("Done!", 0).await?;
+    producer.send_keyless("Done!").await?;
 
     Ok(())
 }
--- a/src/cli/src/consumer/produce/mod.rs
+++ b/src/cli/src/consumer/produce/mod.rs
@@ -102,7 +102,7 @@ impl ProduceOpt {
         if self.kv_mode() {
             self.produce_key_value(producer, string).await?;
         } else {
-            producer.send_record(string, 0).await?;
+            producer.send_keyless(string).await?;
             if self.verbose {
                 println!("[null] {}", string);
             }
--- a/src/cli/src/consumer/produce/mod.rs
+++ b/src/cli/src/consumer/produce/mod.rs
@@ -102,7 +102,7 @@ impl ProduceOpt {
         if self.kv_mode() {
             self.produce_key_value(producer, string).await?;
         } else {
-            producer.send_keyless(string).await?;
+            producer.send_record(string, 0).await?;
             if self.verbose {
                 println!("[null] {}", string);
             }
--- a/src/client/src/producer.rs
+++ b/src/client/src/producer.rs
@@ -57,6 +57,19 @@ impl TopicProducer {
         Ok(())
     }
 
+    /// Sends a plain record with no key to this producer's Topic.
+    ///
+    /// The partition that the record will be sent to will be chosen in a round-robin fashion.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use fluvio::{TopicProducer, FluvioError};
+    /// # async fn example(producer: &TopicProducer) -> Result<(), FluvioError> {
+    /// producer.send("Key", "Value").await?;
+    /// # Ok(())
+    /// # }
+    /// ```
     #[instrument(
         skip(self, value),
         fields(topic = %self.topic),
@@ -99,6 +112,7 @@ impl TopicProducer {
                     partition
                 }
             };
+            println!("Assigned partition: {}", partition);
 
             let key = key.map(|it| DefaultAsyncBuffer::new(it));
             let record = DefaultAsyncBuffer::new(value);
@@ -172,6 +186,8 @@ impl TopicProducer {
 fn partition_siphash(key: &[u8], partition_count: i32) -> i32 {
     use std::hash::{Hash, Hasher};
     use std::convert::TryFrom;
+
+    assert!(partition_count >= 0, "Partition must not be less than zero");
     let mut hasher = SipHasher::new();
     key.hash(&mut hasher);
     let hashed = hasher.finish();
@@ -182,11 +198,16 @@ fn partition_siphash(key: &[u8], partition_count: i32) -> i32 {
 
 #[cfg(test)]
 mod tests {
-    use super::*;
-
     #[test]
     fn test_u64_mod_i32() {
         assert_eq!(i32::MAX as u64, 2147483647_u64);
-        let some_u64 = 2u64 ^ 63;
+        let some_u64 = 2u64.pow(63);
+        let some_u64_mod_i32 = some_u64 % (i32::MAX as u64);
+        assert!(some_u64_mod_i32 < i32::MAX as u64);
+        let some_i32 = some_u64_mod_i32 as i32;
+        println!(
+            "u64: {}, u64_mod: {}, i32: {}",
+            some_u64, some_u64_mod_i32, some_i32
+        );
     }
 }
--- a/src/client/src/producer.rs
+++ b/src/client/src/producer.rs
@@ -122,29 +122,7 @@ impl TopicProducer {
         .await?;
 
         // Create one request per SPU leader
-        let mut requests: Vec<(SpuId, DefaultProduceRequest)> =
-            Vec::with_capacity(partitions_by_spu.len());
-        for (leader, partitions) in partitions_by_spu {
-            let mut request = DefaultProduceRequest::default();
-
-            let mut topic_request = DefaultTopicRequest::default();
-            topic_request.name = self.topic.clone();
-
-            for (partition, records) in partitions {
-                let mut partition_request = DefaultPartitionRequest::default();
-                partition_request.partition_index = partition;
-                partition_request
-                    .records
-                    .batches
-                    .push(DefaultBatch::new(records));
-                topic_request.partitions.push(partition_request);
-            }
-
-            request.acks = 1;
-            request.timeout_ms = 1500;
-            request.topics.push(topic_request);
-            requests.push((leader, request));
-        }
+        let requests = assemble_requests(&self.topic, partitions_by_spu);
 
         for (leader, request) in requests {
             let mut spu_client = self.pool.create_serial_socket_from_leader(leader).await?;
@@ -204,6 +182,38 @@ async fn group_by_spu(
     }
 
     Ok(map)
+}
+
+fn assemble_requests(
+    topic: &str,
+    partitions_by_spu: HashMap<i32, HashMap<i32, Vec<DefaultRecord>>>,
+) -> Vec<(SpuId, DefaultProduceRequest)> {
+    let mut requests: Vec<(SpuId, DefaultProduceRequest)> =
+        Vec::with_capacity(partitions_by_spu.len());
+
+    for (leader, partitions) in partitions_by_spu {
+        let mut request = DefaultProduceRequest::default();
+
+        let mut topic_request = DefaultTopicRequest::default();
+        topic_request.name = topic.to_string();
+
+        for (partition, records) in partitions {
+            let mut partition_request = DefaultPartitionRequest::default();
+            partition_request.partition_index = partition;
+            partition_request
+                .records
+                .batches
+                .push(DefaultBatch::new(records));
+            topic_request.partitions.push(partition_request);
+        }
+
+        request.acks = 1;
+        request.timeout_ms = 1500;
+        request.topics.push(topic_request);
+        requests.push((leader, request));
+    }
+
+    requests
 }
 
 /// A trait for defining a partitioning strategy for key/value records.
@@ -384,5 +394,96 @@ mod tests {
         );
 
         Ok(())
+    }
+
+    #[test]
+    fn test_assemble_requests() {
+        let partitions_by_spu = {
+            let mut pbs = HashMap::new();
+
+            let spu_0 = {
+                let mut s0_partitions = HashMap::new();
+                let partition_0_records = vec![DefaultRecord::new("A"), DefaultRecord::new("B")];
+                let partition_1_records = vec![DefaultRecord::new("C"), DefaultRecord::new("D")];
+                s0_partitions.insert(0, partition_0_records);
+                s0_partitions.insert(1, partition_1_records);
+                s0_partitions
+            };
+
+            let spu_1 = {
+                let mut s1_partitions = HashMap::new();
+                let partition_2_records = vec![DefaultRecord::new("E"), DefaultRecord::new("F")];
+                let partition_3_records = vec![DefaultRecord::new("G"), DefaultRecord::new("H")];
+                s1_partitions.insert(2, partition_2_records);
+                s1_partitions.insert(3, partition_3_records);
+                s1_partitions
+            };
+            pbs.insert(0, spu_0);
+            pbs.insert(1, spu_1);
+            pbs
+        };
+
+        let requests = assemble_requests("TOPIC", partitions_by_spu);
+        assert_eq!(requests.len(), 2);
+
+        // SPU 0
+        {
+            let (spu0, request) = requests.iter().find(|(spu, _)| *spu == 0).unwrap();
+            assert_eq!(*spu0, 0);
+            assert_eq!(request.topics.len(), 1);
+            let topic_request = request.topics.get(0).unwrap();
+            assert_eq!(topic_request.name, "TOPIC");
+            assert_eq!(topic_request.partitions.len(), 2);
+
+            let partition_0_request = topic_request.partitions.get(0).unwrap();
+            assert_eq!(partition_0_request.partition_index, 0);
+            assert_eq!(partition_0_request.records.batches.len(), 1);
+            let batch = partition_0_request.records.batches.get(0).unwrap();
+            assert_eq!(batch.records().len(), 2);
+            let record_0_0 = batch.records().get(0).unwrap();
+            assert_eq!(record_0_0.value.as_ref(), b"A");
+            let record_0_1 = batch.records().get(1).unwrap();
+            assert_eq!(record_0_1.value.as_ref(), b"B");
+
+            let partition_1_request = topic_request.partitions.get(1).unwrap();
+            assert_eq!(partition_1_request.partition_index, 1);
+            assert_eq!(partition_1_request.records.batches.len(), 1);
+            let batch = partition_1_request.records.batches.get(0).unwrap();
+            assert_eq!(batch.records().len(), 2);
+            let record_1_0 = batch.records().get(0).unwrap();
+            assert_eq!(record_1_0.value.as_ref(), b"C");
+            let record_1_1 = batch.records().get(1).unwrap();
+            assert_eq!(record_1_1.value.as_ref(), b"D");
+        }
+
+        // SPU 1
+        {
+            let (spu1, request) = requests.iter().find(|(spu, _)| *spu == 1).unwrap();
+            assert_eq!(*spu1, 1);
+            assert_eq!(request.topics.len(), 1);
+            let topic_request = request.topics.get(0).unwrap();
+            assert_eq!(topic_request.name, "TOPIC");
+            assert_eq!(topic_request.partitions.len(), 2);
+
+            let partition_0_request = topic_request.partitions.get(0).unwrap();
+            assert_eq!(partition_0_request.partition_index, 2);
+            assert_eq!(partition_0_request.records.batches.len(), 1);
+            let batch = partition_0_request.records.batches.get(0).unwrap();
+            assert_eq!(batch.records().len(), 2);
+            let record_0_0 = batch.records().get(0).unwrap();
+            assert_eq!(record_0_0.value.as_ref(), b"E");
+            let record_0_1 = batch.records().get(1).unwrap();
+            assert_eq!(record_0_1.value.as_ref(), b"F");
+
+            let partition_1_request = topic_request.partitions.get(1).unwrap();
+            assert_eq!(partition_1_request.partition_index, 3);
+            assert_eq!(partition_1_request.records.batches.len(), 1);
+            let batch = partition_1_request.records.batches.get(0).unwrap();
+            assert_eq!(batch.records().len(), 2);
+            let record_1_0 = batch.records().get(0).unwrap();
+            assert_eq!(record_1_0.value.as_ref(), b"G");
+            let record_1_1 = batch.records().get(1).unwrap();
+            assert_eq!(record_1_1.value.as_ref(), b"H");
+        }
     }
 }
--- a/src/client/src/producer.rs
+++ b/src/client/src/producer.rs
@@ -15,6 +15,7 @@ use dataplane::record::DefaultAsyncBuffer;
 use crate::FluvioError;
 use crate::spu::SpuPool;
 use crate::sockets::SerialFrame;
+use fluvio_types::SpuId;
 
 /// An interface for producing events to a particular topic
 ///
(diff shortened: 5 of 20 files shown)


sehz:
Can you link this to issue?

examples/01-produce-key-value/src/main.rs
```diff
@@ -23,17 +23,22 @@
...
+        .try_init();
+
     if let Err(e) = produce_key_value().await {
         println!("Produce error: {:?}", e);
     }
 }
 
 async fn produce_key_value() -> Result<(), fluvio::FluvioError> {
-    let producer = fluvio::producer("key-value").await?;
+    let producer = fluvio::producer("twop").await?;
```
simlay, line 36: `twop`?

nicholastmosher: Whoops! I changed that for testing out two partitions. I'll switch it back

src/client/src/producer.rs
```diff
@@ -96,82 +180,131 @@ impl TopicProducer {
...
+/// partitions in the current Topic. It must map each key from the input
+/// slice into a partition stored at the same index in the output Vec.
+///
+/// It is up to the implementor to decide how the keys get mapped to
+/// partitions. This includes deciding what partition to assign to records
+/// with no keys (represented by `None` values in the keys slice).
+///
+/// See [`SiphashRoundRobinPartitioner`] for a reference implementation.
+trait Partitioner {
+    fn partition(&self, keys: &[Option<&[u8]>], partition_count: i32) -> Vec<i32>;
```
sehz, line 207: Is partitioner stateless or stateful?

src/client/src/producer.rs
```diff
@@ -96,82 +180,131 @@ impl TopicProducer {
...
+    use super::*;
+
+    #[test]
+    fn test_u64_mod_i32() {
+        assert_eq!(i32::MAX as u64, 2147483647_u64);
+        let some_u64 = 2u64.pow(63);
+        let some_u64_mod_i32 = some_u64 % (i32::MAX as u64);
+        assert!(some_u64_mod_i32 < i32::MAX as u64);
+        let some_i32 = some_u64_mod_i32 as i32;
+        println!(
```
sehz, line 269: prefer not to have println in test code.  if need to diagnose internal state, put as debug or should do another assert

src/client/src/producer.rs
```diff
@@ -96,82 +180,131 @@ impl TopicProducer {
...
+            "u64: {}, u64_mod: {}, i32: {}",
+            some_u64, some_u64_mod_i32, some_i32
+        );
+    }
+
+    /// Ensure that feeding keyless records one-at-a-time does not assign the same partition
+    #[test]
+    fn test_round_robin_individual() {
```
sehz, line 277: can you change this test to match what is intended to be? like `test_round_robin_unique_partition`

src/client/src/producer.rs
```diff
@@ -96,82 +180,131 @@ impl TopicProducer {
...
+        let partitioner = SiphashRoundRobinPartitioner::new();
+        let partition_count = 10;
+
+        let mut prev_partition = -1;
+        for _ in 0..100 {
+            let p = partitioner.partition(&[None], partition_count)[0];
+            println!("Partition: {}", p);
+            assert_ne!(p, prev_partition);
```
sehz, line 285: not clear behavior of this test. it might be easier if you test individual state

src/client/src/producer.rs
```diff
@@ -96,82 +180,131 @@ impl TopicProducer {
...
+        let partitioner = SiphashRoundRobinPartitioner::new();
+        let partition_count = 11;
+
+        // A batch of 10 records with no keys
+        let batch: Vec<Option<&[u8]>> = (0..10).map(|_| None).collect();
+
+        let mut prev_partition = -1;
+        for _ in 0..10 {
+            let ps = partitioner.partition(&batch, partition_count);
+
```
sehz, line 302: again, test behavior are hard to understand

src/client/src/sync/controller.rs
```diff
@@ -97,7 +95,7 @@ where
                 break;
             }
 
-            select! {
+            tokio::select! {
```
sehz, line 98: not fan of adding prefix

src/client/src/sync/controller.rs
```diff
@@ -111,21 +109,20 @@ where
                             match update_result {
                                 Ok(update) => {
                                     if let Err(err) = self.process_updates(update).await {
-                                        error!("processing updates: {}", err);
+                                        error!("Error processing updates: {}", err);
```
sehz, line 112: this is already error, so marking as error seems to be redundant.  However, welcome making error clear...

src/client/src/sync/controller.rs
```diff
@@ -111,21 +109,20 @@ where
...
                                 },
                                 Err(err) => {
-                                    error!("Error decoding metadata {} update response: {}",S::LABEL,err);
+                                    error!("Error decoding metadata update response, skipping: {}", err);
                                 }
                             }
                         },
                         Some(Err(err)) => {
-                            error!("{} error receiving, end, {}", S::LABEL, err);
+                            error!("Error receiving response, ending: {}", err);
```
sehz, line 121: ditto

src/client/src/sync/controller.rs
```diff
@@ -137,55 +134,33 @@ where
...
-                .map(|change| {
-
-                    match change.header {
-
-                        MsgType::UPDATE => LSUpdate::Mod(change.content),
-                        MsgType::DELETE => LSUpdate::Delete(change.content.key_owned()),
-
-                    }
-                }).collect();
+        debug!(count = updates.all.len(), "Syncing updates:");
```
sehz, line 147: can you comment on what changes happen here?

src/client/src/sync/mod.rs
```diff
@@ -1,164 +1,159 @@
...
+
+use crate::FluvioError;
+use crate::metadata::core::Spec;
+use crate::metadata::store::LocalStore;
+use crate::metadata::store::DualEpochMap;
+use crate::metadata::store::MetadataStoreObject;
+use crate::metadata::spu::SpuSpec;
+use crate::metadata::core::MetadataItem;
+
+pub(crate) type CacheMetadataStoreObject<S> = MetadataStoreObject<S, AlwaysNewContext>;
```
sehz, line 21: Can you minimize change here so it's to easier to see diff? If more make it cleaner, can you do that in separate PR?

sehz (changes requested):

src/client/src/producer.rs
nicholastmosher: The partitioner should have the opportunity to be stateful. In the case of `SiphashRoundRobinPartitioner`, it keeps track of the partition index which should be used next in the round-robin selection.

src/client/src/producer.rs
nicholastmosher: Sure thing, I'll clean this up and add some assertions that are clearer

src/client/src/producer.rs
sehz: Why is partition count need to be passed every time?

src/client/src/producer.rs
```diff
@@ -48,7 +59,32 @@ impl TopicProducer {
...
+    /// # async fn example(producer: &TopicProducer) -> Result<(), FluvioError> {
+    /// producer.send("Key", "Value").await?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[instrument(
+        skip(self, value),
+        fields(topic = %self.topic),
+    )]
+    pub async fn send_keyless<V>(&self, value: V) -> Result<(), FluvioError>
```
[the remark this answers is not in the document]
nicholastmosher: This is meant for the use case of: I want to send one record with no key. Previously, we would use `send_record`, but the problem with that is that it asked for an explicit partition number, which we no longer want to do. This is essentially the same function as `send_record` but minus the `partition` argument.

src/client/src/producer.rs
```diff
@@ -62,21 +98,69 @@ impl TopicProducer {
...
+            let mut partition_request = DefaultPartitionRequest::default();
+
+            partition_request.partition_index = partition;
+            partition_request
+                .records
+                .batches
+                .push(DefaultBatch::new(records));
+            topic_request.name = self.topic.clone();
+            topic_request.partitions.push(partition_request);
+            request.acks = 1;
```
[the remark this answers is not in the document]
nicholastmosher: This is how the producer request was already being built (see the removed diff in `send_records_raw`. If we want to change this we should do it in a new issue

src/client/src/producer.rs
nicholastmosher: That is a good point, we could add the `partition_count` as a piece of state in the implementor of `Partitioner`. I think one relevant question is: Is it possible for the partition_count to change, and if so, how often is that likely to ever happen? If it is very likely to happen frequently, it might make sense to keep it as an argument. If it is unlikely to happen, we may want to add an auxiliary function to the trait, something like `fn resize_partitions(&self, partition_count: i32)`. I think using an argument is just simpler though, but let me know what you think.

src/client/src/sync/mod.rs
nicholastmosher: I remember removing an inline module, I'll go back and undo that to reduce noise in the diff.

src/client/src/sync/controller.rs
nicholastmosher: Sure. Previously, there was a really big if-statement like

```rust
if !updates.all.is_empty() {
    // ...
}
```

which was causing a lot of rightward drift. So all I did was remove that if-statement and shift everything inside back to the left, and at the top I added the opposite if-statement to short-circuit the function:

```rust
if updates.all.is_empty() {
    // ...
    return Ok(());
}
```

So it just made the code easier to read while I was trying to figure out what was going wrong.

src/client/src/producer.rs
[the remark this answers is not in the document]
sehz: let's not add more API unless it is absolutely necessary.

src/client/src/producer.rs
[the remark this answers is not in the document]
sehz: Got it.  Let's create issue to allow developers to configure producer and profile. We probably need producer builder or something like that.

src/client/src/producer.rs
sehz: We should make it general.  Kafka partitioning uses combination of key, values and metadata.  We need metadata in case we support rack assignment

src/client/src/producer.rs
```diff
@@ -96,82 +180,128 @@ impl TopicProducer {
...
+    fn partition(&self, keys: &[Option<&[u8]>], partition_count: i32) -> Vec<i32>;
+}
 
-        debug!("connect to replica leader at: {}", spu_client);
+/// A [`Partitioner`] which combines hashing and round-robin partition assignment
+///
+/// - Records with keys get their keys hashed with siphash
+/// - Records without keys get assigned to partitions using round-robin
+struct SiphashRoundRobinPartitioner {
+    index: Arc<RwLock<i32>>,
```
sehz, line 215: why this need to be in the lock?  A producer doesn't need to be thread safe nor shared

sehz (changes requested):

src/client/src/producer.rs
nicholastmosher: I see. In that case, I think what we should do is make the trait start out like this:

```rust
trait Partitioner {
    fn partition(&self, keys: &[Option<&[u8]>]) -> i32; // Note removal of partition_count
}
```

Then, what we can do is leave it up to the individual Partitioner implementations to require the particular type of configuration they need in order to make decisions about how to go about partitioning. So for our initial partitioner, we would have:

```rust
struct SiphashRoundRobinPartitioner {
    partition_count: i32,
    // ...
}

impl SiphashRoundRobinPartitioner  {
    pub fn new(partition_count: i32) -> Self { ... }
}
```

And later on we could update that constructor to take some more general type, e.g. something like `ClusterConfig` that may have more info about the topics and partitions.

Additionally, in the future we could add a new function to the `Partitioner` trait that allows us to send configuration updates to the implementors:

```rust
trait Partitioner<T=()> {
    fn partition(...) { ... }
    fn update_config(&mut self, config: T);
}
```

src/client/src/producer.rs
nicholastmosher: > A producer doesn't need to be thread safe nor shared

Well, right now all of our producer's methods take self by immutable reference `&self`:

- `fn send<K, V>(&self, ...)`
- `fn send_keyless<V>(&self, ...)`
- `fn send_all<K, V>(&self, ...)`

And `TopicProducer` actually does implement Send and Sync. So in order to have mutable state we need to put the interior mutability somewhere. I am actually reworking some of this based on your earlier comment, let me ping you back on this when I have tweaked it a bit.

nicholastmosher:
Ok, newest changes:

- Updated `Partitioner` trait to look like the following:

```rust
trait Partitioner {
    fn partition(&mut self, keys: &[Option<&[u8]>]) -> Vec<i32>;
    fn update_config(&mut self, config: PartitionerConfig);
}
```

- Added `PartitionerConfig` to hold information such as `partition_count` in a way that lets us expand in the future
- Moved interior mutability out of `Partitioner` itself and up into the `TopicProducer` level

src/client/src/producer.rs
sehz: Why does this need to take array of keys? Partitioning only needed for per record. So it could be like 
```
trait Partitioner {
    fn partition(&self, key: Option<Bytes>,value: Bytes) -> i32; 
    fn update_config(&mut self, config: T
}
```

src/client/src/producer.rs
```diff
@@ -62,21 +100,76 @@ impl TopicProducer {
...
+            request.acks = 1;
+            request.timeout_ms = 1500;
+            request.topics.push(topic_request);
+
+            requests.push((partition, request));
+        }
+
+        for (partition, request) in requests {
+            let replica = ReplicaKey::new(&self.topic, partition);
+            let mut spu_client = self.pool.create_serial_socket(&replica).await?;
```
sehz, line 169: This seems inefficient, we are creating new client event thought they are same.  Should cache them per SPU given partition

src/client/src/producer.rs
nicholastmosher: Right. The reason is because I had originally designed the Partitioner to use interior mutability, i.e. if it needed to keep track of something like a round-robin index, it would keep it in an `Arc<Mutex<i32>>` or something. I did not want to need to lock and unlock a mutex for every key, so I made the API accept a batch of keys at once.

However, now that we're doing mutability at a higher level, I can change this back to one key at a time 👍

src/client/src/producer.rs
nicholastmosher: Yes, I was thinking about that as well. In order to make that happen, I'll essentially need a map of `SpuId -> [PartitionId]`, so that I could create one request per SPU rather than per partition. I will look today to see how to put that together.

nicholastmosher: Actually, I want a mapping of `PartitionId -> LeaderId`.

I am looking through the `TopicSpec` metadata to try to figure out how to get this mapping, and I have a few questions. 

- I see we can either have an `Assigned` or `Computed` topic, and it seems like the `PartitionMaps` in the  Assigned topic might give me the information that I need
  - I see the `PartitionMaps` is essentially a `[PartitionMap { id: PartitionId, replicas: [ SpuId ] } ]`
  - Does the index into `PartitionMaps` represent the leader ID? It doesn't seem that this would make sense, because then there could only be one partition assigned to a leader.
  - Does the index into `PartitionMaps` actually represent the Partition ID, and the thing called `id: PartitionId` actually represent the leader's SPU ID? I think that would make more sense because then multiple partitions could be assigned to the same leader
- How do we handle Computed topics?

@sehz let me know if I'm on the right track or if I'm way off

src/client/src/producer.rs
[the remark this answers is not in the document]
nicholastmosher: Ok, so if we do not include this `send_keyless` API, then the only way to send a single record with no key is to do:

```rust
producer.send_all(Some((None, "Hello!"))).await?;
```

which I think is much more difficult to write and certainly not at all intuitive if you are a user searching for that functionality. If we do include `send_keyless`, users can use it to do the same thing:

```rust
producer.send_keyless("Hello!").await?;
```

So in this case I would actually argue that a function like `send_keyless` is actually very necessary. I am open to suggestions if you have any other preferences for the name of it though.

src/client/src/producer.rs
sehz: Topics PartitionMap represents allocations. It tells which replica (unit of storage) is assigned to SPU. It doesn't tell you about leader.  That is tracked in the `Partition` spec.

sehz: so topic list tells you where replicas are:
```
flvd topic list
 NAME   TYPE      PARTITIONS  REPLICAS  IGNORE-RACK  STATUS                   REASON 
 test   computed      1          1                   resolution::provisioned   
 test2  computed      1          2                   resolution::provisioned   
```

but partition tells you which are leader:
```
flvd partition list
 TOPIC  PARTITION  LEADER  REPLICAS  RESOLUTION  HW    LEO   LSR  FOLLOWER OFFSETS 
 test2  0          5001    [5002]    Online      1     1     0    [ReplicaStatus { spu: 5002, hw: 0, leo: 0 }] 
 test   0          5001    []        Online      1000  1000  0    [] 
```

src/client/src/producer.rs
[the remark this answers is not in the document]
sehz: Hmm. we still have `send` right? so it can be:
```
producer.send(None,"Hello!").await?;
```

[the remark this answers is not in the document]
nicholastmosher: You cannot pass `None` to send. This is the API for `send`:

```rust
    pub async fn send<K, V>(&self, key: K, value: V) -> Result<(), FluvioError>
    where
        K: Into<Vec<u8>>,
        V: Into<Vec<u8>>,
    { ... }
```

So you must pass both a key and a value as arguments to send, you cannot use None. I have also tried other designs for this API but there does not seem to be a good solution for an optional key. This is the reason we need a separate API like `send_keyless`.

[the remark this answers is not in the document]
sehz: Let's move creating new or updating existing API as another issue which doesn't belong here.

For handling none,  I suggest we use custom wrapper struct rather than using `Vec<u8>` directory.  This way we have full control of conversion.

src/client/src/producer.rs
```diff
@@ -96,82 +200,103 @@ impl TopicProducer {
...
+/// partitions in the current Topic. It must map each key from the input
+/// slice into a partition stored at the same index in the output Vec.
+///
+/// It is up to the implementor to decide how the keys get mapped to
+/// partitions. This includes deciding what partition to assign to records
+/// with no keys (represented by `None` values in the keys slice).
+///
+/// See [`SiphashRoundRobinPartitioner`] for a reference implementation.
+trait Partitioner {
+    fn partition(&mut self, key: Option<&[u8]>) -> i32;
```
sehz, line 227: See previous comment, this should take on value as well

src/client/src/producer.rs
```diff
@@ -62,21 +101,86 @@ impl TopicProducer {
...
+                .spec;
+            let leader = partition_spec.leader;
+            let key = key.map(|it| DefaultAsyncBuffer::new(it));
+            let record = DefaultAsyncBuffer::new(value);
+            let record = DefaultRecord::from((key, record));
+            map.entry(leader)
+                .or_insert_with(HashMap::new)
+                .entry(partition)
+                .or_insert_with(Vec::new)
+                .push(record);
```
sehz, line 152: it is important to add unit test for this

sehz (changes requested):
Good progress.  See other comments regarding moving keyless api as separate issue

sehz:
Also, test with replication factor = 2

src/client/src/producer.rs
nicholastmosher: - Which previous comment are you referring to?
- Why should the partitioner accept the value for the record? What is an example use-case that you are imagining?

sehz: It was in previous comment:  https://kafka.apache.org/24/javadoc/?org/apache/kafka/clients/producer/Partitioner.html.  Kafka produce takes on value which make sense.   There is no need to limit to key only.

nicholastmosher: Oh interesting, it looks like they have partitioning strategies that just ignore the key and look at the value contents to find a defined partition https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/producer/UniformStickyPartitioner.java#L25-L36

src/client/src/producer.rs
nicholastmosher: Added! I factored out the `group_by_spu` function and wrote a pretty good test for it, let me know what you think

sehz:
Can you add this to changelog?

examples/00-produce/src/main.rs
```diff
@@ -32,9 +32,9 @@ async fn main() {
 async fn produce() -> Result<(), fluvio::FluvioError> {
     let producer = fluvio::producer("simple").await?;
 
-    let record = "Hello, Fluvio!";
-    producer.send_record(&record, 0).await?;
-    println!("{}", &record);
+    let value = "Hello, Fluvio!";
+    producer.send_keyless(value).await?;
```
sehz, line 36: `keyless` api is still in this PR.  Please move them to separate PR

examples/03-echo/src/main.rs
```diff
@@ -138,13 +138,13 @@ async fn main() {
...
 
-    for i in 0..10 {
+    for i in 0..10u32 {
         println!("Sending record {}", i);
         producer
             .send(format!("Key {}", i), format!("Value {}", i))
             .await?;
     }
-    producer.send_record("Done!", 0).await?;
+    producer.send_keyless("Done!").await?;
```
sehz, line 147: ditto

src/cli/src/consumer/produce/mod.rs
```diff
@@ -102,7 +102,7 @@ impl ProduceOpt {
         if self.kv_mode() {
             self.produce_key_value(producer, string).await?;
         } else {
-            producer.send_record(string, 0).await?;
+            producer.send_keyless(string).await?;
```
sehz, line 105: ditto

src/client/src/producer.rs
```diff
@@ -96,82 +195,219 @@ impl TopicProducer {
...
+        let buffer: Vec<u8> = Vec::from(buffer.as_ref());
+        self.send_keyless(buffer).await?;
+        Ok(())
+    }
+}
 
-        let spu_client = self.pool.create_serial_socket(&replica).await?;
+async fn group_by_spu(
+    topic: &str,
+    partitions: &StoreContext<PartitionSpec>,
```
sehz, line 212: doesn't need dependencies to `StoreContext`, it just need `LocalStore` which assume cache are populated

src/client/src/producer.rs
```diff
@@ -62,21 +103,79 @@ impl TopicProducer {
...
+            topic_request.name = self.topic.clone();
+
+            for (partition, records) in partitions {
+                let mut partition_request = DefaultPartitionRequest::default();
+                partition_request.partition_index = partition;
+                partition_request
+                    .records
+                    .batches
+                    .push(DefaultBatch::new(records));
+                topic_request.partitions.push(partition_request);
```
sehz, line 165: unit test?

src/client/src/producer.rs
```diff
@@ -96,82 +195,219 @@ impl TopicProducer {
...
+
+        assert!(
+            partition_2
+                .iter()
+                .any(|record| record.value.as_ref() == b"F"),
+            "SPU 1/Partition 2 should contain record F",
+        );
+
+        Ok(())
     }
```
sehz, line 412: need test with replication

tests/runner/src/tests/smoke/produce.rs
```diff
@@ -131,7 +131,7 @@ pub async fn produce_message_with_api(
             let message = generate_message(offset, &test_case);
             let len = message.len();
             info!("trying send: {}, iteration: {}", topic_name, i);
-            producer.send_record(message, 0).await.unwrap_or_else(|_| {
+            producer.send_keyless(message).await.unwrap_or_else(|_| {
```
sehz, line 134: ditto

tests/runner/src/tests/producer_stress.rs
```diff
@@ -82,7 +82,7 @@ pub async fn run(client: Arc<Fluvio>, mut test_case: TestCase) -> TestResult {
         for (i, p) in producers.iter().enumerate() {
             let message = format!("producer-{} line-{} {}", i, n, long_str.clone());
 
-            p.send_record(message.clone(), 0)
+            p.send_keyless(message)
```
sehz, line 85: ditto

tests/runner/src/tests/concurrent/producer.rs
```diff
@@ -18,6 +18,6 @@ pub async fn producer(fluvio: Arc<Fluvio>, option: ConcurrentTestCase, digests:
         let record = rand_record();
         let record_digest = hash_record(&record);
         digests.send(record_digest).unwrap();
-        producer.send_record(&record, PARTITION).await.unwrap();
+        producer.send_keyless(record).await.unwrap();
```
sehz, line 21: ditto

sehz (changes requested):
Making good progress.   Couple of minor changes needed

src/client/src/producer.rs
nicholastmosher: Factored out `assemble_requests` method and wrote a test for it

src/client/src/producer.rs
nicholastmosher: I think replication tests are more fitting as an integration test or e2e test. Let's make another issue for it

src/client/src/producer.rs
nicholastmosher: The StoreContext is the type with the `lookup_by_key` method which I'm using to implement this. It is not obvious to me how to go about this with LocalStore, and what is written here is working well

sehz: Actually, this may be good.  If live SPU is not found, it can wait.  Ok. let's leave as is

src/client/src/producer.rs
```diff
@@ -96,82 +148,354 @@ impl TopicProducer {
...
+    }
 
-        let spu_client = self.pool.create_serial_socket(&replica).await?;
+    Ok(map)
+}
 
-        debug!("connect to replica leader at: {}", spu_client);
+fn assemble_requests(
+    topic: &str,
+    partitions_by_spu: HashMap<i32, HashMap<i32, Vec<DefaultRecord>>>,
```
sehz, line 189: Change `i32` to `SpuId`

src/client/src/producer.rs
```diff
@@ -96,82 +148,354 @@ impl TopicProducer {
...
+/// partitions in the current Topic. It must map each key from the input
+/// slice into a partition stored at the same index in the output Vec.
+///
+/// It is up to the implementor to decide how the keys get mapped to
+/// partitions. This includes deciding what partition to assign to records
+/// with no keys (represented by `None` values in the keys slice).
+///
+/// See [`SiphashRoundRobinPartitioner`] for a reference implementation.
+trait Partitioner {
+    fn partition(&mut self, key: Option<&[u8]>, value: &[u8]) -> i32;
```
sehz, line 231: change i32 to SpuId

src/client/src/producer.rs
```diff
@@ -96,82 +148,354 @@ impl TopicProducer {
...
+struct PartitionerConfig {
+    partition_count: i32,
+}
+
+/// A [`Partitioner`] which combines hashing and round-robin partition assignment
+///
+/// - Records with keys get their keys hashed with siphash
+/// - Records without keys get assigned to partitions using round-robin
+struct SiphashRoundRobinPartitioner {
+    index: i32,
```
sehz, line 244: change to `SpuId`

sehz (changes requested):

src/client/src/producer.rs
```diff
@@ -62,21 +78,57 @@ impl TopicProducer {
...
+            });
+
+        // Calculate the partition for each entry
+        // Use a block scope to ensure we drop the partitioner lock
+        let records_by_partition = {
+            let mut iter = Vec::new();
+            let mut partitioner = self.partitioner.lock().await;
+            partitioner.update_config(partition_config);
+
+            for record in entries {
```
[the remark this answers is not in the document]
nicholastmosher: I tried but I actually cannot. The body of this loop uses the `partitioner`, so if that body is moved into a closure, it then borrows the `partitioner` for longer than allowed (because iterators are lazy and using `.map` simply saves the closure object)

src/client/src/producer.rs
nicholastmosher: I don't think this is a `SpuId`, this is the Partition. Is there a `PartitionId` type somewhere?

src/client/src/producer.rs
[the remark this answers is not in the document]
sehz: k. fair enough

src/client/src/producer.rs
nicholastmosher: There is `fluvio_types::PartitionId` which is `i32`, I can use that

src/client/src/producer.rs
nicholastmosher: Changed to PartitionId ;)

sehz (approved):
LGTM.  Nice work!

PR #985, opened 2021-04-21
[Merged by Bors] - Add RecordKey API for sending records without keys

### LINKED ISSUE #977: Add Producer API for sending keyless records
After #965 lands, we will have deprecated the `TopicProducer::send_record` API, which allowed users to send a record without specifying a key for it. This API was deprecated because send_record accepted a partition number as one of the arguments, and we decided that for now, users should not be able to manually specify partitions.

This leaves a gap in our API for a particular use-case, which is sending a single record with no key. Without `send_record`, the only way to send a record with no key involves using the `send_all` function, which forces users to wrap their single record in some kind of iterator before passing it to the producer, e.g. like this:

```rust
producer.send_all(Some( (None, "My record data with no key") )).await?;
```

I believe that this is a workaround that we should not force users to deal with, as it is non-obvious and non-ergonomic. I think we need to introduce a new API method for the "single record with no key" use-case. I originally proposed a `send_keyless` API in #965 (discussion at https://github.com/infinyon/fluvio/pull/965#discussion_r615967714), but we decided to exclude it from that PR in order to discuss it more, which is what this issue is for.

For reference, the `send_keyless` API that I would like to introduce looks like this:

```rust
impl TopicProducer {
    pub async fn send_keyless<V>(&self, value: V) -> Result<(), FluvioError>
    where
        V: Into<Vec<u8>> { ... }
}
```

So the questions I would like to resolve in this discussion are:

1) Do we believe that a new API to fill this use-case is necessary? (I believe it is),
2) Is there a different API that better serves this use-case (I have not thought of one)?
3) Are we happy with the name `send_keyless` or would we prefer something else?

I think it makes sense to put a timeline on this. If nobody has any strong reasons why we should NOT introduce this API, I would like to add this issue to our roadmap as an action item by this Wednesday (4/21/21) to give enough time to ship it with the `0.7.4` release this Thursday.

---

# Discussion Results

After debating the API, we've decided to implement the following:

```rust
pub struct RecordKey(RecordKeyInner);

impl RecordKey {
    pub const NULL: Self = Self(RecordKeyInner::Null);
}

enum RecordKeyInner {
    Null,
    Key(RecordData),
}

impl<K: Into<Vec<u8>>> From<K> for RecordKey {
    fn from(k: K) -> Self {
        Self(RecordKeyInner::Key(RecordData::from(k)))
    }
}

/// A type to hold the contents of a record's value.
pub struct RecordData(bytes::Bytes);

impl<V: Into<Vec<u8>>> From<V> for RecordData {
    fn from(v: V) -> Self {
        let value: Vec<u8> = v.into();
        Self(value.into())
    }
}

fn send<K, V>(key: K, value: V)
    where
        K: Into<RecordKey>,
        V: Into<RecordData>,
{ ... }

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn test_conversion() {
        send(RecordKey::NULL, vec![2]);
        send(vec![3],vec![4]);
        send("Hello", "world");
        send("Hello".to_string(), "World".to_string());
    }
}
```

nicholastmosher:
Closes #977.

- Adds `RecordKey` API to `TopicProducer`
- Updates CLI and examples to use `send` with `RecordKey::Null`

### DIFF
--- a/src/client/src/producer.rs
+++ b/src/client/src/producer.rs
@@ -62,7 +62,7 @@ impl TopicProducer {
     pub async fn send<K, V>(&self, key: K, value: V) -> Result<(), FluvioError>
     where
         K: Into<RecordKey>,
-        V: Into<RecordValue>,
+        V: Into<RecordData>,
     {
         let record_key = key.into();
         let record_value = value.into();
@@ -77,7 +77,7 @@ impl TopicProducer {
     pub async fn send_all<K, V, I>(&self, records: I) -> Result<(), FluvioError>
     where
         K: Into<RecordKey>,
-        V: Into<RecordValue>,
+        V: Into<RecordData>,
         I: IntoIterator<Item = (K, V)>,
     {
         let topics = self.pool.metadata.topics();
@@ -91,11 +91,11 @@ impl TopicProducer {
 
         let entries = records
             .into_iter()
-            .map::<(RecordKey, RecordValue), _>(|(k, v)| (k.into(), v.into()))
+            .map::<(RecordKey, RecordData), _>(|(k, v)| (k.into(), v.into()))
             .map(|(key, value)| {
                 let key = match key.0 {
                     RecordKeyInner::Null => None,
-                    RecordKeyInner::Key(key) => Some(DefaultAsyncBuffer::new(key)),
+                    RecordKeyInner::Key(key) => Some(DefaultAsyncBuffer::new(key.0)),
                 };
                 let value = DefaultAsyncBuffer::new(value.0);
                 DefaultRecord::from((key, value))
@@ -245,20 +245,19 @@ impl RecordKey {
 
 enum RecordKeyInner {
     Null,
-    Key(bytes::Bytes),
+    Key(RecordData),
 }
 
 impl<K: Into<Vec<u8>>> From<K> for RecordKey {
     fn from(k: K) -> Self {
-        let key: Vec<u8> = k.into();
-        Self(RecordKeyInner::Key(key.into()))
+        Self(RecordKeyInner::Key(RecordData::from(k)))
     }
 }
 
 /// A type to hold the contents of a record's value.
-pub struct RecordValue(bytes::Bytes);
+pub struct RecordData(bytes::Bytes);
 
-impl<V: Into<Vec<u8>>> From<V> for RecordValue {
+impl<V: Into<Vec<u8>>> From<V> for RecordData {
     fn from(v: V) -> Self {
         let value: Vec<u8> = v.into();
         Self(value.into())
--- a/src/client/src/producer.rs
+++ b/src/client/src/producer.rs
@@ -61,10 +61,19 @@ impl TopicProducer {
     )]
     pub async fn send<K, V>(&self, key: K, value: V) -> Result<(), FluvioError>
     where
-        K: Into<Vec<u8>>,
-        V: Into<Vec<u8>>,
+        K: Into<RecordKey>,
+        V: Into<RecordValue>,
     {
-        self.send_all(Some((Some(key.into()), value))).await?;
+        let record_key = key.into();
+        let record_value = value.into();
+
+        let maybe_key = match record_key {
+            RecordKey::Null => None,
+            RecordKey::Key(key) => Some(key),
+        };
+        let value = record_value.0;
+
+        self.send_all(Some((maybe_key, value))).await?;
         Ok(())
     }
 
@@ -214,6 +223,44 @@ fn assemble_requests(
     }
 
     requests
+}
+
+/// A key for determining which partition a record should be sent to.
+///
+/// This type is used to support conversions from any other type that
+/// may be converted to a `Vec<u8>`, while still allowing the ability
+/// to explicitly state that a record may have no key (`RecordKey::Null`).
+///
+/// # Examples
+///
+/// ```
+/// # use fluvio::{TopicProducer, FluvioError, RecordKey};
+/// # async fn example(producer: &TopicProducer) -> Result<(), FluvioError> {
+/// producer.send("Hello", String::from("World!")).await?;
+/// producer.send(RecordKey::Null, "World!").await?;
+/// # Ok(())
+/// # }
+/// ```
+pub enum RecordKey {
+    Null,
+    Key(Vec<u8>),
+}
+
+impl<K: Into<Vec<u8>>> From<K> for RecordKey {
+    fn from(k: K) -> Self {
+        let key: Vec<u8> = k.into();
+        Self::Key(key)
+    }
+}
+
+/// A type to hold the contents of a record's value.
+pub struct RecordValue(Vec<u8>);
+
+impl<V: Into<Vec<u8>>> From<V> for RecordValue {
+    fn from(v: V) -> Self {
+        let value: Vec<u8> = v.into();
+        Self(value)
+    }
 }
 
 /// A trait for defining a partitioning strategy for key/value records.
--- a/src/client/src/producer.rs
+++ b/src/client/src/producer.rs
@@ -66,14 +66,7 @@ impl TopicProducer {
     {
         let record_key = key.into();
         let record_value = value.into();
-
-        let maybe_key = match record_key {
-            RecordKey::Null => None,
-            RecordKey::Key(key) => Some(key),
-        };
-        let value = record_value.0;
-
-        self.send_all(Some((maybe_key, value))).await?;
+        self.send_all(Some((record_key, record_value))).await?;
         Ok(())
     }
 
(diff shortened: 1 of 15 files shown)


src/client/src/producer.rs
```diff
@@ -61,10 +61,19 @@ impl TopicProducer {
     )]
     pub async fn send<K, V>(&self, key: K, value: V) -> Result<(), FluvioError>
     where
-        K: Into<Vec<u8>>,
-        V: Into<Vec<u8>>,
+        K: Into<RecordKey>,
+        V: Into<RecordValue>,
     {
-        self.send_all(Some((Some(key.into()), value))).await?;
+        let record_key = key.into();
```
sehz, line 67: Instead of just modifying this.  We should update `send_all` to use `RecordKey` and `RecordValue`.

src/client/src/producer.rs
```diff
@@ -216,6 +225,44 @@ fn assemble_requests(
...
+/// # use fluvio::{TopicProducer, FluvioError, RecordKey};
+/// # async fn example(producer: &TopicProducer) -> Result<(), FluvioError> {
+/// producer.send("Hello", String::from("World!")).await?;
+/// producer.send(RecordKey::Null, "World!").await?;
+/// # Ok(())
+/// # }
+/// ```
+pub enum RecordKey {
+    Null,
+    Key(Vec<u8>),
```
sehz, line 246: This should be `Bytes` to make it easier to translate to Record Type

src/client/src/producer.rs
```diff
@@ -216,6 +225,44 @@ fn assemble_requests(
...
+
+impl<K: Into<Vec<u8>>> From<K> for RecordKey {
+    fn from(k: K) -> Self {
+        let key: Vec<u8> = k.into();
+        Self::Key(key)
+    }
+}
+
+/// A type to hold the contents of a record's value.
+pub struct RecordValue(Vec<u8>);
```
sehz, line 257: Same here

sehz (changes requested):

src/client/src/producer.rs
nicholastmosher: Hmm I don't think we can actually do this. If you look at `send_all`:

```rust
    pub async fn send_all<K, V, I>(&self, records: I) -> Result<(), FluvioError>
    where
        K: Into<Vec<u8>>,
        V: Into<Vec<u8>>,
        I: IntoIterator<Item = (Option<K>, V)>,
    { ... }
```

Specifically this part: `I: IntoIterator<Item = (Option<K>, V)>`

This means that `send_all` is using Option::None to represent "no key" already. If we were to try to change it like this:

```rust
    pub async fn send_all<K, V, I>(&self, records: I) -> Result<(), FluvioError>
    where
        K: Into<RecordKey>,
        V: Into<RecordValue>,
        I: IntoIterator<Item = (Option<K>, V)>,
    { ... }
```

Then we would have an odd set of possible inputs for the key items:

```rust
let records = vec![ (None, "World") ];
let records = vec![ (Some(RecordKey::Null), "World") ]; // This means the same as above
let records = vec![ ("Hello", "World") ];
```

I think this would make things confusing since there would be more than one correct way to do things, which is good to avoid. Because of this, I don't think we should change the signature of `send_all`.

src/client/src/producer.rs
nicholastmosher: I have looked at how to go about this in the code and I don't think it's the right call. The fact that we use `Bytes` is a detail that is only evident down at the request level, an implementation detail that should not matter to the user. Also, coverting `Vec<u8>` to `Bytes` is cheap, it just takes ownership of the backing buffer.

sehz: Implementation matter because you want to reduce unnecessary layer of abstraction.   Since we are using `Bytes` as internal representation we should propagate much as possible

nicholastmosher: A more concrete reason to use `Vec<u8>` is that since `send` calls `send_all` and `send_all` accepts `Into<Vec<u8>>`, that means that we get all sorts of problems if we use `Bytes` in `send`:

```rust
    pub async fn send<K, V>(&self, key: K, value: V) -> Result<(), FluvioError>
    where
        K: Into<RecordKey>,
        V: Into<RecordValue>,
    {
        let record_key = key.into();
        let record_value = value.into();

        let maybe_key = match record_key { // This would become Option<Bytes>
            RecordKey::Null => None,
            RecordKey::Key(key) => Some(key),
        };
        let value = record_value.0; // This becomes `Bytes`

        // the trait `From<bytes::Bytes>` is not implemented for `Vec<u8>`
        self.send_all(Some((maybe_key, value))).await?;
        Ok(())
    }
```

src/client/src/producer.rs
sehz: This works:
```
use super::*;

    use dataplane::bytes::Bytes;

    pub enum RecordKey {
        Null,
        Key(Bytes),
    }
    
    impl<K: Into<Vec<u8>>> From<K> for RecordKey 
    {
        fn from(key: K) -> Self {
            let vec: Vec<u8> = key.into();
            Self::Key(vec.into())
        }
    }
    
    pub struct RecordValue(Bytes);
    
    impl<T: Into<Vec<u8>>> From<T> for RecordValue {
        fn from(val: T) -> Self {
            let vec: Vec<u8> = val.into();
            Self(vec.into())
        }
    }

    pub async fn send_all<K, V, I>(records: I) -> Result<(), FluvioError>
    where
        K: Into<RecordKey>,
        V: Into<RecordValue>,
        I: IntoIterator<Item = (K,V)>,
    { 
        Ok(())
     }


    async fn test() {
        send_all(vec![ (RecordKey::Null, "World") ]).await;
        send_all(vec![ ("Hello", "World")]).await;
    }
```

nicholastmosher: The problem with this solution is that it would be a breaking change. Changing `K: Into<Vec<u8>>` to `K: Into<RecordKey>` is fine because of the `impl<K: Into<Vec<u8>>> From<K> for RecordKey`, but the breaking change comes from this change:

- `I: IntoIter<Item = (Option<K>, V)>` into
- `I: IntoIter<Item = (K, V)>`

So if we wanted to adopt that change to the API, it would be a breaking change.

sehz: Since this is new API, we can make breaking changes assuming next release can be done quickly.  I think it is worth it, this will reduce API surface and make it consistent with `send`

nicholastmosher: Ok, but I would still like to bump the fluvio crate from `0.7.0` to `0.8.0` so that cargo cannot accidentally confuse the two

nicholastmosher:
Alright, I have made one notable change:

Originally, I was going to use this definition for RecordKey:

```rust
pub enum RecordKey {
    Null,
    Key(Bytes),
}
```

But I remembered that even the inner details of a public enum are public, meaning that any future change to the structure or contents of this enum would be a breaking change. So I changed it to be a tuple struct instead:

```rust
pub struct RecordKey(RecordKeyInner);

impl RecordKey {
    pub const NULL: Self = Self(RecordKeyInner::Null);
}

enum RecordKeyInner {
    Null,
    Key(Bytes),
}
```

RecordKey still has the same `impl<K: Into<Vec<u8>>> From<K> for RecordKey`, so the calling patterns for `send` with keys is the same. The only difference now is using the `RecordKey::NULL` constant rather than the `RecordKey::Null` enum variant.

I think this solution is pretty good, it gives us the flexibility in the future to add or change data in the RecordKey type.

ajhunyady:
I'm not an expert here, but I feel you had a simple elegant solution, which now became wrapped into another inner object which makes it harder to understand for a`maybe later` use case which may never happen. Not sure it's the right compromise.

nicholastmosher:
@ajhunyady keep in mind that with the change I just made, the only difference that would be seen from the user's perspective is between using the following for null-key records:

- `producer.send(RecordKey::Null, "Hello").await?`
- `producer.send(RecordKey::NULL, "Hello").await?`

I'd argue that this is a negligible change from the user's perspective, but the extra flexibility to change things down the line is invaluable to us as developers of this API.

sehz:
> Alright, I have made one notable change:
> 
> Originally, I was going to use this definition for RecordKey:
> 
> ```rust
> pub enum RecordKey {
>     Null,
>     Key(Bytes),
> }
> ```
> 
> But I remembered that even the inner details of a public enum are public, meaning that any future change to the structure or contents of this enum would be a breaking change. So I changed it to be a tuple struct instead:
> 
> ```rust
> pub struct RecordKey(RecordKeyInner);
> 
> impl RecordKey {
>     pub const NULL: Self = Self(RecordKeyInner::Null);
> }
> 
> enum RecordKeyInner {
>     Null,
>     Key(Bytes),
> }
> ```
> 
> RecordKey still has the same `impl<K: Into<Vec<u8>>> From<K> for RecordKey`, so the calling patterns for `send` with keys is the same. The only difference now is using the `RecordKey::NULL` constant rather than the `RecordKey::Null` enum variant.
> 
> I think this solution is pretty good, it gives us the flexibility in the future to add or change data in the RecordKey type.

Let's move conversation to issue

nicholastmosher:
@sehz does this look ready to you?

sehz (approved):
Looks good.  But we need to hold off merging to master until other major changes are ready

PR #989, opened 2021-04-22
Initial stuff for wasm32 browser support 

### LINKED ISSUE #933: Spike: Prototype Fluvio wasm client in browser, identify blockers
There are many cases for which it would be excellent to have a browser-compatible wasm client for Fluvio. This would allow us to write applications that produce and consume directly from the client to the cluster. However, there are some challenges that we foresee that we don't know how difficult they will be to overcome. For example:

- Can we use websockets as a transparent drop-in replacement for TCP sockets?
- Can we share much of the client code we already have, or will we need a parallel crate just for wasm?
- Can we implement the same client interface using wasm-compatible APIs?

We should spend some time figuring out which potential problems are real, and which ones can be worked through at this time. We need to know whether we can accomplish this in a reasonable amount of time with the current ecosystem or whether we need to wait for more developments to catch up.

simlay:
This should close #933. It depends on infinyon/future-aio#55. 

TODO:
- [x] Actually do more than just use `unimplemented!()` macros
- [ ] Test it end to end
- [ ] SPU backend support.
- [ ] Remove changes to fluvio-socket.


src/client/src/admin.rs
```diff
@@ -6,9 +6,19 @@ use dataplane::core::Encoder;
...
+use fluvio_socket::AllMultiplexerSocket as FluvioMultiplexerSocket;
+
+#[cfg(not(target_arch = "wasm32"))]
+use fluvio_future::native_tls::AllDomainConnector as FluvioConnector;
+
+#[cfg(target_arch = "wasm32")]
+use crate::websocket::{
+    WebSocketConnector as FluvioConnector,
+    MultiplexerWebsocket as FluvioMultiplexerSocket,
+};
```
sehz, line 20: Instead of directly using WebSocketConnector, It probably more flexible long term to move that to `FluvioSocket`.  In that case, Client would not needed to be exposed to WASM dependencies.

src/client/src/sockets.rs
```diff
@@ -268,13 +280,13 @@ impl SerialFrame for VersionedSerialSocket {
     }
 
     /// send and wait for reply serially
-    async fn send_receive<R>(&mut self, request: R) -> Result<R::Response, FlvSocketError>
```
sehz, line 271: If we move WASM connection to `FluvioSocket`, we can still keep FlvSocketError

src/client/src/websocket.rs
```diff
@@ -0,0 +1,215 @@
...
+/*
+#[derive(Error, Debug)]
+pub enum JsError {
+    #[error(transparent)]
+    Empty
+}
+*/
+
+#[derive(Clone)]
+pub enum WebSocketConnector {
```
sehz, line 27: Let's move this to `FluvioSocket`

sehz (changes requested):
Great work.  See my comment about re-using FluvioSocket/WASM

src/client/.vim
```diff
@@ -0,0 +1 @@
+../socket/.vim
```
simlay, line 1: remove this.

src/client/src/sockets.rs
```diff
@@ -175,15 +196,19 @@ impl ClientConfig {
     }
 
     pub(crate) async fn connect(self) -> Result<VersionedSocket, FluvioError> {
-        let socket =
-            FluvioSocket::connect_with_connector(&self.addr, self.connector.as_ref()).await?;
+        #[cfg(unix)]
+        let socket = FluvioSocket::connect_with_connector(&self.addr, self.connector.as_ref()).await?;
```
simlay, line 200: This will be fixed with https://github.com/infinyon/future-aio/issues/66

src/client/src/sync/controller.rs
```diff
@@ -12,12 +12,13 @@ use event_listener::{Event, EventListener};
 
 use dataplane::core::Encoder;
 use dataplane::core::Decoder;
-use fluvio_socket::AsyncResponse;
```
simlay, line 15: revert all changes to this file.

src/client/src/sync/mod.rs
```diff
@@ -130,7 +130,6 @@ mod context {
                         format!("timed out searching metadata {} failed due to timeout: {} ms",S::LABEL,*MAX_WAIT_TIME),
                     ))
                 }
-
```
simlay, line 133: revert.

src/socket/.vim/settings.json
```diff
@@ -0,0 +1,7 @@
+{
```
simlay, line 1: remove this

src/client/src/config/config.rs
```diff
@@ -15,10 +15,18 @@ use std::fs::create_dir_all;
 use thiserror::Error;
 
 use tracing::debug;
-use dirs::home_dir;
+
 use serde::Deserialize;
 use serde::Serialize;
 
+#[cfg(not(target_arch = "wasm32"))]
```
simlay, line 22: #1115 fixes this.

sehz:
Close this?

PR #994, opened 2021-04-22
[Merged by Bors] - Bump crate versions and VERSION for 0.7.4

nicholastmosher:


src/auth/Cargo.toml
```diff
@@ -22,7 +22,7 @@ tracing = "0.1.21"
 tracing-futures = "0.2.4"
 x509-parser = "0.9.1"
 
-fluvio-controlplane-metadata = { version = "0.7.0", path = "../controlplane-metadata" }
+fluvio-controlplane-metadata = { version = "0.8.0", path = "../controlplane-metadata" }
```
sehz, line 25: reason for updating to `0.7.0`?

sehz (changes requested):

src/auth/Cargo.toml
nicholastmosher: Yes, we changed `TopicSpec::partitions() -> Option<i32>` to `TopicSpec::partitions() -> i32` which is a breaking change

sehz: Got it

sehz (approved):

PR #1000, opened 2021-04-23
[Merged by Bors] - Add permissions.contents: read to GitHub workflows

nicholastmosher:
Based on a new [blog from GitHub], it is apparent that the default setup for `GITHUB_TOKEN` should not be considered secure. It seems that the default settings have the following problems:

- When using `actions/checkout`, the `GITHUB_TOKEN` becomes implicitly available to every subsequent action used in the workflow.
- Even if not using `actions/checkout`, any third-party action can ask for the `GITHUB_TOKEN` and freely receive it, with read/write permissions to your repository
- By setting `permissions.contents: read`, the `GITHUB_TOKEN` that is provided in the workflow is restricted from full read/write access on all resources to strictly read-only on the contents of the repository.

[Blog from GitHub]: https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,8 @@
 name: CI
 
+permissions:
+  contents: read
+
 on:
   push:
     branches:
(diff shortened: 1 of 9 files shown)


sehz (approved):
Nice find!

PR #1001, opened 2021-04-23
[Merged by Bors] - Add check for FLUVIO_VERSION in GH release task

nicholastmosher:
Currently, the task `cargo make github-release-upload` uses `FLUVIO_VERSION` and `TARGET` without first checking that they exist. This has been leading to the problem where bad releases are pushed, e.g. a tag called `v` gets pushed because the version that should follow it is missing. This PR:

- Adds a check for FLUVIO_VERSION to the `github-release-upload` task
- Defines the `FLUVIO_VERSION` variable in the release workflow

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,12 +28,14 @@ jobs:
         os: [ubuntu-latest, macos-latest]
         rust: [stable]
     steps:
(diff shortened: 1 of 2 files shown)


sehz (approved):
LGTM

PR #1008, opened 2021-04-24
[Merged by Bors] - Use file name for the external commands (fixes #889)

### LINKED ISSUE #889: CLI: `fluvio --help` should display extension executable names rather than pretty names
Back when we updated `fluvio version` to display all extension versions, I changed the metadata `command` field to `title` to give a pretty display name rather than a command's own executable name. When doing so, I forgot to update the display name of the extensions in the help menu. It now shows:

```
SUBCOMMANDS:
    consume             Read messages from a topic/partition
    produce             Write messages to a topic/partition
    topic               Manage and view Topics
    partition           Manage and view Partitions
    profile             Manage Profiles, which describe linked clusters
    cluster             Install or uninstall Fluvio clusters
    install             Install Fluvio plugins
    update              Update the Fluvio CLI
    version             Print Fluvio version information
    Fluvio Packages     Package publishing and management
    Fluvio Cloud CLI    Cloud Operations
    help                Prints this message or the help of the given
                        subcommand(s)
```

Whereas we should have `package` and `cloud` instead of `Fluvio Packages` and `Fluvio Cloud`. This should not be difficult to fix, we just need to update the help menu printer to use each executable's filename rather than the `title` it gives us.

avinassh:
Ran into this issue myself, I thought I'd send a patch. Here is the new output now:

```shell 
$ cargo run

Fluvio command-line interface

fluvio-cli [FLAGS] [OPTIONS] <SUBCOMMAND>

FLAGS:
        --tls                   Enable TLS
        --enable-client-cert    TLS: use client cert
    -h, --help                  Prints help information

OPTIONS:
    -c, --cluster <host:port>          Address of cluster
        --domain <domain>              Required if client cert is used
        --ca-cert <ca-cert>
            Path to TLS ca cert, required when client cert is enabled

        --client-cert <client-cert>    Path to TLS client certificate
        --client-key <client-key>      Path to TLS client private key
    -P, --profile <profile>

SUBCOMMANDS:
    consume         Read messages from a topic/partition
    produce         Write messages to a topic/partition
    topic           Manage and view Topics
    partition       Manage and view Partitions
    profile         Manage Profiles, which describe linked clusters
    cluster         Install or uninstall Fluvio clusters
    install         Install Fluvio plugins
    update          Update the Fluvio CLI
    version         Print Fluvio version information
    fluvio-cloud    Cloud Operations
    help            Prints this message or the help of the given
                    subcommand(s)
```

### DIFF
--- a/src/cli/src/lib.rs
+++ b/src/cli/src/lib.rs
@@ -185,7 +185,15 @@ impl HelpOpt {
         // Add external command definitions to our own clap::App definition
         let mut app: App = Root::clap();
         for i in &external_commands {
-            app = app.subcommand(SubCommand::with_name(&i.meta.title).about(&*i.meta.description));
+            if let Some(file_name) = i.path.file_name() {
+                app = app.subcommand(
+                    SubCommand::with_name(&*file_name.to_string_lossy())
+                        .about(&*i.meta.description),
+                );
+            } else {
+                app = app
+                    .subcommand(SubCommand::with_name(&*i.meta.title).about(&*i.meta.description));
+            }
         }
 
         // Use clap's help printer, loaded up with external subcommands
--- a/src/cli/src/lib.rs
+++ b/src/cli/src/lib.rs
@@ -185,14 +185,20 @@ impl HelpOpt {
         // Add external command definitions to our own clap::App definition
         let mut app: App = Root::clap();
         for i in &external_commands {
-            if let Some(file_name) = i.path.file_name() {
-                app = app.subcommand(
-                    SubCommand::with_name(&*file_name.to_string_lossy())
+            match i.path.file_name() {
+                Some(file_name) => {
+                    app = app.subcommand(
+                        SubCommand::with_name(
+                            file_name.to_string_lossy().strip_prefix("fluvio-").unwrap(),
+                        )
                         .about(&*i.meta.description),
-                );
-            } else {
-                app = app
-                    .subcommand(SubCommand::with_name(&*i.meta.title).about(&*i.meta.description));
+                    );
+                }
+                None => {
+                    app = app.subcommand(
+                        SubCommand::with_name(&*i.meta.title).about(&*i.meta.description),
+                    );
+                }
             }
         }
 


nicholastmosher (changes requested):
Thanks for the PR @avinassh, it's greatly appreciated!

External subcommands have filenames with the prefix `fluvio-`, such as `fluvio-cloud`, but the name of the subcommand that should show up in the help menu should be everything after the prefix. In the `fluivo-cloud` example, `cloud` should be displayed in help. I think `str::strip_prefix("fluvio-")` would work nicely here.

One nit: could you use a match statement instead of `if-let/else`?

Thanks!

avinassh:
> External subcommands have filenames with the prefix fluvio-, such as fluvio-cloud, but the name of the subcommand that should show up in the help menu should be everything after the prefix

Is this always guaranteed and I can call unwrap on `str::strip_prefix` OR should I handle the `None` case as well?

nicholastmosher:
I think the representation in the help menu is a different issue than this. Fixing it would probably require editing the help menu template used by structopt/clap. Also just FYI, `fluvio-cloud` currently belongs to a private repo which is why you wouldn't have found it

avinassh:
> I think the representation in the help menu is a different issue than this. 

yeah, actually you are right. btw do you could tell me about the other thing:

> Is this always guaranteed and I can call unwrap on str::strip_prefix OR should I handle the None case as well?

nicholastmosher:
@avinassh in the current implementation, it should be guaranteed, yes: https://github.com/infinyon/fluvio/blob/master/src/cli/src/install/mod.rs#L43-L58

I'd say let's move forward using that unwrap, and if anything changes such that this would actually panic, we'll consider those changes a bug and we'll update accordingly at that time.

avinassh:
Nice! I have updated the PR. 

```shell
$ cargo run -- --help

// snipped

SUBCOMMANDS:
    consume      Read messages from a topic/partition
    produce      Write messages to a topic/partition
    topic        Manage and view Topics
    partition    Manage and view Partitions
    profile      Manage Profiles, which describe linked clusters
    cluster      Install or uninstall Fluvio clusters
    install      Install Fluvio plugins
    update       Update the Fluvio CLI
    version      Print Fluvio version information
    cloud        Cloud Operations
    help         Prints this message or the help of the given subcommand(s)
```

nicholastmosher (approved):
Looks great! Thanks @avinassh!

PR #1010, opened 2021-04-24
[Merged by Bors] - Better, more automated release flow

nicholastmosher:
This is not quite done yet, but I wanted to open a PR so we can start reviewing it since I'm sure there will be questions and things to fix. This introduces a new workflow for CI/CD that makes the deployment of "latest" builds faster and more robust. Let me explain the new workflow:

We have two primary workflow files, `ci.yml` and `publish_latest.yml`.

The CI workflow now has two distinct "modes":

- One is for running checks and tests on PRs, this mode runs every time a commit is pushed to a PR and it builds and runs tests in debug mode.
- The other mode gets triggered whenever a commit is pushed to `staging`, i.e. whenever Bors is executing an `r+` build. In this mode, everything is built in release mode, and the `fluvio` and `fluvio-run` artifacts for each target are saved using `actions/upload-artifact@v2`. This means that if Bors pushes a commit to master, we are guaranteed to have good release artifacts already built and saved.

The `publish_latest` workflow is triggered any time a commit is pushed to `master`, i.e. whenever Bors successfully runs CI and build artifacts are saved. This workflow will download the artifacts that were saved by CI and do the following:

- Publish the artifacts to the `dev` GitHub prerelease
- Build a docker image wrapping `fluvio-run` on linux-musl and push to docker registry
- Push the latest helm chart
- Publish the CLI to packages.fluvio.io and bump the latest CLI version.

The final piece to this work is finishing the `release` workflow, which should essentially just re-tag everything from the most recent "latest" publishing. The objective is that this workflow is very minimal and has very little that can go wrong. Previously, our release workflow would actually try to build the release artifacts, making it very slow as well as susceptible to tripping over flaky builds.

---

There is one foreseen obstacle remaining for finishing the release workflow. When building our components, we bake in the current value of the `VERSION` file. When building "latest" versions, we bake in the git commit into the version as well. This was done in order to satisfy a problem with the platform_version check between a client and cluster, but I now believe that was the wrong choice of approach.

The problem this creates for us now is that when we build release artifacts, they will have git commits baked into the VERSION as a prerelease tag of the version. This is something we do not want when we publish a proper release, so this would prevent us from re-using build artifacts saved from a "latest" workflow as a proper release. I have a fix in mind for this, I just need to do it now.


.github/workflows/publish_latest.yml
```diff
@@ -0,0 +1,162 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        artifact: [fluvio, fluvio-run]
+        target: [x86_64-unknown-linux-musl, x86_64-apple-darwin]
+    steps:
+      - name: Login GH CLI
+        run: gh auth login --with-token < <(echo ${{ secrets.GITHUB_TOKEN }})
+      - name: Download artifact
+        run: gh run download -R nicholastmosher/fluvio -n "${{ matrix.artifact }}-${{ matrix.target }}-${{ github.sha }}"
```
sehz, line 46: remove personal id

.github/workflows/publish_latest.yml
```diff
@@ -0,0 +1,162 @@
...
+      - name: Rename artifact
+        run: mv ${{ matrix.artifact }} ${{ matrix.artifact }}-${{ matrix.target }}
+      - name: Publish artifact
+        run: gh release upload -R nicholastmosher/fluvio --clobber dev ./${{ matrix.artifact }}-${{ matrix.target }}
```
sehz, line 50: ditto

sehz, line 50: don't hard code `dev`. see comment about env var

.github/workflows/publish_latest.yml
```diff
@@ -0,0 +1,162 @@
...
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        rust: [stable]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Login GH CLI
+        run: gh auth login --with-token < <(echo ${{ secrets.GITHUB_TOKEN }})
+      - name: Download artifact
+        run: gh run download -R nicholastmosher/fluvio -n "fluvio-run-x86_64-unknown-linux-musl-${{ github.sha }}"
```
sehz, line 65: remove personal id

.github/workflows/publish_latest.yml
```diff
@@ -0,0 +1,162 @@
...
+
+      - name: Login to Docker Hub
+        run: docker login --username=${{ secrets.DOCKER_USERNAME }} --password=${{ secrets.DOCKER_PASSWORD }}
+      - name: Publish latest development Fluvio Image
+        run: |
+          export TAG="$(cat VERSION)-${{ github.sha }}"
+          k8-util/docker/build.sh "${TAG}" "./fluvio-run"
```
sehz, line 72: use `make latest_image`

.github/workflows/publish_latest.yml
```diff
@@ -0,0 +1,162 @@
...
+          version: '0.32.9'
+      - name: Install fluvio-package
+        run: |
+          curl -fsS https://packages.fluvio.io/v1/install.sh | bash
+          ${HOME}/.fluvio/bin/fluvio install fluvio-package
+
+      - name: Download fluvio x86_64-unknown-linux-musl
+        run: |
+          mkdir -p target/x86_64-unknown-linux-musl/release/
+          gh run download -R nicholastmosher/fluvio \
```
sehz, line 121: remove personal id

.github/workflows/publish_latest.yml
```diff
@@ -0,0 +1,162 @@
...
+            -n "fluvio-x86_64-unknown-linux-musl-${{ github.sha }}" \
+            -D target/x86_64-unknown-linux-musl/release/
+
+      - name: Download fluvio x86_64-apple-darwin
+        run: |
+          mkdir -p target/x86_64-apple-darwin/release/
+          gh run download -R nicholastmosher/fluvio \
```
sehz, line 128: personal id

sehz (changes requested):
Excellent work.  Mostly minor points

sehz:
One of the reason tests are run in `RELEASE` mode is that Github VM can be over-loaded and slow which cause issue in some test that assume timing dependencies.

.github/workflows/ci.yml
```diff
@@ -56,34 +56,40 @@ jobs:
...
-            task: "run-all-doc-test"
+            run: make run-all-doc-test
+            publish: false
+          - name: Unit tests
+            run: make build-all-test run-all-unit-test
+            publish: true
         include:
           - os: ubuntu-latest
             sccache-path: /home/runner/.cache/sccache
+            target: x86_64-unknown-linux-musl
```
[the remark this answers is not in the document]
nicholastmosher: We cannot add the targets directly to the matrix because it would mismatch the targets with the build hosts.

For example, if we had the following matrix:

```yaml
matrix:
  os: [ubuntu-latest, macos-latest]
  ...
  target: [x86_64-unknown-linux-musl, x86_64-apple-darwin]
```

Then we would get jobs with the following configurations:

```yaml
- os: ubuntu-latest
  target: x86_64-unknown-linux-musl
- os: ubuntu-latest
  target: x86_64-apple-darwin # This is mismatched with the host OS
- os: macos-latest
  target: x86_64-unknown-linux-musl # This is mismatched with the host OS
- os: macos-latest
  target: x86_64-apple-darwin
```

The include rule is the proper place for the target definition, since we can define which target gets built on which OS. I think it may be possible to add multiple targets for a given OS by defining multiple include rules, like the following:

```yaml
include:
- os: ubuntu-latest
  target: x86_64-unknown-linux-musl
- os: ubuntu-latest
  target: some-other-target-built-on-ubuntu-host
```

.github/workflows/ci.yml
```diff
@@ -56,34 +56,40 @@ jobs:
...
           - os: macos-latest
             sccache-path: /Users/runner/Library/Caches/Mozilla.sccache
```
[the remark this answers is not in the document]
nicholastmosher: What do you mean? This is saying "when the OS is `macos-latest`, the sccache-path is `/Users/runner/Library/Caches/Mozilla.sccache`". What would it even mean to make this conditional?

.github/workflows/ci.yml
```diff
@@ -111,6 +117,13 @@ jobs:
           toolchain: ${{ matrix.rust }}
           profile: minimal
           override: true
+      - name: Install Musl Tools
+        if: ${{ matrix.target == 'x86_64-unknown-linux-musl' }}
+        run: |
+          sudo apt install -y musl-tools
+          sudo ln -s /usr/bin/musl-gcc /usr/local/bin/x86_64-linux-musl-gcc
+          echo "TARGET_CC=musl-gcc" | tee -a $GITHUB_ENV
+          rustup target add x86_64-unknown-linux-musl
```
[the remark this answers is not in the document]
nicholastmosher: I have recently learned that you can define multiple github actions within a single repo (e.g. see https://github.com/bytecodealliance/wasmtime/blob/main/.github/workflows/main.yml#L77), which has made me think about creating a bunch of little re-usable setup actions for things like this. E.g. an action for installing musl tools or an action for installing sccache

.github/workflows/publish_latest.yml
```diff
@@ -0,0 +1,162 @@
...
+jobs:
+  # Bump the `dev` tag to this commit in order to overwrite the `dev` GH prerelease
+  bump_github_release:
+    name: Bump dev tag
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Bump dev tag
+        run: |
+          git tag -f dev
```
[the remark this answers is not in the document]
nicholastmosher: Is there a specific reason to make it a variable? This simply overwrites a single prerelease on the releases page, e.g.

image

When would we want to change this behavior? I am anticipating this workflow to only ever be run automatically upon Bors merge.

.github/workflows/publish_latest.yml
nicholastmosher: `make latest_image` is too specific for what needs to happen here. What it does is build the docker image from the hardcoded file `"./target/x86_64-unknown-linux-musl/$(BUILD_PROFILE)/fluvio-run"`, instead of what we need here, which is to use `./fluvio-run`. That is the reason I rewrote the `k8-util/docker/build.sh` script to take the filename as an argument.

nicholastmosher:
> One of the reason tests are run in `RELEASE` mode is that Github VM can be over-loaded and slow which cause issue in some test that assume timing dependencies.

Are you saying that we should or should not run tests in release mode?

.github/workflows/ci.yml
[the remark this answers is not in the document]
sehz: hmm.  advantage of putting in the script is that it can be executed locally.  this is not showstopper so we can come back and refactor later on

.github/workflows/publish_latest.yml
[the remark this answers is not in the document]
sehz: it's more that we want to centralize it configuration.   not urgent thought

sehz:
> > One of the reason tests are run in `RELEASE` mode is that Github VM can be over-loaded and slow which cause issue in some test that assume timing dependencies.
> 
> Are you saying that we should or should not run tests in release mode?

We should stick with existing behavior for now.  We can change it to debug if there are no behavior consequences.

nicholastmosher:
Alright @sehz, I think I have addressed most of your points. Here's how I would like to proceed:

- I want to merge this PR so that we can work out any last kinks in the publish flow that may have to do with access keys, etc.
- I have removed the changes to `release.yml` and will open a new PR to finish the release workflow. This should amount to just re-tagging things and uploading them, but I can't test it out until this PR is merged.
- One the release workflow is finished, the only manual thing we should need to do on each release is publish crates and run the workflow dispatch for `release.yml`.

Let me know if there is anything about this you'd like to change

sehz (approved):
Sounds good

PR #1012, opened 2021-04-26
[Merged by Bors] - Update release workflow to leverage new publish flow

nicholastmosher:
This PR rewrites the `release.yml` workflow to re-publish the latest `dev` release as a stable release, rather than attempting to re-build all of the artifacts on demand. This should cause the workflow to be more steady and reliable, and to take less time to run.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,171 +4,102 @@ permissions:
   contents: read
 
 on:
-#  push:
-#    # Sequence of patterns matched against refs/tags
-#    tags:
-#      - "v*" # Push events to matching v*, i.e. v1.0, v20.15.10
   workflow_dispatch:
-    inputs:
-      force:
-        required: false
-        description: 'Force push this release'
-      update_script:
-        description: 'Whether to update the install.sh script'
-        required: false
-      test:
-        required: false
-        description: 'Whether to run a test release'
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -4,171 +4,102 @@ permissions:
   contents: read
 
 on:
-#  push:
-#    # Sequence of patterns matched against refs/tags
-#    tags:
```
sehz, line 9: How do you trigger release workflow if you remove tag?

sehz (changes requested):

.github/workflows/release.yml
nicholastmosher: For now, the release workflow will be run manually by us only when we want to push a real release, i.e. at the end of each week. I imagine once we do that a few times and are confident in the workflow, we could add a special pattern for tags to trigger the workflow, so e.g. we can just do

```
$ git tag v0.8.1
$ git push upstream v0.8.1
```

and have the workflow trigger automatically for us

nicholastmosher: Also note that I didn't actually change the behavior of this trigger, the `on: push: tags` block deleted here was already commented out

sehz (approved):
K. Lets try

PR #1013, opened 2021-04-26
[Merged by Bors] - Do GitHub prerelease at end of CI

nicholastmosher:
This is a continuation of the overhaul of our releases workflow. I discovered that we had timing issues with using the `actions/upload-artifact` action because I was doing a hacky workaround instead of using the corresponding `actions/download-artifact` in the same workflow as is intended. This resulted in the "publish" step being unable to download the artifacts because they were unavailable via the API (using the `gh` CLI) when attempting to download those artifacts from the context of a different workflow.

To fix this, I have shuffled around a little bit of the release flow. Now, we publish to the GH `dev` prerelease at the end of the CI workflow, rather than at the beginning of the "publish" workflow. The end-to-end series of events now looks like this:

- Bors merges a PR to `upstream/staging`, kicking off the CI build
  - CI builds and tests the commit
  - As the last step in CI, we upload the release artifacts _to GitHub Releases, NOT just to github artifacts_
- When CI passes, Bors pushes the commit at staging to master
  - This kicks off the "publish" flow, which downloads the release artifacts from GH Releases (dev) and publishes them as "latest" releases to docker, helm, and packages.fluvio.io
- Then finally, we are able to run the release workflow, which will take the latest GH Releease from dev and publish them as a proper release.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,7 +55,7 @@ jobs:
       - name: Check rustfmt
         run: make check-fmt
 
-  tests:
+  build:
     name: ${{ matrix.task.name }} (${{ matrix.os }})
     runs-on: ${{ matrix.os }}
     strategy:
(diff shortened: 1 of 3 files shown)


sehz (approved):

nicholastmosher:
🤞 

bors r+

PR #1023, opened 2021-04-27
[Merged by Bors] - Fix download for release and helm publish

nicholastmosher:
I think I should be able to test this one out directly, without merging it, let's see.


nicholastmosher:
Ok, turns out it needs to be merged to test, @sehz?

sehz (approved):

PR #1028, opened 2021-04-27
Add version compatibility test

### LINKED ISSUE #996: CI - Add forward/backwards compatibility matrix tests for client/cluster
This may probably be a matrix test.

Steps:
- Install a Fluvio cluster "the official way", but we want to set the `VERSION`.
- With the `VERSION` client:
  - Create a topic, produce and consume from topic
- Compile `master`. With the `master` client:
  - Create a *new* topic, and produce and consume.
  - Finally, try to consume then produce to the old topic

---

~I also want to revise the upgrade testing, so we also test that an old client is forward compatible with a new cluster. This only requires adding the topic create, produce, consume pattern using the stable version client.~ Cover forward compatibility in matrix test

tjtelan:
Exercises the public installation flow. The current strategy is to test `latest`, `stable` (providing no version), and `stable - 1`, which is `0.7.3`. We might need to update this matrix with each release.

1. We install a cluster, create a topic and produce/consume data (while cluster version == cli version)
2. Change CLI versions
3. Repeat step 1 with a new topic
4. Verify compatibility with topic from step 1

This is running on `ubuntu-latest` because it is a short test and I didn't want to queue up for the custom runner. However, we're not doing any k8 testing. I figured that was ok because this is not an upgrade test.

Closes #996

Passing run: https://github.com/infinyon/fluvio-perf-test/actions/runs/790216333

### DIFF
--- a/.github/workflows/cross_version_compat.yml
+++ b/.github/workflows/cross_version_compat.yml
@@ -1,0 +1,189 @@
+name: Cluster-Client version compatibility 
+
+on:
+  push:
+    branches: [master]
+  workflow_dispatch:
+jobs:
+  cross_compat:
+    timeout-minutes: 20
+    name: cluster(${{ matrix.cluster_type }}):${{ matrix.cluster_version }} x CLI:${{ matrix.cli_version }} 
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
+        rust: [stable]
+        cluster_type: [local] # fyi: k8 tests are unstable in github-runner
+        # Strategy: latest, stable, stable-1 release
+        cli_version: [latest, stable, 0.7.3]
+        cluster_version: [latest, stable, 0.7.3]
+        exclude:
+          - cli_version: latest
+            cluster_version: stable
+          - cli_version: latest
+            cluster_version: 0.7.3 
+          - cli_version: stable
+            cluster_version: 0.7.3 
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          repository: infinyon/fluvio 
+      - run: helm version
+      - name: Install Rust ${{ matrix.rust }}
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: ${{ matrix.rust }}
+          profile: minimal
+          override: true
+      - name: Install Minikube for Github runner 
+        if: matrix.os == 'ubuntu-latest' 
+        uses: manusa/actions-setup-minikube@v2.3.1
+        with:
+          minikube version: 'v1.16.0'
+          kubernetes version: 'v1.19.2'
+
+      - name: Open tunnel for Github runner
+        if: matrix.os != 'infinyon-ubuntu-bionic'
+        run: |
+          nohup  minikube tunnel --alsologtostderr > /tmp/tunnel.out 2> /tmp/tunnel.out &
+
+      - name: Setup Minikube for Linux
+        if: startsWith(matrix.os, 'infinyon-ubuntu')
+        run: |
+          pkill -f "minikube tunnel" || true
+          minikube delete
+          minikube start --driver=docker --kubernetes-version 1.19.6
+          nohup  minikube tunnel --alsologtostderr > /tmp/tunnel.out 2> /tmp/tunnel.out &
+      - name: Test minikube
+        run: |
+          minikube profile list
+          minikube status
+
+      - name: Install Fluvio for cluster @ (version ${{ matrix.cluster_version }}), 
+        if: matrix.cluster_version != 'stable'
+        uses: nick-invision/retry@v2
+        with:
+          timeout_minutes: 5
+          max_attempts: 3
+          command: |
+            curl -fsS https://packages.fluvio.io/v1/install.sh | VERSION=${{ matrix.cluster_version }} bash
+            ~/.fluvio/bin/fluvio version
+
+      - name: Install Fluvio for cluster @ (version stable), 
+        if: matrix.cluster_version == 'stable'
+        uses: nick-invision/retry@v2
+        with:
+          timeout_minutes: 5
+          max_attempts: 3
+          command: |
+            curl -fsS https://packages.fluvio.io/v1/install.sh | bash
+
+      - name: Start Fluvio cluster (local) @ (version ${{ matrix.cluster_version }}) 
+        if: matrix.cluster_type == 'local'
+        uses: nick-invision/retry@v2
+        with:
+          timeout_minutes: 5
+          max_attempts: 3
+          command: ~/.fluvio/bin/fluvio cluster start --local
+
+      - name: Start Fluvio cluster (k8) @ (version ${{ matrix.cluster_version }}) 
+        if: matrix.cluster_type == 'k8'
+        uses: nick-invision/retry@v2
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/cross_version_compat.yml
```diff
@@ -0,0 +1,189 @@
+name: Cluster-Client version compatibility 
+
+on:
+  push:
+    branches: [master]
+  workflow_dispatch:
```
nicholastmosher, line 6: We need to talk about when we expect to run this. I think tests should be run on `pull_request` or on push to `staging`. Is the expectation that these tests are just data points or would they be prerequisites to merging new code to master?

tjtelan: @sehz Do you have an opinion?

I only intended for them to be data points, and I didn't intend for this to communicate stability or performance.

Until we have a concept of a long-term stable version, the value from this test is early warning from `latest`. A sign that we'll need to minor version bump.

I don't think we'll test every version exhaustively, but the `exclude` map does document where we are incompatible, just as a consequence of having the test pass 100%.

.github/workflows/cross_version_compat.yml
```diff
@@ -0,0 +1,189 @@
...
+        # Strategy: latest, stable, stable-1 release
+        cli_version: [latest, stable, 0.7.3]
+        cluster_version: [latest, stable, 0.7.3]
+        exclude:
+          - cli_version: latest
+            cluster_version: stable
+          - cli_version: latest
+            cluster_version: 0.7.3 
+          - cli_version: stable
+            cluster_version: 0.7.3
```
nicholastmosher, line 27: I think this is going to become a bottleneck for this test. I think we need to come up with a programmatic way of finding the "stable - 1" version. One potential option for this may be to use the `gh` CLI:

```
$ gh release list -R infinyon/fluvio
dev             Pre-release  (dev)             about 1 hour ago
v0.8.0          Latest       (v0.8.0)          about 3 hours ago
v0.7.3                       (v0.7.3)          about 24 days ago
v0.7.2                       (v0.7.2)          about 1 month ago
v0.7.2-alpha.1  Pre-release  (v0.7.2-alpha.1)  about 1 month ago
v0.7.1                       (v0.7.1)          about 1 month ago
v0.7.1-beta.0   Pre-release  (v0.7.1-beta.0)   about 1 month ago
v0.7.1-alpha.0  Pre-release  (v0.7.1-alpha.0)  about 1 month ago
v0.7.0                       (v0.7.0)          about 2 months ago
v0.7.0-beta.1   Pre-release  (v0.7.0-beta.1)   about 2 months ago
v0.7.0-alpha.5  Pre-release  (v0.7.0-alpha.5)  about 2 months ago
v0.7.0-alpha.4  Pre-release  (v0.7.0-alpha.4)  about 2 months ago
v0.7.0-alpha.3  Pre-release  (v0.7.0-alpha.3)  about 2 months ago
v0.6.2-alpha.2  Pre-release  (v0.6.2-alpha.2)  about 2 months ago
v0.7.0-alpha.2  Pre-release  (v0.7.0-alpha.2)  about 2 months ago
v0.7.0-alpha.1  Pre-release  (v0.7.0-alpha.1)  about 2 months ago
v0.6.2-alpha.1  Pre-release  (v0.6.2-alpha.1)  about 2 months ago
v0.6.1          Pre-release  (v0.6.1)          about 3 months ago
v0.6.0                       (v0.6.0)          about 3 months ago
v0.6.0-rc.5     Pre-release  (v0.6.0-rc.5)     about 3 months ago
v0.6.0-rc.4     Pre-release  (v0.6.0-rc.4)     about 3 months ago
v0.6.0-rc.3     Pre-release  (v0.6.0-rc.3)     about 3 months ago
v0.6.0-rc.2     Pre-release  (v0.6.0-rc.2)     about 3 months ago
v0.6.0-rc.1     Pre-release  (v0.6.0-rc.1)     about 3 months ago
v0.6.0-beta.2   Pre-release  (v0.6.0-beta.2)   about 3 months ago
v0.6.0-beta.1   Pre-release  (v0.6.0-beta.1)   about 4 months ago
v0.6.0-alpha.8  Pre-release  (v0.6.0-alpha.8)  about 4 months ago
v0.6.0-alpha.7  Pre-release  (v0.6.0-alpha.7)  about 4 months ago
v0.6.0-alpha.6  Pre-release  (v0.6.0-alpha.6)  about 4 months ago
v0.6.0-alpha.5  Pre-release  (v0.6.0-alpha.5)  about 5 months ago
```

This output seems to be optimized for human-readability, but you can also find this kind of output via the API:

```
❯ gh api repos/infinyon/fluvio/releases | jq '.[].tag_name'
"dev"
"v0.8.0"
"v0.7.3"
"v0.7.2"
"v0.7.2-alpha.1"
"v0.7.1"
"v0.7.1-beta.0"
"v0.7.1-alpha.0"
"v0.7.0"
"v0.7.0-beta.1"
"v0.7.0-alpha.5"
"v0.7.0-alpha.4"
"v0.7.0-alpha.3"
"v0.6.2-alpha.2"
"v0.7.0-alpha.2"
"v0.7.0-alpha.1"
"v0.6.2-alpha.1"
"v0.6.1"
"v0.6.0"
"v0.6.0-rc.5"
"v0.6.0-rc.4"
"v0.6.0-rc.3"
"v0.6.0-rc.2"
"v0.6.0-rc.1"
"v0.6.0-beta.2"
"v0.6.0-beta.1"
"v0.6.0-alpha.8"
"v0.6.0-alpha.7"
"v0.6.0-alpha.6"
"v0.6.0-alpha.5"
```

There is more info available depending on what `jq` filter you use

.github/workflows/cross_version_compat.yml
```diff
@@ -0,0 +1,189 @@
...
+            ~/.fluvio/bin/fluvio version
+
+      - name: Install Fluvio for cluster @ (version stable), 
+        if: matrix.cluster_version == 'stable'
+        uses: nick-invision/retry@v2
+        with:
+          timeout_minutes: 5
+          max_attempts: 3
+          command: |
+            curl -fsS https://packages.fluvio.io/v1/install.sh | bash
```
nicholastmosher, line 80: I feel like this is a workaround for a limitation in the `install.sh` script. I feel like a better approach would be to make `install.sh` recognize `VERSION=stable` properly so that this workflow can use the same scripting and just substitute different values as needed.

.github/workflows/cross_version_compat.yml
sehz: There are 2 types of workflow we should support:
(1) After every dev release (should not block CI, just warning)
(2) Run locally, in the developers machine

2nd workflow can be treated as separate issue (sort of smoke test).

.github/workflows/cross_version_compat.yml
tjtelan: Yep, this is a workaround. Modifying the installation script seems ok. I'll open an issue, assuming we don't want the default behavior shouldn't change. `VERSION=stable` will have the same behavior as not setting `VERSION` at all.

sehz:
@tjtelan is still valid?

tjtelan:
Yes, but this PR is out of date. This test case is more about a smoke test for CLI version x Cluster version. The upgrade test doesn't quite test the same thing, but there is a lot of overlap.

I think for that reason alone, I might just close this and open an issue to redesign the test case.

PR #1047, opened 2021-05-03
Use nodeports for local kubernetes environment

### LINKED ISSUE #668: Use `nodePort` rather than `tunnel` for accessing services in the minikube
We use [minikube tunnel](https://minikube.sigs.k8s.io/docs/handbook/accessing/) for exposing SC/SPU services to outside cluster.  It however has unfortunately downside that it's difficult to setup and maintain due to different ways to set up depends on OS.  

Alternative is to use `nodeport` which minikube supports.   It also has benefit of being more portable to other K8 setup such as [kind](https://kind.sigs.k8s.io) or [k3](https://rancher.com/products/k3s/).  

This issue is about experimentation to see if it is possible to migrate to `nodeport`

nicholastmosher:
Closes #668.

- Add `useNodePorts` configuration to Fluvio helm chart
- Discover host_ip from SC pod during installation

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2572,9 +2572,7 @@ dependencies = [
 
 [[package]]
 name = "k8-types"
-version = "0.2.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


sehz:
Can close this?

PR #1051, opened 2021-05-04
[Merged by Bors] - feat: Improved SmartStream API using proc macro

### LINKED ISSUE #1033: Design: Smart Stream WASM procedural macro
Currently writing SmartStream filter requires writing boiler plate code such as below:

```rust
use fluvio_smartstream_wasm as smart_stream;

use smart_stream::dataplane::record::DefaultRecord;
use smart_stream::dataplane::core::{Decoder, Encoder};

extern "C" {
    fn copy_records(putr: i32, len: i32);
}

#[no_mangle]
#[allow(clippy::missing_safety_doc)]
pub unsafe fn filter(ptr: *mut u8, len: usize) -> i32 {
    use std::io::Cursor;

    let input_data = Vec::from_raw_parts(ptr, len, len);
    let mut records: Vec<DefaultRecord> = vec![];

    if let Err(_err) = records.decode(&mut Cursor::new(input_data), 0) {
        return -1;
    };

    // filter only
    let filter_records: Vec<DefaultRecord> = records
        .into_iter()
        .filter(|r| {
            let value = String::from_utf8_lossy(r.value().as_ref());
            value.contains('a')
        })
        .collect();

    // encode back
    let mut out = vec![];
    if let Err(_err) = filter_records.encode(&mut out, 0) {
        return -1;
    }

    let out_len = out.len();

    let ptr = out.as_mut_ptr();
    std::mem::forget(out);

    copy_records(ptr as i32, out_len as i32);

    filter_records.len() as i32
}

```

It should be possible to generate boiler plate code using procedure macro to simplify such:

```rust
#[smart_stream(filter)]
fn filter(records: Vec<DefaultRecord>) => Vec<DefaultRecord> {
    records
        .into_iter()
        .filter(|r| {
            let value = String::from_utf8_lossy(r.value().as_ref());
            value.contains('a')
        })
        .collect();
}
```

nicholastmosher:
Closes #1033.

- Adds a `#[smartstream]` attribute that generates glue code to pass data to and from WASM
- Detects attribute arguments `#[smartstream(filter)]` and `#[smartstream(map)]`
- Adds `SimpleRecord` that may be passed to user functions, while hiding internal DefaultRecord details


src/smartstream/README.md
```diff
@@ -0,0 +1,74 @@
...
+name = "fluvio-wasm-filter"
+version = "0.1.0"
+authors = ["Fluvio Contributors <team@fluvio.io>"]
+edition = "2018"
+
+[lib]
+crate-type = ['cdylib']
+
+[dependencies]
+fluvio-smartstream = { path = "../../", features = ["derive"] }
```
sehz, line 29: can remove `derive` if this is default features

src/smartstream/derive/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "fluvio-smartstream-derive"
+version = "0.1.0"
+authors = ["Nick Mosher <nicholastmosher@gmail.com>"]
```
sehz, line 4: fluvio contributors

sehz: it's also missing other standard attributes

src/smartstream/derive/src/ast.rs
```diff
@@ -0,0 +1,61 @@
...
+                    _ => None,
+                })
+                .next()
+                .ok_or_else(|| SynError::new(args[0].span(), "Missing SmartStream type"))?;
+
+        Ok(ss_type)
+    }
+}
+
+pub struct SmartStreamFn {
```
sehz, line 51: should use lifetime ?

src/smartstream/derive/src/ast.rs
```diff
@@ -0,0 +1,61 @@
...
+    pub name: Ident,
+    pub func: ItemFn,
+}
+
+impl SmartStreamFn {
+    pub fn from_ast(func: ItemFn) -> SynResult<Self> {
+        let name = func.sig.ident.clone();
```
sehz, line 58: no need for clone here

src/smartstream/examples/filter/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+name = "fluvio-wasm-filter"
+version = "0.1.0"
+authors = ["Fluvio Contributors <team@fluvio.io>"]
+edition = "2018"
+
+[lib]
+crate-type = ['cdylib']
+
+[dependencies]
+fluvio-smartstream = { path = "../../", features = ["derive"] }
```
sehz, line 11: remove `derive` with default features

src/smartstream/examples/filter/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+use fluvio_smartstream::{smartstream, SimpleRecord};
+
+#[smartstream(filter)]
+pub fn my_filter(record: &SimpleRecord) -> bool {
+    let value = String::from_utf8_lossy(record.value.as_ref());
```
sehz, line 5: no need for intermediate value

src/smartstream/examples/map/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+name = "fluvio-wasm-map"
+version = "0.1.0"
+authors = ["Fluvio Contributors <team@fluvio.io>"]
+edition = "2018"
+
+[lib]
+crate-type = ['cdylib']
+
+[dependencies]
+fluvio-smartstream = { path = "../../", features = ["derive"] }
```
sehz, line 11: remove derive

src/smartstream/examples/map/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+use fluvio_smartstream::{smartstream, SimpleRecord};
+
+#[smartstream(map)]
+pub fn my_map(mut record: SimpleRecord) -> SimpleRecord {
```
sehz, line 4: hmm.  not sure having mutation in the mapping

src/smartstream/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+    #[allow(clippy::missing_safety_doc)]
+    pub unsafe fn dealloc(ptr: *mut u8, size: usize) {
+        let data = Vec::from_raw_parts(ptr, size, size);
+
+        std::mem::drop(data);
+    }
+}
+
+/// A trait for any type that may construct itself from a DefaultRecord.
+pub trait FromRecord: Sized {
```
sehz, line 37: Why can't use TryFrom?

src/smartstream/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+    type Error: std::error::Error;
+    fn from_record(record: &DefaultRecord) -> Result<Self, Self::Error>;
+}
+
+pub trait IntoRecord {
```
sehz, line 42: Why not From?

src/smartstream/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+pub enum VoidError {}
+impl fmt::Display for VoidError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        writeln!(f, "This error is impossible")
+    }
+}
+impl std::error::Error for VoidError {}
+
+#[derive(Debug)]
+pub struct SimpleRecord {
```
sehz, line 56: Why creating another abstraction instead of using DefaultRecord?

sehz (changes requested):

src/smartstream/examples/Cargo.toml
```diff
@@ -1,6 +1,8 @@
 [workspace]
-members = ["test"]
-
+members = [
+    "filter",
+    "map",
+]
```
sehz, line 6: Please make sure to run all `test_stream_fetch`

smart_filter/Makefile
```diff
@@ -1,5 +0,0 @@
-build_test:
```
sehz, line 1: at least some version of this should be kept.

nicholastmosher: I was able to add a `.cargo/config.toml` file with `[build] target = "wasm32-unknown-unknown"` so that cargo build will behave correctly by default, so no need for makefile

src/smartstream/examples/filter/src/lib.rs
nicholastmosher: Do you mean not using String? Or not assigning 'value' and just doing everything inline?

src/smartstream/examples/map/src/lib.rs
nicholastmosher: This isn't a mutable reference, it is an owned SimpleRecord that we are binding as mutable. So we are given an owned value, we edit it, and then return it. We are not referencing something else and editing something behind the scenes.

src/smartstream/src/lib.rs
nicholastmosher: I was taking inspiration from how actix-web's extractors work https://github.com/actix/actix-web/blob/master/src/extract.rs#L17

I think it would be really cool to eventually support something like this:

```rust
#[derive(Serialize, Deserialize)]
struct MyData {
    name: String,
    number: u32,
}

#[smartstream(filter)]
fn find_bob(data: Json<MyData>) -> bool {
    data.inner.name == "bob"
}
```

But I imagine that in order to do that we will need more control over the conversion traits

src/smartstream/src/lib.rs
nicholastmosher: DefaultRecord is just Record<DefaultAsyncBuffer>, and the Record type also contains data like the preamble, which contains fields that seem to be internal to the protocol and should probably not be edited by users. So SimpleRecord represents all the data from DefaultRecord that we actually want users to be able to interact with

src/smartstream/src/lib.rs
sehz: Yes,  we should support for may conversion but it is out of scope for this issue

nicholastmosher: Right. I'm just saying that is the rationale for why I used a custom trait rather than TryFrom/From

src/smartstream/src/lib.rs
sehz: My previous feedback was that we can certainly hide preamble field which should issue.  We should not create unnecessary abstraction when we don't know what this will be in final form.  

Also, some of the preamble will be needed for time based  filtering.

src/smartstream/derive/src/ast.rs
nicholastmosher: Without the clone this does not compile, because I am storing both func and name in a struct, and taking the ident from the func.sig is a partial move

src/smartstream/derive/src/ast.rs
nicholastmosher: I'm not familiar with all of the conventions. How would you have written this?

src/smartstream/examples/Cargo.toml
nicholastmosher: Is this something that is set up already? Wouldn't we need to make new tests for the smart streams since they will edit the contents?

src/smartstream/derive/src/ast.rs
sehz: fair enough

src/smartstream/derive/src/ast.rs
sehz: pub struct SmartStreamFn<'a> {
    ...
}

see: https://github.com/infinyon/node-bindgen/blob/master/nj-derive/src/ast/class.rs

src/smartstream/examples/Cargo.toml
nicholastmosher: I figured out what you meant, it's all fixed

nicholastmosher:
@sehz this PR is passing required tests and is ready to do. Are there any more blocking issues on this?

sehz:
You didn't address comments why another abstraction is needed instead of using `DefaultRecord`

nicholastmosher:
I suppose it boils down to the fact that `DefaultRecord` is a rather low-level type and very strongly tied to the implementation of inner-workings of the system. I tend to try to be very conservative about exposing types like that publicly because it creates a large risk of introducing breaking changes in the future. By creating a new type `SimpleRecord` with the bare-minimum information needed to make this use-case work, we can absorb the shock of any internal modifications by simply updating the conversion functions between DefaultRecord and SimpleRecord.

sehz:
`DefaultRecord` is not suppose to be low level.  It should be canonical representation.  I would rather fix any issue rather than come up with another abstraction which would be more to maintaint

sehz:
Since there is no consensus on `map`, let's reduce to implementing `filter` first.

nicholastmosher:
@sehz I have rebased this on master with the new Records API, and I have removed the `#[smartstream(map)]` until we can come to a decision on that. Is there anything else you'd like to see changed here or can we merge?

sehz:
Failure in CI

src/smartstream/README.md
```diff
@@ -0,0 +1,74 @@
...
+    let value = String::from_utf8_lossy(record.value.as_ref());
+    value.contains('z')
+}
+```
+
+This filter will keep only records whose contents contain the letter `z`.
+
+### Mapping
+
+Follow the same project setup as the filtering example, except use
```
sehz, line 49: Remove map

sehz (changes requested):
minor nit

src/smartstream/README.md
nicholastmosher: Ah yep

nicholastmosher:
The failure in CI is only the unstable tests

sehz:
build failures: `Run make run-unstable-test
7
make: *** No rule to make target 'build_filter_wasm', needed by 'run-unstable-test'.  Stop.
8
Post Run actions/checkout@v2

Makefile
```diff
@@ -168,7 +168,7 @@ run-all-unit-test: test_smartstreams
 	cargo test -p fluvio-storage $(RELEASE_FLAG) $(TARGET_FLAG)
 	make test-all -C src/protocol	
 
-run-unstable-test:	build_filter_wasm
+run-unstable-test:	test_smartstreams
```
sehz, line 171: don't run smart `smartstream` test for integration test.  Only test with ignore should be run.  This is why there is failure

nicholastmosher: Oh, hm. It just looked like the `build_filter_wasm` was there previously which tested the smart streams so I thought I was just updating this. Doesn't one of the tests require the .wasm file from the old smart_filter crate?

sehz: Yes, smart filter is dependent but you don't need to run tests in those crates

nicholastmosher: Ok, I'll make an independent task for building

sehz:
integration test is failing. Can you run that locally?

Makefile
```diff
@@ -157,17 +157,20 @@ check-clippy: install-clippy
 	cargo +$(RUSTV) check --all --all-targets --all-features --tests $(VERBOSE_FLAG)
 	cargo +$(RUSTV) clippy --all --all-targets --all-features --tests $(VERBOSE_FLAG) -- -D warnings -A clippy::upper_case_acronyms
 
+build_smartstreams:
+	rustup target add wasm32-unknown-unknown
+	pushd src/smartstream/examples; cargo build --release; popd || true
```
sehz, line 162: Don't build everything.  Build specific package as done in previous version:
```
build_test:
	cargo build --release --target wasm32-unknown-unknown --package fluvio-filter-test
```

nicholastmosher: I went ahead and added the `--package` flag, but two things to note:

- Since this task changes directories into `src/smartstream/examples` using pushd, it was already only building the filter, and
- I added a `src/smartstream/examples/.cargo/config.toml` that ensures the examples are always built with `wasm32-unknown-unknown`, which is fine because that examples directory is its own workspace

nicholastmosher:
I ran `make run-unstable-test` locally and everything passed. Only reason CI is failing is due to timeout

nicholastmosher:
Ok at this point it is back to showing the path error since I moved the filter example. Which is weird, because I am running these tests locally and it is working just fine. I added more info to the panic message to hopefully figure out what is going wrong.

Makefile
```diff
@@ -157,17 +157,20 @@ check-clippy: install-clippy
...
+build_smartstreams:
+	rustup target add wasm32-unknown-unknown
+	pushd src/smartstream/examples; cargo build --release --package fluvio-wasm-filter; popd || true
 
-build_filter_wasm:
-	rustup target add wasm32-unknown-unknown 
-	make -C smart_filter build_test
+test_smartstreams:
+	rustup target add wasm32-unknown-unknown
+	pushd src/smartstream/derive; cargo test; popd || true
```
sehz, line 166: prefer use of `make -C` instead of trying to manipulate directory

sehz:
CI fail running locally, you probably had something cached:
```

---- replication::replica_test::test_replication2_new_records_sync stdout ----
thread 'replication::replica_test::test_replication2_new_records_sync' panicked at 'leader: IoError(Os { code: 2, kind: NotFound, message: "No such file or directory" })', src/spu/src/replication/mod.rs:150:18
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- services::public::stream_fetch::test::test_stream_filter_fetch_sync stdout ----
thread 'services::public::stream_fetch::test::test_stream_filter_fetch_sync' panicked at 'Unable to read file /home/ubuntu/fluvio/src/smartstream/examples/target/wasm32-unknown-unknown/release/fluvio_wasm_filter.wasm: Os { code: 2, kind: NotFound, message: "No such file or directory" }', src/spu/src/services/public/stream_fetch.rs:663:29

---- services::public::stream_fetch::test::test_stream_filter_max_sync stdout ----
thread 'services::public::stream_fetch::test::test_stream_filter_max_sync' panicked at 'Unable to read file /home/ubuntu/fluvio/src/smartstream/examples/target/wasm32-unknown-unknown/release/fluvio_wasm_filter.wasm: Os { code: 2, kind: NotFound, message: "No such file or directory" }', src/spu/src/services/public/stream_fetch.rs:663:29

```

sehz (changes requested):

.github/workflows/ci.yml
```diff
@@ -199,7 +199,7 @@ jobs:
           override: true
       - name: Run integration tests
         run:  make run-unstable-test
-        timeout-minutes: 10
+        timeout-minutes: 20
```
sehz, line 202: revert back to 10 with fix of build script

nicholastmosher:
Geez it is just really hard to debug a problem that you can't produce locally 😓

I will try using the inner Makefile but I don't think that is going to help the build failure. Can you try running the tests with the latest branch and tell me what output you see?

sehz:
Here is problem.   This is reason why you should use `make -C` not rely on shell script
```
make build_smartstreams
rustup target add wasm32-unknown-unknown
info: component 'rust-std' for target 'wasm32-unknown-unknown' is up to date
pushd src/smartstream/examples; cargo build --release --package fluvio-wasm-filter; popd || true
/bin/sh: 1: pushd: not found
error: package ID specification `fluvio-wasm-filter` did not match any packages
/bin/sh: 1: popd: not found
```

nicholastmosher:
Huh. TIL that pushd and popd are not universally available. I am putting together a make -C solution

sehz:
classic: https://hairysun.com/why-doesnt-pushd-popd-work-in-my-ubuntu-shell-script.html

sehz (approved):
Great work.  Good to merge

nicholastmosher:
Thanks :)

bors r+

nicholastmosher:
bors -

I just realized that due to the underlying setup of the extern fn for filters, if users write a function named exactly `filter` then it will fail to compile for seemingly no reason. I had meant to try to fix that before this gets merged...

image

nicholastmosher:
Ok, just made a last-minute change to fix the name conflict problem. When `#[smartstream]` expands, it creates a definition of `extern "C" fn filter`, which was causing a conflict if users named their own function `filter`. To avoid this, I put the `extern "C" fn filter` inside a `mod __system {}` block so it is not in the same namespace as the user-defined code. This does not interfere with the extern linking because no_mangle functions ignore namespacing when they are compiled to library symbols.

PR #1055, opened 2021-05-05
[Merged by Bors] - Relax platform version requirement for upgrade check

### LINKED ISSUE #1054: Relax platform version comparision
We have logic in the client library to check whether the cluster it has connected to meets a set minimum version. This is in order to make sure that the components that are communicating have the same protocol, etc. However, this version checking is being too specific, which is causing breakages when we are testing out "latest" builds:

```
Existing platform version is different than chart version platform=0.8.1 chart_version=0.8.1-f78f77f947886b894268f9918a369db1aefe1cd6
```

We should relax the version comparison logic to only check the major.minor.patch version compatibility.

nicholastmosher:
Closes #1054.

- Now compares expected platform version to actual platform version using just `major.minor.patch` version components
- Should prevent upgrade error when using a "latest" platform version including a commit hash

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1916,6 +1916,7 @@ dependencies = [
  "prettytable-rs",
  "proc-macro2",
  "quote",
+ "semver 0.11.0",
  "serde",
  "serde_json",
  "structopt",
--- a/src/cli/src/lib.rs
+++ b/src/cli/src/lib.rs
@@ -143,7 +143,8 @@ impl RootCmd {
                 profile.process(out).await?;
             }
             Self::Cluster(cluster) => {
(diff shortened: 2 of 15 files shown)


sehz (changes requested):
Looks good but I don't see any test associated with it?

nacardin (approved):
This works

sehz:
doc test is failing

sehz (approved):
Looks good for now.  Not sure about testing version checking?

nicholastmosher:
@sehz I extracted the version checking into a function and added some tests 👍

sehz (approved):

PR #1058, opened 2021-05-05
[Merged by Bors] - First pass at making release workflow idempotent

### LINKED ISSUE #1056: Make the release step idempotent
Related to #1003 

This step should fail only if there are infrastructure problems.

We should check for the existence of the version we want to release first, and then perform any of the steps

Updating stable branch will not arrive in this PR, due to the testing required and the urgency of this fix

---

This will also fix an issue encountered from the last release.
The release step does not push a plain version helm chart. It still includes the git hash.

tjtelan:
We now check for the existence of artifacts before doing any pushing or retagging.

Also, we can give the action a git commit which it will use as a base (with respect to the `master` branch). This will allow us to re-run the release job in the event of environmental failures.

We also fix the helm chart naming, which used to include git hash, but should not during Release. 

Resolves #1056

Example runs:
Fail, on account of being unable to push helm chart - https://github.com/infinyon/fluvio-gh-actions-dev/actions/runs/814121338

This is an override to the helm chart check, just to show off the bump stable version step - Don't worry about the bump stable fluvio job not depending on release helm job. I needed to test the idempotency of the bump job.
https://github.com/infinyon/fluvio-gh-actions-dev/actions/runs/813947722

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -23,10 +23,10 @@ jobs:
     name: Setup workflow
     runs-on: ubuntu-latest
     outputs:
-      version: ${{ steps.version_step.outputs.VERSION }}
-      target_sha: ${{ steps.version_step.outputs.GIT_SHA }}
-      latest_tag: ${{ steps.docker_step.outputs.LATEST_TAG }}
-      release_tag: ${{ steps.docker_step.outputs.RELEASE_TAG }}
+      VERSION: ${{ steps.version_step.outputs.VERSION }}
+      TARGET_SHA: ${{ steps.version_step.outputs.GIT_SHA }}
+      LATEST_TAG: ${{ steps.docker_step.outputs.LATEST_TAG }}
+      RELEASE_TAG: ${{ steps.docker_step.outputs.RELEASE_TAG }}
     steps:
       - name: Set target sha and Fluvio version
         id: version_step
@@ -57,7 +57,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions: write-all
     env:
-      version: ${{ needs.setup_job.outputs.version }
+      VERSION: ${{ needs.setup_job.outputs.VERSION }
     steps:
       - name: Login GH CLI
         run: gh auth login --with-token < <(echo ${{ secrets.GITHUB_TOKEN }})
@@ -65,7 +65,7 @@ jobs:
         id: release_check
         continue-on-error: true
         run: |
-          gh release download -R infinyon/fluvio "v${{ env.version }}"
+          gh release download -R infinyon/fluvio "v${{ env.VERSION }}"
       # if the check fails, then continue with release
       - name: Download artifacts from dev
         if: ${{ steps.release_check.outcome == 'failure' }}
@@ -73,7 +73,7 @@ jobs:
       - name: Create GH Release
         if: ${{ steps.release_check.outcome == 'failure' }}
         run: |
-          gh release create -R infinyon/fluvio --title="v${{ env.version }}" -F ./CHANGELOG.md "v${{ env.version }}" \
+          gh release create -R infinyon/fluvio --title="v${{ env.VERSION }}" -F ./CHANGELOG.md "v${{ env.VERSION }}" \
             fluvio-x86_64-unknown-linux-musl \
             fluvio-x86_64-apple-darwin \
             fluvio-run-x86_64-unknown-linux-musl \
@@ -85,15 +85,15 @@ jobs:
     needs: [setup_job]
     runs-on: ubuntu-latest
     env:
-      version: ${{ needs.setup_job.outputs.version }}
-      release_tag: ${{ needs.setup_job.outputs.release_tag }}
-      latest_tag: ${{ needs.setup_job.outputs.latest_tag }}
+      VERSION: ${{ needs.setup_job.outputs.VERSION }}
+      RELEASE_TAG: ${{ needs.setup_job.outputs.RELEASE_TAG }}
+      LATEST_TAG: ${{ needs.setup_job.outputs.LATEST_TAG }}
     steps:
       - name: Attempt to pull image tag in docker registry
         id: docker_check
         continue-on-error: true
         run: |
-          if docker pull ${{ env.release_tag }}; then
+          if docker pull ${{ env.RELEASE_TAG }}; then
             if [[ -z ${FORCE_RELEASE} ]]; then
               echo "Image tag already exists"
             else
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -5,82 +5,175 @@ permissions:
...
+      target_sha: ${{ steps.version_step.outputs.GIT_SHA }}
+      latest_tag: ${{ steps.docker_step.outputs.LATEST_TAG }}
+      release_tag: ${{ steps.docker_step.outputs.RELEASE_TAG }}
+    steps:
+      - name: Set target sha and Fluvio version
+        id: version_step
+        run: |
+          if [[ -z "${USE_COMMIT}" ]]; then
+            echo "VERSION=$(curl -fsS https://raw.githubusercontent.com/infinyon/fluvio/${{ github.sha }}/VERSION)" >> $GITHUB_ENV
+            echo "::set-output name=VERSION::$(curl -fsS https://raw.githubusercontent.com/infinyon/fluvio/${{ github.sha }}/VERSION)"
```
nicholastmosher, line 36: You can probably avoid two curls here by saving to a variable in this block

.github/workflows/release.yml
```diff
@@ -5,82 +5,175 @@ permissions:
...
+        id: docker_step
+        run: |
+            echo "::set-output name=LATEST_TAG::infinyon/fluvio:${{ env.VERSION }}-${{ env.GIT_SHA }}"
+            echo "::set-output name=RELEASE_TAG::infinyon/fluvio:${{ env.VERSION }}"
+
+  # Check for Github Release
   release_github:
-    name: Create GitHub Release
+    name: Release Fluvio to GitHub Release
+    needs: [setup_job]
```
nicholastmosher, line 54: Are you sure that environment settings are passed from one job to another? I'm pretty sure it works that way for _steps_, but I have not yet tried this with a job.needs

sehz (approved):

.github/workflows/release.yml
tjtelan: Yeah, I'm sure after some trial and error. I linked some jobs from my dev repo to test it out.

I believe I found that exported vars don't pass from job to job. However, you can use `set-output` to pass data between jobs.

.github/workflows/release.yml
```diff
@@ -119,13 +215,28 @@ jobs:
...
       # Enable this when we are confident in the workflow
-#      - name: Bump stable branch
-#        run: |
-#          git checkout stable
-#          git rebase origin/master
-#          git push origin stable
+      - name: Bump stable branch
+        run: |
+          git fetch
+          CURRENT_STABLE=$(git show-ref | grep refs/remotes/origin/stable | awk '{print $1}')
```
nicholastmosher, line 229: Could this be simplified to `git rev-parse origin/stable`?

.github/workflows/release.yml
```diff
@@ -5,82 +5,177 @@ permissions:
...
+      version: ${{ steps.version_step.outputs.VERSION }}
+      target_sha: ${{ steps.version_step.outputs.GIT_SHA }}
+      latest_tag: ${{ steps.docker_step.outputs.LATEST_TAG }}
+      release_tag: ${{ steps.docker_step.outputs.RELEASE_TAG }}
+    steps:
+      - name: Set target sha and Fluvio version
+        id: version_step
+        run: |
+          if [[ -z "${USE_COMMIT}" ]]; then
+            GITHUB_VERSION=$(curl -fsS https://raw.githubusercontent.com/infinyon/fluvio/${{ github.sha }}/VERSION)
```
nicholastmosher, line 35: I don't think this will work unless you use `export GITHUB_VERION=...`

.github/workflows/release.yml
```diff
@@ -5,82 +5,177 @@ permissions:
...
+            echo "VERSION=${GITHUB_VERSION}" >> $GITHUB_ENV
```
nicholastmosher, line 36: I've started to prefer the style `echo "VAR=value" | tee -a $GITHUB_ENV` so that the value being set is printed to stdout and you can verify it in the logs

.github/workflows/release.yml
```diff
@@ -5,82 +5,177 @@ permissions:
...
+      # if the check fails, then continue with release
       - name: Download artifacts from dev
+        if: ${{ steps.release_check.outcome == 'failure' }}
         run: gh release download -R infinyon/fluvio dev
       - name: Create GH Release
+        if: ${{ steps.release_check.outcome == 'failure' }}
         run: |
-          export VERSION="$(cat VERSION)"
-          gh release create -R infinyon/fluvio --title="v${VERSION}" -F ./CHANGELOG.md "v${VERSION}" \
+          gh release create -R infinyon/fluvio --title="v${{ needs.setup_job.outputs.version }}" -F ./CHANGELOG.md "v${{ needs.setup_job.outputs.version }}" \
```
nicholastmosher, line 74: Can you assign this to a variable once and use that variable? e.g.

```
export VERSION="${{ needs.setup_job.outputs.version }}"
...
```

.github/workflows/release.yml
tjtelan: I don't need the var to leave the job, this works for reusing in the if/else context. I'm using the `set-output` to export to other jobs.. This does work: https://github.com/infinyon/fluvio-gh-actions-dev/runs/2507309943?check_suite_focus=true

.github/workflows/release.yml
tjtelan: That's a good idea to clean up. I carried the pattern over to all the separate jobs.

.github/workflows/release.yml
nicholastmosher: Right, but even just in regular bash scripting, a variable does not typically persist longer than a single command, right? Like if I did `RUST_LOG=debug fluvio version; fluvio version` only the first one would have the env variable. I'd need to do `export RUST_LOG=debug; fluvio version; fluvio version` in order for it to span multiple commands

.github/workflows/release.yml
```diff
@@ -56,22 +56,24 @@ jobs:
     needs: [setup_job]
     runs-on: ubuntu-latest
     permissions: write-all
+    env:
+      version: ${{ needs.setup_job.outputs.version }
```
nicholastmosher, line 60: Can you use caps for env variables? VERSION

.github/workflows/release.yml
```diff
@@ -82,12 +84,16 @@ jobs:
     name: Release Docker Image
     needs: [setup_job]
     runs-on: ubuntu-latest
+    env:
+      version: ${{ needs.setup_job.outputs.version }}
+      release_tag: ${{ needs.setup_job.outputs.release_tag }}
+      latest_tag: ${{ needs.setup_job.outputs.latest_tag }}
```
nicholastmosher, line 90: can you capitalize these?

.github/workflows/release.yml
tjtelan: That's correct. I guess the reason it doesn't matter in this case is that I'm not spawning a child process where I need these vars to be visible. You can try this in your shell too.

```
$ UNEXPORTED="true"
$ echo "Can I see the var? ${UNEXPORTED}"
Can I see the var? true
```

tjtelan: I just went ahead and added the export bc it doesn't hurt, nor does it change the behavior.

PR #1073, opened 2021-05-11
[Merged by Bors] - Stabilize upgrade testing in CI

### LINKED ISSUE #1066: Upgrade Test Failure
https://github.com/infinyon/fluvio/runs/2529081045?check_suite_focus=true

tjtelan:
Fixes #1066

We've found more stability in CI by slowing down calls to the `fluvio` cluster. (The test output prints cpu load avgs, which are frequently overscheduled.)

Upgrade test is driven by shell script `test/upgrade-test.sh`. 

---

We've found more stability in CI by slowing down calls to the `fluvio` cluster. (The test output prints cpu load avgs, which are frequently overscheduled.)

Test data and checksums are generated prior to cluster installation.

We test 3 versions: Last stable, current stable and then we compile and install the repo version.

---

For each version, we create a new topic, produce and consume then verify content w/ checksum. Then we verify functionality old topics after upgrade with a consume, produce, consume + checksum verification.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -463,15 +463,39 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [infinyon-ubuntu-bionic]
+        os: [ubuntu-latest]
         rust: [stable]
+        include:
+          - os: ubuntu-latest
+            sccache-path: /home/runner/.cache/sccache
+            target: x86_64-unknown-linux-musl
     env:
       FLV_SOCKET_WAIT: 600
       FLV_CLUSTER_MAX_SC_VERSION_LOOP: 120
       FLV_CLUSTER_MAX_SC_NETWORK_LOOP: 60
       FLV_TEST_CONSUMER_WAIT: 300000
+      RUST_BACKTRACE: full
+      RUSTC_WRAPPER: sccache
+      RUSTV: ${{ matrix.rust }}
+      TARGET: ${{ matrix.target }}
+      SCCACHE_CACHE_SIZE: 2G
+      SCCACHE_DIR: ${{ matrix.sccache-path }}
+      # SCCACHE_RECACHE: 1 # Uncomment this to clear cache, then comment it back out
     steps:
       - uses: actions/checkout@v2
+
+      - name: Install sccache (ubuntu-latest)
+        if: matrix.os == 'ubuntu-latest'
+        env:
+          LINK: https://github.com/mozilla/sccache/releases/download
+          SCCACHE_VERSION: 0.2.13
+        run: |
+          SCCACHE_FILE=sccache-$SCCACHE_VERSION-x86_64-unknown-linux-musl
+          mkdir -p $HOME/.local/bin
+          curl -L "$LINK/$SCCACHE_VERSION/$SCCACHE_FILE.tar.gz" | tar xz
+          mv -f $SCCACHE_FILE/sccache $HOME/.local/bin/sccache
+          echo "$HOME/.local/bin" | tee -a $GITHUB_PATH
+
       - run: helm version
       - name: Install Rust ${{ matrix.rust }}
         uses: actions-rs/toolchain@v1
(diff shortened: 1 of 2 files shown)


sehz:
@nicholastmosher Please review this as well

.github/workflows/ci.yml
```diff
@@ -463,22 +463,85 @@ jobs:
...
+          path: ${{ matrix.sccache-path }}
+          key: ${{ runner.os }}-sccache-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-sccache-
+
+      - name: Install Minikube for Github runner 
+        if: matrix.os == 'ubuntu-latest' 
+        uses: manusa/actions-setup-minikube@v2.3.1
+        with:
+          minikube version: 'v1.16.0'
```
sehz, line 537: try latest `1.18.1`

.github/workflows/ci.yml
```diff
@@ -463,22 +463,85 @@ jobs:
...
+          kubernetes version: 'v1.19.2'
```
sehz, line 538: we are using 1.19.6

sehz (changes requested):
Looks good.  Couple of minor comments

.github/workflows/ci.yml
```diff
@@ -490,57 +553,31 @@ jobs:
...
+        env:
+          TEST_DATA_BYTES: 10000
+        uses: nick-invision/retry@v2
+        with:
+          timeout_minutes: 45
+          max_attempts: 3
+          command: |
+            export PATH=~/.fluvio/bin:$PATH
+            cd tests
+            ./upgrade-test.sh `gh release -R infinyon/fluvio list | head -3 | tail -2 | awk '{print $1}' | sed 's/v//' | tac | xargs`
```
nicholastmosher, line 575: This can probably be simplified to

```
gh api repos/infinyon/fluvio/releases -q '.[2,1].tag_name' | sed 's/v//'
0.8.1
0.8.2
```

nicholastmosher: Also instead of using backticks for inline commands it is better practice to use `$(...)`

nicholastmosher:
It seems like a lot of this upgrade-test script can be abstracted into bash functions, I'd maybe try to make a function for running tests on a topic, where the topic is given as an argument, for example. I also definitely recommend reading through https://kfirlavi.herokuapp.com/blog/2012/11/14/defensive-bash-programming/ and using some tips from there

.github/workflows/ci.yml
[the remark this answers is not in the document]
tjtelan: Added a comment. The purpose of the one-liner is to get the stable and stable-1 versions. And we pass them into the script to configure the CLI and cluster versions we are going to test.

tests/upgrade-test.sh
```diff
@@ -0,0 +1,262 @@
...
+# We create new topic per round.
+# We'll produce and consume from all test topics per round.
+# Content is verified via checksum
+
+set -exu
+#set -e
+
+echo command: $0 $*
+
+STABLE_MINUS_ONE=${1:?Starting cluster version [pos 1]}
```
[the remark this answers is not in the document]
tjtelan: The purpose was just to have some error thrown when the input validation (`:?`) fails. If no version is passed for argv[1] (or it is null) then we halt execution and throw the error. I changed the message to be a bit more descriptive

https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02

tjtelan:
> It seems like a lot of this upgrade-test script can be abstracted into bash functions, I'd maybe try to make a function for running tests on a topic, where the topic is given as an argument, for example. I also definitely recommend reading through https://kfirlavi.herokuapp.com/blog/2012/11/14/defensive-bash-programming/ and using some tips from there

Definitely hear you, but I'm not sure what is gained out of extra abstraction in this specific internal use-case. The suggestion is much bigger than the problem being solved.

Checksums for each new write to a topic are calculated w/r/t the number of versions we plan on testing, and refactoring into a more stateful or functional style is not going to be trivial in Bash, (or easier to read).

In my opinion, flattening the cli x cluster convolution seems worth the effort only if we plan to test more than 3 versions, like we support in this test.

sehz:
I think this PR is ready to go.  @nicholastmosher agree?

nicholastmosher (approved):
LGTM

sehz:
@nicholastmosher @tjtelan Once this is merged. This should be put as mandatory for Bors?

nicholastmosher:
@sehz I think it eventually should be, but let's maybe give this a week or two to run as non-required so that we can be sure that things are working as expected, then we can promote it to required

PR #1074, opened 2021-05-11
[Merged by Bors] - feat: Add batch mode for CLI producer

### LINKED ISSUE #915: Update CLI Producer to send records in batches
Since we now have a batch producer client API, we should leverage it in the CLI producer in cases where we are reading directly from files and we have an estimated line count. We should not use it when reading from stdin or piped files, since we do not know when they will end and may not have a chance to flush records before closing.

This adds a new flag to the `fluvio produce` command: `--batch-size=X`, which will allow the user to specify the number of records that should be sent in a single request. The default batch size is 50.

nicholastmosher:
Closes #915 

- When using `fluvio produce topic -f <file>`, the producer will send lines in batches of 50 by defualt
- The `fluvio produce topic --batch-size=X` argument can be used to adjust the size of the batch

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,17 +1,21 @@
 # Release Notes
+
 ## Unreleased
-* Added builder for fluvio_storage::config::ConfigOption. (#1076)[https://github.com/infinyon/fluvio/pull/1076]
+* Added builder for fluvio_storage::config::ConfigOption. ([#1076](https://github.com/infinyon/fluvio/pull/1076))
+* Use batch record sending in CLI producer ([#915](https://github.com/infinyon/fluvio/issues/915))
 
 ## Platform Version 0.8.2 - 2020-05-06
 * Fix Replication fail over with duplication ([#1052](https://github.com/infinyon/fluvio/pull/1052))
 * Relax platform version requirement for upgrade check ([#1055](https://github.com/infinyon/fluvio/pull/1055))
 * Update logic for finding latest package release ([#1061](https://github.com/infinyon/fluvio/pull/1061))
+
 ## Platform Version 0.8.1 - 2020-05-03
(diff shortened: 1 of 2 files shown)


nicholastmosher:
@sehz how does this look?

src/cli/src/consumer/produce/mod.rs
```diff
@@ -56,12 +56,19 @@ impl ProduceOpt {
 
         match &self.file {
             Some(path) => {
-                let mut reader = BufReader::new(File::open(path)?);
-                self.produce_lines(&mut producer, &mut reader).await?;
+                let reader = BufReader::new(File::open(path)?);
+                let lines: Vec<_> = reader.lines().filter_map(|it| it.ok()).collect();
+                let batches: Vec<_> = lines.chunks(50).collect();
```
sehz, line 61: I think it is possible to combine `lines`,`batches` into a single iterator

sehz (changes requested):
LGTM.  Couple of nits

src/cli/src/consumer/produce/mod.rs
nicholastmosher: I believe that `chunks` is a method on slices, not on iterators. So I needed to collect into a Vec first to make this work

sehz (changes requested):
add to change log

CHANGELOG.md
```diff
@@ -1,15 +1,20 @@
 # Release Notes
 
+## Platform Version 0.8.3 - 2020-05-13
```
sehz, line 3: Need to rebase to get latest changelog.

nicholastmosher: Ah, got it

sehz (approved):

Issue #1075, opened 2021-05-11
ApiVersion response starts with 4 extra bytes


nacardin:
Example code https://github.com/nacardin/fluvio/blob/20818f8f502473c80e01349008c1115c40bebd76/src/spu/src/liveness_check.rs

simlay:
To reproduce part of this bug, append this to the end of [`dataplane-protocolsrc/versions.rs`](https://github.com/infinyon/fluvio/blob/master/src/dataplane-protocol/src/versions.rs)
```rust
#[cfg(test)]
mod broke_decoding {

    use super::*;
    use fluvio_protocol::api::ResponseMessage;
    use std::io::Cursor;
    #[test]
    fn test_response_decoding() {
        let input_msg = ResponseMessage {
            correlation_id: 1,
            response: ApiVersionsResponse {
                error_code: ErrorCode::None,
                api_keys: vec![
                    ApiVersionKey {
                        api_key: 1001,
                        min_version: 0,
                        max_version: 0,
                    },
                    ApiVersionKey {
                        api_key: 1002,
                        min_version: 1,
                        max_version: 1,
                    },
                    ApiVersionKey {
                        api_key: 1003,
                        min_version: 0,
                        max_version: 0,
                    },
                    ApiVersionKey {
                        api_key: 1004,
                        min_version: 0,
                        max_version: 0,
                    },
                    ],
                    platform_version: PlatformVersion(
                        String::from("0.8.2"),
                    ),
            },
        };
        let bytes = input_msg.as_bytes(0);
        assert!(bytes.is_ok());
        let bytes = bytes.unwrap();
        println!("ENCODED BYTES ARE: {:?}", bytes.to_vec());

        // Logs from sc
        let bytes_from_sc_logs : Vec<u8> = vec![0, 0, 0, 41, 0, 0, 0, 1, 0, 0, 0, 0, 0, 4, 3, 233, 0, 0, 0, 0, 3, 234, 0, 1, 0, 1, 3, 235, 0, 0, 0, 0, 3, 236, 0, 0, 0, 0, 0, 5, 48, 46, 56, 46, 50];
        assert_eq!(bytes, bytes_from_sc_logs);

        // logs from echo example on host
[...]

PR #1081, opened 2021-05-13
[Merged by Bors] - feat: generic implementation of tokio_util::codec::Encoder for FluvioEncoder

morenol:
Fixes https://github.com/infinyon/fluvio/issues/1075

* Changes the implementation of ResponseMessage and RequestMessage encoder to not include their lengths so the encode-decoder is symmetric now.
* 
* Added to `FluvioCodec` the capability to handle `FluvioEncoder` types. 

* Add implementations of FluvioEncoder to &T: FluvioEncoder.

### DIFF
--- a/src/protocol/Cargo.toml
+++ b/src/protocol/Cargo.toml
@@ -17,8 +17,8 @@ store = ["fluvio-future", "fluvio-protocol-api", "log", "bytes"]
 [dependencies]
 fluvio-protocol-core = { version = "0.3.0", path = "fluvio-protocol-core" }
 fluvio-protocol-derive = { version = "0.2.0", path = "fluvio-protocol-derive", optional = true }
-fluvio-protocol-api = { version = "0.3.0", path = "fluvio-protocol-api", optional = true }
-fluvio-protocol-codec = { version = "0.3.0", path = "fluvio-protocol-codec", optional = true }
+fluvio-protocol-api = { version = "0.4.0", path = "fluvio-protocol-api", optional = true }
+fluvio-protocol-codec = { version = "0.3.1", path = "fluvio-protocol-codec", optional = true }
 fluvio-future = { version = "0.3.0", optional = true }
 log = { version = "0.4.8", optional = true }
 bytes = { version = "1.0.0", optional = true }
@@ -27,6 +27,6 @@ bytes = { version = "1.0.0", optional = true }
 flv-util = { version = "0.5.2" }
 fluvio-protocol-derive = { version = "0.2.0", path = "fluvio-protocol-derive" }
 fluvio-protocol-core = { version = "0.3.0", path = "fluvio-protocol-core" }
-fluvio-protocol-api = { version = "0.3.0", path = "fluvio-protocol-api" }
+fluvio-protocol-api = { version = "0.4.0", path = "fluvio-protocol-api" }
 log = { version = "0.4.8" }
 fluvio-future = { version = "0.3.0", features = ["subscriber"]}
--- a/src/protocol/Cargo.toml
+++ b/src/protocol/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "fluvio-protocol"
 edition = "2018"
-version = "0.5.0"
+version = "0.5.1"
 authors = ["Fluvio Contributors <team@fluvio.io>"]
 description = "Fluvio streaming protocol"
 repository = "https://github.com/infinyon/fluvio-protocol"
--- a/src/protocol/fluvio-protocol-api/src/api.rs
+++ b/src/protocol/fluvio-protocol-api/src/api.rs
@@ -1,5 +1,6 @@
 use std::default::Default;
 use std::io::Error as IoError;
+use std::io::ErrorKind;
 use std::fs::File;
 use std::io::Cursor;
 use std::path::Path;
@@ -9,6 +10,7 @@ use std::fmt;
 use std::convert::TryFrom;
 
 use log::debug;
+use log::trace;
 
 use crate::core::Decoder;
 use crate::core::Encoder;
@@ -52,6 +54,17 @@ pub trait ApiMessage: Sized + Default {
 
         let data = buffer.to_vec();
         let mut src = Cursor::new(&data);
+
+        let mut size: i32 = 0;
+        size.decode(&mut src, 0)?;
+        trace!("decoded request size: {} bytes", size);
+
+        if src.remaining() < size as usize {
+            return Err(IoError::new(
+                ErrorKind::UnexpectedEof,
+                "not enought bytes for request message",
+            ));
+        }
 
         Self::decode_from(&mut src)
     }
--- a/src/protocol/fluvio-protocol-api/src/api.rs
+++ b/src/protocol/fluvio-protocol-api/src/api.rs
@@ -1,6 +1,5 @@
 use std::default::Default;
 use std::io::Error as IoError;
-use std::io::ErrorKind;
 use std::fs::File;
 use std::io::Cursor;
 use std::path::Path;
@@ -9,7 +8,6 @@ use std::fmt::Debug;
 use std::fmt;
 use std::convert::TryFrom;
 
-use log::trace;
 use log::debug;
 
 use crate::core::Decoder;
@@ -54,17 +52,6 @@ pub trait ApiMessage: Sized + Default {
 
         let data = buffer.to_vec();
         let mut src = Cursor::new(&data);
-
-        let mut size: i32 = 0;
-        size.decode(&mut src, 0)?;
-        trace!("decoded request size: {} bytes", size);
-
-        if src.remaining() < size as usize {
-            return Err(IoError::new(
-                ErrorKind::UnexpectedEof,
-                "not enought bytes for request message",
-            ));
-        }
 
         Self::decode_from(&mut src)
     }
--- a/src/protocol/fluvio-protocol-api/src/response.rs
+++ b/src/protocol/fluvio-protocol-api/src/response.rs
@@ -1,6 +1,7 @@
 use std::fs::File;
 use std::io::Cursor;
 use std::io::Error as IoError;
+use std::io::ErrorKind;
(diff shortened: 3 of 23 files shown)


src/protocol/fluvio-protocol-api/src/response.rs
```diff
@@ -68,17 +67,6 @@ where
 
         let mut src = Cursor::new(&data);
 
-        let mut size: i32 = 0;
```
morenol, line 71: The test is passing with and without this. I will add a test for this, but first I have to check where is done the encode_to_file or something like that to check if it includes the size or not.

src/protocol/fluvio-protocol-core/src/encoder.rs
```diff
@@ -33,6 +33,20 @@ pub trait Encoder {
         buf.put_slice(&out);
         Ok(buf.freeze())
     }
+
+    /// Return a [`bytes::Bytes`] representation of the object that follows the
+    /// [Kafka Protocol](https://kafka.apache.org/protocol#The_Messages_ListOffsets).
+    /// First 4 bytes are size of the message.  Then total buffer = 4 + message content
+    fn as_kafka_frame_bytes(&self, version: Version) -> Result<Bytes, Error> {
```
morenol, line 40: Maybe as_kafka_message is a better name?

sehz:
@morenol is this ready for review then?

morenol:
> @morenol is this ready for review then?

@sehz  Yes, I only have a couple of questions in my comments but I think that it is ready for review

src/protocol/fluvio-protocol-codec/src/codec.rs
```diff
@@ -101,110 +102,146 @@ mod test {
...
+                // over raw data, the length has to be send.
+                let mut len_buf = vec![];
+                let len = buf.len() as i32;
+                len.encode(&mut len_buf, 0).expect("encoding");
+                sink.send(Bytes::from(len_buf)).await.expect("sending");
+
+                // split buf into two segments, decode should reassembly them
+                let buf2 = buf.split_off(5);
+                sink.send(Bytes::from(buf)).await.expect("sending");
+                fluvio_future::timer::sleep(time::Duration::from_millis(10)).await;
```
sehz, line 180: this is not quite right.  this should test partial codec delivery.  so can't user sink here but use tcp_stream directly

src/protocol/fluvio-protocol-codec/src/codec.rs
```diff
@@ -101,110 +102,146 @@ mod test {
...
+                sink.send(Bytes::from(buf2)).await.expect("sending");
+            }
+        }
+        fluvio_future::timer::sleep(time::Duration::from_millis(50)).await;
+        debug!("finishing. terminating server");
+        Ok(())
+    }
+
+    async fn run_client(data: Vec<u8>, addr: &SocketAddr) -> Result<(), Error> {
```
sehz, line 189: add test for non bytes sending thru codec.  for example such as string, i64

src/protocol/fluvio-protocol-core/src/encoder.rs
sehz: we are deprecating `kafka`.  change to `fluvio_frame_bytes`.   don't think this should belong here but in the codec

src/protocol/fluvio-protocol-core/src/encoder.rs
```diff
@@ -147,6 +161,32 @@ where
     }
 }
 
+/// Ignore encoding for raw bytes.
+impl Encoder for Bytes {
```
sehz, line 165: there should not be special treatment for bytes

src/socket/src/sink.rs
```diff
@@ -71,7 +71,7 @@ impl FluvioSink {
     where
         RequestMessage<R>: FlvEncoder + Default + Debug,
     {
-        let bytes = req_msg.as_bytes(0)?;
+        let bytes = req_msg.as_kafka_frame_bytes(0)?;
```
sehz, line 74: since codec understand, Fluvio encoder they can be sent directly

sehz (changes requested):
Great progress! Few changes requires.
Also, should update bump up creates version for codec, add to changelog

sehz:
@simlay Can you test if your websocket works with this?  You should be able to remove work around for 4 bytes issue

src/protocol/fluvio-protocol-api/src/response.rs
sehz: it should work as is since it is not dealing with transport

src/protocol/fluvio-protocol-core/src/encoder.rs
morenol: This was added just in order to allow the sending of raw data (used for partial sends). For that case `fluvio_frame_bytes` should not append the length at the beginning.

src/protocol/fluvio-protocol-api/src/response.rs
morenol: My main doubt is because in fluvio/src/protocol/src/store.rs, it looks like file_encode() adds the length first. But I don't know if that is related to this. If that is true, this change should be removed and also the change in api.rs and request.rs related to this.

src/socket/src/sink.rs
```diff
@@ -71,28 +71,22 @@ impl FluvioSink {
...
+        (&mut self.inner).send(req_msg).await?;
         Ok(())
     }
 
     /// as server, send back response
     pub async fn send_response<P>(
         &mut self,
         resp_msg: &ResponseMessage<P>,
-        version: Version,
+        _version: Version,
```
morenol, line 83: This is no longer used with this change. Is that a problem?

sehz: Version need to be used during encoding and decoding

sehz:
This looks great! @simlay This is ready. Can you test your websocket changes with this?

However, there is problem with codec's version change since it is exported by wrapper crate.   There are 2 ways we can solve this:
1) Keep old version of FluvioCodec until it is deprecated.  
2) Use similar trick here:  https://github.com/dtolnay/semver-trick.

sehz:
Nice!  In that case, you don't have to bump up major version for codec.  Just change to `0.3.1`

src/protocol/fluvio-protocol-api/src/api.rs
```diff
@@ -9,7 +8,6 @@ use std::fmt::Debug;
 use std::fmt;
 use std::convert::TryFrom;
 
-use log::trace;
 use log::debug;
```
sehz, line 12: this need version bump as well since behavior is changed.  not sure how to keep compatible with previous version.  Idea?

sehz (changes requested):

src/protocol/fluvio-protocol-api/src/api.rs
morenol: I made the bump. Not sure, now the encoding-decoding is symmetric for response and request in any case I think that it is a breaking change.

sehz: Ok.  Based on [trick](https://github.com/dtolnay/semver-trick), let's try this:
* bump up `fluivo-protocol` minor version.
* add fluvio-protocol-api v0.3.0 as dependency. 

in the `fluvio-protocol/src/lib`: change:

```
#[cfg(feature = "api")]
pub mod api {
    pub use fluvio_protocol_api::*;
}

```

to 
```
#[cfg(feature = "api")]
pub mod api {
    pub use fluvio_protocol_api_v2::*;
}
#[cfg(feature = "api")]
pub mod api_v2 {
    pub use fluvio_protocol_api::*;
}
```

where in the `cargo.toml`
```
...
fluvio-protocol-api = { version = "0.3.0", path = "fluvio-protocol-api", optional = true }
fluvio-protocol-api_v2 = { version = "0.2.0",  optional = true, package = "fluvio-protocol-api }
```

And update fluvio_socket to use v2

src/socket/src/sink.rs
```diff
@@ -9,7 +9,7 @@ use async_mutex::Mutex;
 use async_mutex::MutexGuard;
```
sehz, line 9: minor version update for for this crate

simlay: I think we shouldn't do that in this PR to avoid a merge conflict of mine. I can't find a specific announcement bot the [`async-mutex`](https://crates.io/crates/async-mutex) github points to [`async-lock`](https://crates.io/crates/async-lock). I know because the wasm client needs the latest a more recent version of `async-lock` and results in https://github.com/rust-lang/rust/issues/48564 which was fixed in various parts and added to

sehz: @simlay  How about if I update to use async-mutex to async-lock as first step?

simlay: Sure. I'll do that.

simlay: Done in https://github.com/infinyon/fluvio/pull/1086

sehz:
can you rebase from master to changes by @simlay ?

sehz:
@simlay can you try now?

src/protocol/Cargo.toml
```diff
@@ -17,8 +17,8 @@ store = ["fluvio-future", "fluvio-protocol-api", "log", "bytes"]
 [dependencies]
```
sehz, line 17: can you bump up this to 0.5.1 and ensure other dependents crate to this version

src/socket/src/sink.rs
```diff
@@ -9,7 +9,7 @@ use async_lock::Mutex;
 use async_lock::MutexGuard;
```
sehz, line 9: bump this crate to `0.8..1` and ensure all dependent create to use this version?

sehz (changes requested):

sehz:
go thru all users of  `fluvio-protocol` and `fluvio-socket` to ensure they are using latest version

sehz (approved):
Looks great!  Excellent work

PR #1086, opened 2021-05-14
[Merged by Bors] - Updated async-mutex to async-lock

simlay:
This is a subtask of #989.

### DIFF
--- a/src/client/Cargo.toml
+++ b/src/client/Cargo.toml
@@ -30,7 +30,7 @@ serde = { version = "1.0.110", features = ['derive'] }
 serde_json = "1.0.53"
 async-channel = "1.1.0"
 event-listener = "2.5.1"
-async-mutex = "1.2.0"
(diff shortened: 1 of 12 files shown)


src/client/Cargo.toml
```diff
@@ -30,7 +30,7 @@ serde = { version = "1.0.110", features = ['derive'] }
 serde_json = "1.0.53"
```
sehz, line 30: bump up version for client

src/socket/Cargo.toml
```diff
@@ -20,7 +20,7 @@ cfg-if = "1.0.0"
 bytes = "1.0.0"
 once_cell = "1.4.1"
```
sehz, line 21: bump up version for socket

sehz (changes requested):
minor nits

sehz (approved):

PR #1088, opened 2021-05-15
[Merged by Bors] - Clean up dataplane::record module with cleaner types and interfaces

nicholastmosher:
This PR will help our Record type and its helper types be much more user-friendly and allow us to stop wrapping the dataplane::Record API with other types in order to make the user-facing API look nicer. Here's a breakdown of changes:

- Remove `DefaultRecord` and `DefaultBatch` in favor of using default generics, e.g. `Record<B = RecordData>`
- Push down the `RecordKey` and `RecordData` types from the client producer module into dataplane
- Replace `DefaultAsyncBuffer` with `RecordData`
- Re-export `Record`, `RecordKey`, and `RecordData` in client's `producer` module (therefore no breaking changes in client)
- Massage conversion implementations for `Record`, `RecordKey` and `RecordData` in order to make the types more user-friendly

These changes were made specifically in order to support a nicer SmartStream API for authors of new SmartStreams. This will allow us to expose these dataplane types directly and have them be much more friendly and usable, rather than needing to wrap the lower-level API with sugar to present to users, like SimpleRecord was used for before.

### DIFF
--- a/src/dataplane-protocol/src/batch.rs
+++ b/src/dataplane-protocol/src/batch.rs
@@ -15,9 +15,14 @@ use crate::derive::Encode;
 
 use crate::Offset;
 use crate::Size;
-use crate::record::DefaultRecord;
+use crate::record::Record;
 
-pub type DefaultBatchRecords = Vec<DefaultRecord>;
+#[allow(deprecated)]
+#[deprecated(since = "0.6.0", note = "Use 'VecBatch' instead")]
+pub type DefaultBatchRecords = Vec<crate::record::DefaultRecord>;
+
+#[allow(deprecated)]
+#[deprecated(since = "0.6.0", note = "Use 'Batch' instead")]
 pub type DefaultBatch = Batch<DefaultBatchRecords>;
 
 pub trait BatchRecords: Default + Debug + Encoder + Decoder {
@@ -27,7 +32,9 @@ pub trait BatchRecords: Default + Debug + Encoder + Decoder {
     }
 }
 
-impl BatchRecords for DefaultBatchRecords {}
+pub type MemoryBatch = Vec<Record>;
+
+impl BatchRecords for MemoryBatch {}
 
 /// size of the offset and length
 pub const BATCH_PREAMBLE_SIZE: usize = size_of::<Offset>()     // Offset
@@ -36,10 +43,7 @@ pub const BATCH_PREAMBLE_SIZE: usize = size_of::<Offset>()     // Offset
 pub const BATCH_FILE_HEADER_SIZE: usize = BATCH_PREAMBLE_SIZE + BATCH_HEADER_SIZE;
 
 #[derive(Default, Debug)]
-pub struct Batch<R>
-where
-    R: BatchRecords,
-{
+pub struct Batch<R = MemoryBatch> {
     pub base_offset: Offset,
     pub batch_len: i32, // only for decoding
     pub header: BatchHeader,
@@ -117,10 +121,34 @@ where
     }
 }
 
-impl DefaultBatch {
-    /// Create a new batch from a Vec of DefaultRecords
-    pub fn new(records: Vec<DefaultRecord>) -> Self {
-        let mut batch = DefaultBatch::default();
+impl Batch {
+    /// Create a new empty batch
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// add new record, this will update the offset to correct
+    pub fn add_record(&mut self, mut record: Record) {
+        let last_offset_delta = if self.records.is_empty() {
+            0
+        } else {
+            self.records.len() as Offset
+        };
+        record.preamble.set_offset_delta(last_offset_delta);
+        self.header.last_offset_delta = last_offset_delta as i32;
+        self.records.push(record)
+    }
+
+    /// computed last offset which is base offset + number of records
+    pub fn computed_last_offset(&self) -> Offset {
+        self.get_base_offset() + self.records.len() as Offset
+    }
+}
+
+impl<T: Into<MemoryBatch>> From<T> for Batch {
+    fn from(records: T) -> Self {
+        let records = records.into();
+        let mut batch = Self::default();
 
         let records: Vec<_> = records
             .into_iter()
@@ -135,23 +163,6 @@ impl DefaultBatch {
         let len = batch.records.len() as i32;
         batch.header.last_offset_delta = if len > 0 { len - 1 } else { len };
         batch
-    }
-
-    /// add new record, this will update the offset to correct
-    pub fn add_record(&mut self, mut record: DefaultRecord) {
-        let last_offset_delta = if self.records.is_empty() {
-            0
-        } else {
-            self.records.len() as Offset
-        };
-        record.preamble.set_offset_delta(last_offset_delta);
-        self.header.last_offset_delta = last_offset_delta as i32;
-        self.records.push(record)
-    }
-
-    /// computed last offset which is base offset + number of records
-    pub fn computed_last_offset(&self) -> Offset {
-        self.get_base_offset() + self.records.len() as Offset
     }
 }
 
@@ -254,14 +265,14 @@ pub const BATCH_HEADER_SIZE: usize = size_of::<i32>()     // partition leader ep
 
 #[cfg(test)]
 mod test {
-
+    use super::*;
     use std::io::Cursor;
     use std::io::Error as IoError;
 
     use crate::core::Decoder;
     use crate::core::Encoder;
-    use crate::record::{DefaultRecord, DefaultAsyncBuffer};
-    use crate::batch::DefaultBatch;
+    use crate::record::{Record, RecordData};
+    use crate::batch::Batch;
     use super::BatchHeader;
     use super::BATCH_HEADER_SIZE;
 
@@ -274,11 +285,11 @@ mod test {
     #[test]
     fn test_encode_and_decode_batch() -> Result<(), IoError> {
         let value = vec![0x74, 0x65, 0x73, 0x74];
-        let record = DefaultRecord {
-            value: DefaultAsyncBuffer::new(value),
+        let record = Record {
+            value: RecordData::from(value),
             ..Default::default()
         };
-        let mut batch = DefaultBatch::default();
+        let mut batch = Batch::<MemoryBatch>::default();
         batch.records.push(record);
         batch.header.first_timestamp = 1555478494747;
         batch.header.max_time_stamp = 1555478494747;
@@ -286,7 +297,7 @@ mod test {
         let bytes = batch.as_bytes(0)?;
         println!("batch raw bytes: {:#X?}", bytes.as_ref());
 
-        let batch = DefaultBatch::decode_from(&mut Cursor::new(bytes), 0)?;
+        let batch = Batch::<MemoryBatch>::decode_from(&mut Cursor::new(bytes), 0)?;
         println!("batch: {:#?}", batch);
 
         let decoded_record = batch.records.get(0).unwrap();
@@ -317,16 +328,16 @@ mod test {
 
     #[test]
     fn test_records_offset() {
-        let mut batch = DefaultBatch::default();
+        let mut batch = Batch::<MemoryBatch>::default();
         assert_eq!(batch.get_last_offset_delta(), 0);
 
-        batch.add_record(DefaultRecord::default());
+        batch.add_record(Record::default());
         assert_eq!(batch.get_last_offset_delta(), 0);
 
-        batch.add_record(DefaultRecord::default());
+        batch.add_record(Record::default());
         assert_eq!(batch.get_last_offset_delta(), 1);
 
-        batch.add_record(DefaultRecord::default());
+        batch.add_record(Record::default());
         assert_eq!(batch.get_last_offset_delta(), 2);
 
         assert_eq!(
@@ -358,15 +369,15 @@ mod test {
 
     #[test]
     fn test_batch_records_offset() {
-        let mut comparison = DefaultBatch::default();
-        comparison.add_record(DefaultRecord::default());
-        comparison.add_record(DefaultRecord::default());
-        comparison.add_record(DefaultRecord::default());
+        let mut comparison = Batch::<MemoryBatch>::default();
+        comparison.add_record(Record::default());
+        comparison.add_record(Record::default());
+        comparison.add_record(Record::default());
 
-        let batch_created = DefaultBatch::new(vec![
-            DefaultRecord::default(),
-            DefaultRecord::default(),
-            DefaultRecord::default(),
+        let batch_created = Batch::from(vec![
+            Record::default(),
+            Record::default(),
+            Record::default(),
         ]);
 
         for i in 0..3 {
@@ -381,7 +392,7 @@ mod test {
                     .get(i)
                     .expect("get record")
                     .get_offset_delta(),
-                "Creating a DefaultBatch from a Vec gave wrong delta",
+                "Creating a Batch from a Vec gave wrong delta",
             )
         }
 
--- a/src/dataplane-protocol/src/batch.rs
+++ b/src/dataplane-protocol/src/batch.rs
@@ -18,11 +18,11 @@ use crate::Size;
 use crate::record::Record;
 
 #[allow(deprecated)]
-#[deprecated(since = "0.6.0", note = "Use 'VecBatch' instead")]
+#[deprecated(since = "0.5.1", note = "Use 'MemoryRecords' instead")]
 pub type DefaultBatchRecords = Vec<crate::record::DefaultRecord>;
 
 #[allow(deprecated)]
-#[deprecated(since = "0.6.0", note = "Use 'Batch' instead")]
+#[deprecated(since = "0.5.1", note = "Use 'Batch' instead")]
 pub type DefaultBatch = Batch<DefaultBatchRecords>;
 
 pub trait BatchRecords: Default + Debug + Encoder + Decoder {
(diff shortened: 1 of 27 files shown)


sehz:
CI jobs are failing

sehz (changes requested):

src/dataplane-protocol/src/batch.rs
```diff
@@ -36,10 +33,7 @@ pub const BATCH_PREAMBLE_SIZE: usize = size_of::<Offset>()     // Offset
 pub const BATCH_FILE_HEADER_SIZE: usize = BATCH_PREAMBLE_SIZE + BATCH_HEADER_SIZE;
 
 #[derive(Default, Debug)]
-pub struct Batch<R>
-where
-    R: BatchRecords,
-{
+pub struct Batch<R = Vec<Record>> {
```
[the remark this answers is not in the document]
nicholastmosher: Sure. So we had `type DefaultBatch = Batch<Vec<Record>>` prior to this. Everywhere in the codebase that we were using "Batch", we were actually using `DefaultBatch`. When using `struct Batch<R = Vec<Record>>`, all it says is that if somebody uses a type annotation like `fn get_batch(batch: Batch)`, then it is automatically interpreted to mean `fn get_batch(batch: Batch<Vec<Record>>)`. However, it does not force you to _always_ use that default generic parameter. If you want, you could manually specify the generic, such as `fn get_batch(batch: Batch<HashSet<Record>>)` (for example).

So this does not "force" the struct to use this generic, rather it just allows you to optionally omit the generic at the type-site and it will be defaulted to `Vec<Record>`.

An interesting example of this pattern of usage is with the Vec type itself: https://doc.rust-lang.org/stable/src/alloc/vec/mod.rs.html#374-377

It was recently given a new type signature, `struct Vec<T, A: Allocator = Global>`. This new type parameter does not actually affect the day-to-day usage of the type (because you are not required to manually provide the extra generic), but it optionally allows you to specify a custom allocator, e.g. if you wanted to use Vec in an embedded device.

src/dataplane-protocol/src/batch.rs
```diff
@@ -15,10 +15,7 @@ use crate::derive::Encode;
 
 use crate::Offset;
 use crate::Size;
-use crate::record::DefaultRecord;
-
-pub type DefaultBatchRecords = Vec<DefaultRecord>;
-pub type DefaultBatch = Batch<DefaultBatchRecords>;
```
sehz, line 21: We still need to keep this for compatibility, can mark as deprecated

src/dataplane-protocol/src/batch.rs
```diff
@@ -15,10 +15,7 @@ use crate::derive::Encode;
```
sehz, line 15: need to bump up minor version and make sure all dependent crates use new version

src/dataplane-protocol/src/batch.rs
```diff
@@ -27,7 +24,7 @@ pub trait BatchRecords: Default + Debug + Encoder + Decoder {
     }
 }
 
-impl BatchRecords for DefaultBatchRecords {}
+impl BatchRecords for Vec<Record> {}
```
sehz, line 27: I would like to create alias for `Vec<Records>` this might be new type in future.

src/dataplane-protocol/src/batch.rs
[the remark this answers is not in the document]
sehz: Sounds good.  Didn't realize you could provide default type.  This make it reading code easier.  Good find

src/dataplane-protocol/src/batch.rs
```diff
@@ -155,6 +135,27 @@ impl DefaultBatch {
     }
 }
 
+impl<T: Into<Vec<Record>>> From<T> for Batch {
```
sehz, line 138: use type alias for `Vec<Record>` don't use directly

src/dataplane-protocol/src/batch.rs
```diff
@@ -274,19 +275,19 @@ mod test {
...
     fn test_encode_and_decode_batch() -> Result<(), IoError> {
         let value = vec![0x74, 0x65, 0x73, 0x74];
-        let record = DefaultRecord {
-            value: DefaultAsyncBuffer::new(value),
+        let record = Record {
+            value: RecordData::from(value),
             ..Default::default()
         };
-        let mut batch = DefaultBatch::default();
+        let mut batch = Batch::<Vec<Record>>::default();
```
sehz, line 282: use alias instead of directing ref

src/dataplane-protocol/src/batch.rs
```diff
@@ -274,19 +275,19 @@ mod test {
...
         batch.records.push(record);
         batch.header.first_timestamp = 1555478494747;
         batch.header.max_time_stamp = 1555478494747;
 
         let bytes = batch.as_bytes(0)?;
         println!("batch raw bytes: {:#X?}", bytes.as_ref());
 
-        let batch = DefaultBatch::decode_from(&mut Cursor::new(bytes), 0)?;
+        let batch = Batch::<Vec<Record>>::decode_from(&mut Cursor::new(bytes), 0)?;
```
sehz, line 290: use alias

src/dataplane-protocol/src/batch.rs
```diff
@@ -317,16 +318,16 @@ mod test {
 
     #[test]
     fn test_records_offset() {
-        let mut batch = DefaultBatch::default();
+        let mut batch = Batch::<Vec<Record>>::default();
```
sehz, line 321: ditto

src/dataplane-protocol/src/batch.rs
```diff
@@ -358,15 +359,15 @@ mod test {
...
-        let mut comparison = DefaultBatch::default();
-        comparison.add_record(DefaultRecord::default());
-        comparison.add_record(DefaultRecord::default());
-        comparison.add_record(DefaultRecord::default());
-
-        let batch_created = DefaultBatch::new(vec![
-            DefaultRecord::default(),
-            DefaultRecord::default(),
-            DefaultRecord::default(),
+        let mut comparison = Batch::<Vec<Record>>::default();
```
sehz, line 362: use alias

src/dataplane-protocol/src/record.rs
```diff
@@ -20,67 +19,76 @@ use crate::core::Version;
...
-pub trait Records {}
-
-#[derive(Default)]
-pub struct DefaultAsyncBuffer(Bytes);
+enum RecordKeyInner {
+    Null,
+    Key(RecordData),
+}
 
-impl DefaultAsyncBuffer {
```
sehz, line 44: need to preserve `DefaultAsyncBuffer` for compatibility.  Mark as Deprecated

src/dataplane-protocol/src/record.rs
```diff
@@ -20,67 +19,76 @@ use crate::core::Version;
...
-    {
-        DefaultAsyncBuffer(val.into())
+impl<K: Into<Vec<u8>>> From<K> for RecordKey {
+    fn from(k: K) -> Self {
+        Self(RecordKeyInner::Key(RecordData::from(k)))
     }
+}
+
+#[derive(Default)]
+pub struct RecordData(Vec<u8>);
```
sehz, line 71: Please don't change this implementation without discussion.  In fact, you can make this as generic as
`pub struct RecordData<D=Bytes>(D)`.  That way can optimize for other use cases.

src/spu/src/smart_stream/filter.rs
```diff
@@ -203,15 +203,15 @@ impl SmartFilter {
...
+    pub fn filter(&self, slice: AsyncFileSlice, max_bytes: usize) -> Result<Batch, Error> {
         use std::os::unix::io::AsRawFd;
 
         let fd = slice.as_raw_fd();
 
         let mut batch_iterator =
             FileBatchIterator::new(fd, slice.position() as i64, slice.len() as i64);
 
-        let mut filter_batch = DefaultBatch::default();
+        let mut filter_batch = Batch::<Vec<Record>>::default();
```
sehz, line 214: use alias

src/storage/src/batch.rs
```diff
@@ -170,10 +168,7 @@ where
 }
 
 // stream to iterate batch
-pub struct FileBatchStream<R>
-where
-    R: Default + Debug,
-{
+pub struct FileBatchStream<R = Vec<Record>> {
```
sehz, line 171: use alias

src/storage/src/mut_records.rs
```diff
@@ -412,7 +413,7 @@ mod tests {
         let bytes = read_bytes_from_file(&test_file).expect("read bytes");
         assert_eq!(bytes.len(), write_size, "incorrect size for write");
         debug!("read ok");
-        let batch = DefaultBatch::decode_from(&mut Cursor::new(bytes), 0)?;
+        let batch = Batch::<Vec<Record>>::decode_from(&mut Cursor::new(bytes), 0)?;
```
sehz, line 416: use alias

src/storage/src/mut_records.rs
```diff
@@ -467,7 +468,7 @@ mod tests {
         let bytes = read_bytes_from_file(&test_file).expect("read bytes");
         assert_eq!(bytes.len(), write_size, "incorrect size for write");
 
-        let batch = DefaultBatch::decode_from(&mut Cursor::new(bytes), 0)?;
+        let batch = Batch::<Vec<Record>>::decode_from(&mut Cursor::new(bytes), 0)?;
```
sehz, line 471: ditto

src/storage/src/replica.rs
```diff
@@ -457,7 +457,7 @@ mod tests {
         debug!("using test file: {:#?}", test_file);
         let bytes = read_bytes_from_file(&test_file)?;
 
-        let batch = DefaultBatch::decode_from(&mut Cursor::new(bytes), 0)?;
+        let batch = Batch::<Vec<Record>>::decode_from(&mut Cursor::new(bytes), 0)?;
```
sehz, line 460: alias

src/storage/src/replica.rs
```diff
@@ -595,7 +595,7 @@ mod tests {
         let seg2_file = replica_dir.join(TEST_SE2_NAME);
         let bytes = read_bytes_from_file(&seg2_file)?;
 
-        let batch = DefaultBatch::decode_from(&mut Cursor::new(bytes), 0)?;
+        let batch = Batch::<Vec<Record>>::decode_from(&mut Cursor::new(bytes), 0)?;
```
sehz, line 598: alias

sehz (changes requested):
While using default type parameter maybe good trick I think reduce code readability. Let me sit on that

sehz:
Most of changes seems to be superficial changes and not necessary.  Let's reduce this into smaller PR to focus on 2 relevant aspect:
*  Migrating RecordKey and RecordData into `data-protocol` crates
*  Remove direct access to header field.

This should be good enough to make progress on SmartStream.

src/dataplane-protocol/src/record.rs
nicholastmosher: Hmm I don't know whether we should make this type generic, it may make things more complicated than necessary. I'll revert it back to Bytes for now (because changing it later should not be a breaking change), but I do think that Vec<u8> is a better choice for this based on the use-cases I know about for this.

src/dataplane-protocol/src/record.rs
nicholastmosher: I believe that due to other changes in this PR that we will need a breaking change either way. Are you referring to crate compatibility or protocol compatibility? Because for either of those cases:

- Crate compatibility is already being broken, so it would not hurt to remove this type, and
- Protocol compatibility should be maintained even though we have different type names (i.e. the Encode and Decode implementations are the same).

The only reason I'm pushing back on this is that I have not only renamed this but I've also changed a bunch of the conversions and done other breaking things to this type, and the only good way to "preserve compatibility" would be to undo all of those changes and have that code hang out in a module somewhere where it's probably never going to be used. I think it is a better choice to just lean into the breaking change in this case.

src/dataplane-protocol/src/batch.rs
nicholastmosher: Ok, but just out of curiosity, what sort of compatibility are you talking about? At the type-level, `DefaultBatch` and `Batch<R = Vec<Record>>` are equivalent, one can be substituted for another. Plus, all of the instances of `DefaultBatch` have been removed. So, I believe we have the following:

- This PR makes breaking changes and we are bumping up the crate version to reflect that
- We are also updating the dependencies to use the new version and fixing breakages
- And, crates using old versions of this crate should still be able to interop just fine, since the type resolution should be equivalent.

So I _think_ that removing these shouldn't cause any harm, but if you want I can keep these around for awhile as deprecated until we're sure.

nicholastmosher:
@sehz I've reduced this PR to just moving RecordKey and RecordData. I'll open a new PR for making header and preamble crate-private

src/dataplane-protocol/src/batch.rs
```diff
@@ -27,7 +24,17 @@ pub trait BatchRecords: Default + Debug + Encoder + Decoder {
...
-impl BatchRecords for DefaultBatchRecords {}
+#[allow(deprecated)]
+#[deprecated(since = "0.6.0", note = "Use VecBatch instead")]
+pub type DefaultBatchRecords = Vec<crate::record::DefaultRecord>;
+
+#[allow(deprecated)]
+#[deprecated(since = "0.6.0", note = "Use plain Batch instead")]
+pub type DefaultBatch = Batch<DefaultBatchRecords>;
+
+pub type VecBatch = Vec<Record>;
```
sehz, line 35: I would prefer name as something like`MemoryBatch`,'BytesBatch` or something like that to connote that this Records is represented in Memory not in file slice or some other format.   `Vec` is just current implementation but should not assume that.  We should discourage to rely on Vec

src/dataplane-protocol/src/record.rs
```diff
@@ -20,67 +19,82 @@ use crate::core::Version;
...
+pub struct RecordData(Bytes);
 
+impl RecordData {
     pub fn len(&self) -> usize {
         self.0.len()
     }
+}
 
-    /// Check if value is binary content
-    pub fn is_binary(&self) -> bool {
```
sehz, line 57: why is this API removed?

sehz (changes requested):
From looking at the code,  there is trade for adding default record type which simplifies certain aspects but add more complexity in using it which I don't think it is worth it. 

Please revert back changes:
```
pub struct Record&lt;B = RecordData&gt; {    pub preamble: RecordHeader,    pub key: Option&lt;B&gt;,    pub value: B,    pub headers: i64,}
```

However we can debate use of `DefaultRecords`, `DefaultBatch` which could be renamed as something else.

src/dataplane-protocol/src/record.rs
nicholastmosher: From what I can tell it is completely unused

src/dataplane-protocol/src/batch.rs
nicholastmosher: I like `MemoryBatch` 👍

src/dataplane-protocol/src/record.rs
nicholastmosher: I reverted any changes that should be breaking, so this can stay

nicholastmosher:
@sehz I think I have made all the changes requested, let me know if I missed something. I added back any methods that were previously removed, so there should no longer be any breaking changes. There are type aliases that re-export the types under their old names, with a deprecated tag so they show warnings to users. I will change new version to 0.5.1 rather than 0.6.0

src/dataplane-protocol/src/record.rs
nicholastmosher: Here's a summary of changes after latest discussion:

- Renamed DefaultAsyncBuffer to RecordData
- Added a type alias from DefaultAsyncBuffer to RecordData so that any crates importing DefaultAsyncBuffer will still work
  - Added a #[deprecated]` tag to the DefaultAsyncBuffer type alias so those users get warnings.

src/dataplane-protocol/src/batch.rs
nicholastmosher: Bumped version to 0.5.1

src/dataplane-protocol/src/batch.rs
```diff
@@ -15,9 +15,14 @@ use crate::derive::Encode;
 
 use crate::Offset;
 use crate::Size;
-use crate::record::DefaultRecord;
+use crate::record::Record;
 
-pub type DefaultBatchRecords = Vec<DefaultRecord>;
+#[allow(deprecated)]
+#[deprecated(since = "0.6.0", note = "Use 'VecBatch' instead")]
```
sehz, line 21: should be MemoryBatch. and 0.6.0->0.5.1

src/dataplane-protocol/src/batch.rs
```diff
@@ -15,9 +15,14 @@ use crate::derive::Encode;
...
+pub type DefaultBatchRecords = Vec<crate::record::DefaultRecord>;
+
+#[allow(deprecated)]
+#[deprecated(since = "0.6.0", note = "Use 'Batch' instead")]
```
sehz, line 25: ditto

src/dataplane-protocol/src/record.rs
```diff
@@ -20,33 +19,70 @@ use crate::core::Version;
 use crate::derive::Decode;
 use crate::derive::Encode;
 
-use crate::batch::DefaultBatch;
+use crate::batch::Batch;
 use crate::Offset;
 
-pub type DefaultRecord = Record<DefaultAsyncBuffer>;
+#[deprecated(since = "0.6.0", note = "Use 'Record' instead")]
```
sehz, line 25: ditto

sehz (changes requested):

src/dataplane-protocol/src/batch.rs
```diff
@@ -27,7 +32,9 @@ pub trait BatchRecords: Default + Debug + Encoder + Decoder {
     }
 }
 
-impl BatchRecords for DefaultBatchRecords {}
+pub type MemoryBatch = Vec<Record>;
```
sehz, line 35: This is not really batch.  It's sets of records.  So it should been MemoryRecords

nicholastmosher: `MemoryRecords` it is 👍

sehz (approved):
LGTM.  Nice changes

nicholastmosher:
Thanks!

bors r+

nicholastmosher:
bors ping

PR #1103, opened 2021-05-19
[Merged by Bors] - Remove controllerSvc from SPG spec

nacardin:
This is not used anywhere


src/controlplane-metadata/src/spg/k8/spec.rs
```diff
@@ -56,8 +56,6 @@ pub struct SpuTemplate {
     #[serde(skip_serializing_if = "Option::is_none")]
```
sehz, line 56: crate version number need to be updated and all effected crates should use new version otherwise this will cause build issue

sehz (changes requested):

src/controlplane-metadata/src/spg/k8/spec.rs
nacardin: bumped crate versions

sehz (approved):

PR #1106, opened 2021-05-20
[Merged by Bors] - Fix SPU pool issue with wasm client

### LINKED ISSUE #1097: spu_pool once cell work around for wasm32
Subtask of #989.

simlay:
Closes #1097.

### DIFF
--- a/src/client/src/fluvio.rs
+++ b/src/client/src/fluvio.rs
@@ -2,10 +2,10 @@ use std::convert::TryFrom;
 use std::sync::Arc;
 
 use tracing::debug;
-use once_cell::sync::OnceCell;
+use tokio::sync::OnceCell;
+//use once_cell::sync::OnceCell;
 
 use fluvio_socket::{SharedMultiplexerSocket, MultiplexerSocket};
-use fluvio_future::task::run_block_on;
 use fluvio_future::net::DomainConnector;
 //use fluvio_future::native_tls::AllDomainConnector;
 use semver::Version;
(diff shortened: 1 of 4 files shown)


src/client/src/fluvio.rs
```diff
@@ -83,13 +82,20 @@ impl Fluvio {
...
+        // This is done this way because the difference in `run_block_on` for wasm vs non-wasm
+        // Currently, the wasm `block_on` call does not return `T` where as the non-wasm call does
+        // return `T` for a future.
+        // source: https://github.com/async-rs/async-std/blob/35f768166436112db97224e823b4ee610c81d6d6/src/task/block_on.rs#L28-L44
+        // One option is to use https://github.com/Yoeori/async-oncecell but it doesn't seem like
+        // there's a lot of activity on that project.
+        if let Some(pool) = self.spu_pool.get() {
+            Ok(pool.clone())
+        } else {
+            let pool = Arc::new(SpuPool::start(self.config.clone(), self.socket.clone()).await?);
```
sehz, line 95: Problem is that this is not atomic (two concurrent init could cause duplicate init of pooling).  There is: https://docs.rs/tokio/1.5.0/src/tokio/sync/once_cell.rs.html#270-320.

sehz (changes requested):

src/client/src/fluvio.rs
simlay: Ah. Yes. We can try it.

sehz (approved):
Nice!

PR #1111, opened 2021-05-20
[Merged by Bors] - Fluvio socket wasm support

simlay:
Co-PR to https://github.com/infinyon/future-aio/pull/70.

The `clone` calls will not be liked by clippy but the websocket connector returns `String` which does not implement `Copy`.

I think the build will fail due to https://github.com/infinyon/future-aio/pull/69.

### DIFF
--- a/src/client/Cargo.toml
+++ b/src/client/Cargo.toml
@@ -21,7 +21,6 @@ tracing = "0.1.19"
 tracing-futures = "0.2.4"
 futures-util = "0.3.6"
 bytes = "1.0.1"
-dirs = "1.0.2"
 toml = "0.5.5"
 async-rwlock = "1.1.0"
 base64 = { version = "0.13.0" }
@@ -45,6 +44,9 @@ fluvio-spu-schema = { version = "0.6.0", path = "../spu-schema" }
 fluvio-socket = { path = "../socket", version = "0.8.1" }
 fluvio-protocol = { path = "../protocol", version = "0.5.1" }
 dataplane = { version = "0.5.1", path = "../dataplane-protocol", package = "fluvio-dataplane-protocol" }
+
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+dirs = "1.0.2"
 
 [dev-dependencies]
 async-std = { version = "1.6.4", default-features = false }
--- a/src/client/src/config/config.rs
+++ b/src/client/src/config/config.rs
@@ -15,7 +15,14 @@ use std::fs::create_dir_all;
 use thiserror::Error;
 
 use tracing::debug;
+#[cfg(not(target_arch = "wasm32"))]
 use dirs::home_dir;
+
+#[cfg(target_arch = "wasm32")]
+fn home_dir() -> Option<PathBuf> {
+    None
+}
+
 use serde::Deserialize;
 use serde::Serialize;
 
--- a/src/client/src/sockets.rs
+++ b/src/client/src/sockets.rs
@@ -119,7 +119,7 @@ impl ClientConfig {
     }
 
     pub(crate) async fn connect(self) -> Result<VersionedSocket, FluvioError> {
-        debug!("Connecting to {:?}", self.addr);
+        debug!(add = %self.addr, "Connection to");
         let socket = FluvioSocket::connect_with_connector(&self.addr, self.connector.as_ref()).await?;
         VersionedSocket::connect(socket, Arc::new(self)).await
     }
--- a/src/client/src/sockets.rs
+++ b/src/client/src/sockets.rs
@@ -120,7 +120,8 @@ impl ClientConfig {
 
     pub(crate) async fn connect(self) -> Result<VersionedSocket, FluvioError> {
         debug!(add = %self.addr, "Connection to");
-        let socket = FluvioSocket::connect_with_connector(&self.addr, self.connector.as_ref()).await?;
+        let socket =
+            FluvioSocket::connect_with_connector(&self.addr, self.connector.as_ref()).await?;
         VersionedSocket::connect(socket, Arc::new(self)).await
     }
 
--- a/src/client/src/sockets.rs
+++ b/src/client/src/sockets.rs
@@ -128,8 +128,7 @@ impl ClientConfig {
             addr
         };
         debug!("Connecting to {:?}", addr);
-        let socket =
-            FluvioSocket::connect_with_connector(&addr, self.connector.as_ref()).await?;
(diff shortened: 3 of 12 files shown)


sehz:
will merge once https://github.com/infinyon/future-aio/pull/70 is merged  into master

src/client/src/sockets.rs
```diff
@@ -119,8 +119,17 @@ impl ClientConfig {
...
 
     pub(crate) async fn connect(self) -> Result<VersionedSocket, FluvioError> {
+        let addr = self.addr.clone();
+        // When we have websocket proxy in place, this won't be needed.
+        #[cfg(target_arch = "wasm32")]
+        let addr = if addr == "localhost:9010" {
+            String::from("ws://localhost:3001")
+        } else {
+            addr
+        };
```
simlay, line 129: @sehz Thoughts on what to do in place of this?

src/socket/src/sink.rs
```diff
@@ -210,15 +210,15 @@ impl ExclusiveFlvSink {
...
+        self.fd.clone()
     }
 }
 
 impl Clone for ExclusiveFlvSink {
     fn clone(&self) -> Self {
         Self {
             inner: self.inner.clone(),
-            fd: self.fd,
+            fd: self.fd.clone(),
```
simlay, line 221: Clippy doesn't like this but it's needed for wasm as the `fd` is a `String` (doesn't implement `Copy`) in wasm32.

src/client/src/config/config.rs
```diff
@@ -15,7 +15,14 @@ use std::fs::create_dir_all;
 use thiserror::Error;
 
 use tracing::debug;
+#[cfg(not(target_arch = "wasm32"))]
 use dirs::home_dir;
+
+#[cfg(target_arch = "wasm32")]
+fn home_dir() -> Option<PathBuf> {
```
simlay, line 22: This is a work around in place until #1115 is done.

src/client/src/sockets.rs
sehz: This can passed as addr.  There is no need for hard coding

simlay: Yeah, so the reason this is a needed work around is that the client connects to fluvio-sc via a webscocket, then once trying to get a Partition Consumer stream or a producer, the SC gives some metadata about the SPU to connect to which is `localhost:9010` for a local cluster.

This isn't exposed to the client.

When we add a real websocket proxy rather than running `websocat -v --binary ws-l:127.0.0.1:3000 tcp:127.0.0.1:9003`, it needs to do something smarter.

simlay: I removed it from the PR but I'm still using it in my local branch.

src/client/Cargo.toml
```diff
@@ -46,6 +45,9 @@ fluvio-socket = { path = "../socket", version = "0.8.1" }
 fluvio-protocol = { path = "../protocol", version = "0.5.1" }
 dataplane = { version = "0.5.1", path = "../dataplane-protocol", package = "fluvio-dataplane-protocol" }
```
sehz, line 47: need to bump up fluvio-future to use `DefaultDomainConnector`

src/client/src/sockets.rs
```diff
@@ -119,8 +119,8 @@ impl ClientConfig {
     }
 
     pub(crate) async fn connect(self) -> Result<VersionedSocket, FluvioError> {
-        let socket =
-            FluvioSocket::connect_with_connector(&self.addr, self.connector.as_ref()).await?;
+        debug!("Connecting to {:?}", self.addr);
```
sehz, line 122: use object format:  `debug!( add = %self.addr, "connection to")`

src/socket/src/socket.rs
```diff
@@ -7,8 +7,7 @@ use fluvio_protocol::api::RequestMessage;
 use fluvio_protocol::api::ResponseMessage;
 
 use fluvio_future::net::{
-    BoxReadConnection, BoxWriteConnection, ConnectionFd, DefaultTcpDomainConnector,
-    TcpDomainConnector, TcpStream,
+    BoxReadConnection, BoxWriteConnection, ConnectionFd, DefaultDomainConnector, TcpDomainConnector,
```
sehz, line 10: TcpDomainConnector should be only imported for Unix?

sehz (changes requested):
Nice work!  Surprised to see Multiplexor works now without code changes.   Few minor nits

sehz:
Doc test fails

src/socket/src/socket.rs
simlay: I think I forgot to rename this to `SocketDomainConnector` or something in https://github.com/infinyon/future-aio/pull/70. `DefaultConnector is a `Box<dyn TcpDomainConnector>`.

sehz:
Looks good but doc tests seems to failing

simlay:
> Looks good but doc tests seems to failing.

The errors are kinda big but I think it's due to the change in `spawn` from https://github.com/infinyon/future-aio/pull/69

```
 error[E0277]: `dyn TcpDomainConnector` cannot be shared between threads safely
  --> src/spu/src/control_plane/dispatcher.rs:73:9
   |
73 |         spawn(self.dispatch_loop());
   |         ^^^^^ `dyn TcpDomainConnector` cannot be shared between threads safely
   | 
  ::: /mnt/local_disk/actions/.cargo/registry/src/github.com-1ecc6299db9ec823/async-std-1.9.0/src/task/spawn.rs:28:29
   |
28 |     F: Future<Output = T> + Send + 'static,
   |                             ---- required by this bound in `fluvio_future::task::spawn`
```

sehz:
Hmm. Looks like I screwed up. let me fix this

sehz:
Actually, fluvio master works fine against release version.  So this is caused by: https://github.com/infinyon/future-aio/commit/6409c8ce70e605d51952e6a043b442b6a17e0d52

sehz:
weird. now everything is ok.  try `cargo update` and try running all tests

sehz:
I found elegant way to add conditional constrain to: `TcpDomainConnector`.  Will patch `fluvio-future`

sehz:
done.  this should pass now

sehz:
clippy is failing.  we need to add another trait to `fluvio_future`, something like: 
```
trait CloneConnectionFd {
      fn clone(&self) -> Self;
}

impl CloneConnectionFd  for std::os::unix::io::RawFd {
       fn clone(&self) -> Self {
            self.copy
      }
}

....
```

sehz (approved):
Finally. nice work

PR #1118, opened 2021-05-24
[Merged by Bors] - Add OpenTelemetry/Jaeger tracing backend

nicholastmosher:
This is the first step towards improving our observability within Fluvio. This adds a jaeger backend to our tracing setup, so you can now run a jaeger instance and collect traces.

- Adds Jaeger backend to tracing subscriber for SC and SPU
- Adds more tracing instrumentation to the SPU

To try this out for yourself, run Jaeger in a docker container with the following command:

```
docker run -d -p6831:6831/udp -p6832:6832/udp -p16686:16686 -p14268:14268 jaegertracing/all-in-one:latest
```

Then open up `http://localhost:16686` in your browser. Try running some Fluvio commands to interact with the cluster, then view traces sorted by `fluvio-spu` or `fluvio-sc`.

I don't know what we want the ultimate scope of this PR to be. It could potentially be merged now, but do we want to explore any of the following first?

- Add more instrumentation to SC? I think that'd be a good idea
- Add instrumentation to the client, i.e. Fluvio CLI? If yes, we would want to make sure it is optional and probably disabled by default
- Try adding prometheus metrics? There are examples showing how this works, but I believe it requires creating a metrics route on an HTTP server, so it would really only make sense for the SC and SPU. This is probably out of scope for this.

### DIFF
--- a/src/runner/Cargo.toml
+++ b/src/runner/Cargo.toml
@@ -17,7 +17,21 @@ name = "fluvio-run"
 path = "src/bin/main.rs"
 doc = false
 
+[features]
+default = ["telemetry"]
+telemetry = [
+    "tracing-opentelemetry",
+    "opentelemetry",
+    "opentelemetry-jaeger",
+]
+
 [dependencies]
+tracing = "0.1"
+tracing-subscriber = "0.2"
+tracing-opentelemetry = { version = "0.10", optional = true }
+opentelemetry = { version = "0.11", optional = true }
+opentelemetry-jaeger = { version = "0.10", optional = true }
+
 structopt = { version = "0.3.16", default-features = false }
 thiserror = "1.0.20"
 semver = "0.11.0"
--- a/src/runner/Cargo.toml
+++ b/src/runner/Cargo.toml
@@ -18,7 +18,7 @@ path = "src/bin/main.rs"
 doc = false
 
 [features]
-default = ["telemetry"]
+default = []
 telemetry = [
     "tracing-opentelemetry",
     "opentelemetry",
--- a/src/runner/src/bin/main.rs
+++ b/src/runner/src/bin/main.rs
@@ -1,15 +1,15 @@
 use structopt::StructOpt;
 use fluvio_run::RunCmd;
 
-use tracing::Level;
-use tracing_subscriber::{Registry, prelude::*};
-use tracing_subscriber::filter::Directive;
-
 fn main() -> Result<(), Box<dyn std::error::Error>> {
     let cmd: RunCmd = RunCmd::from_args();
 
     #[cfg(feature = "telemetry")]
     let _jaeger_guard = {
+        use tracing::Level;
+        use tracing_subscriber::{Registry, prelude::*};
+        use tracing_subscriber::filter::Directive;
+
         let service_name = match cmd {
             RunCmd::SC(_) => "fluvio-sc",
             RunCmd::SPU(_) => "fluvio-spu",
--- a/src/runner/src/bin/main.rs
+++ b/src/runner/src/bin/main.rs
@@ -1,10 +1,41 @@
 use structopt::StructOpt;
-use fluvio_future::task::run_block_on;
 use fluvio_run::RunCmd;
 
+use tracing::Level;
+use tracing_subscriber::{Registry, prelude::*};
+use tracing_subscriber::filter::Directive;
+
 fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let cmd: RunCmd = RunCmd::from_args();
+
+    #[cfg(feature = "telemetry")]
+    let _jaeger_guard = {
+        let service_name = match cmd {
+            RunCmd::SC(_) => "fluvio-sc",
+            RunCmd::SPU(_) => "fluvio-spu",
+            _ => "fluvio-run",
+        };
+
+        let (tracer, guard) = opentelemetry_jaeger::new_pipeline()
+            .with_service_name(service_name)
+            .install()
+            .unwrap();
+
+        Registry::default()
+            .with(
+                tracing_subscriber::EnvFilter::from_default_env()
+                    .add_directive(Directive::from(Level::DEBUG)),
+            )
+            .with(tracing_subscriber::fmt::layer())
+            .with(tracing_opentelemetry::layer().with_tracer(tracer))
+            .init();
+
+        guard
+    };
+
+    #[cfg(not(feature = "telemetry"))]
     fluvio_future::subscriber::init_tracer(None);
-    let cmd = RunCmd::from_args();
-    run_block_on(cmd.process())?;
+
+    cmd.process()?;
     Ok(())
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -215,7 +215,7 @@ dependencies = [
  "httparse",
  "lazy_static",
  "log",
- "pin-project",
+ "pin-project 1.0.7",
 ]
 
 [[package]]
@@ -1552,7 +1552,7 @@ dependencies = [
  "log",
  "native-tls",
  "openssl",
- "pin-project",
+ "pin-project 1.0.7",
  "thiserror",
  "tracing",
 ]
@@ -1579,7 +1579,7 @@ dependencies = [
  "nix 0.20.0",
  "openssl",
  "openssl-sys",
- "pin-project",
+ "pin-project 1.0.7",
  "pin-utils",
  "thiserror",
  "tracing",
@@ -1694,10 +1694,15 @@ dependencies = [
  "fluvio-future 0.3.3",
  "fluvio-sc",
  "fluvio-spu",
+ "opentelemetry",
+ "opentelemetry-jaeger",
  "semver 0.11.0",
  "serde_json",
  "structopt",
  "thiserror",
+ "tracing",
+ "tracing-opentelemetry",
+ "tracing-subscriber",
 ]
 
 [[package]]
@@ -1809,7 +1814,7 @@ dependencies = [
  "futures-util",
  "log",
  "once_cell",
- "pin-project",
+ "pin-project 1.0.7",
  "thiserror",
  "tokio",
  "tokio-util",
@@ -2492,7 +2497,7 @@ dependencies = [
  "httparse",
  "httpdate",
  "itoa",
- "pin-project",
+ "pin-project 1.0.7",
  "tokio",
  "tower-service",
  "tracing",
@@ -2556,6 +2561,12 @@ dependencies = [
  "wasm-bindgen",
  "web-sys",
 ]
+
+[[package]]
+name = "integer-encoding"
+version = "1.1.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "48dc51180a9b377fd75814d0cc02199c20f8e99433d6762f650d39cdbbd3b56f"
 
 [[package]]
 name = "inventory"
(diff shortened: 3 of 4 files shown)


sehz:
linked to issue?

src/protocol/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "fluvio-protocol"
 edition = "2018"
-version = "0.5.1"
+version = "0.5.2"
```
sehz, line 4: this has not been published yet, so we can still keep 0.5.1

src/protocol/fluvio-protocol-derive/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio-protocol-derive"
-version = "0.2.1"
+version = "0.2.2"
```
sehz, line 3: latest published version is `0.2.0`

src/protocol/fluvio-protocol-derive/src/de.rs
```diff
@@ -17,6 +17,10 @@ pub(crate) fn generate_decode_trait_impls(input: &DeriveItem) -> TokenStream {
             let (impl_generics, ty_generics, where_clause) = kf_struct.generics.split_for_impl();
             quote! {
                 impl #impl_generics fluvio_protocol::Decoder for #ident #ty_generics #where_clause {
+                    #[tracing::instrument(
+                        skip(self, src, version),
+                        fields(type_name = std::any::type_name::<#ident #ty_generics>()),
+                    )]
```
sehz, line 23: this is too low level to add instrument?

src/protocol/fluvio-protocol-derive/src/ser.rs
```diff
@@ -16,12 +16,14 @@ pub(crate) fn generate_encode_trait_impls(input: &DeriveItem) -> TokenStream {
             let size_field_tokens = parse_struct_props_size(&kf_struct.props, &ident);
             quote! {
                 impl #impl_generics fluvio_protocol::Encoder for #ident #ty_generics #where_clause {
+                    #[tracing::instrument(skip(self, dest, version))]
```
sehz, line 19: same.  let's not add instrument to this crate

src/runner/Cargo.toml
```diff
@@ -18,6 +18,12 @@ path = "src/bin/main.rs"
 doc = false
 
 [dependencies]
+tracing = "0.1"
+tracing-subscriber = "0.2"
+tracing-opentelemetry = "0.10"
+opentelemetry = "0.11"
+opentelemetry-jaeger = "0.10"
```
sehz, line 25: make jaeger as optional

src/runner/src/bin/main.rs
```diff
@@ -1,10 +1,27 @@
...
+
+    let (tracer, _uninstall) = opentelemetry_jaeger::new_pipeline()
+        .with_service_name(service_name)
+        .install()
+        .unwrap();
+
+    Registry::default()
+        .with(tracing_subscriber::EnvFilter::new("DEBUG"))
+        .with(tracing_subscriber::fmt::layer())
+        .with(tracing_opentelemetry::layer().with_tracer(tracer))
```
sehz, line 22: make this as optional configuration.

src/spu/src/replication/leader/leaders_state.rs
```diff
@@ -43,17 +43,20 @@ impl<S> ReplicaLeadersState<S> {
 
 impl<S> ReplicaLeadersState<S> {
     /// get clone of state
+    #[instrument(skip(self, replica))]
```
sehz, line 46: remove here

src/spu/src/services/public/stream_fetch.rs
```diff
@@ -449,10 +449,12 @@ pub mod publishers {
         }
 
         // get next stream id
+        #[instrument(skip(self))]
         fn next_stream_id(&self) -> u32 {
             self.stream_id.fetch_add(1, SeqCst)
         }
 
+        #[instrument(skip(self))]
```
sehz, line 457: remove

src/storage/src/replica.rs
```diff
@@ -37,20 +37,24 @@ impl Unpin for FileReplica {}
 impl ReplicaStorage for FileReplica {
     type Config = ConfigOption;
 
+    #[instrument(skip(replica, config))]
     async fn create(replica: &ReplicaKey, config: Self::Config) -> Result<Self, StorageError> {
         Self::create(replica.topic.clone(), replica.partition as u32, 0, config).await
     }
 
+    #[instrument(skip(self))]
```
sehz, line 45: remove

sehz (changes requested):
Careful on adding instrument otherwise it will lead to generation of too many low level spans. Instrument should be added
* API call
* Long duration async, typically network or long storage (for example, sending status)

src/protocol/Cargo.toml
nicholastmosher: Sure, I just needed to use this to force Cargo to compile my local changes, e.g. for fluvio-protocol-derive

src/protocol/fluvio-protocol-derive/src/de.rs
nicholastmosher: Part of the benefit of putting this here is that, since we manually control the Encoding/Decoding, we may want visibility into the performance of the encode/decode implementations for various pieces of data in case something is accidentally taking a long time when it shouldn't. That was my rationale at least. If you think it is too low-level I can remove it

sehz: encode/decode is at `ns` resolution (this will be be compiler optimized) which is too low level which was not designed for tracing.

src/spu/src/services/public/stream_fetch.rs
nicholastmosher: The original reason I put this in this area because of the async mutexes involved. I was thinking it would be a good way to spot any problems if we get any high lock contention or anything like that.

src/runner/src/bin/main.rs
nicholastmosher: Hmm I am trying, but it seems like it is not working when I make things optional, even if the feature is enabled.

nicholastmosher: Just figured out the solution to this, PR is updated

nicholastmosher:
I have named the feature for enabling/disabling opentelemetry+jaeger "telemetry". Let me know if you'd prefer a different name for that, e.g. just "jaeger" or something

src/client/src/sockets.rs
```diff
@@ -214,7 +218,9 @@ impl VersionedSerialSocket {
         // send request & save response
         self.socket.send_and_receive(req_msg).await
     }
+
     /// create new request based on version
+    #[instrument(skip(self, request, version))]
```
sehz, line 223: this is low level.  either put trace or remove it

src/protocol/fluvio-protocol-api/src/request.rs
```diff
@@ -85,10 +85,12 @@ where
     }
 
     #[allow(unused)]
+    #[instrument(skip(self, response))]
```
sehz, line 88: put trace mode

src/protocol/fluvio-protocol-api/src/request.rs
```diff
@@ -85,10 +85,12 @@ where
...
     pub fn new_response(&self, response: R::Response) -> ResponseMessage<R::Response> {
         Self::response_with_header(&self.header, response)
     }
 
+    #[instrument(skip(header, response))]
```
sehz, line 93: trace mode

src/protocol/fluvio-protocol-api/src/request.rs
```diff
@@ -97,6 +99,7 @@ where
     }
 
     #[allow(unused)]
+    #[instrument(skip(self, src, version))]
```
sehz, line 102: trace mode

src/protocol/fluvio-protocol-api/src/request.rs
```diff
@@ -132,6 +136,7 @@ impl<R> Decoder for RequestMessage<R>
 where
     R: Request,
 {
+    #[instrument(skip(self, src, version))]
     fn decode<T>(&mut self, src: &mut T, version: Version) -> Result<(), IoError>
```
sehz, line 140: trace mode

src/protocol/fluvio-protocol-api/src/request.rs
```diff
@@ -150,6 +155,7 @@ where
         self.header.write_size(version) + self.request.write_size(self.header.api_version())
     }
 
+    #[instrument(skip(self, out, version))]
```
sehz, line 158: trace mode

src/runner/Cargo.toml
```diff
@@ -17,7 +17,21 @@ name = "fluvio-run"
 path = "src/bin/main.rs"
 doc = false
 
+[features]
+default = ["telemetry"]
```
sehz, line 21: let's not put this under default feature

src/runner/src/bin/main.rs
```diff
@@ -1,10 +1,47 @@
 use structopt::StructOpt;
-use fluvio_future::task::run_block_on;
 use fluvio_run::RunCmd;
 
+use tracing::Level;
+use tracing_subscriber::{Registry, prelude::*};
+use tracing_subscriber::filter::Directive;
+
 fn main() -> Result<(), Box<dyn std::error::Error>> {
-    fluvio_future::subscriber::init_tracer(None);
```
sehz, line 6: make jaeger optional.  if `telemetry` feature is not defined then default to previous behavior.

src/runner/src/bin/main.rs
```diff
@@ -1,10 +1,47 @@
...
-    let cmd = RunCmd::from_args();
-    run_block_on(cmd.process())?;
+    let cmd: RunCmd = RunCmd::from_args();
+
+    #[cfg(feature = "telemetry")]
```
sehz, line 11: use if_config to simplify

src/sc/src/controllers/partitions/controller.rs
```diff
@@ -70,6 +70,7 @@ impl PartitionController {
         // info!("spu controller is terminated");
     }
 
+    #[instrument(skip(self, listener), name = "partition_controller_sync")]
```
sehz, line 73: not fan of this renaming

src/sc/src/controllers/partitions/reducer.rs
```diff
@@ -79,6 +79,7 @@ impl PartitionReducer {
         }
     }
 
+    #[instrument(skip(self, updates))]
```
sehz, line 82: let's put as separate PR.  One for adding feature flag.  And create separate PR for each effected modules.

sehz (changes requested):
This is too big of PR.  Lets' separate into individual PR.  First should be only about putting Jaegar collector.  Then let's do individual PR for each crates

nicholastmosher:
Minimized the PR to only include Jaeger/Opentelemetry backend. Instrumentation has been broken down by crate and I'll introduce that in further PRs.

src/runner/src/bin/main.rs
nicholastmosher: I think if_cfg would actually be about the same as this setup, and it adds an external dependency. I'd like to leave it unless you have an overwhelmingly strong opinion about it

sehz (approved):
Can merge after release

sehz:
will review rest of PR once this is landed otherwise hard to see changes

Issue #986, opened 2021-04-21
`PartitionConsumer::stream` should not block while there are no records


nicholastmosher:
If I create a new empty topic and run the following code, it will block indefinitely:

```rust
let consumer = fluvio::consumer("my-topic", 0).await?;

// This call blocks until a record appears in the partition
let mut stream = consumer.stream(Offset::beginning()).await?;
```

This is not desirable. The expected behavior would be for the stream object to be returned as soon as a connection to the partition leader is created. Waiting should only happen if we explicitly await on fetching elements on the stream, e.g. here:

```rust
while let Some(Ok(record)) = stream.next().await { ... }
```

PR #1120, opened 2021-05-25
[Merged by Bors] - Initial stuff for Fluvio::connect_with_connector

simlay:
I also added in a wasm test. I've got mixed feelings about where it is now.

### DIFF
--- a/src/client/src/fluvio.rs
+++ b/src/client/src/fluvio.rs
@@ -65,6 +65,12 @@ impl Fluvio {
     /// ```
     pub async fn connect_with_config(config: &FluvioConfig) -> Result<Self, FluvioError> {
         let connector = DomainConnector::try_from(config.tls.clone())?;
+        Self::connect_with_connector(connector, config).await
+    }
+    pub async fn connect_with_connector(
+        connector: DomainConnector,
+        config: &FluvioConfig,
+    ) -> Result<Self, FluvioError> {
         let config = ClientConfig::new(&config.endpoint, connector);
         let inner_client = config.connect().await?;
         debug!("connected to cluster at: {}", inner_client.config().addr());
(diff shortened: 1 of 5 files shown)


sehz:
rustfmt is failing

sehz (approved):
LGTM

sehz:
local cluster test is failing

simlay:
> local cluster test is failing

"Expected browser binary location, but unable to find binary in default location, no 'moz:firefoxOptions.binary' capability provided, and no binary flag set on the command line"

@nacardin is firefox or chrome installed on the custom runner?

sehz:
Suggest separating integration test as 2nd PR so code changes can go thru

src/client/src/fluvio.rs
```diff
@@ -65,6 +65,12 @@ impl Fluvio {
     /// ```
     pub async fn connect_with_config(config: &FluvioConfig) -> Result<Self, FluvioError> {
         let connector = DomainConnector::try_from(config.tls.clone())?;
+        Self::connect_with_connector(connector, config).await
+    }
+    pub async fn connect_with_connector(
+        connector: DomainConnector,
+        config: &FluvioConfig,
+    ) -> Result<Self, FluvioError> {
```
nicholastmosher, line 73: Why are we adding this? And more importantly, what is the usage pattern here? Are we expecting typical users to leverage this or is it more of an implementation detail for the WASM library that users will not directly use?

sehz (commented):
This allows client to use custom networking.  For example:  unix socket, websocket, QUIC, etc.

src/client/src/fluvio.rs
```diff
@@ -204,3 +210,108 @@ fn check_platform_compatible(cluster_version: &Version) -> Result<(), FluvioErro
...
+        let topic = "wasm-test-produce-consume".to_string();
+        /*
+        let err = admin.create(topic, false, TopicSpec::default()).await;
+        tracing::error!("ERROR: {:?}", err);
+        assert!(err.is_ok())
+        */
+        let producer = client.topic_producer(topic.clone()).await;
+        assert!(producer.is_ok());
+        let producer = producer.unwrap();
+        let send = producer.send("foo", "bar").await;
```
simlay, line 288: This is a work around because #986 is still an issue.

sehz:
will merge this after release

sehz:
Move  changelog for this to 0.8.4

PR #1134, opened 2021-05-26
fix: install k8_types dependency with the 'core' feature

morenol:
Trying to compile a crate that depends on this failed with:

```rust
5 | use fluvio_stream_model::k8_types::core::secret::{SecretHeader, SecretSpec};
  |                                    ^^^^ could not find `core` in `k8_types`
```

That is behind the [core](https://github.com/infinyon/k8-api/blob/e4b0a0929b01fd0e7c3a9ec268c63dcfb09d7bf1/src/k8-types/src/lib.rs#L5) feature config.
So, it looks like that this feature is needed in order to use that. Not sure why it is failing for me and not in the CI.

### DIFF
--- a/src/stream-model/Cargo.toml
+++ b/src/stream-model/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "fluvio-stream-model"
 edition = "2018"
-version = "0.5.1"
+version = "0.5.2"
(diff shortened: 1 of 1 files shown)


sehz:
because it `stream-model` doesn't export `core` of `k8-types`.  You should import `k8-types` directly

PR #1135, opened 2021-05-27
[Merged by Bors] - Fix placement of the producer CLI > prompt

nicholastmosher:
Before this PR, the producer would not print `>` before the first stdin prompt, making it appear that the CLI was hanging:

A screenshot of a terminal showing the producer does not print a caret before the user is expected to type

After the first caret prompt, it also mistakenly printed `>` before the Ok! acknowledgement.

Now, it has been corrected to print `>` before each stdin prompt, but not before each Ok!

A screenshot of a terminal where the producer CLI is printing carets correctly for each prompt

### DIFF
--- a/src/cli/src/consumer/produce/mod.rs
+++ b/src/cli/src/consumer/produce/mod.rs
@@ -68,6 +68,9 @@ impl ProduceOpt {
             }
             None => {
                 let mut lines = BufReader::new(std::io::stdin()).lines();
+                if self.interactive_mode() {
+                    eprint!("> ");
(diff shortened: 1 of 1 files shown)


sehz:
Rebase please

nicholastmosher:
Sure, but is there a specific reason to rebase? Bors takes care of merging if there are no conflicts, and it looks like there aren't any here. Just FYI

sehz:
CI was failing

sehz (approved):

PR #1148, opened 2021-06-03
[Merged by Bors] - feat: Add 'fluvio produce --raw' for sending whole files

### LINKED ISSUE #1139: feat: ability to upload raw binary file
It seems like we can't upload binary file probably because of separator:
```
fluvio produce test -f example_filter.wasm 
```
this should result in a single record but creates many records:
```
 TOPIC               PARTITION  LEADER  REPLICAS  RESOLUTION  HW     LEO    LSR  FOLLOWER OFFSETS 
 test                0          0       []        Online      74     74     0    [] 

```

nicholastmosher:
Closes #1139.

- Adds `fluvio produce <topic> --raw` flag to indicate that input should be sent as one full record, rather than broken down by lines.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Release Notes
 
+## Unreleased
+* Add `--raw` flag to `fluvio produce` for sending whole file input ([#1149](https://github.com/infinyon/fluvio/pull/1148))
+
 ## Platform Version 0.8.4 - 2020-05-29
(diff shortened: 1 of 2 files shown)


sehz (changes requested):
minor nit.  Can you add to changelog ?

sehz (changes requested):
This should go into 0.8.4

nicholastmosher:
Oh whoops

sehz (approved):

PR #1151, opened 2021-06-04
Separate unstable tests from CI

### LINKED ISSUE #1078: Separate CI workflow from unstable test workflow
Currently we have a whole host of tests in `ci.yml` that are unstable and regularly failing, but which are not required in order for us to accept PRs and merge things into master. However, since we are actively developing jobs that are failing in the `ci.yml` file, our GitHub build badge is always failing. The badge statuses are specific to a particular workflow file, so if we take all of the unstable test jobs and put them into a different workflow file such as `unstable.yml`, then our CI badge should properly reflect that it passes. We can always promote unstable tests to `ci.yml` in the future when they're ready.

nicholastmosher:
Closes #1078.

- Remove unstable tests from CI. From now on, the CI workflow should _always_ pass (i.e. green check) in master
- Move those unstable tests into a new workflow file, `ci-unstable.yml`.

### DIFF
--- a/.github/workflows/ci-unstable.yml
+++ b/.github/workflows/ci-unstable.yml
@@ -1,0 +1,260 @@
+name: Unstable CI
+
+permissions:
+  contents: read
+
+on:
+  push:
+  pull_request:
+    branches: [master]
+  workflow_dispatch:
+    inputs:
+      verbose:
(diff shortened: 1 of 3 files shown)


nicholastmosher:
Ah, need to push on upstream, closing.

PR #1161, opened 2021-06-06
[Merged by Bors] - Allow setting ingress when not using NodePort

nacardin:
Currently errors out with `SPU service missing NodePort` when service is of type `ClusterIP`


tjtelan (approved):
It took me a moment to see the importance this change. We only need to check for a NodePort if that's the configured load balancer type. Otherwise you'll always error on the NodePort check.

LGTM.

sehz:
Can you add this to changelog?

PR #1162, opened 2021-06-07
[Merged by Bors] - Optimize upgrade test by caching Docker image between jobs

### LINKED ISSUE #1090: optimize upgrade test
Currently,  upgrade tests take too long (now it takes long as 30min) since it is building image in the GitHub VM which is very slow.   However, build image is already built using another job `k8_cluster_test`.   So here is step to optimize:
* Save either binaries or image (without publishing to docker hub)  in the GitHub artifact
* Re-use in the `upgrade_test`.
* Once this is done, we can move step: `Publish image` from `k8_cluster_test` into separate job.

nicholastmosher:
Closes #1090.

- Creates an independent `build_image` step to build our Docker image and upload it to GH artifacts
- Makes k8_cluster_test and k8_upgrade_test depend on `build_image` and download the cached image to run tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -432,7 +432,7 @@ jobs:
       - name: Run smoke-test-k8-tls-root
         timeout-minutes: 5
         run: |
-          make UNINSTALL=noclean smoke-test-k8-tls-root-nobuild
+          make TEST_BIN=./flv-test UNINSTALL=noclean smoke-test-k8-tls-root-nobuild
 
       - name: Publish image
         if: ${{ env.RELEASE }} == 'true'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -432,17 +432,7 @@ jobs:
       - name: Run smoke-test-k8-tls-root
         timeout-minutes: 5
         run: |
-          kubectl create configmap authorization --from-file=POLICY=./src/sc/test-data/auth_config/policy.json --from-file=SCOPES=./src/sc/test-data/auth_config/scopes.json
-          FLV_SPU_DELAY=5 ./flv-test \
-            smoke \
-            --spu 2 \
-            --tls \
-            --develop \
-            --client-log warn --server-log fluvio=debug \
-            --authorization-config-map authorization \
-            --keep-cluster \
-            -- \
-            --producer-iteration=1000
+          make UNINSTALL=noclean smoke-test-k8-tls-root-nobuild
 
       - name: Publish image
         if: ${{ env.RELEASE }} == 'true'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,33 +29,7 @@ jobs:
         with:
           access_token: ${{ secrets.GITHUB_TOKEN }}
 
-  rustfmt:
-    name: Rustfmt (${{ matrix.os }})
-    runs-on: ${{ matrix.os }}
-    strategy:
-      matrix:
-        os: [ubuntu-latest]
-        rust: [stable]
-    env:
-      RUST_BACKTRACE: full
-      RUSTC_WRAPPER: sccache
-      SCCACHE_CACHE_SIZE: 300M
-      SCCACHE_DIR: /home/runner/.cache/sccache
-      SCCACHE_IDLE_TIMEOUT: 0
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -294,20 +398,54 @@ jobs:
...
-          make RELEASE=true TARGET=x86_64-unknown-linux-musl  k8-setup
-      - name: Make image
-        run: make RELEASE=true minikube_image
+          make FLUVIO_BIN=./fluvio RELEASE=true TARGET=x86_64-unknown-linux-musl k8-setup
+
       - name: Run smoke-test-k8-tls-root
         timeout-minutes: 5
         run: |
-          make RELEASE=true TARGET=x86_64-unknown-linux-musl UNINSTALL=noclean smoke-test-k8-tls-root
+          kubectl create configmap authorization --from-file=POLICY=./src/sc/test-data/auth_config/policy.json --from-file=SCOPES=./src/sc/test-data/auth_config/scopes.json
```
sehz, line 435: don't duplicate makefile steps otherwise this will cause issue when makefile is changed

.github/workflows/ci.yml
```diff
@@ -441,14 +533,35 @@ jobs:
...
           minikube status
-
       - name: Setup for upgrade test
         run: |
           gh auth login --with-token < <(echo ${{ secrets.GITHUB_TOKEN }})
           curl -fsS https://packages.fluvio.io/v1/install.sh | bash
 
-      - name: Start sccache server
-        run: sccache --start-server
+      # Download artifacts
```
sehz, line 541: I am concerned this seems to be duplicated from other steps. can it re-use same steps?

sehz (changes requested):

.github/workflows/ci.yml
nicholastmosher: I was going to ask about this. The Make recipe that was run here previously was `make smoke-test-k8-tls-root`, but in the dependencies of that Make task, it also executes `build_test` which executes `build_cluster` and `build_cli`. So what ends up happening is a full rebuild of all our components, when what we actually want is to just run the version of `flv-test` that we just downloaded from the artifact cache.

I will see if there is a way to prevent running the dependencies for the Make recipe, but this was done specifically to prevent a full rebuild.

.github/workflows/ci.yml
nicholastmosher: I don't think there's a way to abstract a group of steps unfortunately. I think this is about as compact as this can be

nicholastmosher:
Opened https://github.com/infinyon/fluvio/pull/1165 to resolve some of the Makefile duplication. Merge this after that

nicholastmosher:
@sehz this looks ready to go

sehz (approved):
LGTM.  Still it seems take >20m

nicholastmosher:
Yes, unfortunately the bottleneck is now just the plain-old compile speed. The critical path is now just:

- Compile the binaries `fluvio`, `fluvio-run`, `flv-test`
- Build the docker image with `fluvio-run`
- Run the integration tests (k8_cluster_tests, cluster_upgrade_tests)

.github/workflows/ci.yml
```diff
@@ -95,6 +150,14 @@ jobs:
       # SCCACHE_RECACHE: 1 # Uncomment this to clear cache, then comment it back out
     steps:
       - uses: actions/checkout@v2
+
+      # If this job is being run by Bors (it was pushed to staging),
+      # then build and run in release mode
+      - name: Set RELEASE mode
+        if: github.event_name == 'push' && github.ref == 'refs/heads/staging'
+        run: |
+          echo "RELEASE=true" | tee -a $GITHUB_ENV
```
tjtelan, line 159: Does this value need to persist between jobs? It seems like you check that this var is set from the `k8_cluster_test` job.

You can do that with `::set_output`. Check out the Release workflow for example.

nicholastmosher: Hmm you're right. I think I can just go back to checking `github.event_name == 'push' && github.ref == 'refs/heads/staging'`

I think I was trying to be slick and just check RELEASE because it better describes what we we're really interested in, but it's probably better to not try to be fancy, it'd probably make things more brittle

Good catch by the way, nice eye

PR #1169, opened 2021-06-14
[Merged by Bors] - Build tests when not in CI

nicholastmosher:
GitHub Actions CI always sets the env variable `CI=true`, so this change leverages that for running smoke-tests from the Makefile. 

- In development mode (i.e. no CI=true), the `make smoke-test` family of commands will build tests as needed.
- In CI mode (i.e. CI=true), `make smoke-test` will not run any build commands, and use cached artifacts instead


sehz (approved):
Nice job!

PR #1170, opened 2021-06-14
Consumer wasm api

nicholastmosher:
Closes https://github.com/infinyon/fluvio-client-wasm/issues/32

- Adds `SmartStreamModule` type for WASM binaries given to the consumer by users
- Turn `ConsumerConfig` into a proper builder that can take WASM by binary or base64 (for web use-cases)

### DIFF
--- a/src/client/src/consumer.rs
+++ b/src/client/src/consumer.rs
@@ -133,8 +133,10 @@ impl PartitionConsumer {
     /// # use fluvio::{PartitionConsumer, FluvioError, Offset, ConsumerConfig};
     /// # async fn example(consumer: &PartitionConsumer) -> Result<(), FluvioError> {
     /// // Use custom fetching configurations
-    /// let fetch_config = ConsumerConfig::default()
-    ///     .with_max_bytes(1000);
+    /// let config = ConsumerConfig::builder()
+    ///     .max_bytes(1000)
+    ///     .build()
+    ///     .unwrap();
     ///
     /// let response = consumer.fetch_with_config(Offset::beginning(), fetch_config).await?;
     /// for batch in response.records.batches {
@@ -277,9 +279,11 @@ impl PartitionConsumer {
     /// # async fn example(consumer: &PartitionConsumer) -> Result<(), FluvioError> {
     /// use futures::StreamExt;
     /// // Use a custom max_bytes value in the config
-    /// let fetch_config = ConsumerConfig::default()
-    ///     .with_max_bytes(1000);
-    /// let mut stream = consumer.stream_with_config(Offset::beginning(), fetch_config).await?;
+    /// let config = ConsumerConfig::builder()
+    ///     .max_bytes(1000)
+    ///     .build()
+    ///     .unwrap();
+    /// let mut stream = consumer.stream_with_config(Offset::beginning(), config).await?;
     /// while let Some(Ok(record)) = stream.next().await {
     ///     let key: Option<String> = record.key().map(|key| String::from_utf8_lossy(key).to_string());
     ///     let value = String::from_utf8_lossy(record.value());
@@ -329,9 +333,11 @@ impl PartitionConsumer {
     /// # async fn example(consumer: &PartitionConsumer) -> Result<(), FluvioError> {
     /// use futures::StreamExt;
     /// // Use a custom max_bytes value in the config
-    /// let fetch_config = ConsumerConfig::default()
-    ///     .with_max_bytes(1000);
-    /// let mut stream = consumer.stream_batches_with_config(Offset::beginning(), fetch_config).await?;
+    /// let config = ConsumerConfig::builder()
+    ///     .max_bytes(1000)
+    ///     .build()
+    ///     .unwrap();
+    /// let mut stream = consumer.stream_batches_with_config(Offset::beginning(), config).await?;
     /// while let Some(Ok(batch)) = stream.next().await {
     ///     for record in batch.records() {
     ///         let key = record.key.as_ref().map(|key| String::from_utf8_lossy(key.as_ref()).to_string());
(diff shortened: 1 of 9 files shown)


sehz:
CI failing?

sehz:
Can we not require bump version? Can deprecated existing API

src/client/src/consumer.rs
```diff
@@ -410,12 +416,13 @@ impl PartitionConsumer {
             .lookup_version(DefaultStreamFetchRequest::API_KEY)
             .unwrap_or((WASM_MODULE_API - 1) as i16);
 
-        if !config.wasm_module.is_empty() {
-            if stream_fetch_version >= WASM_MODULE_API as i16 {
-                stream_request.wasm_module = config.wasm_module;
-            } else {
+        if let Some(wasm) = config.smartstream_filter {
+            println!("USING SMARTSTREAM WITH WASM SIZE: {}", wasm.binary.len());
```
simlay, line 420: I think you want `debug!`.

nicholastmosher: Ah, I actually need to remove that, good catch

nicholastmosher:
@sehz I reverted deleting the old methods and I marked them as deprecated. Changed crate bump as `0.8.5 -> 0.8.6`. I also think I fixed CI, but we'll see. CHANGELOG has also been updated. If CI passes this should be good to go.

src/client/src/consumer.rs
```diff
@@ -564,38 +571,159 @@ static MAX_FETCH_BYTES: Lazy<i32> = Lazy::new(|| {
     max_bytes
 });
 
+#[derive(Debug, Clone)]
+pub struct SmartStreamModule {
```
sehz, line 575: Instead of creating another wrapper, rather use existing fluvio data type `RecordData` instead of bytes.  This will also come in handy when storing WASM bytes in the Fluvio later on.

src/client/src/consumer.rs
```diff
@@ -564,38 +571,159 @@ static MAX_FETCH_BYTES: Lazy<i32> = Lazy::new(|| {
...
+impl SmartStreamModule {
+    /// Instantiate a SmartStream from raw WASM binary
+    pub fn from_binary<T: Into<Vec<u8>>>(binary: T) -> Self {
+        Self {
+            binary: binary.into(),
+        }
+    }
+
+    /// Instantiate a SmartStream from a base64-encoded string
+    pub fn from_base64<S: Into<String>>(string: S) -> Result<Self, FluvioError> {
```
sehz, line 588: This seems too premature to add as supporting API.  This conversion can be done by caller.

src/client/src/consumer.rs
```diff
@@ -564,38 +571,159 @@ static MAX_FETCH_BYTES: Lazy<i32> = Lazy::new(|| {
...
     /// Maximum number of bytes to be fetched at a time.
+    #[deprecated(since = "0.8.6", note = "Use 'ConsumerConfig::builder()' instead")]
     pub fn with_max_bytes(mut self, max_bytes: i32) -> Self {
         self.max_bytes = max_bytes;
         self
     }
 
     /// set wasm filter
+    #[deprecated(since = "0.8.6", note = "Use 'ConsumerConfig::builder()' instead")]
     pub fn with_wasm_filter(mut self, bytes: Vec<u8>) -> Self {
```
sehz, line 670: use RecordData

sehz (changes requested):

src/client/src/consumer.rs
nicholastmosher: I think it would not be a good idea to overload `RecordData` to mean two different things. I also think that it is actually a good idea to have this `SmartStreamModule` type because we can expand on it in the future, e.g. validate that it is correct WASM before instantiating it. If you would prefer not to have a new type for this right now, I can turn it back into a `Vec<u8>`, but I think using `RecordData` is a bad idea.

src/client/src/consumer.rs
nicholastmosher: See my comment above

sehz:
https://github.com/infinyon/fluvio-client-wasm/issues/32 need further work.   We need further design work on how to configure and pass WebAssembly.  So we should close this for now

nicholastmosher:
@sehz I minimized this some more based on discussion today, let me know if it looks alright.

sehz:
CI failing?

sehz (changes requested):
Given this seems only slight better API but may add more tech debt, is this PR necessary? I would rather tackle this when we need to start adding compression or other features.

SmartStream config may need it's own builder (SmartStreamConfigBuilder?)

src/client/src/consumer.rs
```diff
@@ -565,33 +572,67 @@ static MAX_FETCH_BYTES: Lazy<i32> = Lazy::new(|| {
...
+    ///     .unwrap();
+    /// ```
+    #[builder(default = "*MAX_FETCH_BYTES")]
     pub(crate) max_bytes: i32,
+    #[builder(setter(skip))]
     pub(crate) isolation: Isolation,
-    wasm_module: Vec<u8>,
+    /// A WASM module to use as a SmartStream filter
+    #[builder(setter(into, strip_option), default)]
+    smartstream_filter: Option<Vec<u8>>,
```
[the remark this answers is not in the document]
nicholastmosher: Option makes more sense here IMO. Seeing this in the struct as `smartstream_filter: Vec<u8>` makes me think that the field will always be populated, whereas an `Option<Vec<u8>>` semantically lets me know that I should expect to check whether a value is present or not. Note that this also does not have any runtime cost, because any type containing a pointer (like vec does) inside an Option will be niche-optimized: https://stackoverflow.com/questions/16504643/what-is-the-overhead-of-rusts-option-type

nicholastmosher:
I noticed that now that I stopped using the `Rc<RefCell<...>>` pattern in fluvio-client-wasm that these changes aren't necessary any more. We can revisit config in the future.

PR #1171, opened 2021-06-16
[Merged by Bors] - Give user feedback when consuming in interactive mode

### LINKED ISSUE #700: CLI: `fluvio consume` should give user feedback

nicholastmosher:
Closes #700 

When using the Consumer CLI in "interactive mode", the CLI will now print a little message about how it is going to consume.

image

The most important one here is in the case that no flags are given, e.g. `fluvio consume one`. This was confusing to the user in the past because it would appear to just hang as if it were loading, when it was actually consuming from the end of the partition and just waiting for records to arrive. Now there's a message to say that.


sehz:
conflict

nicholastmosher:
@sehz should be ready now

sehz (approved):
nice improvement

nicholastmosher:
This seems to be spurious CI failure, going to try again

image

bors r+

PR #1179, opened 2021-06-18
[Merged by Bors] - Relax mutability requirements of FluvioAdmin client

### LINKED ISSUE #1178: Relax mutability requirements for Fluvio Admin client
Right now, the methods on the FluvioAdmin client all take `&mut self` reference:

```rust
impl AdminClient {
    pub async fn create<...>(&mut self, ...) { ... }
    pub async fn delete<...>(&mut self, ...) { ... }
    pub async fn list<...>(&mut self, ...) { ... }
}
```

However, after digging around a little bit, I discovered that these methods do not actually require mutable access to any internal state (everything is already synchronized), so we can actually relax these methods to take `&self`.

```rust
impl AdminClient {
    pub async fn create<...>(&self, ...) { ... }
    pub async fn delete<...>(&self, ...) { ... }
    pub async fn list<...>(&self, ...) { ... }
}
```

This will be helpful for use-cases like the `fluvio-client-wasm` wrapper around the Admin API, because we can reduce the wrapper code from `Rc<RefCell<FluvioAdmin>>` (which is currently causing panics due to multiple borrowing of RefCell) down to `Rc<FluvioAdmin>`.

nicholastmosher:
Closes #1178. Will help to prevent bugs like https://github.com/infinyon/fluvio-client-wasm/issues/42

- Changes `&mut self` to `&self` for `create`, `delete`, and `list` methods on `FluvioAdmin`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Platform Version 0.8.5 - UNRELEASED
 * Add unstable Admin Watch API for topics, partitions, and SPUs ([#1136](https://github.com/infinyon/fluvio/pull/1136))
 * Make recipes for smoke tests no longer build by default, helps caching. ([#1165](https://github.com/infinyon/fluvio/pull/1165))
+* Relax requirement of `FluvioAdmin` methods from `&mut self` to `&self`. ([#1178](https://github.com/infinyon/fluvio/pull/1178))
  
 ## Platform Version 0.8.4 - 2020-05-29
 * Don't hang when check for non exist topic. ([#697](https://github.com/infinyon/fluvio/pull/697))
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1337,7 +1337,7 @@ dependencies = [
 
 [[package]]
 name = "fluvio"
-version = "0.8.5"
+version = "0.8.6"
 dependencies = [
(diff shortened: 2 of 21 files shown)


sehz (changes requested):
Is this backwards compatible or requires major bump?

nicholastmosher:
I believe this is totally backwards compatible. Passing a `&mut T` where a `&T` is expected is totally valid. The only thing that will change for callers is they may get a warning if they do the following:

```rust
let mut admin = fluvio.admin().await; // Warning: `admin` does not need to be mutable
admin.list::<TopicSpec>(vec![]).await?;
```

nicholastmosher:
Let me see how hard it's going to be to fix clippy

sehz:
Can you put 1.53 clippy fix as separate PR?

sehz (approved):
LGTM.  Nice improvement!

morenol (approved):

nicholastmosher:
Should pass now that clippy fixes are merged.

bors r+

nicholastmosher:
Dang, I think something wonky happened with the macos job and caused it to take a really long time. Let's see if that was a fluke. The workflow was like 2 minutes from being done when Bors timed out 😭 

bors r+

Issue #1177, opened 2021-06-17
Add a SmartStream filter example using Regex


nicholastmosher:
It would be nice to demonstrate a SmartStream accepting or rejecting records based on using a Regex to search for substrings in the Record. I'm imagining a regex that searches for `XXX-XX-XXXX` (imagine scrubbing social-security numbers), and rejecting matching records.

@ajhunyady feel free to post any example scenarios you think would be good to show off with regex.

ajhunyady:
@nicholastmosher, nice use case!

PR #1174, opened 2021-06-16
[Merged by Bors] - Smartstream map

nicholastmosher:
Closes #1172 

This is updated since our first iteration of SmartStreams with filter. Users can write a mapping SmartStream like the following now:

```rust
use fluvio_smartstream::{smartstream, Record, RecordData};

#[smartstream(map)]
fn map(record: &Record) -> (Option<RecordData>, RecordData) {
    let key = record.key.cloned();
    let mut value = Vec::from(record.value.as_ref());

    value.make_ascii_uppercase();
    (key, value.into())
}
```

This gives the following results:

```
flvd consume test -B
one
two
Three
fouR
```

```
❯ flvd consume test -B --smart-stream=examples/target/wasm32-unknown-unknown/release/fluvio_wasm_map.wasm
ONE
TWO
THREE
FOUR
```

PR #1183, opened 2021-06-21
[Merged by Bors] - Smartstream map regex

nicholastmosher:
This is a continuation of #1177 that uses a `#[smartstream(map)]` rather than a filter. This converts any instance of a string that looks like a social-security number into a scrubbed version `***-**-****`. It needed to be a PR because it depends on the SmartStream Map API #1174 to be merged first.

image

### DIFF
--- a/src/smartstream/examples/map_regex/src/lib.rs
+++ b/src/smartstream/examples/map_regex/src/lib.rs
@@ -1,0 +1,15 @@
+use regex::Regex;
+use once_cell::sync::Lazy;
+use fluvio_smartstream::{smartstream, Result, Record, RecordData};
+
+static SSN_RE: Lazy<Regex> = Lazy::new(|| Regex::new(r"\d{3}-\d{2}-\d{4}").unwrap());
+
+#[smartstream(map)]
+pub fn map(record: &Record) -> Result<(Option<RecordData>, RecordData)> {
+    let key = record.key.clone();
+
+    let string = std::str::from_utf8(record.value.as_ref())?;
+    let output = SSN_RE.replace_all(string, "***-**-****").to_string();
+
+    Ok((key, output.into()))
+}
--- a/src/smartstream/examples/Cargo.lock
+++ b/src/smartstream/examples/Cargo.lock
@@ -97,16 +97,16 @@ dependencies = [
  "flv-util",
  "futures-util",
  "once_cell",
- "semver 1.0.3",
+ "semver 1.0.4",
  "thiserror",
  "tracing",
 ]
 
 [[package]]
 name = "fluvio-future"
-version = "0.3.6"
+version = "0.3.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b0212c5fab32e362715ed8f08424e57aaf764a4ce206d6b6bf61fe23bd819371"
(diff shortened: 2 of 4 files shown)


sehz:
Can you link this to issue?

nicholastmosher:
@sehz this has been rebased and I just re-tested it to make sure everything is good. If CI passes it's ready to go

src/smartstream/examples/map_regex/src/lib.rs
```diff
@@ -0,0 +1,20 @@
+use fluvio_smartstream::{smartstream, Record, RecordData};
+use regex::Regex;
+
+#[smartstream(map)]
+pub fn map(record: &Record) -> (Option<RecordData>, RecordData) {
+    let key = record.key.clone();
+
+    let string_result = std::str::from_utf8(record.value.as_ref());
```
sehz, line 8: So maybe we should have conversion to string like this: `let string_result: String  = record.value.try_into()?`.
This also indicates that we need to return Result: `Result<(Option<RecordData>, RecordData)>` where Result will be smart stream result type.

src/smartstream/examples/map_regex/src/lib.rs
```diff
@@ -0,0 +1,20 @@
...
+    let string = match string_result {
+        Ok(s) => s,
+        Err(_) => return (key, record.value.clone()),
+    };
+
+    let social_security_regex = Regex::new(r"\d{3}-\d{2}-\d{4}").unwrap();
```
sehz, line 14: this should be declared as constant or outside function.  Creating regex every time in the loop is costly

sehz (changes requested):

src/smartstream/examples/map_regex/src/lib.rs
nicholastmosher: I thought about that, I was originally going to put this into a `once_cell::sync::Lazy`, but then I didn't know whether the Wasm module was re-instantiated every time a record was passed or not. If we put this into a `static Lazy<Regex>` will it be preserved between executions? I think it will at least for a batch at a time now that I think about it. I'll go ahead and add that.

src/smartstream/examples/map_regex/src/lib.rs
nicholastmosher: If we made the return type `-> Result<(Option<RecordData>, RecordData)>`, what would you expect the value in the stream to be when an `Err` is returned?

Example:

```
Topic: integers =          [16, 9, 4, 1, 0, -1]
Consumer squareroot.wasm = [4, 3, 2, 1, 0, ?]
```

If we had a consumer stream that took the square root of all the input numbers, then we can imagine our `squareroot` function returns `Err` when given a number less than 0. How would we want that to be represented in the output stream?

---

In the `filter`-style SmartStreams, the way I handled this was by early-returning `false` whenever some step failed, essentially filtering out all malformatted data. For Map I sort of ended up following the same approach, where if any step fails we simply return the input data unmodified. It does seem like we might want to have a better way of handling this.

nicholastmosher: Error handling strategy tracked in #1198

nicholastmosher:
@sehz I have updated this with changes from error handling, which were what was blocking this. This should be ready to merge whenever 👍

sehz (approved):
Looks great!

PR #1188, opened 2021-06-24
[Merged by Bors] - Clean up stream_fetch code

nicholastmosher:
This is just a small cleanup task as I am working on aggregate SmartStreams. Right now all of the SmartStream engine stuff is in the `filter.rs` module in the SPU, but if we are going to be reusing it (e.g. in `aggregate.rs`) then we should move this stuff up one level to be more "common" code.

There should not be any changes in here except for moving the location of this code.

### DIFF
--- a/src/spu/src/services/public/stream_fetch.rs
+++ b/src/spu/src/services/public/stream_fetch.rs
@@ -395,7 +395,7 @@ impl StreamFetchHandler {
             }
         } else {
             debug!("empty records, skipping");
-            return Ok((offset.isolation(&self.isolation), false));
+            Ok((offset.isolation(&self.isolation), false))
         }
     }
 }
--- a/src/spu/src/services/public/stream_fetch.rs
+++ b/src/spu/src/services/public/stream_fetch.rs
@@ -169,96 +169,86 @@ impl StreamFetchHandler {
         let mut leader_offset_receiver = self.leader_state.offset_listener(&self.isolation);
 
         let mut counter: i32 = 0;
-        let mut consumer_offset: Option<Offset> = if !consumer_wait {
-            // since we don't need to wait for consumer, can move consumer to same offset as last read
-            Some(last_read_offset)
-        } else {
-            None // offset for consumer
-        };
+        // since we don't need to wait for consumer, can move consumer to same offset as last read
+        let mut consumer_offset: Option<Offset> = (!consumer_wait).then(|| last_read_offset);
 
         loop {
             counter += 1;
-
             debug!(counter, ?consumer_offset, last_read_offset, "waiting");
 
             select! {
-
                 _ = self.end_event.listen() => {
                     debug!("end event has been received, terminating");
                     break;
                 },
 
-
                 changed_consumer_offset = self.consumer_offset_listener.listen() => {
+                    if changed_consumer_offset == INIT_OFFSET {
+                        continue;
+                    }
 
+                    if changed_consumer_offset >= last_read_offset {
+                        debug!(
+                            changed_consumer_offset,
+                            last_read_offset,
+                            "consume caught up"
+                        );
+                        consumer_offset = Some(changed_consumer_offset);
+                        continue;
+                    }
 
-                    if changed_consumer_offset != INIT_OFFSET  {
-
-                        // consume hasn't read all offsets, need to send back gaps
-                        if changed_consumer_offset < last_read_offset {
-                            debug!(
-                                changed_consumer_offset,
-                                last_read_offset,
-                                "need send back"
-                            );
-                            let (offset,wait) = self.send_back_records(changed_consumer_offset,module.as_ref()).await?;
-                            last_read_offset = offset;
-                            if wait {
-                                consumer_offset = None;
-                                debug!(
-                                    last_read_offset,
-                                    "wait for consumer"
-                                );
-                            } else {
-                                consumer_offset = Some(last_read_offset);   // no need wait for consumer, skip it
-                            }
-
-                        } else {
-                            debug!(
-                                changed_consumer_offset,
-                                last_read_offset,
-                                "consume caught up"
-                            );
-                            consumer_offset = Some(changed_consumer_offset);
-                        }
+                    // consume hasn't read all offsets, need to send back gaps
+                    debug!(
+                        changed_consumer_offset,
+                        last_read_offset,
+                        "need send back"
+                    );
+                    let (offset, wait) = self.send_back_records(changed_consumer_offset, module.as_ref()).await?;
+                    last_read_offset = offset;
+                    if wait {
+                        consumer_offset = None;
+                        debug!(
+                            last_read_offset,
+                            "wait for consumer"
+                        );
+                    } else {
+                        consumer_offset = Some(last_read_offset);   // no need wait for consumer, skip it
                     }
                 },
 
                 // received new offset from leader
                 leader_offset_update = leader_offset_receiver.listen() => {
-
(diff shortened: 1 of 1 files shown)


src/spu/src/services/public/stream_fetch.rs
```diff
@@ -169,96 +169,86 @@ impl StreamFetchHandler {
...
                 _ = self.end_event.listen() => {
                     debug!("end event has been received, terminating");
                     break;
                 },
 
-
                 changed_consumer_offset = self.consumer_offset_listener.listen() => {
+                    if changed_consumer_offset == INIT_OFFSET {
+                        continue;
+                    }
```
nicholastmosher, line 188: In this section, I am reducing right-drift by:
- Inverting the conditional to the large `if` statement
- Using `continue` in the new `if` block to short-circuit the following code (equivalent to not entering the original `if` body).

src/spu/src/services/public/stream_fetch.rs
```diff
@@ -169,96 +169,86 @@ impl StreamFetchHandler {
...
 
+                    if changed_consumer_offset >= last_read_offset {
+                        debug!(
+                            changed_consumer_offset,
+                            last_read_offset,
+                            "consume caught up"
+                        );
+                        consumer_offset = Some(changed_consumer_offset);
+                        continue;
```
nicholastmosher, line 197: Again reducing rightward-drift by using `continue`.

This is the `else` body from https://github.com/infinyon/fluvio/pull/1188/files#diff-5bcc5ec4673d6a5c8633195a7360c3402f873769c1faa90cf627b242179d2241L217-L222

src/spu/src/services/public/stream_fetch.rs
```diff
@@ -169,96 +169,86 @@ impl StreamFetchHandler {
...
+                            debug!(delay_consumer_offset = leader_offset_update);
                             last_read_offset = leader_offset_update;
-                        }
-                    } else {
-                        // we don't know consumer offset, so we delay
-                        debug!(delay_consumer_offset = leader_offset_update);
-                        last_read_offset = leader_offset_update;
+                            continue;
+                        },
+                    };
```
nicholastmosher, line 231: Here, reducing right-drift by using a match statement to unwrap `last_consumer_offset`. If the value is `None`, just `continue`.

src/spu/src/services/public/stream_fetch.rs
```diff
@@ -312,103 +302,100 @@ impl StreamFetchHandler {
...
-                    filter
-                        .filter(records.raw_slice(), self.max_bytes as usize)
-                        .map_err(|err| {
-                            IoError::new(ErrorKind::Other, format!("filter err {}", err))
-                        })?
-                };
+        if file_partition_response.records.len() == 0 {
+            debug!("empty records, skipping");
+            return Ok((offset.isolation(&self.isolation), false));
+        }
```
nicholastmosher, line 308: Reducing right-drift again. This is the contents of what used to be this `else` block: https://github.com/infinyon/fluvio/pull/1188/files#diff-5bcc5ec4673d6a5c8633195a7360c3402f873769c1faa90cf627b242179d2241L409

src/spu/src/services/public/stream_fetch.rs
```diff
@@ -312,103 +302,100 @@ impl StreamFetchHandler {
...
+            inner_sink
+                .encode_file_slices(&response_msg, self.header.api_version())
+                .await?;
+
+            drop(inner_sink);
+
+            debug!(read_time_ms = %now.elapsed().as_millis(),"finish sending back records");
 
-            Ok((offset.isolation(&self.isolation), false))
+            Ok((offset.isolation(&self.isolation), true))
```
nicholastmosher, line 398: This diff is misleading, I didn't just change a true to false. This `true` is actually from this line in the original source:

https://github.com/infinyon/fluvio/pull/1188/files#diff-5bcc5ec4673d6a5c8633195a7360c3402f873769c1faa90cf627b242179d2241L406

src/spu/src/services/public/stream_fetch.rs
```diff
@@ -312,103 +302,100 @@ impl StreamFetchHandler {
...
+                        filter_response,
+                    );
 
-                trace!("sending back file fetch response msg: {:#?}", response_msg);
+                trace!("sending back filter respone: {:#?}", filter_response_msg);
 
                 let mut inner_sink = self.sink.lock().await;
                 inner_sink
-                    .encode_file_slices(&response_msg, self.header.api_version())
+                    .send_response(&filter_response_msg, self.header.api_version())
```
nicholastmosher, line 364: This diff is also misleading. This call to `inner_sink.send_response` corresponds to the following call in the original:

https://github.com/infinyon/fluvio/pull/1188/files#diff-5bcc5ec4673d6a5c8633195a7360c3402f873769c1faa90cf627b242179d2241L371-L373

nicholastmosher:
I think a lesson here is that unindenting code confuses git diff by a lot 😅

sehz:
CI failures?

nicholastmosher:
I do not know why CI is failing, I am unable to replicate on my machine. Let me try re-running the action

nicholastmosher:
The error we were seeing here comes from a subset of the changes I was trying to make, so I minimized this PR to only contain the cleanup that is not problematic. I am going to separately work on the other changes and bring them back in another PR once I've figured out the problem. If this passes CI it should be ready to go.

nicholastmosher:
@sehz this should be ready now

src/spu/src/services/public/stream_fetch.rs
sehz: not if this is clearer.  I suggest leave this alone.  If you want to update this, I think need unit test to cover

src/spu/src/services/public/stream_fetch.rs
sehz: Hmm. again, I suggest leave alone unless you have unit test to cover all path

sehz (changes requested):

nicholastmosher:
If you don't think the current tests cover these paths then I'd prefer to just close this. I don't want to get sidetracked writing unit tests for this since I think it would take a long time for me to fully grasp the scenario setup.

sehz:
other parts of code changes looks good and helpful.  Suggest just revert part that needs unit test

nicholastmosher:
@sehz I reverted the bigger changes that would have needed unit tests, how does this look now?

src/spu/src/services/public/stream_fetch.rs
```diff
@@ -331,53 +320,52 @@ impl StreamFetchHandler {
...
-                    debug!(next_offset, "filter, no records send back, skipping");
-                    false
-                };
+                let mut inner_sink = self.sink.lock().await;
+                inner_sink
+                    .send_response(&filter_response_msg, self.header.api_version())
+                    .await?;
 
-                Ok((next_offset, consumer_wait))
+                Ok((next_offset, true))
```
sehz, line 368: it's hard to follow change here.  let's revert this sections

sehz (changes requested):

src/spu/src/services/public/stream_fetch.rs
nicholastmosher: Reverted. The original reason I put this here is because up above, we check if `consumer_wait` is `false` and return if so, meaning that if we get to this point in the code, `consumer_wait` must be `true`. But it doesn't make a difference if we keep using the variable.

sehz (approved):

PR #1189, opened 2021-06-25
[Merged by Bors] - Extract SmartStreamEngine and Module from filter

nicholastmosher:
This is a bit of cleanup that was originally in #1188 but I have now split out. I must have fat-fingered something when I cleaned this up the first time because it caused the stream_fetch tests to fail before. I don't know what I changed but now these changes pass all of the tests locally for me, so I imagine they will pass CI (I hope).


sehz (approved):
LGTM

PR #1190, opened 2021-06-28
[Merged by Bors] - Extract FileBatchIterator from filter

nicholastmosher:
This is part of the SmartStream aggregator work. Right now there is code in `filter.rs` that will need to be shared with the aggregator, namely the `FileBatchIterator` and related types. I've moved these up to their own module.


sehz (approved):
LGTM

nicholastmosher:
Got the following error in CI, hoping it is spurious failure. Seems to be a problem downloading Zig dependency

image

bors r+

PR #1194, opened 2021-07-01
[Merged by Bors] - Smartstream aggregator

### LINKED ISSUE #1173: Add `#[smartstream(aggregate)]` API
Another type of SmartStream we want to support is one for aggregations (folding). My thinking right now is to just provide an "accumulator Record" to the SmartStream fold function on each iteration, and have the fold function return the new aggregate Record. For example:

```rust
use fluvio_smartstream::{smartstream, Record, RecordData};

#[smartstream(aggregate)]
fn aggregate(accumulator: &Record, next: &Record) -> RecordData {
    todo!()
}
```

Let's imagine that our input stream is a set of records that are UTF-8 strings containing numbers we want to add together

```
topic-numbers:
["1", "2", "3", "4", "5", "6", "7", "8", "9"]
```

We want to write an aggregation so that our output stream gives the following:

```
["1", "3", "6", "10", "15", "21", "28", "36", "45"]
```

The implementation of this SmartStream would look like the following:

```rust
use fluvio_smartstream::{smartstream, Record, RecordData};

#[smartstream(aggregate)]
fn aggregate(accumulator: &Record, next: &Record) -> RecordData {
    let acc_value = Vec::from(accumulator.value.as_ref());

    // Parse the accumulator's number
    let acc_number_result = std::str::from_utf8(acc_value)
        .and_then(|s| s.parse::<u32>());
    let acc_number: u32 = match acc_number_result {
        Ok(n) => n,
        Err(_) => return acc_value.into(),
    };

    // Parse the current record's number
    let next_number_result = std::str::from_utf8(next.value.as_ref())
        .and_then(|s| s.parse::<u32>());
    let next_number = match next_number_result {
        Ok(n) => n,
        Err(_) => return acc_value.into(),
    };

    let next_acc_number = acc_number + next_number;
    let next_acc_string = format!("{}", next_acc_number);
    next_acc_string.into()
}
```

---

## CLI

In order to support aggregation, we'll make the following changes to the CLI:

```
fluvio consume <topic> --smart-stream // DEPRECATE
fluvio consume <topic> --filter=filter.wasm
fluivo consume <topic> --map=map.wasm
fluvio consume <topic> --aggregate=aggregator.wasm --initial=initial.txt
```

nicholastmosher:
Closes #1173

### DIFF
--- a/src/cli/src/consume/mod.rs
+++ b/src/cli/src/consume/mod.rs
@@ -88,7 +88,7 @@ pub struct ConsumeOpt {
 
     /// (Optional) Path to a file to use as an initial accumulator value with --aggregate
     #[structopt(long)]
-    pub accumulator: Option<PathBuf>,
+    pub initial: Option<PathBuf>,
 }
 
 impl ConsumeOpt {
@@ -136,7 +136,7 @@ impl ConsumeOpt {
             builder.wasm_map(buffer);
         }
 
-        match (&self.aggregate, &self.accumulator) {
+        match (&self.aggregate, &self.initial) {
             (Some(wasm_path), Some(acc_path)) => {
                 let wasm = std::fs::read(wasm_path)?;
                 let acc = std::fs::read(acc_path)?;
--- a/src/cli/src/consume/mod.rs
+++ b/src/cli/src/consume/mod.rs
@@ -141,8 +141,15 @@ impl ConsumeOpt {
                 let acc = std::fs::read(acc_path)?;
                 builder.wasm_aggregate(wasm, acc);
             }
+            (Some(wasm_path), None) => {
+                let wasm = std::fs::read(wasm_path)?;
+                builder.wasm_aggregate(wasm, Vec::new());
+            }
+            (None, Some(_)) => {
+                println!("In order to use --accumulator, you must also specify --aggregate");
+                return Ok(());
+            }
             (None, None) => (),
-            _ => println!("Need both aggregator WASM and Accumulator"),
         }
 
         let consume_config = builder.build()?;
--- a/src/cli/src/consume/mod.rs
+++ b/src/cli/src/consume/mod.rs
@@ -86,6 +86,7 @@ pub struct ConsumeOpt {
     #[structopt(long, group("smartstream"))]
     pub aggregate: Option<PathBuf>,
 
+    /// (Optional) Path to a file to use as an initial accumulator value with --aggregate
     #[structopt(long)]
     pub accumulator: Option<PathBuf>,
 }
--- a/src/cli/src/consume/mod.rs
+++ b/src/cli/src/consume/mod.rs
@@ -81,6 +81,13 @@ pub struct ConsumeOpt {
     /// Path to a SmartStream map wasm file
     #[structopt(long, group("smartstream"))]
     pub map: Option<PathBuf>,
+
+    /// Path to a WASM file for aggregation
+    #[structopt(long, group("smartstream"))]
+    pub aggregate: Option<PathBuf>,
+
+    #[structopt(long)]
+    pub accumulator: Option<PathBuf>,
 }
 
 impl ConsumeOpt {
@@ -126,6 +133,16 @@ impl ConsumeOpt {
             let buffer = std::fs::read(map_path)?;
             debug!(len = buffer.len(), "read filter bytes");
             builder.wasm_map(buffer);
+        }
+
+        match (&self.aggregate, &self.accumulator) {
+            (Some(wasm_path), Some(acc_path)) => {
+                let wasm = std::fs::read(wasm_path)?;
+                let acc = std::fs::read(acc_path)?;
+                builder.wasm_aggregate(wasm, acc);
+            }
+            (None, None) => (),
+            _ => println!("Need both aggregator WASM and Accumulator"),
         }
 
         let consume_config = builder.build()?;
--- a/src/dataplane-protocol/src/smartstream.rs
+++ b/src/dataplane-protocol/src/smartstream.rs
@@ -106,4 +106,11 @@ mod encoding {
             fmt::Debug::fmt(self, f)
         }
     }
+
+    /// Used to pass both the accumulator and records to an Aggregate smartstream
+    #[derive(Debug, Default, Encoder, Decoder)]
+    pub struct Aggregate {
+        pub accumulator: Vec<u8>,
+        pub records: Vec<u8>,
+    }
 }
--- a/src/dataplane-protocol/src/smartstream.rs
+++ b/src/dataplane-protocol/src/smartstream.rs
@@ -1,5 +1,5 @@
 pub use encoding::{
-    SmartStreamRuntimeError, SmartStreamType, SmartStreamBaseInput, SmartStreamAggregateInput,
+    SmartStreamRuntimeError, SmartStreamType, SmartStreamInput, SmartStreamAggregateInput,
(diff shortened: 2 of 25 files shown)


src/spu/src/services/public/stream_fetch.rs
```diff
@@ -955,4 +1006,107 @@ mod test {
...
+        BatchProducer::builder()
+            .records(records)
+            .record_generator(Arc::new(generate_aggregate_record))
+            .build()
+            .expect("batch")
+            .records()
+    }
+
+    #[fluvio_future::test(ignore)]
+    async fn test_stream_aggregate_fetch() {
```
nicholastmosher, line 1025: @sehz what do you think about this test?

src/spu/src/services/public/stream_fetch.rs
```diff
@@ -955,4 +1006,107 @@ mod test {
...
+
+            let mut accumulator = "0".repeat(100);
+
+            for i in 0..10 {
+                assert_eq!(batch.records()[i].value().as_ref(), accumulator.as_bytes());
+                assert_eq!(batch.records()[i].get_offset_delta(), i as i64);
+
+                accumulator.push_str(&(i + 1).to_string().repeat(100));
+            }
+        }
```
sehz, line 1108: this looks good for initial is empty.
need test cases for:
* initial is not empty
* max_bytes is applied
* interleaving producer

nicholastmosher:
@sehz I've got this all rebased against master, including the error handling changes. Would you like to review again?

src/spu/src/services/public/stream_fetch.rs
```diff
@@ -521,10 +458,76 @@ impl StreamFetchHandler {
 
                 debug!(read_time_ms = %now.elapsed().as_millis(),"finish sending back records");
 
-                Ok((offset.isolation(&self.isolation), true))
+                Ok((read_end_offset.isolation(&self.isolation), true))
             }
         }
     }
+
+    async fn send_processed_response(
```
sehz, line 466: add tracing to make it easier to debug

src/spu/src/services/public/stream_fetch.rs
```diff
@@ -521,10 +458,76 @@ impl StreamFetchHandler {
...
+        batch: Batch,
+        smartstream_error: Option<SmartStreamRuntimeError>,
+    ) -> Result<(Offset, bool), SocketError> {
+        type DefaultPartitionResponse = FetchablePartitionResponse<RecordSet>;
+
+        let error_code = match smartstream_error {
+            Some(error) => ErrorCode::SmartStreamError(SmartStreamError::Runtime(error)),
+            None => file_partition_response.error_code,
+        };
+        debug!("Filter error code output: {:#?}", &error_code);
```
sehz, line 479: either use `tracing` or implement debug trace for error code, so it can be like this:
```
debug!(error_code, "filter error code");
```

src/cli/src/consume/mod.rs
```diff
@@ -81,6 +81,13 @@ pub struct ConsumeOpt {
     /// Path to a SmartStream map wasm file
     #[structopt(long, group("smartstream"))]
     pub map: Option<PathBuf>,
+
+    /// Path to a WASM file for aggregation
+    #[structopt(long, group("smartstream"))]
+    pub aggregate: Option<PathBuf>,
+
+    #[structopt(long)]
+    pub accumulator: Option<PathBuf>,
```
sehz, line 90: is this value or function? I believe this is initial value

sehz (changes requested):
Looks good.  Several minor issues to address

src/cli/src/consume/mod.rs
nicholastmosher: The `--aggregate` is the function, and `--accumulator` is the starting value

sehz: `accumulator` implies it is sort of functional.  Suggest: `initial-value` or something like that.  For example, the rust fold use `init`:  https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.fold

src/dataplane-protocol/src/smartstream.rs
```diff
@@ -24,6 +36,36 @@ mod encoding {
...
+    // indicate the numbers of records, and negative values indicate various
+    // types of errors.
+    //
+    // THEREFORE, THE DISCRIMINANTS FOR ALL VARIANTS ON THIS TYPE MUST BE NEGATIVE
+    #[repr(i32)]
+    #[derive(thiserror::Error, Debug, Clone, PartialEq, Encoder, Decoder)]
+    #[fluvio(encode_discriminant)]
+    pub enum SmartStreamInternalError {
+        #[error("encountered unknown error during Smartstream processing")]
+        UnknownError = -1,
```
sehz, line 54: is there any reason why this is negative?

src/smartstream/derive/src/generator/filter.rs
```diff
@@ -21,15 +21,13 @@ pub fn generate_filter_smartstream(func: &SmartStreamFn) -> TokenStream {
                 let input_data = Vec::from_raw_parts(ptr, len, len);
                 let mut smartstream_input = fluvio_smartstream::dataplane::smartstream::SmartStreamInput::default();
                 if let Err(_err) = fluvio_smartstream::dataplane::core::Decoder::decode(&mut smartstream_input, &mut std::io::Cursor::new(input_data), 0) {
-                    // return fluvio_smartstream::ENCODING_ERROR;
-                    return -11;
+                    return fluvio_smartstream::dataplane::smartstream::SmartStreamInternalError::DecodingBaseInput as i32;
```
sehz, line 24: this is verbose. can you import `SmartStreamInternalError` at top?

sehz (changes requested):
couple of minor points

src/dataplane-protocol/src/smartstream.rs
nicholastmosher: I commented why right above on this struct 😂 

```
    // The presence of one of these errors most likely indicates a logic bug.
    // This error type is `#[repr(i32)]` because these errors are returned
    // as the raw return type of a Smartstream WASM function, i.e. the return
    // type in `extern "C" fn filter(ptr, len) -> i32`. Positive return values
    // indicate the numbers of records, and negative values indicate various
    // types of errors.
    //
    // THEREFORE, THE DISCRIMINANTS FOR ALL VARIANTS ON THIS TYPE MUST BE NEGATIVE
```

src/smartstream/derive/src/generator/filter.rs
nicholastmosher: It is bad practice to put imports in derive macros, because then those imports will pollute the user's code when they get imported. If you look at the macro expansions for a lot of popular crates they all use full crate paths like this.

src/dataplane-protocol/src/smartstream.rs
sehz: Got it.   Sorry didn't see that

src/smartstream/derive/src/generator/filter.rs
sehz: Not necessary.  You can put that in the beginning of the fn or block.

nicholastmosher: Ah, looking at this closer, we would not be mixing generated code into the user code anyway so it should be fine 👍

sehz (approved):
Excellent.   This will add nice new capability to SmartStream

PR #1203, opened 2021-07-06
[Merged by Bors] - move Mac as Tier2 Platform

sehz:
* Add lld-11 installation for ubuntu LTS 18.04
* Move Mac to Tier2 CI build


nicholastmosher:
Ah, this is because you removed `macos-latest` from the matrix but not from the `include` block

image

If you remove these three lines it should work

.github/workflows/ci-tier2.yml
```diff
@@ -0,0 +1,179 @@
...
+          - os: macos-latest
+            rust: stable
+            sccache-path: /Users/runner/Library/Caches/Mozilla.sccache
+            target: x86_64-apple-darwin
+            task:
+              name: Clippy
+              run: make check-clippy
+          - os: ubuntu-latest
+            sccache-path: /home/runner/.cache/sccache
+            target: x86_64-unknown-linux-musl
```
nicholastmosher: This should all be able to be gotten rid of, since rustfmt, clippy, and ubuntu-latest are run on Tier 1 CI

.github/workflows/publish.yml
```diff
@@ -94,7 +94,7 @@ jobs:
           ${HOME}/.fluvio/bin/fluvio package publish \
             --version="${VERSION}+${{ github.sha }}" \
             target/x86_64-unknown-linux-musl/release/fluvio-run \
-            target/x86_64-apple-darwin/release/fluvio-run
+         #   target/x86_64-apple-darwin/release/fluvio-run
```
nicholastmosher, line 97: Should this be deleted?

.github/workflows/ci.yml
```diff
@@ -261,6 +228,77 @@ jobs:
           name: flv-test-${{ matrix.target }}
           path: target/${{ matrix.target }}/release/flv-test
 
+
+  run_platform_unit_test:
```
nicholastmosher, line 232: Can we use a build matrix rather than copying this job?

sehz: Unfortunately, it mislead to wrong job name.    Instead, let's try to use custom action

PR #1208, opened 2021-07-07
[Merged by Bors] - Move Docker Push to after tests have passed in CI

### LINKED ISSUE #1205: CI: Docker publishing should not happen in the Cluster Test
see: https://github.com/infinyon/fluvio/runs/3004264042?check_suite_focus=true.

It should be happen as last step in the workflow when all other steps are successful. But currently, it is publishing even thought other test `Kubernetes cluster upgrade test` is still happening

nicholastmosher:
Closes #1205.

- Removes `docker push` commands from `k8_cluster_test`
- Adds a `docker_push` job that runs after all test jobs, this is now where `docker push` happens

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -485,15 +485,6 @@ jobs:
         run: |
           make FLUVIO_BIN=./fluvio TEST_BIN=./flv-test UNINSTALL=noclean smoke-test-k8-tls-root
 
-      - name: Publish image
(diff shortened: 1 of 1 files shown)


sehz:
rebase

sehz (approved):

nicholastmosher:
This one will be a slow one to test/fix, but here we go

bors r+

Issue #1039, opened 2021-04-29
Add Windows to CI


tjtelan:
Github actions offers 2 types of Windows runners. Only one of them has WSL (v1, but I'm not sure what the limitations are at this point)

https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md

I'm not sure what the shape of this initial solution will look like.
- Will the codebase compile in Windows?
- Will a server run in Windows runner in CI?
- Do I need a separate test for using the Linux binary in WSL?

sehz:
Fluvio does't have to compile nor run in Windows.  Just have to run in WSL

tjtelan:
Resolved by #1295

PR #1218, opened 2021-07-09
[Merged by Bors] - Initial windows support

simlay:
Depends on https://github.com/infinyon/future-aio/pull/85. Part of #1039.

But am not sure a better pattern. Thoughts?


Cargo.toml
```diff
@@ -55,3 +55,6 @@ debug-assertions = true
 overflow-checks = true
 incremental = false
 codegen-units = 256
+[patch.crates-io]
+#fluvio-future = { path = "../future-aio/" }
+fluvio-future = { git = "https://github.com/simlay/future-aio", branch = "add-windows-to-ci" }
```
sehz, line 60: we should use non forked version of future-aid or published version

src/auth/src/x509/authenticator.rs
```diff
@@ -51,10 +54,15 @@ impl X509Authenticator {
         tcp_stream: &TcpStream,
         authorization_request: AuthRequest,
     ) -> Result<bool, IoError> {
+        #[cfg(unix)]
```
sehz, line 57: use [connection_fd](https://github.com/infinyon/future-aio/blob/master/src/net/mod.rs)

src/client/Cargo.toml
```diff
@@ -48,7 +48,12 @@ dataplane = { version = "0.5.1", path = "../dataplane-protocol", package = "fluv
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 dirs = "1.0.2"
+
+[target.'cfg(unix)'.dependencies]
 fluvio-spu-schema = { version = "0.6.1", path = "../spu-schema", features = ["file"] }
+[target.'cfg(windows)'.dependencies]
+fluvio-spu-schema = { version = "0.6.1", path = "../spu-schema" }
```
sehz, line 55: we do need file support otherwise can't load it config

src/socket/src/socket.rs
```diff
@@ -121,5 +121,15 @@ cfg_if::cfg_if! {
                 Self::from_stream(Box::new(tcp_stream.clone()),Box::new(tcp_stream), fd)
             }
         }
+    } else if #[cfg(windows)] {
```
sehz, line 124: if you use connection_fd, this is not needed

sehz (changes requested):

src/auth/src/x509/authenticator.rs
simlay: I think this will make it cleaner. https://github.com/infinyon/future-aio/pull/80/files#diff-4c6a8ac4ba774ab01f57abae00bd7caef7c65b5bf09404c3f9a68b93fa06843bR60-R73

sehz: So may be we should get PR80 on `future-aio` merge first?

src/client/Cargo.toml
simlay: You mean the config in `~/.fluvio/config` (or whatever in windows)? I think that part works here.

sehz: ok

sehz (changes requested):
Looks good, just add to change log

sehz (approved):
Nice. Great work

simlay:
bors try

PR #1219, opened 2021-07-09
[Merged by Bors] - Upgrade to new wasmtime API

### LINKED ISSUE #1200: SPU: upgrade to Wasmtime 0.28.0
To convert SmartStream engine multi-thread capable by using: https://github.com/alexcrichton/rfcs-2/blob/new-api/accepted/new-api.md

nicholastmosher:
Closes #1200 

This upgrades to the new Wasmtime API that better supports Send + Sync and multithreaded environments. This should make it easier to finish up other smartstream work

### DIFF
--- a/src/spu/src/services/public/stream_fetch.rs
+++ b/src/spu/src/services/public/stream_fetch.rs
@@ -24,8 +24,9 @@ use fluvio_types::event::offsets::OffsetChangeListener;
 
 use crate::core::DefaultSharedGlobalContext;
 use crate::replication::leader::SharedFileLeaderState;
-use crate::smart_stream::{SmartStreamModule, SmartStreamEngine};
 use publishers::INIT_OFFSET;
+use wasmtime::{Engine, Module};
+use crate::smart_stream::filter::SmartFilter;
 
 /// Fetch records as stream
 pub struct StreamFetchHandler {
@@ -40,7 +41,7 @@ pub struct StreamFetchHandler {
     consumer_offset_listener: OffsetChangeListener,
     leader_state: SharedFileLeaderState,
     stream_id: u32,
-    sm_engine: SmartStreamEngine,
+    sm_engine: Engine,
     sm_bytes: Vec<u8>,
 }
 
@@ -96,7 +97,7 @@ impl StreamFetchHandler {
                 consumer_offset_listener: offset_listener,
                 stream_id,
                 leader_state: leader_state.clone(),
-                sm_engine: SmartStreamEngine::new(),
+                sm_engine: Engine::default(),
                 sm_bytes,
                 max_fetch_bytes,
             };
@@ -149,14 +150,14 @@ impl StreamFetchHandler {
         // and can't be send across Send
         let module = if !self.sm_bytes.is_empty() {
             Some(
-                self.sm_engine
-                    .create_module_from_binary(&self.sm_bytes)
-                    .map_err(|err| -> FlvSocketError {
+                Module::from_binary(&self.sm_engine, &self.sm_bytes).map_err(
+                    |err| -> FlvSocketError {
                         FlvSocketError::IoError(IoError::new(
                             ErrorKind::Other,
                             format!("module loading error {}", err),
                         ))
-                    })?,
+                    },
+                )?,
             )
         } else {
             None
@@ -273,7 +274,7 @@ impl StreamFetchHandler {
     async fn send_back_records(
         &mut self,
         offset: Offset,
-        module_option: Option<&SmartStreamModule>,
+        module_option: Option<&Module>,
     ) -> Result<(Offset, bool), FlvSocketError> {
         let now = Instant::now();
         let mut file_partition_response = FilePartitionResponse {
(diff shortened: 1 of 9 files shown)


src/spu/src/smart_stream/filter.rs
```diff
@@ -5,74 +5,62 @@ use std::io::Cursor;
...
 use crate::smart_stream::file_batch::FileBatchIterator;
 
 const FILTER_FN_NAME: &str = "filter";
 type FilterFn = TypedFunc<(i32, i32), i32>;
 
-impl SmartStreamModuleInner {
-    pub fn create_filter(&self) -> Result<SmartFilter> {
-        let callback = Arc::new(RecordsCallBack::new());
-        let callback2 = callback.clone();
+pub struct SmartFilter {
```
sehz, line 19: rename to `SmartStreamFilter`

src/spu/src/smart_stream/memory.rs
```diff
@@ -10,12 +10,16 @@ const MEMORY: &str = "memory";
 
 /// Copy a byte array into an instance's linear memory
 /// and return the offset relative to the module's memory.
-pub fn copy_memory_to_instance(bytes: &[u8], instance: &Instance) -> Result<isize, Error> {
+pub fn copy_memory_to_instance(
+    store: &mut Store<()>,
```
sehz, line 14: why need to pass store?

src/spu/src/smart_stream/mod.rs
```diff
@@ -1,92 +1,11 @@
...
-use crate::smart_stream::filter::SmartFilter;
+use std::sync::Mutex;
+use anyhow::Result;
+use wasmtime::{Memory, Store};
 
 mod memory;
 pub mod filter;
 pub mod file_batch;
 
-pub struct SmartStreamEngine(Engine);
```
sehz, line 13: Prefer to keep abstraction instead of exposing it directly

src/spu/src/smart_stream/mod.rs
```diff
@@ -95,22 +14,10 @@ pub struct RecordsMemory {
...
-            {
-                debug!(wasm_mem_len = self.len, "copying from wasm");
-                let mut bytes = vec![0u8; self.len as u32 as usize];
-                bytes.copy_from_slice(data);
-                bytes
-            } else {
-                vec![]
-            }
-        }
+    fn copy_memory_from(&self, store: &mut Store<()>) -> Result<Vec<u8>> {
```
sehz, line 17: why is store needed?

sehz (changes requested):
Thanks for migration efforts.  Few questions....

src/spu/src/smart_stream/memory.rs
nicholastmosher: In the new wasmtime API, a `&mut Store<T>` is required for calls to:

- `instance.get_memory`
- `instance.get_func`
- `alloc.call`
- `memory.data_ptr`

since all of this information is now tracked explicitly in a Store rather than implicitly through Rc inside these objects

src/spu/src/smart_stream/mod.rs
nicholastmosher: In the new API, a bunch of the wasmtime functions now take a `&mut Store<T>` where they didn't need to before. In this case, it is used in `self.memory.read(store, ...)`

src/spu/src/smart_stream/mod.rs
nicholastmosher: I can revert this but I think it won't do anything other than add noise where it is used, since the wasmtime API calls require `&Engine` directly.

nicholastmosher:
@sehz any other questions on this?

src/spu/src/services/public/stream_fetch.rs
```diff
@@ -149,14 +151,14 @@ impl StreamFetchHandler {
         // and can't be send across Send
         let module = if !self.sm_bytes.is_empty() {
             Some(
-                self.sm_engine
-                    .create_module_from_binary(&self.sm_bytes)
-                    .map_err(|err| -> FlvSocketError {
+                Module::from_binary(&self.sm_engine.0, &self.sm_bytes).map_err(
```
sehz, line 154: let's keep existing API

src/spu/src/services/public/stream_fetch.rs
```diff
@@ -310,9 +312,10 @@ impl StreamFetchHandler {
 
                 debug!("creating smart filter");
                 let filter_batch = {
-                    let filter = module.create_filter().map_err(|err| {
-                        IoError::new(ErrorKind::Other, format!("creating filter {}", err))
-                    })?;
+                    let mut filter =
```
sehz, line 315: let's keep existing API

sehz (changes requested):
Lets keep existing API much as possible.
also add to CHANGELOG.

sehz (approved):
LGTM.  Thanks

nicholastmosher:
Oh it worked, woohoo! 🎉

PR #1226, opened 2021-07-12
Review: Smartstream error collection and response

### LINKED ISSUE #1198: Error reporting for SmartStreams
The current API for SmartStreams does not provide much room for error management. Right now we have Map and Filter implemented with these APIs:

```rust
fn filter(record: &Record) -> bool {}
fn map(record: &Record) -> (Option<RecordData>, RecordData) {} // (Key, Value)
```

Both of these return types are "infallible", meaning that an implementor _must_ provide a "successful" return value. This is easy to hand-wave with Filter, because if something fails you can "just return false". But this approach will not support debugging very well, and furthermore it does not account for the possibility of failures in Map.

For example, in a Map, we may be expecting input records to be JSON, but receive records that do not parse correctly. We should have a way to indicate when errors occur and what the errors were.

So it seems right away that the APIs should look more like this:

```rust
fn filter(record: &Record) -> Result<bool, SmartStreamError> {} // Where the details of SmartStreamError are TBD
fn map(record: &Record) -> Result<(Option<RecordData>, RecordData), SmartStreamError> {}
```

> Note, another approach would be with a trait, e.g. `-> Result<bool, impl SmartStreamError>` for Filter

Another question that needs answering is what to do when a SmartStream panics. We will need to investigate how wasmtime propagates panics (trapping mechanism) and how to communicate information to the host during a panic.

We will have some constraints on the `SmartStreamError` type (or trait), based on the following limitations:

- Passing values over the WASM boundary essentially requires serializing to bytes and back
- This implies that we should have some type of bound like `SmartStreamError: Serialize + Deserialize` (either using serde or another encoding scheme such as `fluvio_protocol::{Encode, Decode}`

### Design Update

As of now, here is the plan for the error handling interface:

- Create a type `fluvio_smartstream::Result` which is aliased as `type Result<T> = eyre::Result<T>`
  - This is used because it implements `From<E> where E: std::error::Error`, i.e. any error type we are likely to see
- Create a wrapper `struct SmartStreamResult<T> where T: Encode + Decode` which wraps a `Result<T, String>`
  - When we get the `eyre::Result<T>` back from the SmartStream call, if it is `Err` we convert it to a Stringified representation of the error and wrap that in `SmartStreamResult<T>`
  - We also implement `SmartStreamResult<T>: Encode + Decode` so we can serialize it to pass across the WASM boundary.

---

The next question after figuring out how to capture and collect errors is how to report them. Some options are:

- Upon error, end the consumer stream with the error as the last item in the stream
- Upon error, inject the serialized error into the data stream and continue data processing
- Create a dedicated "errors" topic where errors are sent to be viewed later, preserving the original stream for "successful" data only

I think at this stage, the best course of action would be the first of these, as it will be the simplest to implement and begin to collect feedback for further use-cases. I imagine we will want to revisit this at some time in the future.

---

## Action Items

- [x] Design a type/trait `SmartStreamError` (name flexible) for reporting errors
- [x] Implement "error delivery" by halting a consumer stream and terminating it with the responsible error value

nicholastmosher:
Closes #1198, follows #1219

Opening as a draft for review. This PR has the following:

- Adds new types to the `StreamFetchRequest` to help distinguish between requests with different types of smartstreams
  - `SmartStreamPayload`
  - `SmartStreamKind`
  - `SmartStreamWasm`
- Adds a `SmartStreamError` type built for catching any kind of `std::error::Error` (using eyre)
- WIP: Adds an `Option<SmartStreamError>` into the partition response for delivering errors back to client

This is pretty quick-and-dirty so far, I am expecting to need to rearrange some things before we want to merge, suggestions welcome.


sehz:
Suggest to split this PR once #1219 is merged

nicholastmosher:
Sure 👍 I expect we will need to break this PR up into multiple, but it'd be nice to first prove that everything works end-to-end before trying to commit the pieces.

nicholastmosher:
This can be tested using the `filter_odd_error` smartstream example. It filters out odd numbers, and returns an error if:
- An input is not a UTF-8 string
- An input is not an int

An example flow for checking errors:

```
cd src/smartstream/examples
cargo build --release
cd -
flvd topic create two
flvd consume two -B --smart-stream=src/smartstream/examples/target/wasm32-unknown-unknown/release/fluvio_wasm_filter_odd.wasm

Then go produce some data
```

Some notes:

- Right now, the error message will not appear if you send messages 1 by 1 interactively, I have to figure out why
- However, sending a bunch of records _and then_ starting the consumer will prompt the error

src/client/src/error.rs
```diff
@@ -30,6 +30,8 @@ pub enum FluvioError {
         cluster_version: Version,
         client_minimum_version: Version,
     },
+    #[error("User's SmartStream failed with the following error: {0}")]
+    SmartStreamUserError(String),
```
sehz, line 34: Do we need increase semversion ?

src/dataplane-protocol/src/error_code.rs
```diff
@@ -6,47 +6,72 @@
 
 use flv_util::string_helper::upper_cammel_case_to_sentence;
 use fluvio_protocol::{Encoder, Decoder};
+use crate::smartstream::SmartStreamUserError;
 
 // -----------------------------------
 // Error Definition & Implementation
 // -----------------------------------
```
sehz, line 14: we should bump major version

src/dataplane-protocol/src/smartstream.rs
```diff
@@ -0,0 +1,41 @@
...
+
+mod encoding {
+    use crate::record::Record;
+    use fluvio_protocol::{Encoder, Decoder};
+    use std::fmt;
+
+    /// A type used to capture and serialize errors from within a SmartStream
+    #[derive(Debug, Default, Clone, PartialEq, Encoder, Decoder)]
+    pub struct SmartStreamUserError {
+        rendered: String,
```
sehz, line 14: we should at least model error similar to: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/RuntimeError

src/smartstream/derive/src/generator/filter.rs
```diff
@@ -0,0 +1,64 @@
...
+            pub unsafe fn filter(ptr: *mut u8, len: usize) -> i32 {
+                // DECODING
+                extern "C" {
+                    fn copy_records(putr: i32, len: i32);
+                }
+
+                let input_data = Vec::from_raw_parts(ptr, len, len);
+                let mut records: Vec<fluvio_smartstream::dataplane::record::Record> = vec![];
+                if let Err(_err) = fluvio_smartstream::dataplane::core::Decoder::decode(&mut records, &mut std::io::Cursor::new(input_data), 0) {
+                    return -1;
```
sehz, line 24: why -1 ? at least define constant

src/smartstream/derive/src/generator/filter.rs
```diff
@@ -0,0 +1,64 @@
...
+                            output.error = Some(fluvio_smartstream::dataplane::smartstream::SmartStreamUserError::from(e));
+                            break;
+                        }
+                    }
+                }
+
+                // ENCODING
+                let mut out = vec![];
+                if let Err(_) = fluvio_smartstream::dataplane::core::Encoder::encode(&mut output, &mut out, 0) {
+                    return -1;
```
sehz, line 51: use constant

src/smartstream/derive/src/generator/map.rs
```diff
@@ -0,0 +1,65 @@
...
+
+                // DECODING
+                extern "C" {
+                    fn copy_records(putr: i32, len: i32);
+                }
+
+                let input_data = Vec::from_raw_parts(ptr, len, len);
+                let mut records: Vec<fluvio_smartstream::dataplane::record::Record> = vec![];
+                if let Err(_err) = fluvio_smartstream::dataplane::core::Decoder::decode(&mut records, &mut std::io::Cursor::new(input_data), 0) {
+                    return -1;
```
sehz, line 25: use constant

src/smartstream/derive/src/generator/map.rs
```diff
@@ -0,0 +1,65 @@
...
+                            output.error = Some(fluvio_smartstream::dataplane::smartstream::SmartStreamUserError::from(e));
+                            break;
+                        }
+                    }
+                }
+
+                // ENCODING
+                let mut out = vec![];
+                if let Err(_) = fluvio_smartstream::dataplane::core::Encoder::encode(&mut output, &mut out, 0) {
+                    return -1;
```
sehz, line 52: use constant

src/smartstream/examples/filter/src/lib.rs
```diff
@@ -1,12 +1,7 @@
...
-    let string = match str_result {
-        Ok(s) => s,
-        _ => return false,
-    };
-
-    string.contains('a')
+pub fn filter(record: &Record) -> Result<bool> {
+    let string = std::str::from_utf8(record.value.as_ref())?;
+    Ok(string.contains('a'))
 }
```
sehz, line 7: nice example!

src/spu-schema/src/server/stream_fetch.rs
```diff
@@ -43,10 +48,138 @@ where
...
+    Map,
+}
+
+impl Default for SmartStreamKind {
+    fn default() -> Self {
+        Self::Filter
+    }
+}
+
+impl Encoder for SmartStreamKind {
```
sehz, line 78: Can you migrate to use derive instead of custom encoder now?

src/spu-schema/src/server/stream_fetch.rs
```diff
@@ -43,10 +48,138 @@ where
...
+}
+
+/// Different possible representations of WASM modules.
+///
+/// In a fetch request, a WASM module may be given directly in the request
+/// as raw bytes.
+///
+// TODO ... or, it may be named and selected from the WASM store.
+#[derive(Debug)]
+pub enum SmartStreamWasm {
```
sehz, line 126: migrate to use derive

src/spu/src/services/public/stream_fetch.rs
```diff
@@ -373,7 +403,13 @@ impl StreamFetchHandler {
                     .await?;
 
                 Ok((next_offset, consumer_wait))
-            } else {
+            }
+            Some(SmartStream::Map(map)) => {
+                warn!("HANDLING FETCH, GOT MAP");
+                todo!()
```
sehz, line 409: should be error

src/spu/src/smart_stream/filter.rs
```diff
@@ -117,9 +118,14 @@ impl SmartStreamFilter {
                 .and_then(|m| m.copy_memory_from(&mut self.store).ok())
                 .unwrap_or_default();
             debug!(out_filter_bytes = bytes.len());
+
             // this is inefficient for now
-            let mut records: MemoryRecords = vec![];
-            records.decode(&mut Cursor::new(bytes), 0)?;
+            let mut output = SmartStreamOutput::default();
+            output.decode(&mut Cursor::new(bytes), 0)?;
+            warn!("FILTER OUTPUT: {:?}", &output);
```
sehz, line 125: why warn?

sehz (changes requested):
Generally looks good. Few minor comments.  
Need comprehensive integration tests

src/spu/src/smart_stream/filter.rs
nicholastmosher: Ah sorry, left over from debugging, I'll remove

src/smartstream/derive/src/generator/filter.rs
nicholastmosher: This is copy-pasted from your original filtering code :P I'll make a constant

src/dataplane-protocol/src/smartstream.rs
nicholastmosher: So I sort of tried to do something like that where all of the info is captured in a structural way, but unfortunately there is a good deal of magic with the `eyre` crate that actually captures this information, and it does not have an interface that allows you to extract that info in a structured way, only as Display or Debug text.

I was thinking that in the future we could dig down and just learn the primitives that eyre uses to e.g. capture the location of an error, but that is a bit out of scope for the time being. My imagined upgrade path for this would be:

- Add a new variant to the top-level SmartStreamError, `SmartStreamError::UserErrorV2(...)`
- Write `SmartStreamUserErrorV2` that can capture structured errors using the same primitives `eyre` uses

PR #1229, opened 2021-07-12
[Merged by Bors] - Remove sccache to see if perf is impacted

nicholastmosher:
Let's compare with current CI workflow to see if anything actually runs slower without sccache.


sehz:
Looks like removing `sccache`  either improved or no impact at all.  Like it!

nicholastmosher:
@sehz do we want to try merging this and worse case we can revert it? It'll be hard to test the "release mode" workflow otherwise

sehz:
Can we always revert. let's merge it!

sehz (approved):

PR #1234, opened 2021-07-13
[Merged by Bors] - Generalize target type in fluivo-package-index

nicholastmosher:
This, followed by changes to `fluvio-packages`, will unblock us from publishing binaries from more targets than just `x86_64-unknown-linux-musl` and `x86_64-apple-darwin`.

### DIFF
--- a/src/package-index/src/target.rs
+++ b/src/package-index/src/target.rs
@@ -1,6 +1,7 @@
 use std::fmt;
 use serde::{Serialize, Deserialize};
 use crate::Error;
+use std::borrow::Cow;
 
 const PACKAGE_TARGET: &str = env!("PACKAGE_TARGET");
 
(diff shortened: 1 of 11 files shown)


src/package-index/src/target.rs
```diff
@@ -14,36 +15,57 @@ pub fn package_target() -> Result<Target, Error> {
...
+///
+/// # Example
+///
+/// ```
+/// # use fluvio_index::Target;
+/// let target: Target = "x86_64-unknown-linux-musl".parse().unwrap();
+/// ```
+#[derive(Debug, Clone, Hash, PartialEq, Eq, Serialize)]
+#[serde(transparent)]
+pub struct Target(Cow<'static, str>);
```
simlay, line 32: Hmm, I kinda like the enum of this better. I understand why you've done it this way though. Can one use the `transparent` and `rename` with serde at the sametime?

simlay (commented):
This seems okay. I like enums more than a transparent string target. It might be worth fighting serde a little bit if you can use `rename` with a can work with a `transparent` container for say `Other(Cow<'static, str>)` as the unknown target type.

There's a few ownership changes that are unclear why they're needed but I'm okay with giving clippy and rustc the wheel.

src/package-index/src/target.rs
nicholastmosher: What is it about the enum that you like better? With the enum representation we would need to add variants for each new target we want to support. In either case, the "serialized" representation is the same, it is transparently represented as a string that is just the name of the target.

What do you mean about `rename`? I'm no longer using `rename` in this new version.

sehz: @simlay ?

sehz: Given that we are going to add more variants, I don't think we can encode as Enum

simlay: Sounds good. Let's keep as is.

sehz (approved):

PR #1239, opened 2021-07-14
[Merged by Bors] - Update release flow to zip artifacts

nicholastmosher:
This should enable the release flow to handle arbitrary targets as we add them. However, it's also the type of change that is really impossible to test unless we actually try it.


sehz (approved):
can't really see if this works or not but let's try

PR #1243, opened 2021-07-15
Build CLI for aarch64-apple-darwin

nacardin:
Requires `macos-11` runner, waiting for GH approval


simlay:
This uses zig to actually build the CLI, right? I think that the rust client (and therefor the other language clients) may need https://github.com/infinyon/future-aio/pull/80 to build the releases of the clients.

PR #1250, opened 2021-07-15
[Merged by Bors] - Parallelize CI for fluvio, fluvio-run, flv-test

nicholastmosher:
This seems to reduce the CI run time (in debug mode, not on `staging`) from about 17-18 mins to about 11 mins

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,34 +29,53 @@ jobs:
   # build binaries. use release for staging
   # this requires check and test
   build_binaries:
-    name: Build binaries for Target (${{ matrix.rust_target }}) on ${{ matrix.os }}
+    name: Build ${{ matrix.binary }} for ${{ matrix.target }} (${{ matrix.os }})
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        rust_target:
+        target:
 #          - aarch64-unknown-linux-musl
 #          - armv7-unknown-linux-gnueabihf
           - x86_64-apple-darwin
           - x86_64-unknown-linux-musl
         rust: [stable]
+        binary: [fluvio]
         include:
-#          - rust_target: aarch64-unknown-linux-musl
-#            os: ubuntu-latest
-#            binaries: [fluvio]
-#          - rust_target: armv7-unknown-linux-gnueabihf
-#            os: ubuntu-latest
-#            binaries: [fluvio]
-          - rust_target: x86_64-apple-darwin
-            os: macos-latest
-            binaries: [fluvio]
-          - rust_target: x86_64-unknown-linux-musl
-            os: ubuntu-latest
-            binaries: [fluvio, fluvio-run, flv-test]
+#          - os: ubuntu-latest
+#            rust: stable
+#            target: aarch64-unknown-linux-musl
+#            binary: fluvio
+#            run: make build-cli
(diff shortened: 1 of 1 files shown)


nicholastmosher:
Since some of the job names were changed, it took one execution to refresh the rust-cache. I expect when we run release builds on `staging` it will also take one execution to reset the cache and we will see build time improvements.

.github/workflows/ci.yml
```diff
@@ -29,34 +29,53 @@ jobs:
   # build binaries. use release for staging
   # this requires check and test
   build_binaries:
-    name: Build binaries for Target (${{ matrix.rust_target }}) on ${{ matrix.os }}
+    name: Build ${{ matrix.binary }} for ${{ matrix.target }} (${{ matrix.os }})
```
sehz, line 32: like to keep `rust_target` in order not to get confused with other target

.github/workflows/ci.yml
```diff
@@ -29,34 +29,53 @@ jobs:
...
+#          - os: ubuntu-latest
+#            rust: stable
+#            target: armv7-unknown-linux-gnueabihf
+#            binary: fluvio
+#            run: make build-cli
+          - os: macos-latest
+            rust: stable
+            target: x86_64-apple-darwin
+            binary: fluvio
+            run: make build-cli
```
sehz, line 58: this should only contain matrix configuration not how to perform steps

.github/workflows/ci.yml
```diff
@@ -83,57 +102,41 @@ jobs:
...
     strategy:
       matrix:
         os: [ubuntu-latest]
         rust: [stable]
-        rust_target: [x86_64-unknown-linux-gnu]
-        check: [clippy, test, unstable]
+        target: [x86_64-unknown-linux-gnu]
+        check:
+          - name: Rustfmt
+            run: make check-fmt
```
sehz, line 126: don't add runs here

sehz (changes requested):

.github/workflows/ci.yml
nicholastmosher: The problem is that the old way is really ugly and unweildy, with several disadvantages:

- Each instance of the build job has X steps, one for each binary that might be built, but all except 1 of those are disabled, making the job screen really cluttered and ugly, e.g. here on a target where we only want `fluvio`:

image

- This problem gets worse for every new binary we might want to add to the workflow

When we set up the job this way, it takes much less code and it also displays in the nicest way in the actions UI. I really don't see any disadvantages to doing things this way, it makes it much easier.

nicholastmosher: This is opposed to the newer, flatter layout where each job is only instantiated with steps that it actually uses:

image

nicholastmosher:
My goal with structuring the CI is to make it so that, as much as possible, the only thing we should need to do in the future is update the `matrix` section, with minimal changes to the job steps themselves.

.github/workflows/ci.yml
sehz: that's implementation details.  important point is that `matrix` is configuration.  It should not contain how to perform configuration.

nicholastmosher:
@sehz I took the make commands back out of the matrix

sehz (approved):
LGTM

PR #1251, opened 2021-07-15
[Merged by Bors] - Run CD test after publish workflow in 'stable' and 'latest' mode

nicholastmosher:
This basically sets the "composite action test" to be run only _after_ the `publish.yml` job finishes, and to run it twice:

- Once for `stable`
- Once for `latest`

Hopefully that will give us a good indication of how the end-to-end flow is working for both the active CLI and the staged-up-to-be-released CLI

See the workflow working here: https://github.com/nicholastmosher/fluvio/runs/3081131127?check_suite_focus=true

### DIFF
--- a/actions/action-install-fluvio-cluster.sh
+++ b/actions/action-install-fluvio-cluster.sh
@@ -14,7 +14,7 @@ REPO_VERSION="$(curl -sSf https://raw.githubusercontent.com/infinyon/fluvio/mast
 CHART_VERSION="${REPO_VERSION}-${GITHUB_SHA}"
 
 #
-# Install Fluvio System Charts
+# Install Fluvio Cluster
 #
 
 # Install Local Fluvio Cluster
@@ -22,8 +22,7 @@ if [ "$CLUSTER_TYPE" = "local" ]; then
 
     # If VERSION is equal to exactly "latest", use LATEST channel
     if [ "${VERSION}" == "latest" ]; then
-        fluvio cluster start --rust-log $RUST_LOG --develop --local --spu $SPU_NUMBER --chart-version="${REPO_VERSION}-c963500f9d985a1a42b67380bf7bb683cdace1d7"
-#        fluvio cluster start --rust-log $RUST_LOG --develop --local --spu $SPU_NUMBER --chart-version="${CHART_VERSION}"
+        fluvio cluster start --rust-log $RUST_LOG --develop --local --spu $SPU_NUMBER --chart-version="${CHART_VERSION}"
     else
         fluvio cluster start --rust-log $RUST_LOG --develop --local --spu $SPU_NUMBER
     fi
--- a/actions/action-install-fluvio-cluster.sh
+++ b/actions/action-install-fluvio-cluster.sh
@@ -17,20 +17,6 @@ CHART_VERSION="${REPO_VERSION}-${GITHUB_SHA}"
 # Install Fluvio System Charts
 #
 
-# If VERSION is equal to exactly "latest", use LATEST channel
-if [ "${VERSION}" == "latest" ]; then
-    fluvio cluster start --setup --local --sys --chart-version="${REPO_VERSION}-c963500f9d985a1a42b67380bf7bb683cdace1d7"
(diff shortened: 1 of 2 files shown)


actions/action-install-fluvio-cluster.sh
```diff
@@ -10,16 +10,22 @@ curl -sSf https://raw.githubusercontent.com/infinyon/fluvio/master/install.sh |
...
 
+# Install Local Fluvio Cluster
 if [ "$CLUSTER_TYPE" = "local" ]; then
-        # Install Local Fluvio Cluster
+
+    # If VERSION is equal to exactly "latest", use LATEST channel
+    if [ "${VERSION}" == "latest" ]; then
+        fluvio cluster start --rust-log $RUST_LOG --develop --local --spu $SPU_NUMBER --chart-version="${CHART_VERSION}"
+    else
         fluvio cluster start --rust-log $RUST_LOG --develop --local --spu $SPU_NUMBER
```
sehz, line 27: is this what is expected from user?

sehz (changes requested):

actions/action-install-fluvio-cluster.sh
nicholastmosher: So at the end of the `publish.yml` step, a new helm chart called `fluvio-app:VERSION-HASH` is pushed to the chart museum. This corresponds to the new `latest` Fluvio CLI. However, `fluvio cluster start --local` has the VERSION it was compiled with baked-in (in this case it'd be `0.8.5`), and it tries to find a helm chart with just that `fluvio-app:VERSION`, which does not exist (because we have only run `publish.yml`, not `release.yml`).

By the time that the `0.8.5` CLI reaches the user, we will also have an `0.8.5` helm chart, at which point `--chart-version` will not be necessary. However, for the purpose of testing, we need to manually supply the helm chart version because otherwise `fluvio-cluster` will not be able to find it.

nicholastmosher: Oh also to note is that the `--chart-version` flag is only provided when `VERSION=latest`, so the instance of this test that runs on `stable` will run this second command, with no `--chart-version` (i.e. using stable).

sehz: I would expect `curl -fsS https://packages.fluvio.io/v1/install.sh | VERSION=latest bash` to be sufficient.

nicholastmosher: Oh so all you want is to check the installer? Don't want to spin up a cluster to check on after installing?

sehz: Yes but I would expect to use installer.  Right?

nicholastmosher: Right so if I'm a brand-new user the very first thing I do is to use the `install.sh` one-liner. Then, if I was using fluvio locally I'd need to use `fluvio cluster start`. So it depends on how far through this flow we want to test here.

sehz: We should perform exactly same action as user would do.   The action is described here: https://github.com/infinyon/fluvio/blob/master/actions/action-install-fluvio-cluster.sh

nicholastmosher: You just linked to this file?

But also the setup here in `action-install-fluvio-cluster.sh` currently runs extra commands that I think a user would not typically run, such as `fluvio cluster start --sys` and `fluvio cluster check`. One of the things I was doing here was eliminating those lines so the setup flow is exactly the same as the user.

nicholastmosher:
> Yes but I would expect to use installer. Right?

The installer is _only_ used for installing the Fluvio CLI, it is not responsible for starting a cluster. After the CLI is installed, the user uses it to run `fluvio cluster start` and start a cluster.

The open question is: do we want this workflow to just test the installer? Or do we want to test that we can both install the CLI _and_ the cluster?

actions/action-install-fluvio-cluster.sh
sehz: Yes, I agree, we should remove them

sehz:
looks ok but need to ensure CI is passing

sehz (approved):
LGTM

PR #1256, opened 2021-07-16
Fix CI

nacardin:


sehz:
should clear `~/.fluvio`

.github/workflows/ci.yml
```diff
@@ -223,20 +230,24 @@ jobs:
...
           path: ~/.fluvio/extensions
-      - run: chmod +x ~/.fluvio/extensions/fluvio-run
       - name: Download artifact - flv-test
         uses: actions/download-artifact@v2
         with:
           name: flv-test-x86_64-unknown-linux-musl
           path: .
       - name: Print artifacts and mark executable
-        run: ls -la . && chmod +x ./fluvio ./flv-test && ./fluvio version && ./flv-test -h
+        run: ls -la . && chmod +x ./fluvio ./flv-test && ./fluvio -h && ./flv-test -h
```
nicholastmosher, line 239: This will now fail since we are cross-compiling more targets. Attempting to invoke `./fluvio` when it is compiled for `armv7-unknown-linux-gnueabihf` will cause a failure.

nacardin: But, for this test we are specifically downloading `fluvio-x86_64-unknown-linux-musl`

nicholastmosher: Ah I see, for some reason I thought this was in the `build_binareis` job, but I was reading the `needs: build_binaries`. In that case, looks good 👍

.github/workflows/ci.yml
```diff
@@ -335,7 +348,11 @@ jobs:
       - name: Run smoke-test-k8-tls-root
         timeout-minutes: 5
         run: |
-          make FLUVIO_BIN=./fluvio TEST_BIN=./flv-test smoke-test-k8-tls-root
+          make FLUVIO_BIN=./fluvio TEST_BIN=./flv-test UNINSTALL=noclean smoke-test-k8-tls-root
+
+      - name: Clean minikube
+        run: |
```
nicholastmosher, line 354: Should this have `if: always()` so that even if the test fails it still cleans up minikube?

nacardin: Yes, updated both local and k8 cluster tests

.github/workflows/ci.yml
```diff
@@ -195,22 +195,29 @@ jobs:
     needs: build_binaries
     strategy:
       matrix:
-        os: [ubuntu-latest]
```
sehz, line 198: for now, let's stick to ubuntu unless it is proven to be unstable

.github/workflows/ci.yml
```diff
@@ -195,22 +195,29 @@ jobs:
...
+        os: [infinyon-ubuntu-bionic]
         rust: [stable]
     env:
       FLV_SOCKET_WAIT: 600
     steps:
       - uses: actions/checkout@v2
-      - name: Install Minikube for Github runner
```
sehz, line 204: for `ubuntu`, we need this

.github/workflows/ci.yml
```diff
@@ -195,22 +195,29 @@ jobs:
...
+        uses: actions-rs/toolchain@v1
         with:
-          minikube version: "v1.22.0"
-          kubernetes version: "v1.19.6"
-          github token: ${{ secrets.GITHUB_TOKEN }}
-          driver: docker
+          toolchain: ${{ matrix.rust }}
+          profile: minimal
+          override: true
+      - name: Install zig
```
sehz, line 211: we don't need zig since we are not building anything here

.github/workflows/ci.yml
```diff
@@ -195,22 +195,29 @@ jobs:
...
+        run: ./actions/zig-install.sh ${{ matrix.os }}
+      - name: Clean up Fluvio user dir
+        run: rm -rf ~/.fluvio
+      - name: Setup Minikube for Linux
```
sehz, line 215: only need for custom runner

.github/workflows/ci.yml
```diff
@@ -223,20 +230,27 @@ jobs:
...
         run: ls -la . && chmod +x ./fluvio ./flv-test && ./fluvio version && ./flv-test -h
+      - name: Print extension artifacts and mark executable
+        run: cd ~/.fluvio/extensions && ls -la . && chmod +x ./fluvio-run && ./fluvio-run -h
       - name: Setup
         run: ./fluvio cluster start --setup --local --develop
       - name: Run smoke-test
         timeout-minutes: 3
         run: |
           make FLUVIO_BIN=./fluvio TEST_BIN=./flv-test UNINSTALL=noclean smoke-test-tls-root
+      - name: Clean up minikube
```
sehz, line 248: only need for custom runner

.github/workflows/ci.yml
```diff
@@ -284,24 +298,26 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
```
sehz, line 300: same comments

sehz (changes requested):

PR #1257, opened 2021-07-16
[Merged by Bors] - Made 8ks a feature flag to cli and added build-cli-minimimal to makefile

simlay:
This is a sub task to #1218.

### DIFF
--- a/.github/workflows/ci_tier2.yaml
+++ b/.github/workflows/ci_tier2.yaml
@@ -6,7 +6,7 @@ name: CI_Platform_Tier2
 permissions:
   contents: read
 
-concurrency: 
+concurrency:
   group: ${{ 'ci_tier2' }}
   cancel-in-progress: true
 
@@ -26,7 +26,7 @@ env:
   VERBOSE: ${{ github.events.input.verbose }}
 
 jobs:
-  
+
   # Run all checks and unit test. This always run on debug mode
   check:
     name: Check tier2 platform (${{ matrix.rust_target }}) (${{ matrix.check }})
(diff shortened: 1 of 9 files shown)


sehz (changes requested):
LGTM.  Just need to changelog

.github/workflows/ci_tier2.yaml
```diff
@@ -119,6 +119,9 @@ jobs:
       - name: Build CLI
         run: make build-cli
 
+      - name: Build minimal cli
```
sehz, line 122: why do we need to build two different version of CLI here?

src/cli/src/http.rs
```diff
@@ -27,14 +27,16 @@ pub async fn execute(request: Request) -> Result<Response, Error> {
...
     debug!("Opened TLS stream from TCP stream");
     let response = client::connect(tls_stream, request).await?;
 
     debug!(?response, "Http response:");
     Ok(response)
 }
 
-async fn create_tls() -> fluvio_future::native_tls::TlsConnector {
-    fluvio_future::native_tls::TlsConnector::default()
+async fn create_tls() -> fluvio_future::openssl::TlsConnector {
```
sehz, line 38: this should return Result

sehz (changes requested):
Generally looks good. minor nit.
For Arm32 target, use `minimal CLI` instead of full Cli

.github/workflows/ci_tier2.yaml
simlay: Maybe this isn't the right place to put it but I'd like the minimal build or a check in CI. The `consumer` feature flag for the cli is needed for the build to work so I'd like to avoid that for this feature flag.

sehz (approved):
Looks good enough for now.  CI changes for other targets can be done in later PR

PR #1265, opened 2021-07-19
[Merged by Bors] - Update cache keys to prevent conflicting saves

nicholastmosher:
In our CI, our slowest job is "Unit/Doc test", and it appears that our rust-cache setup is failing to cache for that job:

image

I believe this is because this job is part of a matrix and only the first job in that matrix is "reserving" the key to cache with. This PR should give each job in the matrix a unique key so that caching can work properly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,6 +97,8 @@ jobs:
       - name: Install zig
         run: ./actions/zig-install.sh ${{ matrix.os }}
       - uses: Swatinem/rust-cache@v1
+        with:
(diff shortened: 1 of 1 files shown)


nicholastmosher:
See the Unit/Doc test build time changes:

- Original: 16m24s https://github.com/infinyon/fluvio/runs/3104068561?check_suite_focus=true
- With cache keys: 6m40s https://github.com/infinyon/fluvio/pull/1265/checks?check_run_id=3104333135

sehz:
Great performance improvement

sehz (approved):

PR #1274, opened 2021-07-21
Remove dataplane reexport

nicholastmosher:
One of the things that confused me about the codebase for a really long time is the fact that `fluvio-dataplane-protocol` re-exports some items from `fluivo_protocol`, so all around the codebase I would see mixed instances of `use dataplane::derive::Encode` versus `use fluvio_protocol::Encode` and such, causing confusion about where these items actually lived and whether they were even the same.

I think that re-exports are useful at the top-level of a multi-crate project where third-party users are expected to use our code (i.e. the client crate), but I think that it makes more sense for internal code to directly import items from the original crate.

In this PR, I removed the re-exports of `fluvio_protocol` items from `fluvio-dataplane-protocol` and changed all of the `use dataplane` instances of these items to `use fluvio_protocol`. I wanted to do this now, since we are doing major bumps in preparation for `0.9.0` and this is a breaking change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1517,7 +1517,7 @@ dependencies = [
 
 [[package]]
 name = "fluvio-dataplane-protocol"
-version = "0.5.2"
+version = "0.6.0"
 dependencies = [
  "bytes",
  "cfg-if 1.0.0",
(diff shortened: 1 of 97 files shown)


nicholastmosher:
This was more of a cosmetic change that I had hoped to land in 0.9.0, but it didn't so I'm going to close it.

PR #1276, opened 2021-07-21
[Merged by Bors] - Fix derive(Encoder) bug for enums with Unit and Data variants

nicholastmosher:
It turns out that Unit variant enums were not having their `write_size` implementation generated properly. This fixes it by generating write_size arms that report the tag size for Unit variants.

Since we have not yet published fluvio-protocol:0.6.0 I have not bumped the crate version.

### DIFF
--- a/src/protocol/fluvio-protocol-derive/src/ser.rs
+++ b/src/protocol/fluvio-protocol-derive/src/ser.rs
@@ -274,7 +274,15 @@ fn parse_enum_variants_size(
                 };
                 variant_expr.push(arm);
             }
-            _ => (),
+            FieldKind::Unit => {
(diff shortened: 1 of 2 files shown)


nicholastmosher:
@sehz how does this look?

sehz:
is there test for it?

nicholastmosher:
The test is just making sure that this compiles, which it would not have done before:

https://github.com/infinyon/fluvio/pull/1276/files#diff-2480077bb110e24a0dec2c47176b1bbbd13f92d81a252be5917c1ca9cd6fed98R75-R79

sehz (approved):

PR #1281, opened 2021-07-21
[Merged by Bors] - Revert to native tls for cli

### LINKED ISSUE #1273: [Bug]: TLS error in `fluvio install`
We noticed in the `CD_Dev` workflow that the `install.sh` script was failing when using `VERSION=latest`:

https://github.com/infinyon/fluvio/runs/3120104226?check_suite_focus=true

image

After some digging, I figured out that this is a TLS problem introduced in https://github.com/infinyon/fluvio/commit/4ffcaa66376d9bf95232058db4e137d0485be19c when we switched from `fluvio_future::native_tls` to `fluvio_future::openssl_tls`. It affects all http calls in the CLI, which includes `fluvio install` and `fluvio update`.

This was the code change that caused this:

https://github.com/infinyon/fluvio/blob/4ffcaa66376d9bf95232058db4e137d0485be19c/src/cli/src/http.rs#L39-L42

simlay:
I've tested these changes over in #1218 and it works there as well.

Should resolve #1273.

### DIFF
--- a/src/cli/src/http.rs
+++ b/src/cli/src/http.rs
@@ -38,13 +38,12 @@ async fn create_tls() -> fluvio_future::native_tls::TlsConnector {
     fluvio_future::native_tls::TlsConnector::default()
 }
 
-#[fluvio_future::test_async]
-async fn test_web_request() -> Result<(), Error> {
+#[fluvio_future::test]
+async fn test_web_request() {
     use fluvio_index::HttpAgent;
     use http_types::StatusCode;
     let agent = HttpAgent::default();
-    let index = agent.request_index()?;
-    let response = execute(index).await?;
+    let index = agent.request_index().expect("Failed to get request index");
+    let response = execute(index).await.expect("Failed to execute request");
     assert_eq!(response.status(), StatusCode::Ok);
-    Ok(())
 }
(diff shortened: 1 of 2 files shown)


src/cli/src/http.rs
```diff
@@ -26,17 +25,26 @@ pub async fn execute(request: Request) -> Result<Response, CliError> {
...
     Ok(response)
 }
+async fn create_tls() -> fluvio_future::native_tls::TlsConnector {
+    fluvio_future::native_tls::TlsConnector::default()
+}
 
-async fn create_tls(
-) -> Result<fluvio_future::openssl::TlsConnector, fluvio_future::openssl::TlsError> {
-    Ok(fluvio_future::openssl::TlsConnector::builder()?.build())
+#[fluvio_future::test_async]
```
sehz, line 41: use `test` rather than `test_async`.  this way, it doesn't need to return result

sehz (changes requested):
LGTM.  minor nit

src/cli/src/http.rs
simlay: I mean, I kinda like using `Result` as it allows for the use of `?` in the unit test.

sehz: Problem is that it's hard to trace if test fails.
so instead of:
```
  let index = agent.request_index()?;
```
should be like:
```
  let index = agent.request_index().expect("can't find index");
```

simlay: Done.

sehz (approved):
LGMT. Thanks!.  can merge after makefile PR

PR #1289, opened 2021-07-23
[Merged by Bors] - Add apply to stream dispatcher store

### LINKED ISSUE #1294: Add capability to run Apply action on StoreContext
Currently we can create Kubernetes objects in a StoreContext with create_spec, but we cannot define in that function custom metadata for the object. For instance OwnerReference

morenol:
Currently we can create Kubernetes objects in a StoreContext with create_spec, but we cannot define in that function custom metadata for the object. For instance OwnerReference

Fixes #1294


sehz (changes requested):
Looks good.  Need minor things:
* Create Issue for it.
* Add to CHANGELOG

sehz (approved):

PR #1290, opened 2021-07-23
[Merged by Bors] - Fix replication issue with k8

sehz:
Issue was mismatch in the SVC and StatefulSet.  Changes include:
* Fix label
* Update the smoke test to check for replication status
* Add exponential backoff connecting follower to leader


sehz:
@tjtelan I commented out panic in test runner because it was suppressing error.

PR #1298, opened 2021-07-26
Fix typo 'lsr' to 'lrs'

### LINKED ISSUE #1296: Rename LSR to LRS
- https://github.com/infinyon/fluvio/blob/a3b5074add2983180db4693367efe08c88eeb16b/src/controlplane-metadata/src/partition/status.rs#L88
- https://github.com/infinyon/fluvio/blob/a3b5074add2983180db4693367efe08c88eeb16b/src/controlplane-metadata/src/partition/status.rs#L32

nicholastmosher:
Closes #1296 in preparation for the `0.9.0` release. This includes renaming a CRD status field name, is there anything special we need to do to mark that breaking change? It is only a print column name if that matters.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2628,9 +2628,9 @@ dependencies = [
 
 [[package]]
 name = "js-sys"
-version = "0.3.51"
+version = "0.3.52"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "83bdfbace3a0e81a4253f73b49e960b053e396a11012cbd49b9b74d6a2b67062"
+checksum = "ce791b7ca6638aae45be056e068fc756d871eb3b3b10b8efa62d1c9cec616752"
 dependencies = [
  "wasm-bindgen",
 ]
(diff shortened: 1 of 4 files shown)


sehz (approved):

sehz:
reopen because it was revert.  need to ensure we can upgrade from previous version

nicholastmosher:
Added `serde(alias = lsr)` which means that serde will deserialize from either "lsr" (old field name) or "lrs" (new field name).

https://serde.rs/field-attrs.html#alias

nicholastmosher:
bors try

sehz (approved):
merge in after running upgrade test manually: `make upgrade-test`

nicholastmosher:
@sehz what should we do about this? Do we want to close the PR for now and keep an issue open about this?

sehz:
can you make upgrade work transparently?

ajhunyady:
My original ask was to change the "Label" that we are showing the user.
It looks like this became something else.

nicholastmosher:
We have already fixed the label, but it'd be nice to drill all the way down and get the typos out of the codebase. The original idea of this PR was to fix this issue during the major version bump to `0.9.0`, so if we decide to move this forward at all it'd probably need to go in the next major bump to `0.10.0`

PR #1304, opened 2021-07-27
Set owner_reference in input metadata

morenol:
When using the new apply method, the owner_reference is being ignored. This fixes that.

Follow up of https://github.com/infinyon/fluvio/issues/1294


sehz (changes requested):
Can you explain more detail why this is needed?

morenol:
I think that it is not needed

PR #1305, opened 2021-07-27
[Merged by Bors] - Add upgrade test to Makefile

tjtelan:
* Fixed a bunch of relative path issues from Makefile calling test
* Also, actually enable upgrade in CD
* Upgrade test supports latest published with USE_LATEST


sehz (approved):
LGTM

PR #1307, opened 2021-07-27
[Merged by Bors] - Add 'arm-unknown-linux-gnueabihf' to CI builds

### LINKED ISSUE #1160: Platform support: Arm32 for Raspberry pi v3 (Arm7)
Support IOT use cases.

nicholastmosher:
Closes #1160 

This uses `cross` the same way that the armv7 build does

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,11 @@ jobs:
             run: make build-cli
           - os: ubuntu-latest
             rust: stable
+            rust-target: arm-unknown-linux-gnueabihf
+            binary: fluvio
+            run: make build-cli
+          - os: ubuntu-latest
+            rust: stable
             rust-target: armv7-unknown-linux-gnueabihf
             binary: fluvio
             run: make build-cli
@@ -472,6 +477,7 @@ jobs:
         rust-target:
           - aarch64-unknown-linux-musl
           - armv7-unknown-linux-gnueabihf
+          - arm-unknown-linux-gnueabihf
           - x86_64-apple-darwin
           - x86_64-unknown-linux-musl
         artifact: [fluvio]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,17 +46,14 @@ jobs:
             rust: stable
             rust-target: aarch64-unknown-linux-musl
             binary: fluvio
-            run: make build-cli
           - os: ubuntu-latest
             rust: stable
             rust-target: arm-unknown-linux-gnueabihf
             binary: fluvio
-            run: make build-cli
           - os: ubuntu-latest
             rust: stable
             rust-target: armv7-unknown-linux-gnueabihf
             binary: fluvio
-            run: make build-cli
           - os: windows-latest
             rust: stable
             rust-target: x86_64-pc-windows-msvc
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ install_tools_mac:
 	brew install helm
 
 build-cli: install_rustup_target
-ifeq ($(TARGET), armv7-unknown-linux-gnueabihf)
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -46,10 +46,11 @@ install_tools_mac:
 	brew install yq
 	brew install helm
 
-build-cli: install_rustup_target
-ifeq ($(TARGET), armv7-unknown-linux-gnueabihf)
+ifneq (,$(findstring arm, $(TARGET))) # If TARGET contains the substring "arm"
+build-cli: install_cross
```
simlay, line 50: I don't feel all that strongly on this but having a make variable like `CARGO_BUILDER` (not sure on the name) that's `cross` for some builds and is just `cargo` for others may be nice.

nicholastmosher: That might be possible, yeah. Though we do need the recipe to distinguish between installing cross or a rustup target depending, so it seems like this conditional might need to stay

simlay: Yeah. I'm noticing that the makefile conditional language is a bit lacking.

.github/workflows/ci.yml
```diff
@@ -47,6 +47,11 @@ jobs:
             rust-target: aarch64-unknown-linux-musl
             binary: fluvio
             run: make build-cli
+          - os: ubuntu-latest
+            rust: stable
+            rust-target: arm-unknown-linux-gnueabihf
+            binary: fluvio
+            run: make build-cli
```
sehz, line 54: Remove run. Should not have any impl in matrix

Makefile
```diff
@@ -65,8 +66,10 @@ build-test:	build-cluster build-cli
 	cargo build --bin flv-test $(RELEASE_FLAG) $(TARGET_FLAG) $(VERBOSE_FLAG)
 
 install_rustup_target:
-	./build-scripts/install_target.sh
-	
+	rustup target add $(TARGET)
```
sehz, line 69: Why this is removed?

sehz (changes requested):

.github/workflows/ci.yml
nicholastmosher: This was already like this, if we want to change it we should do it in another PR, I don't want to rewrite a CI flow here.

Makefile
nicholastmosher: The only thing `install_target.sh` did was execute `rustup target add` or `cargo install cross`, but I worked that into the Makefile itself since now we have to distinguish between two targets (arm and armv7)

.github/workflows/ci.yml
sehz: Then dont add to it

Makefile
sehz: Keep existing script because it is not easy to add complex condtional

.github/workflows/ci.yml
sehz, line 54: don't add run

Makefile
```diff
@@ -47,7 +47,7 @@ install_tools_mac:
 	brew install helm
 
 build-cli: install_rustup_target
-ifeq ($(TARGET), armv7-unknown-linux-gnueabihf)
```
sehz, line 50: leverage `install_rust_target` to compute `cross` or `cargo` as @simlay as suggested.  Then we can make this simple

sehz (changes requested):

.github/workflows/ci.yml
nicholastmosher: This entire job is structured this way, I am just adding one include rule

image

As I mentioned above, I don't want to rewrite this job as part of this PR

nicholastmosher:
@sehz does this look good?

sehz (approved):
Looks great.  Thanks

nicholastmosher:
bors try

PR #1309, opened 2021-07-27
[Merged by Bors] - Add upgrade test to CI workflow

### LINKED ISSUE #1262: CI: Re-enable Upgrade Test
CI build process has been refactored and cleanup by https://github.com/infinyon/fluvio/issues/1206.   Upgrade test need to be migrated:  see. PR #1260

Upgrade test should work on two cases:

- [ ] upgrade to dev release (image is released as dev).  this only run in master
- [ ] upgrade from branch (in this case image is saved in the GitHub artifact directory) 

The logging should be minimized as default

tjtelan:
Still testing, so there's heavy duplication in CI jobs

The CI job should test in DEBUG mode with the CI artifacts. The CD_Dev job should use the public artifacts

Minor fix, the smoke test is wrapped in `nick-invision/retry@v2` to retry the job if it fails, max 3 attempts. Just to save time rerunning all jobs for CI environmental issues.

Resolves #1262

### DIFF
--- a/.github/workflows/cd_dev.yaml
+++ b/.github/workflows/cd_dev.yaml
@@ -137,7 +137,7 @@ jobs:
           max_attempts: 3
           command: |
             export PATH=~/.fluvio/bin:$PATH
-            make upgrade-test
+            USE_LATEST=true make upgrade-test
 
       - name: Clean minikube
         run: |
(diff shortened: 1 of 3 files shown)


tjtelan:
@sehz it looks like this test catches the after-upgrade consumer issue and the failure is correct

sehz (approved):
Great.  let's disable upgrade in CI to get passed by Boris.  Another PR should fix upgrade issue and re-enable it

PR #1326, opened 2021-07-29
[Merged by Bors] - Update error-code and add smartstream error types

nicholastmosher:
Partial PR from https://github.com/infinyon/fluvio/pull/1226

### DIFF
--- a/src/dataplane-protocol/src/smartstream.rs
+++ b/src/dataplane-protocol/src/smartstream.rs
@@ -1,0 +1,109 @@
+pub use encoding::{SmartStreamRuntimeError, SmartStreamType, SmartStreamInput, SmartStreamOutput};
+
+mod encoding {
+    use std::fmt;
+    use crate::Offset;
+    use crate::record::{Record, RecordData};
+    use fluvio_protocol::{Encoder, Decoder};
+
+    /// A type used to pass data into a SmartStream WASM module
+    #[derive(Debug, Default, Clone, Encoder, Decoder)]
+    pub struct SmartStreamInput {
+        /// The base offset of this batch of records
+        pub base_offset: Offset,
+        /// The records for the SmartStream to process
+        pub record_data: Vec<u8>,
+    }
+
+    /// A type used to return processed records and/or an error from a SmartStream
+    #[derive(Debug, Default, Encoder, Decoder)]
+    pub struct SmartStreamOutput {
+        /// The successfully processed output Records
+        pub successes: Vec<Record>,
+        /// Any runtime error if one was encountered
+        pub error: Option<SmartStreamRuntimeError>,
+    }
+
+    /// A type used to capture and serialize errors from within a SmartStream
+    #[derive(thiserror::Error, Debug, Default, Clone, PartialEq, Encoder, Decoder)]
+    pub struct SmartStreamRuntimeError {
+        /// Error hint: meant for users, not for code
+        pub hint: String,
+        /// The offset of the Record that had a runtime error
+        pub offset: Offset,
+        /// The type of SmartStream that had a runtime error
+        pub kind: SmartStreamType,
+        /// The Record key that caused this error
+        pub record_key: Option<RecordData>,
+        /// The Record value that caused this error
+        pub record_value: RecordData,
+    }
+
+    impl SmartStreamRuntimeError {
+        pub fn new(
+            record: &Record,
+            base_offset: Offset,
+            kind: SmartStreamType,
+            error: eyre::Error,
+        ) -> Self {
+            let hint = format!("{:?}", error);
+            let offset = base_offset + record.preamble.offset_delta();
+            let record_key = record.key.clone();
+            let record_value = record.value.clone();
+            Self {
+                hint,
+                offset,
+                kind,
+                record_key,
+                record_value,
+            }
+        }
+    }
+
+    impl fmt::Display for SmartStreamRuntimeError {
+        fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+            let key = self
+                .record_key
+                .as_ref()
+                .map(display_record_data)
+                .unwrap_or_else(|| "NULL".to_string());
+            let value = display_record_data(&self.record_value);
+            write!(
+                f,
+                "{}\n\n\
+                SmartStream Info: \n    \
+                Type: {}\n    \
+                Offset: {}\n    \
+                Key: {}\n    \
+                Value: {}",
+                self.hint, self.kind, self.offset, key, value,
+            )
+        }
+    }
+
+    fn display_record_data(record: &RecordData) -> String {
+        match std::str::from_utf8(record.as_ref()) {
+            Ok(s) => s.to_string(),
+            _ => format!("Binary: {} bytes", record.as_ref().len()),
+        }
+    }
+
+    #[derive(Debug, Clone, PartialEq, Encoder, Decoder)]
+    pub enum SmartStreamType {
+        Filter,
+        Map,
+    }
+
+    impl Default for SmartStreamType {
+        fn default() -> Self {
+            Self::Filter
+        }
+    }
+
+    impl fmt::Display for SmartStreamType {
+        fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+            // Use Debug for Display to print variant name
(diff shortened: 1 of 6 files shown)


sehz:
Hmm. I don't think SmartStreamError should be include in here.  Or at least separate out from ErrorCode

src/dataplane-protocol/src/smartstream.rs
```diff
@@ -0,0 +1,167 @@
...
+    use crate::record::Record;
+    use fluvio_protocol::{Encoder, Decoder};
+
+    /// A type used to capture and serialize errors from within a SmartStream
+    #[derive(thiserror::Error, Debug, Default, Clone, PartialEq, Encoder, Decoder)]
+    pub struct SmartStreamRuntimeError {
+        /// Error hint: meant for users, not for code
+        pub hint: String,
+        /// The offset of the Record that had a runtime error
+        pub offset: i64,
```
sehz, line 19: use type alias Offset

src/dataplane-protocol/src/smartstream.rs
```diff
@@ -0,0 +1,167 @@
...
+        /// The type of SmartStream that had a runtime error
+        pub kind: SmartStreamType,
+        /// A preview of the Record key that caused this error
+        pub record_key: String,
```
sehz, line 23: This should be RecordData

src/dataplane-protocol/src/smartstream.rs
```diff
@@ -0,0 +1,167 @@
...
+        /// A preview of the Record value that caused this error
+        pub record_value: String,
```
sehz, line 25: Should be RecordData

src/dataplane-protocol/src/smartstream.rs
```diff
@@ -0,0 +1,167 @@
...
+                Offset: {}\n    \
+                Key: {}\n    \
+                Value: {}",
+                self.hint, self.kind, self.offset, self.record_key, self.record_value
+            )
+        }
+    }
+
+    #[derive(Debug, Default, Clone, PartialEq, Encoder, Decoder)]
+    pub struct SmartStreamRuntimeErrorBuilder {
```
sehz, line 44: Why not use DeriveBuilder?

sehz (changes requested):

src/dataplane-protocol/src/smartstream.rs
nicholastmosher: Because there are some little weird requirements on this type, such as it needs to impl `Encoder/Decoder` (because this type gets passed across the WASM boundary) and it just made it easier to have better control over the builder in this specific case.

src/dataplane-protocol/src/smartstream.rs
nicholastmosher: This field is supposed to be a "preview" of the record data in a user-readable way, it is not intended to preserve the exact contents of the record. Notably, this field gets truncated so that it is only 100 columns wide and at most 20 lines long, so that the error type size does not balloon

sehz: this is supposed to be only dealing with types not user display issue. let CLI or other output worry about it. just pass records having problem

src/dataplane-protocol/src/smartstream.rs
nicholastmosher: This isn't intended to be a lossless representation of the Record, but a user-readable repr

src/dataplane-protocol/src/smartstream.rs
sehz: k

src/dataplane-protocol/src/smartstream.rs
sehz: let's not worry about user readability here.

src/dataplane-protocol/src/smartstream.rs
nicholastmosher: We still want this to be truncated to a reasonable size though, right? I don't want to have 1K+ data in these errors

sehz: any record key/value should fit into a single transmission.  this is not something error handling need to worry about it

nicholastmosher: Alright, will use RecordData

src/dataplane-protocol/src/smartstream.rs
```diff
@@ -0,0 +1,135 @@
...
+
+        /// Generate a preview of the Record that failed
+        pub fn record(&mut self, record: &Record) -> &mut Self {
+            self.record_key = record.key.clone();
+            self.record_value = Some(record.value.clone());
+            self
+        }
+
+        /// Produce the SmartStreamRuntimeError
+        pub fn build(&self) -> SmartStreamRuntimeError {
```
sehz, line 99: is there reason why this need to ref instead of self?

src/dataplane-protocol/src/smartstream.rs
```diff
@@ -0,0 +1,135 @@
...
+                hint: self.hint.clone(),
+            }
+        }
+    }
+
+    /// A type used to return processed records and/or an error from a SmartStream
+    #[derive(Debug, Default, Encoder, Decoder)]
+    pub struct SmartStreamOutput {
+        pub successes: Vec<Record>,
+        pub error: Option<SmartStreamRuntimeErrorBuilder>,
```
sehz, line 114: shouldn't this be SmartStreamRunTimeError?

sehz (changes requested):

src/dataplane-protocol/src/smartstream.rs
nicholastmosher: The `SmartStreamOutput` type is the top-level data object that gets serialized from WASM and transmitted to SPU. At that point, we only have partial data of what we want to collect for the runtime error, specifically the `hint` and the `offset_delta`, but the other pieces of data come from the SPU itself. So we use the builder here because it collects half of the data in WASM and half in the SPU, and then it gets converted to the proper `SmartStreamRuntimeError` right before sending the response to the consumer

src/dataplane-protocol/src/smartstream.rs
nicholastmosher: This is generally the pattern used when you have a `&mut Self` chained builder. E.g. from the derive_builder documentation, this is what derive_builder usually generates

image

src/dataplane-protocol/src/smartstream.rs
sehz: So this shouldn't belong in here since this is only relevant internally to SPU-WASM.  This create should be contain that protocol between client and cluster (SPU).

nicholastmosher: Where do you think it should go instead? I put it here because `dataplane-protocol` is the common crate that is imported by both the WASM application code and the SPU, so the types belong here so that they are accessible to both. I can't move this just into the SPU because then the WASM couldn't import these types, and vice-versa, I can't move this type into the `fluvio_smartstream` crate because the SPU couldn't import them there.

This seems to be the only place that works, unless we create yet another crate and I really don't want to do that.

sehz: What about smartstream crate?

nicholastmosher: The smartstream crate is not imported by the SPU, it is only imported by the WASM module itself

nicholastmosher: It is also not something that we would _want_ to import to the SPU, because it exports low-level symbols that are only relevant to the WASM, such as `copy_records` and either `filter` or `map` raw "extern C" functions, which would make no sense to have in the SPU

nicholastmosher: > Probably need new smartstream specific protocol crate

Can we please tackle this in another issue in the future? I am trying so hard to get this V1 shipped

nicholastmosher:
Latest updates:

- We now have both `SmartStreamInput` and `SmartStreamOutput` for passing data in and out of WASM
- We pass the batch base_offset into the SmartStream so that if an error needs to be built it can have the absolute offset by adding the base_offset and the offset_delta
- Got rid of the error builder so now there is just `SmartStreamRuntimeError`

sehz (approved):
LGTM

PR #1333, opened 2021-07-29
[Merged by Bors] - Fix clippy for new 1.54 changes

nicholastmosher:


src/client/src/error.rs
```diff
@@ -1,28 +1,27 @@
...
     #[error("Spu not found: {0}")]
     SPUNotFound(i32),
-    #[error(transparent)]
-    IoError(#[from] IoError),
     #[error("Fluvio socket error")]
-    FlvSocketError(#[from] FlvSocketError),
+    Socket(#[from] SocketError),
     #[error("Fluvio SC schema error")]
-    ApiError(#[from] ApiError),
+    ScSchema(#[from] ApiError),
```
sehz, line 22: change this to ControlPlane

src/client/src/error.rs
```diff
@@ -1,28 +1,27 @@
...
     #[error("Fluvio config error")]
-    ConfigError(#[from] ConfigError),
+    Config(#[from] ConfigError),
```
sehz, line 24: this is too generic. Change this to ProfileConfig

src/storage/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 edition = "2018"
 name = "fluvio-storage"
-version = "0.5.0"
+version = "0.4.0"
```
sehz, line 4: no need to change this version

sehz (changes requested):
Mostly looks good.  Couple of minor changes needed

src/storage/Cargo.toml
nicholastmosher: I just think that since the last published version on crates.io is 0.3.0 that it makes no sense to jump to 0.5.0, I don't see a reason why not to

src/client/src/error.rs
nicholastmosher: I was mostly naming it after the crate it came from, which is `sc-schema`, not controlplane

src/client/src/error.rs
nicholastmosher: What about `ClientConfig`?

src/storage/Cargo.toml
sehz: k

src/client/src/error.rs
sehz: How about `Admin` because that's what error is related to

sehz (changes requested):

nicholastmosher:
What if we kept the name of the variant `ScSchema`, but we make the Display show something like "SC controlplane error"?

sehz:
> What if we kept the name of the variant `ScSchema`, but we make the Display show something like "SC controlplane error"?

`ScSchema` is should be changed it's too implementation sound.  `AdminApi` is more descriptive

sehz (approved):

PR #1344, opened 2021-08-01
[Merged by Bors] - Client updates for smartstream-errors

### LINKED ISSUE #1198: Error reporting for SmartStreams

nicholastmosher:
The last part of the smartstream-error story, client side integration!

Closes #1198 

This adds the following flags to the CLI:

- `fluvio consume --filter=path/to/smartstream.wasm`
- `fluvio consume --map=path/to/smartstream.wasm`

### DIFF
--- a/src/cli/src/consumer/consume/mod.rs
+++ b/src/cli/src/consumer/consume/mod.rs
@@ -74,9 +74,13 @@ pub struct ConsumeOpt {
     )]
     pub output: ConsumeOutputType,
 
-    /// Path to a WASM binary file
-    #[structopt(short, long)]
-    pub smart_stream: Option<PathBuf>,
+    /// Path to a SmartStream filter wasm file
+    #[structopt(long, group("smartstream"))]
+    pub filter: Option<PathBuf>,
+
+    /// Path to a SmartStream map wasm file
+    #[structopt(long, group("smartstream"))]
+    pub map: Option<PathBuf>,
 }
 
 impl ConsumeOpt {
@@ -112,10 +116,16 @@ impl ConsumeOpt {
             builder.max_bytes(max_bytes);
         }
 
-        if let Some(filter_path) = &self.smart_stream {
+        if let Some(filter_path) = &self.filter {
             let buffer = std::fs::read(filter_path)?;
             debug!(len = buffer.len(), "read filter bytes");
             builder.wasm_filter(buffer);
+        }
+
+        if let Some(map_path) = &self.map {
+            let buffer = std::fs::read(map_path)?;
+            debug!(len = buffer.len(), "read filter bytes");
+            builder.wasm_map(buffer);
         }
 
         let consume_config = builder.build()?;
--- a/src/client/src/consumer.rs
+++ b/src/client/src/consumer.rs
@@ -6,8 +6,11 @@ use once_cell::sync::Lazy;
 use futures_util::future::{Either, err};
 use futures_util::stream::{StreamExt, once, iter};
 
-use fluvio_spu_schema::server::stream_fetch::{DefaultStreamFetchRequest, DefaultStreamFetchResponse};
-use dataplane::Isolation;
+use fluvio_spu_schema::server::stream_fetch::{
+    DefaultStreamFetchRequest, DefaultStreamFetchResponse, SmartStreamPayload, SmartStreamWasm,
+    SmartStreamKind, WASM_MODULE_V2_API,
+};
+use dataplane::{Isolation, SmartStreamError};
 use dataplane::ReplicaKey;
 use dataplane::ErrorCode;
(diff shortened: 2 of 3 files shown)


src/cli/src/consumer/consume/mod.rs
```diff
@@ -74,9 +74,13 @@ pub struct ConsumeOpt {
     )]
     pub output: ConsumeOutputType,
 
-    /// Path to a WASM binary file
-    #[structopt(short, long)]
-    pub smart_stream: Option<PathBuf>,
+    /// Path to a SmartStream filter wasm file
+    #[structopt(long, group("smartstream"))]
+    pub filter: Option<PathBuf>,
```
sehz, line 79: This will leads to proliferation of duplicate argument type.   For example, in the future, WASM could be loading not just from path but in the URL.  In that case, we have to change all these types.  Instead, we should decouple WASM from it's module type.  So instead, it's better if we have something like this:

```
pub struct SmartStreamConfig {
        pub module_source: Option<PathBuf>,
        pub module_type:  SmartStreamKind
```

sehz (changes requested):

src/cli/src/consumer/consume/mod.rs
nicholastmosher: This will impact the user experience in the CLI,so @ajhunyady do you have any feedback on this? I think if we went with this approach, the CLI would look like:

```
fluvio consume <topic> --module-source=./smartstream.wasm --module-type=filter
fluvio consume <topic> --module-source=./smartstream.wasm --module-type=map
```

As opposed to what we had previously talked about:

```
fluvio consume <topic> --filter=./smartstream.wasm
fluvio consume <topic> --map=./smartstream.wasm
```

nicholastmosher:
Updated CLI to the following after our discussion:

```
fluvio consume <topic> --filter-module=filter.wasm
fluvio consume <topic> --map-module=map.wasm
```

sehz (approved):
LGTM

PR #1348, opened 2021-08-02
[Merged by Bors] - Add tag resolution to package-index

nicholastmosher:
This allows the Fluvio CLI to install packages whose version references a tag in addition to a specific version

Whereas before we could only use exact versions such as

```
fluvio install fluvio:0.8.4
```

We can now use tag names in the position after `:`

```
fluvio install fluvio:stable
fluvio install fluvio:latest
```


sehz (approved):

sehz:
fix clippy issue before merging into master

PR #1355, opened 2021-08-03
[Merged by Bors] - Fix SmartStream label, update CHANGELOG

nicholastmosher:


src/smartstream/derive/src/generator/map.rs
```diff
@@ -49,7 +49,7 @@ pub fn generate_map_smartstream(func: &SmartStreamFn) -> TokenStream {
                             let error = fluvio_smartstream::dataplane::smartstream::SmartStreamRuntimeError::new(
                                 &record,
                                 smartstream_input.base_offset,
-                                fluvio_smartstream::dataplane::smartstream::SmartStreamType::Filter,
+                                fluvio_smartstream::dataplane::smartstream::SmartStreamType::Map,
```
sehz, line 52: there should be test to catch this issue.

sehz (changes requested):

src/smartstream/derive/src/generator/map.rs
nicholastmosher: I opened an issue, can we handle later? https://github.com/infinyon/fluvio/issues/1356

sehz: this kind of major issue can't be fell thru crack.  let's complete the test

nicholastmosher: Added test

sehz (approved):
LGTM

PR #1387, opened 2021-08-05
[Merged by Bors] - Remove nested consumer module, flatten CLI

### LINKED ISSUE #1308: Remove `consumer` hierarchy from CLI modules and errors
Right now, the non-admin commands of Fluvio are oddly organized as being "consumer" commands, when it would make sense for them to be considered top-level commands. This hierarchy is reflected in the module structure and the error structure of the CLI, which can lead to some confusing error messages such as a "consumer error" when producing:

```
 $ fluvio produce test2                    
Error: 
   0: Consumer Error
   1: Fluvio client error
   2: Topic not found: test2
```

We should:

- [ ] Flatten the CLI module structure, removing the `consumer` grouping for topic, partition, produce, consumer
- [ ] Flatten the error types, removing the ConsumerError type that wraps other CLI errors

nicholastmosher:
Closes #1308


src/cli/src/lib.rs
```diff
@@ -173,6 +179,64 @@ impl RootCmd {
     }
 }
 
+// For some reason this doc string is the one that gets used for the top-level help menu.
+// Please don't change it unless you want to update the top-level help menu "about".
+/// Fluvio command-line interface
+#[derive(StructOpt, Debug)]
+pub enum FluvioCmd {
```
nicholastmosher, line 186: Just FYI this is not all new code, I just moved it here from `src/cli/consumer/mod.rs`

src/cli/src/consume/mod.rs
```diff
@@ -89,7 +89,7 @@ impl ConsumeOpt {
         name = "Consume",
         fields(topic = %self.topic, partition = self.partition),
     )]
-    pub async fn process(self, fluvio: &Fluvio) -> Result<(), ConsumerError> {
+    pub async fn process(self, fluvio: &Fluvio) -> Result<()> {
```
sehz, line 92: reason why this have to be `CliError` ?  It's better to scope to specific Error

sehz (changes requested):

src/cli/src/consume/mod.rs
nicholastmosher: CliError makes the most sense here. `ConsumerError` was a shim type that was overly specific and was actually mis-applied in a lot of places (e.g. an error creating a topic would be considered a "ConsumerError"). We don't really lose anything by removing this extra layer of nesting.

sehz (approved):

PR #1390, opened 2021-08-06
[Merged by Bors] - Remove dead field in SPU

### LINKED ISSUE #1332: Remove dead code in SPU
Need to validate that this is unused and delete it when we have a chance

- [x] https://github.com/infinyon/fluvio/blob/master/src/spu/src/control_plane/dispatcher.rs#L38

nicholastmosher:
Closes #1332 

@sehz there is also more dead code in this file, there is a termination_sender channel that is unused. Do we want to get rid of that, or leave it in case we want to use it in the future?

I'm talking about here: https://github.com/infinyon/fluvio/blob/master/src/spu/src/control_plane/dispatcher.rs#L48-L51


sehz:
can remove it

sehz (approved):

PR #1391, opened 2021-08-06
Restore last_stable_offset to prevent breaking changes

nicholastmosher:
This undoes the removal of `last_stable_offset` from `FetchablePartitionResponse` in an effort to remove breakages from the protocol change.


sehz (changes requested):
need to lower min  PLATFORM  again

nicholastmosher:
We have decided against this due to the semantic version implications of bumping to 0.9.0

PR #1410, opened 2021-08-09
[Merged by Bors] - Recompile SC if VERSION has changed

nicholastmosher:
This will cause the SC to recompile if the VERSION file changes, even if no SC code has changed


sehz (approved):
Good find

Issue #1380, opened 2021-08-05
[Tracking] Fluvio perf test harness integration


tjtelan:
The changes from the performance testing drafts need to be included in the repo but the scope of the branch is too broad for a single [PR](https://github.com/infinyon/fluvio/pull/1252). This issue is tracking each component so we can keep track of their merge status.

Several of these changes should be built with a feature flag. Perhaps split via: performance measurement support, data vis/export, and other cluster clients.

### Create generic Test driver for test instrumentation
- [x] Centralize producer and consumer configuration (orig from Smoke test) so it could be reused in other tests
- [x] Centralize producer send / consumer receive so it could be reused in other tests
- [ ] Configurable producer batch record size - #1538
- [ ] Configurable producer batch time period - #1538
- [x] Record latency of producer traffic
- [x] Record latency of consumer traffic

### Test reporting improvements
- [ ] Structured test data as csv row - #1539
- [x] Default print performance stats
- [x] Fix panic behavior to return more debugging info from panic or assert

### Interpret producer/consumer traffic characteristics, in terms of
- [x] Latency as Percentile(50, 90, 99, 99.9) over test duration
- [x] Throughput over test duration (Mbps x time)

### Rough draft graphs from test
- [ ] Generate simple graphs from test output - #1543
- [ ] CSV data export of test results - #1539
- [ ] Tech debt: Better graphs

### Collect apples-to-apples test data for Fluvio
- [ ] x Kafka - https://github.com/infinyon/fluvio/issues/1544
- [ ] x Pulsar - https://github.com/infinyon/fluvio/issues/1544

Tech debt:
- Performance specific test harness (Design considerations vs flv-test including all the testing functions)
- Measurement sanity tests

tjtelan:
Smaller issues have been created and are now being tracked individually. This tracking issue is no longer necessary. Closing

PR #1417, opened 2021-08-10
[Merged by Bors] - Flv test driver - decouple tests from Fluvio, part 1

tjtelan:
Refactoring codebase to support #1380

* Moving TestResult and TestTimer into their own modules
* Passing the measurements histograms from the test all the way out so we can easily display more statistics (Avg, P50, P90, P99, P999)
* Add TestDriver abstraction so we can support testing against non-Fluvio platforms.


sehz (approved):
LGTM

PR #1434, opened 2021-08-13
[Merged by Bors] - Fix plugin install

nicholastmosher:
Adds a new error variant representing "this package isn't available" for a given package, version, and target. Also catches that error when running `fluvio install` in order to print a nice message to the user instead of crashing. Now, if somebody uses the installer from a platform that is not supported (or they try to specify a VERSION that does not exist), they'll get a message like the following in their installer output:

image


sehz (approved):
LGTM

PR #1437, opened 2021-08-14
[Merged by Bors] - Add post-release test

### LINKED ISSUE #1433: Add Post-release test
The upgrade test uses the `--develop` flag for upgrading, so that the development cluster uses a development docker image.

The gap is we don't quite test that the development CLI will successfully deploy a cluster of the same version until the upgrade test is run again.

https://github.com/infinyon/fluvio/issues/1415#issuecomment-898002407

Post-release, we should test that
* [x] Starting a brand new cluster works
* [x] Upgrade works from the previous stable to the current stable.

tjtelan:
Run workflow after Release
Check the installer
Start a local and k8s cluster
Do a basic upgrade from prev version

Resolves #1433

### DIFF
--- a/.github/workflows/post-release.yml
+++ b/.github/workflows/post-release.yml
@@ -1,0 +1,126 @@
+name: Post-Release
+
+permissions:
+  contents: read
+
+on:
+
+  workflow_run:
+    workflows: ["Release"]
+    types:
+      - completed
+  workflow_dispatch:
+    inputs:
+      alt_version:
+        required: false
+        description: Provide expected semver. If empty VERSION in repo will be used
+        default: ''
+
+env:
+  USE_VERSION: ${{ github.event.inputs.alt_version }}
+
+jobs:
+  installer_check:
+    name: Installer check
+    strategy:
+      matrix:
+        version: [none, stable, semver]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set the expected fluvio version
+        run: |
+          if [[ -z "${{ env.USE_VERSION }}" ]]; then
+            echo "EXPECTED_VERSION=$(cat VERSION)" | tee -a $GITHUB_ENV
+          else
+            echo "EXPECTED_VERSION=${{ github.event.inputs.alt_version }}" | tee -a $GITHUB_ENV
+          fi
+          echo "EXPECTED_VERSION: $EXPECTED_VERSION"
+
+      - name: Curl Install - no version
+        if: matrix.version == 'none'
+        run: curl -fsS https://packages.fluvio.io/v1/install.sh | bash | tee /tmp/installer.version
+      - name: Curl Install - stable 
+        if: matrix.version == 'stable'
+        run: curl -fsS https://packages.fluvio.io/v1/install.sh | VERSION=stable bash | tee /tmp/installer.version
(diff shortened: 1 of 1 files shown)


.github/workflows/post-release.yml
```diff
@@ -0,0 +1,126 @@
...
+          echo "EXPECTED_VERSION: $EXPECTED_VERSION"
+
+      - name: Curl Install - no version
+        if: matrix.version == 'none'
+        run: curl -fsS https://packages.fluvio.io/v1/install.sh | bash | tee /tmp/installer.version
+      - name: Curl Install - stable 
+        if: matrix.version == 'stable'
+        run: curl -fsS https://packages.fluvio.io/v1/install.sh | VERSION=stable bash | tee /tmp/installer.version
+      - name: Curl Install - Version number
+        if: matrix.version == 'semver'
```
sehz, line 46: all 3 steps seems to be same.  only diff is version which can be done in a separate step so action can be written as:
```
curl -fsS https://packages.fluvio.io/v1/install.sh | ${ env.FLUVIO_VERSION}  | tee /tmp/installer.version
```

where `FLUVIO_VERSIOn` can be computed in previous step

.github/workflows/post-release.yml
```diff
@@ -0,0 +1,126 @@
...
+    env:
+      FLV_SOCKET_WAIT: 600
+    steps:
+      - uses: nolar/setup-k3d-k3s@v1
+        if: matrix.cluster_type == 'k8s'
+      - name: Install Fluvio
+        run: curl -fsS https://packages.fluvio.io/v1/install.sh | bash
+      - name: Start local cluster
+        if: matrix.cluster_type == 'local'
+        run: ~/.fluvio/bin/fluvio cluster start --local
```
sehz, line 78: similar to above just compute `--local` using cluster type

sehz (changes requested):

.github/workflows/post-release.yml
tjtelan: I don't think I quite understand the goal you have in mind.

This matrix test is meant to validate the installer script installs the most recent release w/r/t the value of the `VERSION` env var, which is mainly how the installer script selects the version number to download. There are 3 cases I've identified.

The env var is:
* unset
  * The basic one-liner on the Fluvio site doesn't set this var
* set to `stable`
  * This is a release channel to test
* A specific version, e.g `x.y.z`
  * In our release process, this should be defined by the `VERSION` file. (But this is configurable in the action, for testing in between releases)

sehz: My comment is about removing redundant expression: `curl -fsS https://packages.fluvio.io/v1/install.sh | xxx}  | tee /tmp/installer.version` for all 3 cases

.github/workflows/post-release.yml
tjtelan: I don't quite understand the feedback. What am I supposed to compute?

The purpose of this matrix test is to verify the default behavior for starting a cluster, and the two modes (local, or k8) have different CLI commands (the `--local` flag), so I put them in different arms.

This test is one of the main motivators for #1433

sehz (approved):
LGTM. Nice job

PR #1442, opened 2021-08-17
[Merged by Bors] - Do not resolve absolute offsets to negative

### LINKED ISSUE #1210: Update relative offset resolution to be saturating
Right now, using relative offsets `Offset::from_beginning(X)` or `Offset::from_end(X)` when fetching from a partition have undesirable behavior when X is greater than the number of records in that partition. When this happens, a stream request will just hang. Instead, I think that the relative offset resolution should be _saturating_, meaning that:

If `n` is the size of the partition, then

- When we fetch using `Offset::from_end(X)` and `X > n`, then the fetch should resolve to absolute offset `0`
- When we fetch using `Offset::from_beginning(X)` and `X > n`, then the fetch should resolve to absolute offset `X-1`

The end result of this is that no call to `fetch` should ever cause the client to hang.

Related: https://github.com/infinyon/fluvio/issues/1007

nicholastmosher:
Closes #1210

### DIFF
--- a/src/client/src/offset.rs
+++ b/src/client/src/offset.rs
@@ -283,15 +283,23 @@ impl Offset {
             OffsetInner::FromBeginning(offset) => {
                 let replica = ReplicaKey::new(topic, partition);
                 let offsets = fetch_offsets(client, &replica).await?;
-                offsets.start_offset + offset
+                let resolved = offsets.start_offset + offset;
+
+                // Resolved offset should be between 0 and the last offset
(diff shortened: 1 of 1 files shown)


sehz (changes requested):
need unit test

nicholastmosher:
Good suggestion, I found a new bug I would have introduced if not for writing the tests 😅

sehz (approved):
Nice work!

PR #1443, opened 2021-08-17
[Merged by Bors] - Add --tail for reading from the end of a log

### LINKED ISSUE #1216: Add `fluvio consume --tail=X` to consume starting X from the end of a stream
Right now, to consume records starting X from the end of a stream, we use `fluvio consume <topic> --offset="-X"`. We think it would be more ergonomic to switch this up so that:

- `--offset` always takes a positive number, counting from the beginning of the stream, and
- `--tail`, a new command, takes a positive number and counts from the end of the stream

nicholastmosher:
Closes #1216

### DIFF
--- a/src/cli/src/consume/mod.rs
+++ b/src/cli/src/consume/mod.rs
@@ -6,7 +6,6 @@
 
 use std::{io::Error as IoError, path::PathBuf};
 use std::io::ErrorKind;
-use std::convert::TryFrom;
 use tracing::{debug, trace, instrument};
 use structopt::StructOpt;
 use structopt::clap::arg_enum;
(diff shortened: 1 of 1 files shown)


src/cli/src/consume/mod.rs
```diff
@@ -51,9 +48,17 @@ pub struct ConsumeOpt {
     #[structopt(short, long)]
     pub key_value: bool,
 
-    /// Offsets can be positive or negative. (Syntax for negative offset: --offset="-1")
-    #[structopt(short, long, value_name = "integer")]
-    pub offset: Option<i64>,
+    /// Consume records starting X from the beginning of the log (default: 0)
+    #[structopt(short = "B", value_name = "integer", conflicts_with_all = &["offset", "tail"])]
+    pub from_beginning: Option<Option<u32>>,
```
sehz, line 53: why is this option of option?

nicholastmosher: There are three states:

- `fluvio consume <topic>` (no `--from-beginning`): consume from the end (corresponds to `None`)
- `fluvio consume <topic> --from-beginning`: consume from the very beginning (corresponds to `Some(None)`)
- `fluvio consume <topic> --from-beginning=10`: consume starting 10 after the beginning (`Some(Some(offset))`)

nicholastmosher: Same pattern goes for `--tail`

sehz (approved):

PR #1456, opened 2021-08-18
[Merged by Bors] - Defer stream initialization in consumer

### LINKED ISSUE #986: `PartitionConsumer::stream` should not block while there are no records

morenol:
Fixes #986


sehz (approved):
LGTM

PR #1463, opened 2021-08-19
[Merged by Bors] - Update CHANGELOG with items from 0.9.3

nicholastmosher:
Seems like we've been forgetting to update the CHANGELOG in our PR's. We might want to create a pull request template to remind us to do things like changelog and crate version bumps.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,12 @@
 # Release Notes
 
 ## Platform Version 0.9.3 - UNRELEASED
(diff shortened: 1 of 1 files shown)


sehz (approved):

morenol:
Maybe we can add a check like this to the CI: https://github.com/rusty-celery/rusty-celery/blob/41efda696132bff7451b191d185a48129ca1b2e3/.github/workflows/ci.yml#L19

Issue #1450, opened 2021-08-18
Release checklist: `0.9.3`


nicholastmosher:
## Pre-release

Prior to releasing, the release manager should check the following:

Update fluvio website doc (master):

- [x] Review the "getting started" docs.
- [x] Update Any API related docs.
- [x] Review rest of doc to ensure they are up to date.

Other dependent repos:
- [x] Update `infinyon/fluvio-smartstream-template` if needed.

## Release

First Lock master branch (TBD)

- [x] Publish all public crates.
- [x] Update `CHANGELOG.md` with release date.
- [x] Run the `release.yml` workflow on GitHub Actions.
- [x] Wait for workflow to complete successfully and apply `vX.Y.Z` tag to git.
- [x] Create PR from master to stable branch.  Get approval and merge it.
- [x] Update fluvio website to stable.

Unlock Master branch (TBD)

## Post-release

After performing the release, the release manager should do the following in order
to prepare for the next release and announce the current release to the community:

- [x] Bump up the version in the `VERSION` file.
- [x] Add UNRELEASED section for new version to `CHANGELOG.md`.
- [ ] Announce the release on Discord and Twitter.

nicholastmosher:
I went to close this but I figured @simlay should get the satisfaction of checking it off 😄 nice job!

simlay:
:tada:

PR #1327, opened 2021-07-29
[Merged by Bors] - Spu smartstream errors

nicholastmosher:
Second sub-piece of https://github.com/infinyon/fluvio/pull/1226

Review/merge after https://github.com/infinyon/fluvio/pull/1326

PR #1467, opened 2021-08-19
[Merged by Bors] - Initial stuff for crates.io publishing

simlay:
Subtask for #1450.

### DIFF
--- a/src/spu-schema/Cargo.toml
+++ b/src/spu-schema/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio-spu-schema"
-version = "0.7.0"
+version = "0.8.0"
 edition = "2018"
 authors = ["Fluvio Contributors <team@fluvio.io>"]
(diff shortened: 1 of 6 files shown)


src/spu-schema/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio-spu-schema"
-version = "0.7.0"
+version = "0.8.0"
```
sehz, line 3: does it require major bump?

sehz (changes requested):

src/spu-schema/Cargo.toml
simlay: I think so, because `wasm_payload` was added to `StreamFetchRequest` in https://github.com/infinyon/fluvio/pull/1327

sehz: k

sehz (approved):

PR #1469, opened 2021-08-20
Hotfix to fix release ci

simlay:
Most of these edits are from my vim that removes white space at the end of lines.


simlay:
Close in favor of #1470.

Issue #1420, opened 2021-08-10
[Bug]: Consumer using `-d` does not apply SmartStreams


nicholastmosher:
When consuming from a topic, applying a SmartStream will fail if the consume is invoked with `-d, --disable-continuous`. This is because `-d` performs a "fetch" rather than opening a stream. This uses a different underlying API, and the SPU handles this request differently. We have been intending to deprecate the fetch API and move exclusively to a streaming interface, so this would be a good opportunity to do that. Once we set `-d` to use the streaming API (and just stop after a fixed number rather than continue streaming), we should see SmartStreams behave normally again.

PR #1481, opened 2021-08-20
[Merged by Bors] - Deprecate fetch API. Create stream wrapper for --disable-continuous

### LINKED ISSUE #957: Deprecate consumer Fetch API
Right now the consumer has a Stream and a Fetch API, where the Stream is for consuming records and potentially waiting for new ones to become available (e.g. if we are at the end of the topic), and the Fetch API is for a one-time consume of a range of records. I believe we can safely deprecate the Fetch API because the Stream API should be able to accomplish all of the same use-cases as Fetch.

For example, one could use `let records: Vec<_> = stream.take(X).collect().await` instead of fetch.

nicholastmosher:
Closes #957 and #1420.

This deprecates `PartitionConsumer::fetch` and `PartitionConsumer::fetch_with_config` and replaces the "fetch records but don't wait" functionality with a Stream adapter that closes after all known records have been consumed. The way this works is:

- When we run `fluvio consume -B -d`, internally we make a `DefaultOffsetPartitionRequest` which tells us the offset of the last-known record in the partition. If we take the difference between that offset and the starting offset, we find the number of records that should be available to stream right now, without waiting for more records to be produced.
- The new `TakeRecords` stream adapter wraps an inner stream and takes a number of records to fetch, and we give the "number of available records" as this number. After this adapter has witnessed this many records flow through the stream, it closes, returning `Poll::Ready(None)` and causing the consumer to finish.
- One small change in here is that `ConsumerConfig` now has a `disable_continuous` field which is used to determine whether or not to keep the stream open or to close it after all known records have been consumed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Platform Version 0.9.4 - UNRELEASED 
 * Publish docker image for aarch64 #1389 ([#1389](https://github.com/infinyon/fluvio/pull/1389))
 * Do not panic when trying to create topic with space in the name. ([#1448](https://github.com/infinyon/fluvio/pull/1448))
+* Deprecate consumer fetch API ([#957](https://github.com/infinyon/fluvio/issues/957))
 
 ## Platform Version 0.9.3 - 2021-08-19
 * Fix Replication timing. ([#1439](https://github.com/infinyon/fluvio/pull/1439))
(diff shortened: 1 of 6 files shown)


sehz (changes requested):
Nice implementation.  Couple of minor housekeepings.
* bump up version
* add to change log

sehz (approved):

sehz:
clippy failures

PR #1485, opened 2021-08-23
use release mode for building SmartStream examples

sehz:


sehz:
apparently, integration test is depends on release type. closing now.

PR #1486, opened 2021-08-23
[Merged by Bors] - Add template for pull requests

nicholastmosher:
I keep forgetting some upkeeping in PR's and I figured others might have this problem as well, so I created a PR template as a reminder for those handful of things.


sehz (changes requested):
[glommio](https://github.com/DataDog/glommio/blob/master/.github/PULL_REQUEST_TEMPLATE.md) has useful check list item that we could use:

nicholastmosher:
Added some similar sections to this template

sehz (approved):
Nice!

PR #1523, opened 2021-08-27
[Merged by Bors] - Handle stream fetch in a separate thread

### LINKED ISSUE #1522: [Bug]: StreamFetchHandler initialization is done in the conenction handling thread
StreamFetchHandler initialization is done in the thread handling the requests from a client connection. Because of this, the thread does not handle new requests from the same client until the smartstream initialization is done.

morenol:
fixes #1522

### DIFF
--- a/crates/fluvio-spu/src/services/public/stream_fetch.rs
+++ b/crates/fluvio-spu/src/services/public/stream_fetch.rs
@@ -38,6 +38,7 @@ pub struct StreamFetchHandler {
     isolation: Isolation,
     max_bytes: u32,
     max_fetch_bytes: u32,
+
     header: RequestHeader,
     sink: ExclusiveFlvSink,
     end_event: Arc<StickyEvent>,
@@ -628,8 +629,8 @@ mod test {
     use fluvio_socket::{FluvioSocket, MultiplexerSocket};
     use dataplane::{
         Isolation,
+        fetch::DefaultFetchRequest,
         fixture::BatchProducer,
-        produce::{DefaultProduceRequest, DefaultTopicRequest, DefaultPartitionRequest},
         record::{RecordData, Record},
     };
     use dataplane::fixture::{create_batch, TEST_RECORD};
@@ -1560,33 +1561,14 @@ mod test {
             .await
             .expect("create stream");
 
-        let mut produce_request = DefaultProduceRequest::default();
-        let mut topic_request = DefaultTopicRequest {
-            name: topic.to_string(),
-            ..Default::default()
-        };
-        let mut partition_request = DefaultPartitionRequest {
-            partition_index: 0,
-            ..Default::default()
-        };
-        partition_request.records.batches.push(Batch::from(vec![
-            Record::default(),
-            Record::default(),
-            Record::default(),
-            Record::default(),
-        ]));
-        topic_request.partitions.push(partition_request);
-        produce_request.acks = 1;
-        produce_request.timeout_ms = 1500;
-        produce_request.topics.push(topic_request);
+        let fetch_request = DefaultFetchRequest::default();
+        let response = client_socket
+            .send_and_receive(RequestMessage::new_request(fetch_request))
+            .await;
 
-        client_socket
-            .send_and_receive(RequestMessage::new_request(produce_request))
-            .await
-            .expect("produce error");
+        assert!(response.is_ok());
 
         server_end_event.notify();
-
         debug!("terminated controller");
     }
 }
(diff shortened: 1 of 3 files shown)


crates/fluvio-spu/src/services/public/stream_fetch.rs
```diff
@@ -80,6 +95,7 @@ impl StreamFetchHandler {
             debug!("Has WASM payload: {}", msg.wasm_payload.is_some());
 
             let smartstream = if let Some(payload) = msg.wasm_payload {
+                std::thread::sleep(std::time::Duration::from_secs(40));
                 let wasm = &payload.wasm.get_raw()?;
                 let module = sm_engine.create_module_from_binary(wasm).map_err(|err| {
```
morenol, line 100: This operation in the test is very fast so, sometimes we don't notice the error, in cloud it usually takes more than 10 seconds (but sometimes is also fast).

crates/fluvio-spu/src/services/public/stream_fetch.rs
```diff
@@ -80,6 +95,7 @@ impl StreamFetchHandler {
             debug!("Has WASM payload: {}", msg.wasm_payload.is_some());
 
             let smartstream = if let Some(payload) = msg.wasm_payload {
+                std::thread::sleep(std::time::Duration::from_secs(40));
```
sehz, line 98: this will block the thread.  this should be never done in async operations.

crates/fluvio-spu/src/services/public/stream_fetch.rs
```diff
@@ -48,6 +48,21 @@ pub struct StreamFetchHandler {
...
     /// handle fluvio continuous fetch request
+    pub async fn spawn(
+        request: RequestMessage<FileStreamFetchRequest>,
+        ctx: DefaultSharedGlobalContext,
+        sink: ExclusiveFlvSink,
+        end_event: Arc<StickyEvent>,
+    ) -> Result<(), SocketError> {
+        spawn(async move {
+            if let Err(err) = StreamFetchHandler::start(request, ctx, sink, end_event).await {
+                error!("error starting stream fetch handler: {:#?}", err);
```
sehz, line 59: this  need to notify end when error occurs like this:
```
 self.end_event.notify();
```

sehz (changes requested):
better clean solution is put entire module creation and preparation as separate abstraction.

crates/fluvio-spu/src/services/public/stream_fetch.rs
sehz: for testing, you can inject async sleep only if test flag is used with env var.

crates/fluvio-spu/src/services/public/stream_fetch.rs
```diff
@@ -1493,4 +1509,83 @@ mod test {
...
+            wasm_payload: Some(wasm_payload),
+            ..Default::default()
+        };
+
+        let _stream = client_socket
+            .create_stream(RequestMessage::new_request(stream_request), 11)
+            .await
+            .expect("create stream");
+
+        let mut produce_request = DefaultProduceRequest::default();
```
sehz, line 1562: in order to test this, perform concurrent 2 requests (filter, simple fetch).  simple fetch should always return first (by comparing

sehz (changes requested):

crates/fluvio-spu/src/services/public/stream_fetch.rs
```diff
@@ -1493,4 +1510,63 @@ mod test {
...
+            .create_stream(RequestMessage::new_request(stream_request), 11)
+            .await
+            .expect("create stream");
+
+        let fetch_request = DefaultFetchRequest::default();
+        let _response = client_socket
+            .send_and_receive(RequestMessage::new_request(fetch_request))
+            .await
+            .expect("fetch error");
+
```
sehz, line 1568: either put some assertion or remove it since it doesn't  check anything

sehz (changes requested):

crates/fluvio-spu/src/services/public/stream_fetch.rs
morenol: this  panics in case of timeout

sehz: need simple assertion to

sehz (changes requested):
one more thing.  add to changelog

sehz (approved):
LGTM

PR #1531, opened 2021-08-29
[Merged by Bors] - Cleanup fluvio service streams

nicholastmosher:
Cleanup in `fluvio-service`

### DIFF
--- a/crates/fluvio-service/src/server.rs
+++ b/crates/fluvio-service/src/server.rs
@@ -1,6 +1,5 @@
 use std::fmt;
 use std::fmt::Debug;
-use std::io::Error as IoError;
 use std::marker::PhantomData;
 use std::process;
 use std::sync::Arc;
@@ -8,36 +7,14 @@ use std::os::unix::io::AsRawFd;
 
 use futures_util::StreamExt;
 use async_trait::async_trait;
-use tracing::{debug, error, info};
-use tracing::instrument;
-use tracing::trace;
+use tracing::{instrument, debug, error, info};
 
-use fluvio_types::event::SimpleEvent;
 use fluvio_future::net::{TcpListener, TcpStream};
 use fluvio_future::task::spawn;
 use fluvio_protocol::api::ApiMessage;
 use fluvio_protocol::Decoder as FluvioDecoder;
 use fluvio_socket::{FluvioSocket, SocketError};
-
-#[async_trait]
-pub trait SocketBuilder: Clone {
-    async fn to_socket(&self, raw_stream: TcpStream) -> Result<FluvioSocket, IoError>;
-}
-
-#[derive(Debug, Clone)]
-pub struct DefaultSocketBuilder {}
-
-#[async_trait]
-impl SocketBuilder for DefaultSocketBuilder {
-    async fn to_socket(&self, raw_stream: TcpStream) -> Result<FluvioSocket, IoError> {
-        let fd = raw_stream.as_raw_fd();
-        Ok(FluvioSocket::from_stream(
-            Box::new(raw_stream.clone()),
-            Box::new(raw_stream),
-            fd,
-        ))
-    }
-}
+use fluvio_types::event::SimpleEvent;
 
 #[derive(Debug)]
 pub struct ConnectInfo {
@@ -49,7 +26,7 @@ pub struct ConnectInfo {
 /// Request -> Response is type specific
 /// Each response is responsible for sending back to socket
 #[async_trait]
-pub trait FlvService {
+pub trait FluvioService {
     type Request;
     type Context;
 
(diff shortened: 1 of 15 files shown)


crates/fluvio-service/src/server.rs
```diff
@@ -62,159 +39,111 @@ pub trait FlvService {
...
 
-    async fn run_shutdown(self, shutdown_signal: Arc<SimpleEvent>) {
-        match TcpListener::bind(&self.addr).await {
-            Ok(listener) => {
-                debug!(addr = %self.addr, "starting event loop");
-                self.event_loop(listener, shutdown_signal).await;
-            }
+    #[instrument(
+        skip(self, shutdown),
+        fields(addr = %self.addr),
```
sehz, line 87: can remove self and remove add from fields since self implements debug

crates/fluvio-service/src/server.rs
```diff
@@ -62,159 +39,111 @@ pub trait FlvService {
...
-                    break;
-                }
-
-            }
-        }
-
-        debug!("server terminating");
-    }
+        info!("Opened TcpListener, waiting for connections");
+        let mut incoming = listener.incoming().take_until(shutdown.listen_pinned());
```
sehz, line 100: very cool!

crates/fluvio-service/src/server.rs
```diff
@@ -62,159 +39,111 @@ pub trait FlvService {
...
+        info!("Closed TcpListener");
+    }
+
+    #[instrument(skip(stream, context, service))]
+    async fn handle_request(stream: TcpStream, context: C, service: Arc<S>, host: String) {
+        let addr = stream
+            .peer_addr()
+            .map(|addr| addr.to_string())
+            .unwrap_or_else(|_| "".to_owned());
+        debug!(%addr, "Handling request");
```
sehz, line 127: change `addr` to `peer_addr`

crates/fluvio-service/src/server.rs
```diff
@@ -275,8 +208,6 @@ mod test {
         let reply2 = socket.send(&msg2).await.expect("send");
         trace!("received 2nd reply from server: {:#?}", reply2);
         assert_eq!(reply2.response.msg, "hello2");
-
-        shutdown.notify();
```
sehz, line 279: why this is removed?

crates/fluvio-types/src/event.rs
```diff
@@ -40,6 +40,10 @@ impl SimpleEvent {
         listener.await
```
sehz, line 40: bump up version

sehz (changes requested):
Looks great with few minor points

crates/fluvio-service/src/server.rs
nicholastmosher: I think I deleted this by mistake, good catch

crates/fluvio-types/src/event.rs
nicholastmosher: Ah so I wanted to ask, what do you think of renaming this to `StickyEvent`? It's an event because you listen for it, but after it is triggered once it always yields immediately. I think it communicates the behavior better, what do you think?

crates/fluvio-service/src/server.rs
nicholastmosher: I have come to believe that it's best to put only the very minimal information into the tracing spans because it makes logs very noisy. I think that especially since this is at a top-level hierarchy that a lot of other code paths stem from, it'd be good to keep noise down at this point

crates/fluvio-types/src/event.rs
sehz: reasonable but we have to keep old name around and deprecated until next major version

crates/fluvio-service/src/server.rs
sehz: > it is more important to keep consistent so have reliable trace which I found to be critical.  Noise problem can be delegated to tooling/filter

crates/fluvio-service/src/server.rs
```diff
@@ -62,159 +39,108 @@ pub trait FlvService {
...
+            host: host.clone(),
+        };
+
+        let result = service.respond(context, socket, connection_info).await;
+        match result {
+            Ok(_) => {
+                info!(%host, %peer_addr, "Response sent successfully, closing connection");
+            }
+            Err(err) => {
+                error!(%host, %peer_addr, "Failed to send response: {}", err);
```
sehz, line 142: let's keep old log message.   `response` doesn't send response. it just handles request.

sehz (changes requested):
last comment

crates/fluvio-service/src/server.rs
nicholastmosher: Got it. What would you think of changing `respond` to `handle` then?

sehz (approved):
LGTM

PR #1532, opened 2021-08-29
[Merged by Bors] - Cleanup handlers

nicholastmosher:
Cleans up SPU public service handlers, adding better structure to the instrumentation as well (including correlation_id and client_id in spans for all request handlers). This moves the offset_update handler to its own file as well.


sehz:
CI failures

crates/fluvio-spu-schema/src/server/api.rs
```diff
@@ -67,3 +67,27 @@ impl ApiMessage for SpuServerRequest {
         }
     }
 }
+
+impl SpuServerRequest {
+    pub fn correlation_id(&self) -> i32 {
+        match self {
```
sehz, line 73: why is this needed?

crates/fluvio-spu/src/services/public/api_versions.rs
```diff
@@ -45,6 +44,7 @@ pub async fn handle_kf_lookup_version_request(
         UpdateOffsetsRequest::DEFAULT_API_VERSION,
     ));
 
+    trace!("Returning ApiVersionsResponse: {:#?}", &response);
```
sehz, line 47: You can trace on `fluvio_socket`.

crates/fluvio-spu/src/services/public/fetch_handler.rs
```diff
@@ -3,70 +3,111 @@ use tracing::{debug, trace, instrument};
...
 
 use crate::core::DefaultSharedGlobalContext;
+use dataplane::record::FileRecordSet;
 
 /// perform log fetch request using zero copy write
-#[instrument(skip(request, ctx, sink))]
+#[instrument(
+    skip(request, ctx, sink),
+    fields(
+        max_bytes = request.request.max_bytes,
```
sehz, line 19: not fan of adding custom fields.  if we need to do this, it's better to add to debug on request.  make it consistent

crates/fluvio-spu/src/services/public/fetch_handler.rs
```diff
@@ -3,70 +3,111 @@ use tracing::{debug, trace, instrument};
...
+    ),
+)]
 pub async fn handle_fetch_request(
     request: RequestMessage<FileFetchRequest>,
     ctx: DefaultSharedGlobalContext,
     sink: ExclusiveFlvSink,
 ) -> Result<(), SocketError> {
     let (header, fetch_request) = request.get_header_request();
+    trace!("Handling FileFetchRequest: {:#?}", fetch_request);
```
sehz, line 28: handle by `fluvio_socket`

crates/fluvio-spu/src/services/public/fetch_handler.rs
```diff
@@ -3,70 +3,111 @@ use tracing::{debug, trace, instrument};
...
-        }
-
+        let topic_response = handle_fetch_topic(&ctx, &fetch_request, topic_request).await?;
         fetch_response.topics.push(topic_response);
     }
 
     let response =
         RequestMessage::<FileFetchRequest>::response_with_header(&header, fetch_response);
-    trace!("sending back file fetch response: {:#?}", response);
+    trace!("Sending FileFetchResponse: {:#?}", response);
```
sehz, line 38: too many tracing.

sehz (changes requested):
initial comments

crates/fluvio-spu-schema/src/server/api.rs
nicholastmosher: I added this to make it easy to get to correlation_id and client_id and add that information to the span in the main SPU request handler function: https://github.com/infinyon/fluvio/pull/1532/files#diff-2277166e98453fe79d4a9c077d19f29f5e51fbb911900f3671aea595cbf572d3R75-R76

These are actually two critical pieces of information that should be included at the top-level request span, and should help debugging ability significantly, e.g. you can filter for a specific correlation_id when looking at logs to see the flow of events.

sehz: Here is current trace:
```
 handle_request{host="0.0.0.0:9010"}:respond
{socket=Socket(13) _connection=ConnectInfo { host: "0.0.0.0:9010", peer: "127.0.0.1:48528" }}:s
```

We should generalize, 
* socket
* port
* peer
* client_id
* correlation_id

This should be done highest possible to handle any API.

nicholastmosher: I agree, let's look at how we can do that. I just broke this PR into two pieces because something in the second half broke the build, so let's come back to this when I open the next PR with that code

nicholastmosher:
I just broke this PR in half because something in the second half broke the tests. This PR should now be all passing, will fix the second half and open a new PR

nicholastmosher:
Ok nevermind about the breaking the PR in half, I fixed the second half so it's all back in here now.

sehz:
still CI failures

nicholastmosher:
Ah, dang clippy. Should be good now

crates/fluvio-spu-schema/src/server/api.rs
```diff
@@ -67,3 +67,38 @@ impl ApiMessage for SpuServerRequest {
...
+            Self::ApiVersionsRequest(request) => request.header.correlation_id(),
+            Self::ProduceRequest(request) => request.header.correlation_id(),
```
sehz, line 75: So requires we have to implement this for every request.   Instead, do it this before: `decode_with_header`

sehz (changes requested):

crates/fluvio-spu-schema/src/server/api.rs
nicholastmosher: I don't think that we can instrument `decode_with_header` and get the result we want. Instrumentation follows a call stack, not a stream pipeline, which means the top-level instrumentation is at the top-level service handler, i.e. `handle_spu_request` in the SPU public service. I think it's unfortunate that we have to fan out like this since all the enum variants contain `RequestMessage<T>` with different types for `T`. I think this may indicate that there's a better way to structure this request format, but I didn't want to go that deep in trying to change things.

For now, I think this is a fixed-cost thing that we can improve later if we end up adding many more types of requests, but we have not added a new request type in a very long time so I think this is manageable, and I think the benefit to having `correlation_id` and `client_id` in the top-level spans is worth it.

crates/fluvio-spu/src/services/public/api_versions.rs
```diff
@@ -12,13 +12,10 @@ use fluvio_spu_schema::server::update_offset::UpdateOffsetsRequest;
 use fluvio_spu_schema::{ApiVersionsRequest, ApiVersionsResponse};
 
 #[instrument(skip(request))]
-pub async fn handle_kf_lookup_version_request(
+pub async fn handle_api_version_request(
```
sehz, line 15: Let's put this as start of span by removing skip

crates/fluvio-spu/src/services/public/fetch_handler.rs
```diff
@@ -1,72 +1,99 @@
...
 
 use crate::core::DefaultSharedGlobalContext;
+use dataplane::record::FileRecordSet;
+use dataplane::api::ResponseMessage;
 
 /// perform log fetch request using zero copy write
-#[instrument(skip(request, ctx, sink))]
+#[instrument(skip(request, ctx))]
 pub async fn handle_fetch_request(
-    request: RequestMessage<FileFetchRequest>,
```
sehz, line 14: same comment

crates/fluvio-spu-schema/src/server/api.rs
sehz: See my comment below.  If we have span start at the handler, we should not need this

sehz (changes requested):

nicholastmosher:
The reason I did it the way I did was so that we don't have to add the correlation_id and client_id to each and every request handler, which is more code and maintenance than adding it once to the overarching handler (that way we never forget to add it if we add more requests). I don't see how adding these annotations to the individual handlers is better than this.

sehz:
> The reason I did it the way I did was so that we don't have to add the correlation_id and client_id to each and every request handler, which is more code and maintenance than adding it once to the overarching handler (that way we never forget to add it if we add more requests). I don't see how adding these annotations to the individual handlers is better than this.

I didn't actually say we need both.  Certainly, need `correlation_id` to trace at request level and need API as well.  

It would be better if we start with what ideal span should be.

sehz:
And break PR into two different PRs.  Most of changes are ok which is re-arrange things.  Most time consuming part will be settling design of SPAN

nicholastmosher:
@sehz I removed the correlation_id and client_id methods for now, we can revisit soon

crates/fluvio-spu/src/services/public/api_versions.rs
nicholastmosher: Let's revisit this when we talk about spans in more depth

sehz (changes requested):
don't agree this is right span level.  let's take

crates/fluvio-spu-schema/src/server/api.rs
```diff
@@ -67,3 +67,16 @@ impl ApiMessage for SpuServerRequest {
         }
     }
 }
+
+impl SpuServerRequest {
```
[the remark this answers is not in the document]
nicholastmosher: This is not used for the span, it is used for sending the response. One of the things I did in this PR was remove the uses of the `call_service` macro, and to do that I needed direct access to the api_version from each request. So this does not have to do with spans at all.

https://github.com/infinyon/fluvio/pull/1532/files#diff-2277166e98453fe79d4a9c077d19f29f5e51fbb911900f3671aea595cbf572d3R87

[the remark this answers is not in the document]
sehz: why call_service was removed?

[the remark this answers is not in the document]
nicholastmosher: It made it incredibly difficult to reason about the logical flow of the code. Even after reading it several times it was very hard to wrap my head around exactly what happens when we hit that macro. Every time I read it, I'd be like "does this await? does this spawn a task? does this just execute a function? what's the control flow here?" I think it was a bad place to use a macro when a function would have done just fine. I find this version to be much more readable to myself as well as be more readable for fresh eyes on this code.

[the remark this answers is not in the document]
sehz: The purpose of the macro is to put consistent tracing and behavior.  this is not optional

nicholastmosher:
@sehz I undid the changes to `call_service`, how does this look now?

sehz:
clippy failing

nicholastmosher:
@sehz fixed clippy

crates/fluvio-spu/src/services/public/mod.rs
```diff
@@ -1,30 +1,143 @@
...
+use self::produce_handler::handle_produce_request;
+use self::fetch_handler::handle_fetch_request;
+use self::offset_request::handle_offset_request;
+use self::offset_update::handle_offset_update;
+use self::stream_fetch::StreamFetchHandler;
 pub use stream_fetch::publishers::StreamPublishers;
 
-pub(crate) type PublicApiServer =
-    FluvioApiServer<SpuServerRequest, SpuServerApiKey, DefaultSharedGlobalContext, PublicService>;
+pub(crate) type SpuPublicServer = FluvioApiServer<
```
sehz, line 28: Adding SPU prefix  is redundant.  We are in the SPU so there is no confusion here.

crates/fluvio-spu/src/services/public/mod.rs
```diff
@@ -1,30 +1,143 @@
...
-// start server
-pub fn create_public_server(addr: String, ctx: DefaultSharedGlobalContext) -> PublicApiServer {
-    info!(
-        "starting SPU: {} at public service at: {}",
-        ctx.local_spu_id(),
-        addr
-    );
+#[derive(Debug)]
+pub struct SpuPublicService {
+    _0: (), // Prevent construction
```
sehz, line 37: what's point of doing this?

crates/fluvio-spu/src/services/public/mod.rs
```diff
@@ -1,30 +1,143 @@
...
+    }
+
+    pub fn server(addr: String, ctx: DefaultSharedGlobalContext) -> SpuPublicServer {
+        info!(
+            spu_id = ctx.local_spu_id(),
+            %addr,
+            "Starting SPU public service:",
+        );
+
+        FluvioApiServer::new(addr, ctx, SpuPublicService::new())
```
sehz, line 52: This seems like pointless indirection.  Shouldn't better to wrap `FluvioApiServer` ?

sehz (changes requested):

crates/fluvio-spu/src/services/public/mod.rs
nicholastmosher: This is exactly the same code as what used to be called `create_public_server`, I just put it as an associated function on `SpuPublicService` as a convenience

crates/fluvio-spu/src/services/public/mod.rs
nicholastmosher: I don't really like having structs that have no fields, it seems like we should either have a unit struct (which can be constructed out of thin air), or to add one private field to prevent it from being constructed except via the `new` function. This is more future-proof, since it prevents destructuring on the type. There is no downside to doing this.

nicholastmosher:
@sehz thoughts?

crates/fluvio-spu/src/services/public/mod.rs
sehz: They are different because they are import from another modules.  If it is renamed, we have rename other variable as well.

crates/fluvio-spu/src/services/public/mod.rs
sehz: `_0` is not standard Rust idiom.  Still don't see why need to be wrapped.

nicholastmosher: I have seen it used before, different folks use different names but the pattern is the same:
https://rust-unofficial.github.io/patterns/idioms/priv-extend.html#alternative-private-fields-for-structs

crates/fluvio-spu/src/services/public/mod.rs
nicholastmosher: Do you mean if the `SpuPublicServer` type is renamed? I would say the same thing about the `create_public_service` function, I don't think this argument favors one over the other.

sehz: Not seeing any increase in clarity or reduction of code with proposed renaming other than making everything verbose.  Just stick with changes that improve instrumentation or debugging.

nicholastmosher:
@sehz I undid the renaming that I'd done, I put back `create_public_server`

sehz (approved):

PR #1556, opened 2021-09-02
Replace SmartStream enum with a trait

nicholastmosher:
This is the big goal I was trying to achieve yesterday, I had to rewrite it because the first time I wrote it I must have introduced a bug somewhere during shuffling things around, but this version works.

Check out the additions/deletions ratio on this one 😎


sehz (approved):
Nice!

Issue #1544, opened 2021-08-31
fluvio-test - Add other cluster support


tjtelan:
We want to create apples-to-apples test data from our workflows between Pulsar and Kafka clusters.

The same CLI flags for configuring the behavior of a producer and consumer need to be shared. And the tests need to be reasonably blind to the type of cluster it is performing actions against.

At the very least, the producer and consumer handling should be shared across all the tests and not just the smoke test.

PR #1583, opened 2021-09-08
[Merged by Bors] - Add Longevity test to fluvio-test

### LINKED ISSUE #1554: fluvio-test doesn't respect `--topic-name` flag
Simple smoke test with a provided `--topic-name`.
`./target/debug/fluvio-test smoke --topic-name test2 -d -k`

It is expected that the smoke test uses a topic named `test2`. however, it uses the default topic name `test`

```
fluvio topic list
 NAME  TYPE      PARTITIONS  REPLICAS  IGNORE-RACK  STATUS                   REASON 
 test  computed      1          1                   resolution::provisioned   
```

### LINKED ISSUE #1577: Move Hourly Test as part of Fluvio Test

tjtelan:
Based off of Concurrent test
Test is configured to run for a set amount of time (10 min default in hourly test)

Producer and consumer were decoupled from each other wrt the concurrent test. Producer calculates a CRC and embeds into record sent to Fluvio. Consumer re-calculates record and verifies CRC matches.

Topic data loading before test was disabled for now, as it does not behave consistently in CI.

Resolves #1577 
Resolves #1554


crates/fluvio-test-util/test_meta/environment.rs
```diff
@@ -122,8 +131,8 @@ pub struct EnvironmentSetup {
     pub keep_cluster: bool,
 
     /// topic name used
-    #[structopt(short("t"), long, default_value = "topic")]
```
sehz, line 125: why is removed?.

sehz (changes requested):
Overall looks good.  Just a minor question

crates/fluvio-test-util/test_meta/environment.rs
tjtelan: It wasn't completely removed. I moved this default value to `crates/fluvio-test-util/test_runner/test_meta.rs` for #1544 since I needed this fixed to test.

I changed this to Option<String> so I could support a hierarchy for overriding the topic name the test should use and I needed the possibility for the value to be `None`.

1. If we provide `--topic-name`, then the test will use that value, overriding the `fluvio_test` macro
2. Next, if the `fluvio_test` macro providers a topic name, then use that value
3. Otherwise we'll default to the topic name `test`

sehz: seems like this could be clean up. but it is not show stopper. let's merge first

PR #1588, opened 2021-09-09
[Merged by Bors] - Add fluvio-test to dev releases

tjtelan:
Also explicitly opt-out `fluvio-test` from fluvio package publish in the `Publish` and `Release` workflows.

`fluvio-test` is needed in the hourly test workflow for #1577, but we don't want to spend any time building because we already do all of that in `CI`. To avoid rebuilding, we should publish to the `dev` release.


sehz (approved):
LGTM

PR #1590, opened 2021-09-09
Use cargo-xtask pattern to write build configuration in Rust

nicholastmosher:
After reading [this blog on large Rust workspaces](https://matklad.github.io/2021/08/22/large-rust-workspaces.html) I found myself thinking more about the [cargo-xtask](https://github.com/matklad/cargo-xtask/) pattern that was mentioned there and I think it's a really good fit for our project. I went ahead and started playing with it with some really good results.

Basically, the goal of this is to reduce or eliminate ad-hoc scrips and Makefiles we need in order to build the project. This relies on the `alias` feature of the `.cargo/config.toml` file in order to allow us to run `cargo xtask` to build and run the new `xtask` crate (xtask is not a dependency, it is a new crate added to our workspace).

To see what I've already put together in this PR, try out running `cargo xtask`, you should see this:

```
cargo xtask
    Finished dev [unoptimized + debuginfo] target(s) in 0.18s
     Running `target/debug/xtask`
xtask 0.0.0

USAGE:
    xtask <SUBCOMMAND>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

SUBCOMMANDS:
    build                 Build fluvio-cli, fluvio-run, fluvio-test, and all smartstreams
    build-cli             Build fluvio-cli
    build-cluster         Build fluvio-run
    build-test            Build fluvio-test
    build-smartstreams    Build all smartstream examples
    build-image           Build fluvio-run and package it into a docker image
    clippy                Run clippy on all crates
    test                  Run all unit, doc, and integration tests
    test-units            Run all unit tests
    test-docs             Run all doc tests
    test-integration      Run all integration tests
    install-target        Install the toolchain for the given target, via rustup or cross
    help                  Prints this message or the help of the given subcommand(s)
```

I wrote this initial set of subcommands to try to replace some commonly used Make tasks. Some direct replacements are:

```
make build-cli -> cargo xtask build-cli
make build-cluster -> cargo xtask build-cluster
make build-test -> cargo xtask build-test
make build_smartstreams -> cargo xtask build-smartstreams
make check-clippy -> cargo xtask clippy
make run-all-doc-test -> cargo xtask test-docs
make run-all-unit-test -> cargo xtask test-units
make run-integration-test -> cargo xtask test-integration
make install_rustup_target -> cargo xtask install-target
make fluvio_image -> cargo xtask build-image [--minikube] [--k3d] [--kind]
```

I have so far not deleted the corresponding Make tasks, I want to try to gradually test out this workflow before doing anything like that. But my hope is that we can eventually use this pattern to replace Make, cargo-make, and most bash scripts in the repo.

### Abuot `cargo xtask build-image`

I wanted to quickly describe what `build-image` does, because I think it is a much nicer docker workflow than what we had before. Basically, it first builds `fluvio-run`, using the expected behavior with the `--release` flag and also the `--target` flag (behaving the same way cargo does in those cases). It also builds the docker image and exports it as a tar directly adjacent to the binary that was compiled. So for example, for the following invocations, we get the following output:

```
cargo xtask build-image -> target/debug/infinyon-fluvio.tar
cargo xtask build-image --release -> target/release/infinyon-fluvio.tar
cargo xtask build-image --target=x86_64-unknown-linux-musl -> target/x86_64-unknown-linux-musl/debug/infinyon-fluvio.tar
```

I think this is a great workflow because it places the tar in such a way that it is easy to recognize the nature of that docker image (built in debug or release, built for what target). This also makes it easy to import the image into minikube/k3d/kind or to upload the image to a cache or to publish it.


.github/workflows/ci.yml
```diff
@@ -96,15 +96,15 @@ jobs:
 
       - name: Build fluvio
         if: ${{ matrix.binary == 'fluvio' }}
-        run: make build-cli
+        run: cargo xtask build-cli
```
sehz, line 99: this is more verbose than before. can you make alias or reduce boiler plate?

.github/workflows/ci.yml
```diff
@@ -96,15 +96,15 @@ jobs:
...
 
       - name: Build fluvio-run
         if: ${{ matrix.binary == 'fluvio-run' }}
-        run: make build-cluster
+        run: cargo xtask build-cluster
```
sehz, line 103: same

.github/workflows/ci.yml
```diff
@@ -96,15 +96,15 @@ jobs:
...
 
       - name: Build fluvio-test
         if: ${{ matrix.binary == 'fluvio-test' }}
-        run: make build-test
+        run: cargo xtask build-test
```
sehz, line 107: same

.github/workflows/ci.yml
```diff
@@ -192,7 +192,7 @@ jobs:
 
       - name: Build fluvio
         if: ${{ matrix.binary == 'fluvio' }}
-        run: make build-cli
+        run: cargo xtask build-cli
```
sehz, line 195: same

Dockerfile
```diff
@@ -0,0 +1,3 @@
+ARG ARCH=
```
sehz, line 1: What's reason for this change? Is this related to this PR if not, let's create separate PR

crates/xtask/src/build.rs
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    pub fn build_cli(&self) -> Result<PathBuf> {
+        install_target(self.target.as_deref())?;
+        let path = self.build_bin("fluvio")?;
+        Ok(path)
+    }
+
+    pub fn build_cluster(&self) -> Result<PathBuf> {
+        install_target(self.target.as_deref())?;
```
sehz, line 39: code duplication here

crates/xtask/src/build.rs
```diff
@@ -0,0 +1,142 @@
...
+        }
+        let profile = if self.release { "release" } else { "debug" };
+        path.push(profile);
+        path.push(bin);
+
+        Ok(path)
+    }
+
+    pub fn build_smartstreams(&self) -> Result<()> {
+        install_target(Some("wasm32-unknown-unknown"))?;
```
sehz, line 81: make this as constant

crates/xtask/src/build.rs
```diff
@@ -0,0 +1,142 @@
...
+        cmd!(
+            CARGO,
+            "build",
+            "--target=wasm32-unknown-unknown",
```
sehz, line 85: use constant

sehz (changes requested):
Some initial feedback.  
But overall, I don't see how is improvement over makefile.  Rust's strength is type checking and safety and this example doesn't emphasis any of that.   Here, it is hard to see what's dependencies.

Much better use of this is replacement of bash script or small automation.

crates/xtask/src/lib.rs
```diff
@@ -0,0 +1,161 @@
...
+
+#[derive(StructOpt, Debug)]
+#[structopt(settings = &[structopt::clap::AppSettings::DeriveDisplayOrder])]
+pub struct Root {
+    #[structopt(subcommand)]
+    pub cmd: RootCmd,
+}
+
+#[derive(StructOpt, Debug)]
+pub enum RootCmd {
```
sehz, line 26: This defeats purpose of xtask.  It's better if this is separated out

.github/workflows/ci.yml
nicholastmosher: IMO this is a small cost to pay for the customization possible with `xtask`. The way cargo aliases work is that the shortest you can get is `cargo <X>`, you can't make a totally independent alias that's not under `cargo`. The main point is to be able to clone the project and without having any external tools (including make) you can just execute your tasks.

crates/xtask/src/lib.rs
nicholastmosher: Can you expand on this? To me the purpose of xtask is to move all logic associated with our build into Rust, where we can write clean and reusable code to maintain our build process better, and I don't see what `RootCmd` itself has to do with that.

.github/workflows/ci.yml
sehz: ?? you just replacing one tools (make) with another tool (task) not removing it

crates/xtask/src/lib.rs
sehz: Seems like this is overloaded with many task which don't think xtask is designed for.  Ideally, there should many tasks such that they can be composed not a single monolithic super task:

For example, I can see task break down in terms of:
* Building components (CLI, platform, smart stream)
* Building docker image
* Running integration test
* Creating K8 cluster, cleaning
* Publishing
* Releasing
* etc

nicholastmosher:
> Much better use of this is replacement of bash script or small automation

One of the reasons that I pointed out the `cargo xtask build-image` command in particular is because that command could allow us to get rid of `k8-util/docker/build.sh`, which is one of those loose scripts that we have to maintain, so it definitely helps this case you've made

PR #1591, opened 2021-09-09
Added spu local address logic

### LINKED ISSUE #1537: Use Internal Address for Fluvio Config inside a Managed Connector
When running a fluvio client outside k8, the config has an endpoints per profile.
For example, a local k8 cluster has:

```toml
[cluster."127.0.0.1"]
endpoint = "192.168.49.2:9003"

[cluster."127.0.0.1".tls]
tls_policy = "verified"
tls_source = "inline"
```

A given connector will run inside the k8s cluster so the endpoint will not be the same. This needs to also be solved in the spu return.

https://github.com/infinyon/fluvio-connectors/issues/13 is the old ticket for this.

After a discussion, it seems that the most ideal user experience to have a field on [`FluvioConfig`](https://github.com/infinyon/fluvio/blob/927ccf4ce316f89e7f5e25014aa0cc7992f4006a/crates/fluvio/src/config/cluster.rs#L13-L24), to use "internal endpoints" (name undecided). This should be used in [`connect_to_leader` in spu.rs](https://github.com/infinyon/fluvio/blob/927ccf4ce316f89e7f5e25014aa0cc7992f4006a/crates/fluvio/src/spu.rs#L76-L91).

### LINKED ISSUE #1546: Restructure SPU spec to separate cluster endpoint and ingress endpoint
Ingress should be separate from the local endpoint so that the client can choose how to connect. This will allow clients to be deployed within the same K8 cluster.

Renaming from public/private to data/control should clarify their intended usage.

Also, if possible, it would be better to move the endpoint fields from spec to status, as they do not define the desired state of an SPU, rather they are derived from K8 service after creation.

Example:
```
SpuSpec {
    id: 0,
    type: Managed,
    rack: None,
}

SpuStatus {
    resolution: Online
    data_endpoint: {
        host: "fluvio-spu-main-0.fluvio-spg-main",
        port: 9005
    },
    data_ingress_endpoints: [{
        host: "192.168.49.2",
        port: 9005
    }]
    control_endpoint: {
        host: "fluvio-spg-main-0.fluvio-spg-main",
        port: 9006
    },
}
```

simlay:
Closes #1537.
Closes  #1546.

Tried to make this a non-breaking change. Looks like with the optional type for `public_endpoint_local`, it works as expected.

Testing this was a bit of a pain, we ended up spinning up a new pod in the k8s cluster and running the fluvio client with `use_spu_local_address = true` in the config to verify that it was using the correct endpoint.

### DIFF
--- a/crates/fluvio-controlplane-metadata/src/spu/spec.rs
+++ b/crates/fluvio-controlplane-metadata/src/spu/spec.rs
@@ -12,8 +12,6 @@ use std::fmt;
 
 use flv_util::socket_helpers::EndPoint as SocketEndPoint;
 use flv_util::socket_helpers::EndPointEncryption;
-use fluvio_types::defaults::{SPU_PRIVATE_HOSTNAME, SPU_PRIVATE_PORT};
-use fluvio_types::defaults::SPU_PUBLIC_PORT;
 use fluvio_types::SpuId;
 use flv_util::socket_helpers::ServerAddress;
 
@@ -21,7 +19,7 @@ use dataplane::core::{Encoder, Decoder};
 use dataplane::bytes::{Buf, BufMut};
 use dataplane::core::Version;
 
-#[derive(Decoder, Encoder, Debug, Clone, PartialEq)]
+#[derive(Decoder, Encoder, Debug, Clone, PartialEq, Default)]
 #[cfg_attr(
     feature = "use_serde",
     derive(serde::Serialize, serde::Deserialize),
@@ -36,6 +34,9 @@ pub struct SpuSpec {
     pub private_endpoint: Endpoint,
     #[cfg_attr(feature = "use_serde", serde(skip_serializing_if = "Option::is_none"))]
     pub rack: Option<String>,
+
+    #[cfg_attr(feature = "use_serde", serde(skip_serializing_if = "Option::is_none"))]
+    pub public_endpoint_local: Option<Endpoint>,
 }
 
 impl fmt::Display for SpuSpec {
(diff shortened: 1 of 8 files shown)


crates/fluvio-controlplane-metadata/src/spu/spec.rs
```diff
@@ -12,16 +12,14 @@ use std::fmt;
 
 use flv_util::socket_helpers::EndPoint as SocketEndPoint;
```
sehz, line 13: can you bump up version?

crates/fluvio/src/spu.rs
```diff
@@ -78,7 +78,17 @@ impl SpuPool {
 
         debug!("connecting to spu: {}", spu.spec);
         let mut client_config = self.config.with_prefix_sni_domain(spu.key());
-        let spu_addr = spu.spec.public_endpoint.addr();
+
+        let spu_addr =
+            if self.config.use_spu_local_address && spu.spec.public_endpoint_local.is_some() {
```
sehz, line 83: prob want something like: `if let Some(local) = spu.spec.public_endpoint`

sehz (changes requested):
Looks good with couple of nits.  Also add to changelog

crates/fluvio/src/spu.rs
simlay: Yeah, this is kinda exactly what nick and I discussed when I wrote it.

nicholastmosher: You can do a match with a guard, something like:

```rust
match spu.spec.public_endpoint_local {
    Some(endpoint) if self.config.use_spu_local_address => { ... }
    _ => { ... }
}
```

crates/fluvio-controlplane-metadata/src/spu/spec.rs
simlay: Done.

sehz (approved):
LGTM

PR #1617, opened 2021-09-13
Cleanup aliases in `fluvio-spu`

nicholastmosher:
This is some quick followup cleanup I've wanted to do for a while in the SPU. There are some patterns I've noticed that I think can be improved:

- I noticed a bunch of type aliases such as `type DefaultXyz = Xyz<FileReplica>`. I think a better pattern for this is by using a default generic in the struct definition, e.g. `struct Xyz<S = FileReplica>`, that way we can just say `Xyz` when we want `Xyz<FileReplica>`, or we can explicitly give a generic if we want to override that type.
- I also noticed that we had a bunch of `SharedXyz` which almost always was an alias for `type SharedXyz = Arc<Xyz>` I think having the `Shared` aliases all over the place was confusing (and the Arc version is actually shorter to type). Combine this with the fact that Shared only _sometimes_ meant Arc, and other times had different synchronization mechanisms, made it very confusing to reason about the data types.

In this PR I just went through and removed those aliases and substituted the true types in the `fluvio-spu` crate. I think this makes the codebase much clearer and easier to reason about (e.g. knowing you're dealing with an Arc instead of guessing).


sehz:
CI failures

sehz:
type alias is useful way to reduce noise in the code.

nicholastmosher:
I think if it were not possible to use the default generics like `struct Foo<S = FileReplica>` then I would mostly agree, but the default generic helps with readability quite a lot (you can just use `Foo` most places). And then `Arc<T>` is shorter and clearer than `SharedT` IMO.

nicholastmosher:
@sehz the CI failures here have been fixed

PR #1620, opened 2021-09-14
[Merged by Bors] - Add crates publishing to Release

### LINKED ISSUE #1397: Release: Automate crates publishing
As part of release process, need to make sure all crates have been published.  Since we know exactly dependencies between crates. We can perform explicit publishing steps.

tjtelan:
Resolves #1397

Add script to handle serializing the publish of our crates in a specific order so it can be added into the Release flow


tjtelan:
@sehz Can we confirm if the `CARGO_REGISTRY_TOKEN` repo secret is up-to-date? My latest test run is failing in an unexpected way because we expect that we are logged in. I'm using the env var for authentication based on https://doc.rust-lang.org/cargo/reference/config.html#credentials

---

If we are logged in, we should see an error that looks like this:
```
Caused by:
  the remote server responded with an error: crate version `0.2.5` is already uploaded
```

But instead we get an error like this:
```
 Caused by:
  the remote server responded with an error (status 403 Forbidden): must be logged in to perform that action
```

https://github.com/infinyon/fluvio/pull/1620/checks?check_run_id=3593466010

tjtelan:
It appears that the secret is empty? I was able to reproduce the error message when I set the `CARGO_REGISTRY_TOKEN` env var to be empty.

```
$ export CARGO_REGISTRY_TOKEN=
$ echo $CARGO_REGISTRY_TOKEN

$ ./release-scripts/publish-crates.sh
[...]
error: failed to publish to registry at https://crates.io

Caused by:
  the remote server responded with an error (status 403 Forbidden): must be logged in to perform that action
fluvio-types upload check failed. Will try again
~/fluvio
❌ Max number of publish attempts reached
❌ Max attempts: 3
```

But when I set the env var to use my token, it works

```
$ export CARGO_REGISTRY_TOKEN=***
$ ./release-scripts/publish-crates.sh
[...]
error: failed to publish to registry at https://crates.io

Caused by:
  the remote server responded with an error: crate version `0.2.5` is already uploaded
~/fluvio
✅ Publish success after # Attempts: 1
✅ Crates uploaded: 0
```

sehz:
It could be that secrets are only available in master not in PR

sehz (approved):

tjtelan:
> It could be that secrets are only available in master not in PR

Possibly, yes. In that case, I believe that we know what to expect when we try to release.

tjtelan:
bors ping

PR #1622, opened 2021-09-14
[Merged by Bors] - Check status of most recent CD_Dev run

### LINKED ISSUE #1613: Release workflow must check for cd_dev success
release should only run when cd-dev workflow is success.  This can be done by checking status or running workflow.  With latest GitHub actions, custom action can invoke another action

tjtelan:
Resolves #1613

Job passes in this CI run: https://github.com/infinyon/fluvio/actions/runs/1232136512

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -800,8 +800,8 @@ jobs:
       - name: Get status of latest CD_Dev run
         id: cd_dev_check
         run: |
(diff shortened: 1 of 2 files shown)


sehz (commented):
other option is to re-run cd-dev workflow.  we can make cd-dev workflow as action which can be depend on this

sehz (approved):

Issue #1077, opened 2021-05-11
Add automatic batching to Producer API


nicholastmosher:
We want Fluvio's Producer API to be as easy-to-use as possible, while also achieving the highest possible performance. Our current TopicProducer API has several drawbacks that harm usability and make high performance the exception rather than the norm. The number-one problem with the current API is the **lack of automatic batching**.

Let's take a look at some things the current TopicProducer API does poorly, and what things we want to be able to do well.

## What do we do now?
- Currently, `producer.send(...).await?` will make a full round-trip request/response to the SPU.
- This means that every single record is sent as an individual network request, it is _highly inefficient_
- To do any sort of batching, we require the caller to use `producer.send_all(...)`, which has several downsides:
  - It requires the user to memorize one more function and know when it is appropriate to use it (hard to use)
  - It requires the user to write their own buffering logic any time they want to use batching, when batching should be the default in most cases.
- We do not provide any metadata to describe committed records, e.g. Offset or Partition

## What we _should_ be doing

- We should accept individual records from the caller, but automatically perform batching to increase efficiency
- Batching should be configurable with the following options:
  - Linger: A duration of time during which new records will be buffered rather than immediately sent. Every time the linger duration elapses, the producer should flush all buffered records to SPUs.
  - Max Bytes: The maximum size that a batch may become before it _must_ be flushed. Upon accepting a new record, if that record would increase the size of the batch above `max_bytes`, it will trigger a flush to SPU, and the new record will be added to the next batch.
- We should ensure that callers can still detect when their record has been committed or failed to commit
  - We will tackle this in Phase 2
- We should provide access to metadata for committed records, such as Offset and Partition
  - Phase 2
- We should provide the ability to manually flush any buffered records
- Accomplish this with as few breaking API changes as possible

---

## API Changes

The following changes will be made to the Producer API to support batching:

1) The addition of `TopicProducer::flush()` to flush batched records
2) The addition of `ProducerConfig` to specify batching behavior (batch duration and batch size)
3) A new `Fluvio::topic_producer_with_config` method for instantiating a producer with non-default configuration

1) Definition

```rust
impl TopicProducer {
    /// Waits for all records to be flushed from the internal buffer
    pub async fn flush(&self) -> Result<(), FluvioError> { ... }
}
```

2) Definition

```rust
pub struct ProducerConfig {
    pub batch_duration: std::time::Duration,
    pub batch_size: usize, // Max batch size in bytes
}
```

3) Definition

```rust
[...]

Issue #1626, opened 2021-09-14
Producer Phase 2: Receive acknowledgements for committed/failed records


nicholastmosher:
- This `send` should put a record into a queue and return immediately without blocking/waiting
- This should include a `flush` feature which allows waiting until records have been written

### API Proposal

This API introduces the following types:

```rust
struct Producer { ... }
struct ProducerConfig {
    linger: Duration,
    max_bytes: u32,
    acks: AckPolicy
}
/// The number of replica commit acknowledgements needed
enum AckPolicy {
    None, // acks 0
    Leader, // acks 1
    All, // acks all
}
struct ProduceStatus { ... }
impl ProduceStatus {
    /// Wait for metadata about the Record that we tried to send.
    ///
    /// This will be Ok if the record was committed or Err if the batch failed.
    /// This is not available until the producer receives all needed acks from the SPU, so
    /// `.await`ing this will wait until that happens.
    async fn record_metadata(&mut self) -> Result<&RecordMetadata, &FluvioError> { ... }
    /// Wait for metadata about the Batch in which this record was sent.
    ///
    /// This is available as soon as the producer's batched _request_ is ready, which is sooner
    /// than when we will know about the actual commit/RecordMetadata status.
    async fn batch_metadata(&mut self) -> Result<&BatchMetadata, &FluvioError> { ... }
}
struct RecordMetadata {
    /// The committed offset that this record was written to
    pub offset: Offset,
    /// The index of the partition that this record was written to
    pub partition: PartitionId,
}
struct BatchMetadata {
    pub partition: PartitionId, // The PartitionId of the batch the record was assigned to
    pub batch_size: u32, // The number of bytes in the batch the record was assigned to
}
```

The new `Producer` type will be accessed via `fluvio.producer()`, similar to the old `fluvio.topic_producer("topic")` but it will be topic-agnostic. The new `Producer` will be able to send records to any topic, it will not be constrained.

```rust
impl Fluvio {
    fn producer(&self) -> Producer { ... }
    fn producer_with_config(&self, config: ProducerConfig) -> Producer { ... }
}
```

The methods on the new producer include:

```rust
impl Producer {
    pub async fn send<S, K, V>(&self, topic: S, key: K, value: V) -> Result<ProduceStatus, FluvioError>
    where:
        S: Into<String>,
        K: Into<RecordKey>,
        V: Into<RecordData>,
    { ... }

    /// Flush any records that are buffered. Does not return until responses have been received for all batches.
[...]

PR #1631, opened 2021-09-15
Producer batching

nicholastmosher:
Closes https://github.com/infinyon/fluvio/issues/1077.


crates/fluvio/src/producer/dispatcher.rs
```diff
@@ -0,0 +1,219 @@
...
+    /// This will be triggered by the [`Producer`] that starts this dispatcher.
+    shutdown: Arc<StickyEvent>,
+}
+
+enum Event {
+    NewRecord((String, RecordPending<()>)),
+    Timeout(Instant),
+}
+
+impl ProducerDispatcher {
```
sehz, line 48: follow dispatcher pattern in rest of the code.
and use similar flush logic in the SPU: https://github.com/infinyon/fluvio/blob/master/crates/fluvio-storage/src/mut_records.rs

sehz (changes requested):
https://github.com/infinyon/fluvio/issues/1626 needs to be updated with latest design.

crates/fluvio/src/producer/dispatcher.rs
```diff
@@ -0,0 +1,236 @@
...
+    Timeout,
+}
+
+impl Dispatcher {
+    pub(crate) fn new(pool: Arc<SpuPool>, incoming: Receiver<DispatcherMessage>) -> Self {
+        let partitioner = Box::new(SiphashRoundRobinPartitioner::new());
+        let shutdown = StickyEvent::new();
+
+        // Set up the event stream: Select between new records and timeouts
+        let timer =
```
sehz, line 51: look at other dispatcher is done.  it's not good to pass stream in the future like this. it's better to create stream after spawning.  this willl avoid static lifetime

sehz (changes requested):
need integration test.  please add to `fluvio-test`

crates/fluvio/src/producer/mod.rs
```diff
@@ -63,9 +103,19 @@ impl TopicProducer {
...
+        let value = value.into();
+        let record = Record::from((key, value));
+        let pending_record = PendingRecord {
+            partition: (),
+            record,
+            topic: self.topic.clone(),
+        };
+        let msg = DispatcherMessage::Record(pending_record);
+
+        self.inner.dispatcher.send(msg).await.map_err(|_| {
```
sehz, line 116: So what happens to pending records?  Shouldn't this wait if producer msg is in flight?

crates/fluvio/src/producer/mod.rs
```diff
@@ -125,6 +175,46 @@ impl TopicProducer {
...
+    /// for i in 0..100 {
+    ///     producer.send(i.to_string(), format!("Hello, {}", i)).await?;
+    /// }
+    /// producer.flush().await?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub async fn flush(&self) -> Result<(), FluvioError> {
+        // Create a channel for the dispatcher to notify us when flushing is done
+        let (tx, rx) = async_channel::bounded(1);
```
sehz, line 200: so what happens if you try to flush twice?

sehz (changes requested):

crates/fluvio/src/producer/mod.rs
nicholastmosher: This will now wait whenever there are messages in flight. This channel is now a `bounded(1)`, which means that the sender must wait until the dispatcher is ready to receive and buffer a record. This only happens when the dispatcher is not busy waiting on producer responses.

sehz: but channel is 1 sized that means it has 1 pending msg?

sehz: It would be more clean if buffer management and communication to SPU are separated.  Buffer management can be done at time send to make it easier to manage state.

nicholastmosher: Now using a rendesvous channel, so `dispatcher.send` will not complete until the dispatcher actually takes it from the receiver, ensuring there are no "pending" records in limbo.

crates/fluvio/src/producer/mod.rs
nicholastmosher: Flushing twice works as expected. Here we have a unique `bounded(1)` for _each_ call to flush, so they do not interfere with one another.

crates/fluvio/src/producer/dispatcher.rs
nicholastmosher: I've updated the dispatcher since this comment, it now respects `batch_size` and `batch_duration` when flushing

crates/fluvio-test/src/tests/producer_fail/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+        for i in 0..1_000 {
+            producer.send(RecordKey::NULL, i.to_string()).await?;
+        }
+        println!("Send 1000");
+
+        cluster_manager.terminate_spu(leader).expect("terminate");
+        println!("Terminate SPU");
+
+        for i in 0..1_000 {
+            producer.send(RecordKey::NULL, i.to_string()).await?;
```
sehz, line 90: need assertion here

crates/fluvio-test/src/tests/producer_fail/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+        .get_cluster()
+        .expect("cluster")
+        .env_driver()
+        .create_cluster_manager();
+
+    println!("Got cluster manager");
+
+    let result: Result<_, FluvioError> = (|| async move {
+        for i in 0..1_000 {
+            producer.send(RecordKey::NULL, i.to_string()).await?;
```
sehz, line 82: assert

crates/fluvio/src/error.rs
```diff
@@ -45,6 +48,10 @@ To interact with this cluster, please install the matching CLI version using the
     ConsumerConfig(String),
     #[error("Encountered a runtime error in the user's SmartStream: {0}")]
     SmartStreamRuntime(#[from] SmartStreamRuntimeError),
+    #[error("Producer error")]
+    Producer(#[from] ProducerError),
+    #[error("An internal producer error occurred: {0}")]
+    InternalProducerError(String),
```
sehz, line 54: There is nothing user can do.  Just mapped to Other instead of creating another variant

crates/fluvio/src/producer/mod.rs
```diff
@@ -29,355 +35,301 @@ pub struct TopicProducer {
...
+///     ..Default::default()
+/// };
+///
+/// // Use config on this call:
+/// fluvio.topic_producer_with_config("topic", config).await?;
+/// # Ok(())
+/// # }
+/// ```
+#[derive(Debug)]
+pub struct ProducerConfig {
```
sehz, line 63: This is tied to specific batch strategy config

crates/fluvio/src/producer/mod.rs
```diff
@@ -29,355 +35,301 @@ pub struct TopicProducer {
...
-        let record_value = value.into();
-        self.send_all(Some((record_key, record_value))).await?;
-        Ok(())
+        let key = key.into();
+        let value = value.into();
+        let record = Record::from((key, value));
+
+        {
+            // Check if there have been any failures
+            let mut receiver = self.inner.status_receiver.lock().expect("Mutex poisoned");
```
sehz, line 192: should use async lock

crates/fluvio/src/producer/mod.rs
```diff
@@ -29,355 +35,301 @@ pub struct TopicProducer {
...
+                    BatchStatus::InternalError(e) => {
+                        return Err(FluvioError::InternalProducerError(e));
+                    }
+                    // No action to take on success
+                    _ => {}
+                }
+            }
+        }
+
+        let topic_spec = self.pool.lookup_topic(&self.topic).await?.spec;
```
sehz, line 207: this should been already computed. no need to do again for every send call

examples/produce-stress/src/main.rs
```diff
@@ -0,0 +1,24 @@
+//! This example shows how producing many records quickly will cause them to be batched
```
sehz, line 1: should be moved to docs

sehz (changes requested):

crates/fluvio-test/src/tests/producer_fail/mod.rs
nicholastmosher: So I can add an assertion in the first loop (before SPU is terminated), but I can't add an assertion here. The reason is that, ever since the SPU died, we don't know _which_ call to `send` is going to fail, because we don't know when the dispatcher is going to try to flush.

The way this test is set up is that each call to `send` uses `?` so that when an Err crops up _somewhere_, it will get caught by the closure surrounding it. If you look at the end of the test, there is an assertion on the final `Result` that comes out of the closure. Meaning that we are asserting that at least one call to `send` or `flush` failed, but we don't necessarily know which one.

crates/fluvio-test/src/tests/producer_fail/mod.rs
nicholastmosher: I'll add assertions here because they are pre-SPU-failure 👍

crates/fluvio/src/producer/mod.rs
```diff
@@ -29,355 +35,301 @@ pub struct TopicProducer {
...
-        self.send_all(Some((record_key, record_value))).await?;
-        Ok(())
+        let key = key.into();
+        let value = value.into();
+        let record = Record::from((key, value));
+
+        {
+            // Check if there have been any failures
+            let mut receiver = self.inner.status_receiver.lock().expect("Mutex poisoned");
+            while let Ok(status) = (*receiver).try_recv() {
```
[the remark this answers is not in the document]
nicholastmosher: Sure, so each time we call `producer.send()`, we want to return `Err` if something has gone wrong with a batch somewhere. The way we check that is by looking at the `status_receiver` channel, which is where the Dispatcher will send the success/failure status of each batch it flushes.

I'm using `while let Ok(status) = (*receiver).try_recv() { ... }` specifically because I don't want to _wait_ for a status in the channel, I just want to see if there is an Err _already there_. So if there's anything in the channel, it will sift through them, ignoring the successes but bubbling up `Err` if there was a `BatchStatus::Failed` or `BatchStatus::InternalError` in the channel.

crates/fluvio/src/producer/mod.rs
nicholastmosher: Isn't the point of the MetadataStores that the metadata could change? This shouldn't end up calculating anything, just look up the spec in the cache

crates/fluvio/src/producer/mod.rs
nicholastmosher: I can go ahead and do this if you want me to, but I was reading about this yesterday and I think the std mutex should be fine here. We don't need to hold the lock across any `.await` points and there shouldn't be high contention. The tokio guide was basically saying to reach for std::sync::Mutex since it is faster, and only fall back to an async mutex when really needed: https://tokio.rs/tokio/tutorial/shared-state#on-using-stdsyncmutex

Let me know if you want me to switch it, but that's why I had it that way to start

crates/fluvio-test/src/tests/producer_fail/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+        println!("Send 1000");
+
+        cluster_manager.terminate_spu(leader).expect("terminate");
+        println!("Terminate SPU");
+
+        for i in 0..1_000 {
+            producer.send(RecordKey::NULL, i.to_string()).await?;
+        }
+        println!("Send 1000");
+        producer.flush().await?;
```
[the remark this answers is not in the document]
nicholastmosher: see my comment above ☝️

crates/fluvio/src/producer/dispatcher.rs
```diff
@@ -0,0 +1,305 @@
...
+                let span = debug_span!("flush", reason = "timeout");
+                self.partition_pool.flush_all().instrument(span).await?;
+            }
+        }
+
+        Ok(())
+    }
+}
+
+pub(crate) struct PartitionPool {
```
sehz, line 111: So is this a shared instance? This would be problematic.  Why not just for this producer instance?

crates/fluvio/src/producer/dispatcher.rs
```diff
@@ -0,0 +1,305 @@
...
+                    .expect("just inserted, element must exist")
+            }
+        };
+
+        partition_producer.buffer(record, respond_to_caller).await?;
+        Ok(())
+    }
+
+    /// Flush the buffers for all partitions.
+    pub async fn flush_all(&mut self) -> Result<(), FluvioError> {
```
sehz, line 168: this would flush all replicas which is not something desired.  should flush only specific topic

crates/fluvio/src/producer/dispatcher.rs
```diff
@@ -0,0 +1,305 @@
...
+        for result in results {
+            if let Err(e) = result {
+                error!("Internal error broadcasting batch status: {}", e);
+            }
+        }
+        Ok(())
+    }
+}
+
+/// Manages a buffer for a single partition's records.
```
sehz, line 186: where is partition id?

crates/fluvio/src/producer/dispatcher.rs
```diff
@@ -0,0 +1,305 @@
...
+    /// Buffer the given record, flushing if necessary.
+    pub async fn buffer(
+        &mut self,
+        record: AssociatedRecord,
+        respond_to_caller: Sender<ProducerMsg>,
+    ) -> Result<(), FluvioError> {
+        // If this record is too large for the buffer, return it to the producer.
+        if !self.buffer.could_fit(&record) {
+            respond_to_caller
+                .send_async(ProducerMsg::RecordTooLarge(record, self.config.batch_size))
```
sehz, line 227: this is something you don't want async

crates/fluvio/src/producer/mod.rs
sehz: `async_mutex` doesn't use std mutex and much superior

crates/fluvio/src/producer/mod.rs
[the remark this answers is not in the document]
sehz: why not just use futures poll?  or use simple lock that returns (return Option<Error>) this make it harder to understand. similar to: https://github.com/infinyon/fluvio/blob/master/crates/fluvio-spu/src/control_plane/message_sink.rs

sehz (changes requested):

crates/fluvio/src/producer/mod.rs
```diff
@@ -29,355 +35,301 @@ pub struct TopicProducer {
...
+    fn drop(&mut self) {
+        debug!("Notify dispatcher shutdown");
+        self.dispatcher_shutdown.notify();
+    }
 }
 
 impl TopicProducer {
-    pub(crate) fn new(topic: String, pool: Arc<SpuPool>) -> Self {
+    pub(crate) fn new(topic: String, pool: Arc<SpuPool>, config: ProducerConfig) -> Self {
         let partitioner = Box::new(SiphashRoundRobinPartitioner::new());
```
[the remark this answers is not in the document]
nicholastmosher: What do you have in mind when you say "buffering strategy"? Is it a subset of the config options? Or is it more like an algorithm that could be represented by a trait?

crates/fluvio/src/producer/dispatcher.rs
nicholastmosher: Right now it actually is not shared, there is one Dispatcher/PartitionPool per TopicProducer (perhaps we need a different name than PartitionPool). I think we can make it shared at some point in the future

crates/fluvio/src/producer/dispatcher.rs
nicholastmosher: Currently, TopicProducer and Dispatcher are 1:1, so flush_all is just flushing all the partitions for that one topic

crates/fluvio/src/producer/dispatcher.rs
nicholastmosher: There is `ReplicaKey`

crates/fluvio/src/producer/buffer.rs
```diff
@@ -0,0 +1,197 @@
...
+            size: 0,
+            max_size,
+            records: VecDeque::new(),
+        }
+    }
+
+    /// Add a new [`Record`] to the buffer. If this record would cause the buffer to
+    /// exceed the `max_size`, the record is rejected and returned in an `Err`.
+    pub fn push(&mut self, pending: AssociatedRecord) -> Result<(), AssociatedRecord> {
+        let record_size = Encoder::write_size(
```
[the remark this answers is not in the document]
nicholastmosher: `could_fit` is called by the `PartitionProducer` before even trying to push the record. Here's the difference:

- `could_fit` describes whether a record is larger than the `max_size` of a buffer. If this is false, then we could never fit this record in this buffer even if the buffer were empty.
  - In this scenario, our only option is to give the caller an error, in this case we also return the record to them.
- `push` will check if the _adding the record_ to the buffer will make the occupied size of the buffer larger than `max_size`.
  - In this scenario, we can just go ahead and flush the buffer to make room for the new record.

crates/fluvio/src/producer/dispatcher.rs
```diff
@@ -0,0 +1,305 @@
...
+            let incoming = self.commands.take().unwrap();
+            let records = incoming.into_stream().map(Event::Command);
+            let select = futures_util::stream::select(timer, records);
+            select.take_until(self.shutdown.listen_pinned())
+        };
+
+        while let Some(event) = events.next().await {
+            let result = self.handle_event(event).await;
+            if let Err(e) = result {
+                error!("Producer Dispatcher error: {}", e);
```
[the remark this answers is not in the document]
nicholastmosher: What do you think should happen here? I believe that only internal logic errors come through here, everything else is sent through the status broadcaster channel. We could also send this Error through that channel but it isn't really a "Batch status" it is like a dispatcher internal error, not necessarily related to any specific batch.

crates/fluvio/src/producer/assoc.rs
```diff
@@ -0,0 +1,231 @@
...
+#[non_exhaustive]
+pub struct BatchFailure {
+    pub batch: BatchInfo,
+    pub error_code: ErrorCode,
+}
+
+/// A type to build a `ProduceRequest` while associating the [`RecordUid`] of each record.
+#[derive(Debug, Default)]
+pub(crate) struct AssociatedRequest {
+    pub(crate) request: ProduceRequest,
```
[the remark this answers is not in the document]
nicholastmosher: `AssociatedRequest` is a wrapper around `ProduceRequest` that also keeps track of what `RecordIds` of records were added to the request. This is created during the flushing process, right before sending through the socket.

crates/fluvio/src/producer/mod.rs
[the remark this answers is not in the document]
sehz: In future, we probably need support custom buffer/partitioning, so instead of harding coding this, it could be model as Trait.  Or at least minimum, tie configuration with Buffering so it's easier to handle that as separate configurable entity

crates/fluvio/src/producer/assoc.rs
```diff
@@ -0,0 +1,231 @@
...
+        Self::default()
+    }
+
+    pub fn add(&mut self, pending: AssociatedRecord) {
+        let topic = {
+            let maybe_topic = self
+                .request
+                .topics
+                .iter_mut()
+                .find(|t| t.name == pending.replica_key.topic);
```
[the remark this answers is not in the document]
nicholastmosher: Can you elaborate? I'm just taking a record with a given topic and partition and adding it to the right place in the request. There isn't anything as good as the `hash_map.entry()` API for Vec's

nicholastmosher:
> why not just use futures poll? or use simple lock that returns (return Option) this make it harder to understand. similar to: https://github.com/infinyon/fluvio/blob/master/crates/fluvio-spu/src/control_plane/message_sink.rs

Because we don't want the caller to _wait_ if there has not been a batch status reported yet. In that case, the caller just assumes that everything is fine and the records are getting buffered. If it does happen to see a failed BatchStatus, then it will return an Err to the caller

crates/fluvio/src/producer/dispatcher.rs
```diff
@@ -0,0 +1,305 @@
...
+    /// Pool for communicating with the SPUs.
+    spus: Arc<SpuPool>,
+    /// Producer behavior configuration.
+    config: Arc<ProducerConfig>,
+    /// This partition's buffered records.
+    buffer: RecordBuffer,
+    /// A channel for sending the status of batches back to caller.
+    broadcast_status: tokio::sync::broadcast::Sender<BatchStatus>,
+    /// The topic and partition this producer operates for.
+    replica_key: ReplicaKey,
```
sehz, line 197: I think you need state machine here to model behavior of each producer.  they can be:
* Buffering (Ready)
* Sending
* Error

Caller can wait if sending

crates/fluvio/src/producer/dispatcher.rs
```diff
@@ -0,0 +1,305 @@
...
+use dataplane::ReplicaKey;
+
+use crate::FluvioError;
+use crate::spu::SpuPool;
+use crate::producer::{Result, ProducerError};
+use crate::producer::{DispatcherMsg, ProducerConfig, ProducerMsg};
+use crate::producer::assoc::{AssociatedResponse, AssociatedRequest, AssociatedRecord, BatchStatus};
+use crate::producer::buffer::RecordBuffer;
+
+pub(crate) struct Dispatcher {
```
sehz, line 18: Making this into a sequential controller  seems hard to reason about behavior.  For example, what if there is error in one of the producer.  Also, difficult to pipeline.   Rather, make each PartitionProducer as independent controller each with its' independent state machine.

crates/fluvio/src/producer/dispatcher.rs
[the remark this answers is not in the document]
sehz: Why there would be "Internal logic Error"?

crates/fluvio/src/producer/mod.rs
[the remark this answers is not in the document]
nicholastmosher: Partitioning does already have a trait that allows for it to have alternate implementations, it is just not public yet. The only thing hardcoded here is which Partitioner implementation is used (the SiphashRoundRobinPartitioner). I would say this is not part of the scope for this PR, since this is how partitioning worked before.

As far as representing the "batching strategy" as a trait, I think that could potentially be much more effort, and I think it would be best to defer that until we actually need multiple batching strategies and we know what our needs are, rather than trying to include it here where we would be guessing our future needs.

crates/fluvio/src/producer/dispatcher.rs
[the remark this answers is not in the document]
nicholastmosher: I just double checked to be sure, but the only thing that could cause this is if `channel.send()` returned an `Err` internally, which I am not expecting to happen but wanted to print out if it happened anyway. I think this should really only happen if all status receivers are dropped. Our other option would be to silently ignore it, but I think it's a better idea to have a printout so we can at least see if it starts happening.

crates/fluvio/src/producer/dispatcher.rs
nicholastmosher: I think that is something we can certainly do for a v2, but if possible I would really like to get this merged since it is tested and working. I think the structure of the code is also good enough that we can update this in the future when we need to.

crates/fluvio/src/producer/dispatcher.rs
nicholastmosher: Can we defer this? In practice the code already behaves in these states, I don't think that adding something like a state enum would add much value at this point in time.

sehz:
Can we separate out sticky event related code first?

nicholastmosher:
Yes, let me try to remove that

crates/fluvio/src/producer/dispatcher.rs
```diff
@@ -0,0 +1,305 @@
...
+        for (_, partition_producer) in self.pool.iter_mut() {
+            let fut = partition_producer.flush();
+            flush_futures.push(fut);
+        }
+
+        // Flush all concurrently
+        let results = futures_util::future::join_all(flush_futures).await;
+        for result in results {
+            if let Err(e) = result {
+                error!("Internal error broadcasting batch status: {}", e);
```
sehz, line 180: so error is not propagating?

sehz (changes requested):

crates/fluvio/src/producer/dispatcher.rs
```diff
@@ -0,0 +1,305 @@
...
+    }
+
+    /// Buffer the given record, flushing if necessary.
+    pub async fn buffer(
+        &mut self,
+        record: AssociatedRecord,
+        respond_to_caller: Sender<ProducerMsg>,
+    ) -> Result<(), FluvioError> {
+        // If this record is too large for the buffer, return it to the producer.
+        if !self.buffer.could_fit(&record) {
```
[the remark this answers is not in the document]
nicholastmosher: When accepting a record to put into the buffer, if the record would overflow the buffer's `max_size`, it first flushes the buffer then puts the record in.

crates/fluvio/src/producer/dispatcher.rs
nicholastmosher: The `PartitionProducer::flush` call will send any batch-request related errors through the `broadcast_status` channel which then ends up returning to the caller. The `Result` returned by flush itself can only ever be `Err` if there is a problem with the channel itself, which is an internal error. Our options for dealing with that kind of error are:
- Panic,
- Print the error (I'm using this option, with `error!`)
- Do nothing

sehz:
obsolete

PR #1651, opened 2021-09-17
[Merged by Bors] - Generate GH Release notes from changelog

### LINKED ISSUE #1633: Changelog automation for GH Release
We might be waiting for Github's feature re: changelogs in GH Release https://github.com/github/roadmap/issues/248

However, it might be worth reiterating the goals since we've tried to do this several times.

The minimum we need is something that will read from our `CHANGELOG.md` and only use the relevant block to the version number we are releasing.

This typically is the top release. And by convention, in the file it should still be marked as `UNRELEASED` as the data of release. We want to replace that with the current date in the form `YYYY-MM-DD`.

---

Our current release automation uses the whole `CHANGELOG.md` for the new release, and we need to manually edit it to remove the previous blocks. Not difficult, but still manual.

tjtelan:
Automate manual step from release

Take the section that matches the current VERSION and use for release notes

Resolves #1633


sehz (approved):
good to have this automated!

PR #1652, opened 2021-09-17
Move Arc inside StickyEvent, makes Future easier to use

nicholastmosher:
This is a little bit of cleanup for `StickyEvent` that makes it much easier to use with Futures combinators.

Key changes:

- We now use `StickyEvent` instead of `Arc<StickyEvent>` everywhere, the Arc is now inside and StickyEvent is Clone
- The `listen` method now returns a Future bounded by `'static`, rather than one that is bounded by the `&self` lifetime
  - This is the key improvement that helps work with combinators

### DIFF
--- a/crates/fluvio-types/src/event.rs
+++ b/crates/fluvio-types/src/event.rs
@@ -268,6 +268,34 @@ mod test {
     }
 
     #[fluvio_future::test]
+    async fn test_take_until() {
+        use std::pin::Pin;
+        use futures_util::{stream, Stream, StreamExt};
+        use super::StickyEvent;
+
+        let end: StickyEvent = StickyEvent::new();
+        let stream = stream::repeat(9);
+        let _until = stream.take_until(end.listen());
+
+        let stream2 = stream::repeat(9);
+        let _until2 = stream2.take_until(end.listen_pinned());
+
+        // In our real use-case, we need our stream to be `impl Stream + Send + Sync + 'static`
+        fn assert_stream<S: Stream + Send + Sync + 'static>(stream: S) -> S {
+            stream
+        }
+        // The stream needs to be able to be moved, as it will be stored in a struct
+        fn move_it<S>(it: S) -> S {
+            it
+        }
+        let stream3 = stream::repeat(9);
+        let until3 = stream3.take_until(end.listen_pinned());
+        let until3 = assert_stream(until3);
+        let pinned: Option<Pin<Box<dyn Stream<Item = i32> + Send + Sync>>> = Some(Box::pin(until3));
+        let _moved = move_it(pinned);
+    }
+
+    #[fluvio_future::test]
     async fn test_offset_listener_no_wait() {
         let publisher = OffsetPublisher::shared(0);
         let listener = publisher.change_listner();
--- a/crates/fluvio-types/src/event.rs
+++ b/crates/fluvio-types/src/event.rs
@@ -280,19 +280,17 @@ mod test {
         let stream2 = stream::repeat(9);
         let _until2 = stream2.take_until(end.listen_pinned());
 
-        // In our real use-case, we need our stream to be `impl Stream + Send + Sync + 'static`
-        fn assert_stream<S: Stream + Send + Sync + 'static>(stream: S) -> S {
-            stream
-        }
-        // The stream needs to be able to be moved, as it will be stored in a struct
-        fn move_it<S>(it: S) -> S {
-            it
-        }
         let stream3 = stream::repeat(9);
         let until3 = stream3.take_until(end.listen_pinned());
-        let until3 = assert_stream(until3);
-        let pinned: Option<Pin<Box<dyn Stream<Item = i32> + Send + Sync>>> = Some(Box::pin(until3));
-        let _moved = move_it(pinned);
+        let pinned = Box::pin(until3);
+
+        struct DispatcherStateOrSomething {
+            _stream: Option<Pin<Box<dyn Stream<Item = i32>>>>,
+        }
+
+        let _stored = DispatcherStateOrSomething {
+            _stream: Some(pinned),
+        };
     }
 
     #[fluvio_future::test]
--- a/crates/fluvio-types/src/event.rs
+++ b/crates/fluvio-types/src/event.rs
@@ -1,3 +1,4 @@
+use std::fmt;
 use std::sync::atomic::{Ordering, AtomicBool};
 use std::sync::Arc;
 
@@ -10,9 +11,7 @@ const DEFAULT_EVENT_ORDERING: Ordering = Ordering::SeqCst;
 pub use StickyEvent as SimpleEvent;
 
 #[derive(Clone)]
-pub struct StickyEvent {
-    inner: Arc<StickyInner>,
-}
+pub struct StickyEvent(Arc<StickyInner>);
 
 struct StickyInner {
     flag: AtomicBool,
@@ -32,21 +31,19 @@ impl StickyEvent {
     }
 
     pub fn new() -> Self {
-        Self {
-            inner: Arc::new(StickyInner {
-                flag: AtomicBool::new(false),
-                event: Event::new(),
-            }),
-        }
+        Self(Arc::new(StickyInner {
+            flag: AtomicBool::new(false),
+            event: Event::new(),
+        }))
     }
 
     // is flag set
     pub fn is_set(&self) -> bool {
-        self.inner.is_set()
+        self.0.is_set()
     }
 
     pub fn listen(&self) -> impl std::future::Future<Output = ()> {
-        let inner = self.inner.clone();
+        let inner = self.0.clone();
 
         async move {
             if inner.is_set() {
@@ -70,14 +67,22 @@ impl StickyEvent {
     }
 
     pub fn notify(&self) {
-        self.inner.flag.store(true, DEFAULT_EVENT_ORDERING);
-        self.inner.event.notify(usize::MAX);
+        self.0.flag.store(true, DEFAULT_EVENT_ORDERING);
+        self.0.event.notify(usize::MAX);
     }
 }
 
 impl Default for StickyEvent {
     fn default() -> Self {
         Self::new()
+    }
+}
+
+impl fmt::Debug for StickyEvent {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_tuple("StickyEvent")
+            .field(&self.0.is_set())
+            .finish()
     }
 }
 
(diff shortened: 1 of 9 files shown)


crates/fluvio-types/src/event.rs
```diff
@@ -9,47 +9,69 @@ const DEFAULT_EVENT_ORDERING: Ordering = Ordering::SeqCst;
 #[deprecated(since = "0.2.5", note = "use StickyEvent instead")]
 pub use StickyEvent as SimpleEvent;
 
+#[derive(Clone)]
 pub struct StickyEvent {
+    inner: Arc<StickyInner>,
+}
```
sehz, line 15: Use new type pattern: 
```
pub struct SticktEvent(Arc<StickyInner>);
```

crates/fluvio-types/src/event.rs
```diff
@@ -9,47 +9,69 @@ const DEFAULT_EVENT_ORDERING: Ordering = Ordering::SeqCst;
...
+        Self {
+            inner: Arc::new(StickyInner {
+                flag: AtomicBool::new(false),
+                event: Event::new(),
+            }),
+        }
     }
 
     // is flag set
     pub fn is_set(&self) -> bool {
```
sehz, line 44: use `Deref` trait rather than duplicating method

crates/fluvio-types/src/event.rs
```diff
@@ -9,47 +9,69 @@ const DEFAULT_EVENT_ORDERING: Ordering = Ordering::SeqCst;
...
-        self.flag.load(DEFAULT_EVENT_ORDERING)
+        self.inner.is_set()
     }
 
-    pub async fn listen(&self) {
-        if self.is_set() {
-            trace!("before, flag is set");
-            return;
-        }
+    pub fn listen(&self) -> impl std::future::Future<Output = ()> {
```
sehz, line 48: use `Deref` in case, this can be move deleted to inner

crates/fluvio-types/src/event.rs
```diff
@@ -9,47 +9,69 @@ const DEFAULT_EVENT_ORDERING: Ordering = Ordering::SeqCst;
...
 
-    pub fn listen_pinned(&self) -> impl std::future::Future<Output = ()> + '_ {
+    pub fn listen_pinned(&self) -> impl std::future::Future<Output = ()> {
         Box::pin(self.listen())
     }
 
     pub fn notify(&self) {
-        self.flag.store(true, DEFAULT_EVENT_ORDERING);
-        self.event.notify(usize::MAX);
+        self.inner.flag.store(true, DEFAULT_EVENT_ORDERING);
```
sehz, line 73: move this to inner

crates/fluvio-types/src/event.rs
```diff
@@ -9,47 +9,69 @@ const DEFAULT_EVENT_ORDERING: Ordering = Ordering::SeqCst;
...
+                return;
+            }
 
-        listener.await
+            listener.await
+        }
     }
 
-    pub fn listen_pinned(&self) -> impl std::future::Future<Output = ()> + '_ {
+    pub fn listen_pinned(&self) -> impl std::future::Future<Output = ()> {
```
sehz, line 68: move this to inner

sehz (changes requested):

crates/fluvio-types/src/event.rs
nicholastmosher: Alright, I just tried this out and I'm getting the following error:

```
error[E0446]: private type `StickyInner` in public interface
  --> crates/fluvio-types/src/event.rs:75:5
   |
18 | struct StickyInner {
   | ------------------ `StickyInner` declared as private
...
75 |     type Target = StickyInner;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^ can't leak private type
```

I don't really want to expose the `StickyInner` type, can we leave it as-is?

crates/fluvio-types/src/event.rs
nicholastmosher: This actually needs to be implemented on the outer type. If you look in the body here, we create an `async move` block and pass an owned `Arc<StickyInner>` into it, that way the Future has the `'static` bound. We wouldn't be able to do this if we put this impl on `StickyInner`.

crates/fluvio-types/src/event.rs
nicholastmosher: See my comment about Deref, if we can't do deref we shouldn't do this I think

crates/fluvio-types/src/event.rs
sehz: What is we want to achieve here?  not clear how this improve from just simple aliasing.

nicholastmosher:
@sehz Let me back up and show the use-case I need `StickyEvent` for and why it doesn't work as-is.

Basically, I want to be able to use it to stop a stream of events after `shutdown.listen()` triggers. Like this:

```rust
let end: Arc<StickyEvent> = shutdown.clone();
let until = select.take_until(end.listen_pinned());
```

The `take_until` combinator just takes a Future, and when that Future comes back as `Ready`, the stream stops yielding elements.

When I use the existing `StickyEvent`, I get this error:

```
error[E0597]: `end` does not live long enough
  --> crates/fluvio/src/producer/dispatcher.rs:57:39
   |
57 |         let until = select.take_until(end.listen_pinned());
   |                                       ^^^ borrowed value does not live long enough
58 |         let events: Option<Pin<Box<dyn Stream<Item = Event> + Send + Sync>>> =
59 |             Some(Box::pin(until));
   |                  --------------- cast requires that `end` is borrowed for `'static`
...
68 |     }
   |     - `end` dropped here while still borrowed
```

If we look closely at `listen_pinned`, we can see why this happens:

```rust
pub fn listen_pinned(&self) -> impl std::future::Future<Output = ()> + '_ { .. }

// Or, another way to see it:

pub fn listen_pinned<'a>(&'a self) -> impl std::future::Future<Output = ()> + 'a { .. }
```

In order to make this work, what we actually want is this:

```rust
pub fn listen_pinned(&self) -> impl std::future::Future<Output = ()> { .. }

// A.k.a:

pub fn listen_pinned(&self) -> impl std::future::Future<Output = ()> + 'static { .. }
```

The way this PR gets us there is by moving the `Arc` inside the StickyEvent, which lets us clone `Arc<StickyInner>` into the future so it is not bounded by the `&self` lifetime.

sehz:
> This is a little bit of cleanup for `StickyEvent` that makes it much easier to use with Futures combinators.
> 
> 
> 
> Key changes:
> 
> 
> 
> - We now use `StickyEvent` instead of `Arc<StickyEvent>` everywhere, the Arc is now inside and StickyEvent is Clone
> 
> - The `listen` method now returns a Future bounded by `'static`, rather than one that is bounded by the `&self` lifetime
> 
>   - This is the key improvement that helps work with combinators

Let me check

sehz:
I was able to combine stick event with take_until: https://github.com/infinyon/fluvio/pull/1655.  For both `listen`.  What am I missing?  Can you show why this doesn't work?

nicholastmosher:
@sehz I saw your PR with unit tests for StickyEvent, I [added some conditions](https://github.com/sehz/fluvio/pull/1) to it that reproduce the constraints that I'm working with. The test now fails due to `StickyEvent` not living long enough, which is the same error I was seeing before, and the errors this this PR ☝️ fixes

nicholastmosher:
@sehz I added your unit tests here and included my addition with constraints that only pass in this PR and not before:

https://github.com/infinyon/fluvio/pull/1652/files#diff-8a094ddc9e41a4112d8b94963fc33eeb8a8bd66793fe6a1ddc5f4a2de09b9d58R284-R295

crates/fluvio-types/src/event.rs
```diff
@@ -234,6 +267,34 @@ mod test {
...
+
+        let end: StickyEvent = StickyEvent::new();
+        let stream = stream::repeat(9);
+        let _until = stream.take_until(end.listen());
+
+        let stream2 = stream::repeat(9);
+        let _until2 = stream2.take_until(end.listen_pinned());
+
+        // In our real use-case, we need our stream to be `impl Stream + Send + Sync + 'static`
+        fn assert_stream<S: Stream + Send + Sync + 'static>(stream: S) -> S {
```
sehz, line 284: why stream need to be static?

nicholastmosher: Because in the place I'm using the stream, I want to move it into a struct to store it. The `'static` bound is needed in order to make that happen.

sehz: Typical stream doesn't need to have static bound (for example, in the futures crate).  Can you show example?

nicholastmosher: Yes. In the producer batching implementation, this code does not compile unless these changes to StickyEvent are present.

https://github.com/infinyon/fluvio/pull/1631/files#diff-c0b5d4df8eae78fbd44bbef0db4a5e88c024ae666dd804f5c075fd212843e6dcR50-R65

Without these changes, using `shutdown.listen_pinned()` causes an error such as "shutdown does not live long enough". This is because the future it returns is tied to the lifetime of the `shutdown` borrow.

```rust
impl StickyEvent {
    // Old, does not work
    fn listen_pinned<'a>(&'a self) -> impl std::future::Future<Output = ()> + 'a { ... }

    // New, works
    fn listen_pinned(&self) -> impl std::future::Future<Output = ()> + 'static { ... }
}
```

The changes in this PR make this second version of `listen_pinned` possible, because the Future we return holds onto an `Arc<StickyInner>`, which is fully owned and has no outstanding references.

sehz: Can you provide isolated test case? It's hard to understand why this is happening.

nicholastmosher:
@sehz I have clarified the test case.

Here is the test case showing the error before this PR is applied: https://github.com/sehz/fluvio/pull/1/files

Here is the test case showing the code working as expected in this PR: https://github.com/infinyon/fluvio/pull/1652/files#diff-8a094ddc9e41a4112d8b94963fc33eeb8a8bd66793fe6a1ddc5f4a2de09b9d58R283-R293

crates/fluvio-types/src/event.rs
nicholastmosher: See my comment below 👇

sehz: see updated: https://github.com/infinyon/fluvio/pull/1655/files.

Issue with your examples are:
* Boxing,  There is no need to box if you use generics
* Take look at how other streams are constructed: https://github.com/rust-lang/futures-rs/blob/master/futures-util/src/stream/stream/take_until.rs.

Other issue is that why is stream is need to put in the struct? This is only really needed if you creating stream combinator.

nicholastmosher:
This was originally needed for batching producer, but we managed to do without it. I'm going to close it and if we need it later we can revisit.

PR #1653, opened 2021-09-17
[Merged by Bors] - Lock-free partitioner for producer

nicholastmosher:
This is related to the producer batching. I've moved the partitioning code to it's own module and made it lock-free. Here are the key changes:

- The `Partitioner::partition` method now takes `&self` instead of `&mut self`
  - This was done by using an `AtomicI32` instead of a regular i32
- This eliminates the need to use `Mutex<dyn Partitioner>` so now we just use `Box<dyn Partitioner>`

### DIFF
--- a/crates/fluvio/src/producer/partitioning.rs
+++ b/crates/fluvio/src/producer/partitioning.rs
@@ -1,0 +1,99 @@
+use std::sync::atomic::{AtomicI32, Ordering};
+use siphasher::sip::SipHasher;
+use fluvio_types::PartitionId;
+
+/// A trait for defining a partitioning strategy for key/value records.
+///
+/// A Partitioner is given a slice of potential keys, and the number of
+/// partitions in the current Topic. It must map each key from the input
+/// slice into a partition stored at the same index in the output Vec.
+///
+/// It is up to the implementor to decide how the keys get mapped to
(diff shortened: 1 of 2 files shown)


crates/fluvio/src/producer/partitioning.rs
```diff
@@ -0,0 +1,99 @@
...
+        assert_eq!(key2_partition, 1);
+        let key3_partition = partitioner.partition(&config, None, &[]);
+        assert_eq!(key3_partition, 2);
+        let key4_partition = partitioner.partition(&config, None, &[]);
+        assert_eq!(key4_partition, 0);
+        let key5_partition = partitioner.partition(&config, None, &[]);
+        assert_eq!(key5_partition, 1);
+        let key6_partition = partitioner.partition(&config, None, &[]);
+        assert_eq!(key6_partition, 2);
+    }
```
sehz, line 98: need test for concurrency since this use atomics

sehz (changes requested):

crates/fluvio/src/producer/partitioning.rs
nicholastmosher: Added test 👍

sehz (approved):

PR #1654, opened 2021-09-17
[Merged by Bors] - Add spinner in local installer longer steps

morenol:
This adds a spinner for the SC and SPU related steps of the installation process

### DIFF
--- a/crates/fluvio-cluster/src/start/local.rs
+++ b/crates/fluvio-cluster/src/start/local.rs
@@ -6,7 +6,6 @@ use std::time::Duration;
 use fluvio::{Fluvio, FluvioConfig};
 use indicatif::ProgressBar;
 use semver::Version;
-
 use derive_builder::Builder;
 use tracing::{debug, error, info, instrument, warn};
 use once_cell::sync::Lazy;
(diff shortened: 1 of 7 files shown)


crates/fluvio-cluster/src/start/local.rs
```diff
@@ -658,18 +602,20 @@ impl LocalInstaller {
...
+        self.render(InstallProgressMessage::ProfileSet, None);
 
         Ok(format!("local context is set to: {}", local_addr))
     }
 
     #[instrument(skip(self))]
     async fn launch_spu_group(&self, client: SharedK8Client) -> Result<(), LocalInstallError> {
         let count = self.config.spu_replicas;
-        self.render(InstallProgressMessage::LaunchingSPUGroup(count));
+
```
sehz, line 613: Instead of having everyone managing progress bar.  just a create a global progress bar at beginning since everything is a single sequential.    then anytime, we want to make show status, it should be just matter of:
```
self.progress.println("creating management group");    // this will print new line 'creating management group' and show spinner.
self.progress.println("set up profile");   // this will print new line and show spinner
self.progress.println_timer("Setting up SPU {I}",i);   // this will print spinner but display count time
```

sehz (changes requested):

sehz (approved):
Excellent!

Issue #1215, opened 2021-07-08
Add `fluvio diagnostics` for collecting useful debugging information


nicholastmosher:
When debugging a problem with Fluvio, it is useful to have a bunch of information such as:

- Logs
- Tooling info (e.g. `kubectl version`, `helm version`, etc.)
- Host info (uname)
- Profile type
- etc.

Some of this information is given by `fluvio version`, but it'd be nice to have an all-in-one command that produces a `.zip` file that we can collect when something goes wrong, to make it easier to report bugs.

sehz:
Let's roll into `cluster check` command.  We can add different flags `--platform` ``-network` for different check

nicholastmosher:
I think this is still worth having, going to remove the stale label

PR #1671, opened 2021-09-21
[Merged by Bors] - Add 'fluvio diagnostics' for collecting debug information

### LINKED ISSUE #1667: Cluster support logs
We need one step method to gather logs for cluster for diagnose.  It should gather everything in a single zip file.  At minimum:
* logs for SC and SPU
* CRD (Topics, partitions, etc)
* K8 objects, Services, ConfigMap,
* Pod Info
See `save_logs` step in: https://github.com/infinyon/fluvio/blob/master/.github/workflows/ci.yml

We should borrow from what we have in the Github action when failure occurs 

It should be something like:
```
$ fluvio cluster logs
... creating logs <spinner>
save to `fluvio_cluster_0.9.8.tar.gz`
```

nicholastmosher:
Closes #1667 and #1215

Adds `fluvio diagnostics` subcommand which will produce a zip file, `diagnostics-YYYY-MM-DD-hh-mm-ss.zip`. This file contains:

- Logs for any SC or SPU, local or kubernetes
- Kubernetes metadata, including all Fluvio pods/services/statefulsets, and CRD's spu/spg/partition/topic
- Fluvio admin metadata, topic/partition/spu/spg

### DIFF
--- a/crates/fluvio-cluster/src/cli/diagnostics.rs
+++ b/crates/fluvio-cluster/src/cli/diagnostics.rs
@@ -98,7 +98,7 @@ impl DiagnosticsOpt {
         .read()?;
         // Filter for only Fluvio pods
         let pods = pods
-            .split(" ")
+            .split(' ')
             .filter(|pod| pod.contains("fluvio"))
             .collect::<Vec<_>>();
 
@@ -136,7 +136,7 @@ impl DiagnosticsOpt {
         .read()?;
         // Filter for only Fluvio services
         let objects = objects
-            .split(" ")
+            .split(' ')
             .filter(|obj| !filter_fluvio || obj.contains("fluvio"))
             .map(|name| name.trim())
             .collect::<Vec<_>>();
--- a/crates/fluvio-cluster/src/cli/diagnostics.rs
+++ b/crates/fluvio-cluster/src/cli/diagnostics.rs
@@ -51,7 +51,7 @@ impl DiagnosticsOpt {
         let _ = self.copy_fluvio_specs(temp_path).await;
 
         let time = chrono::Local::now().format("%Y-%m-%d-%H-%M-%S").to_string();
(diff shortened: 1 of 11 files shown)


sehz:
I noticed that it is using zip format.  This should be in `.gz` format as they are unix standard

crates/fluvio-cluster/src/cli/diagnostics.rs
```diff
@@ -0,0 +1,208 @@
...
+use structopt::StructOpt;
+use fluvio::config::ConfigFile;
+use duct::cmd;
+use which::which;
+use crate::cli::ClusterCliError;
+
+type Result<T, E = ClusterCliError> = core::result::Result<T, E>;
+
+#[cfg(target_os = "macos")]
+const LOGS_PATH: &str = "/usr/local/var/log/fluvio/";
```
sehz, line 11: should be using same path from cluster config not duplicating it

crates/fluvio-cluster/src/cli/error.rs
```diff
@@ -24,6 +24,9 @@ pub enum ClusterCliError {
     /// An error occurred while communicating with Fluvio
     #[error("Fluvio client error")]
     ClientError(#[from] FluvioError),
+    /// An error occurred finding an executable
+    #[error("error finding executable")]
+    Which(#[from] which::Error),
```
sehz, line 29: This goes into diagnostics error or turn into io error

sehz (changes requested):
* add to change log.
* need to add k8 PVC objects

nicholastmosher:
@sehz this should be ready to go. Now uses `.tar.gz` instead of `.zip`, and extracts into a `./diagnostics/` directory.

sehz (approved):
Excellent!  this will make it easier to collect debugging info fluvio

sehz:
clippy failures

PR #1675, opened 2021-09-22
[Merged by Bors] - Change chart location for local installer

morenol:
Fixes https://github.com/infinyon/fluvio/issues/1647

Running `helm install fluvio-sys ./k8-util/helm/` fails with 
```
Error: validation: chart.metadata is required
```
`helm install fluvio-sys ./k8-util/helm/fluvio-sys` works

### DIFF
--- a/crates/fluvio-cluster/src/cli/start/local.rs
+++ b/crates/fluvio-cluster/src/cli/start/local.rs
@@ -23,14 +23,8 @@ pub async fn process_local(
         .render_checks(true)
         .spu_replicas(opt.spu);
 
-    match opt.k8_config.chart_location {
-        Some(chart_location) => {
-            builder.local_chart(chart_location);
-        }
-        None if opt.develop => {
-            builder.local_chart("./k8-util/helm/fluvio-sys");
-        }
-        _ => (),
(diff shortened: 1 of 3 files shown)


crates/fluvio-cluster/src/cli/start/local.rs
```diff
@@ -28,7 +28,7 @@ pub async fn process_local(
             builder.local_chart(chart_location);
         }
         None if opt.develop => {
-            builder.local_chart("./k8-util/helm");
+            builder.local_chart("./k8-util/helm/fluvio-sys");
```
sehz, line 31: Hmm.  This is no longer correct. Let's just remove this option

sehz (changes requested):

crates/fluvio-cluster/src/cli/start/local.rs
morenol: Do you mean not having a "default" value for local_chart when running --local in --develop?

sehz: Yes. leave it optional

sehz (approved):

PR #1684, opened 2021-09-24
[Merged by Bors] - Compare crates in repo w/ published crates

### LINKED ISSUE #1496: Add job to validate crate version
We need some test to check if a crate's code has been changed, and that the version needs to be increased and published.

Or another way to describe:
A check that compares a published crate with the crate on disk.

* If the crate on crates.io matches the crate on disk, then the check should pass
* If they are different because the crate on disk has changed, then the check should fail.

Ideally, this test would run in `CI` if we can implement w/o building, but if we end up needing to build release artifacts for the test, then it might belong in `CD_Dev`

tjtelan:
Catch if we forget to version bump a crate

Fixes #1496

Not yet handling changes only to a crate's `Cargo.toml` due to crates.io re-formatting the original Cargo.toml. This is an edge case.

This test probably belongs in `CD_Dev` but keeping it in `CI` for now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -796,3 +796,25 @@ jobs:
     steps:
       - name: Done
         run: echo "Done!"
+
+  test_crate_versions:
+    name: Verify if crate versions need bump
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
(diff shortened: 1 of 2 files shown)


sehz (approved):
Nice job!

PR #1693, opened 2021-09-27
[Merged by Bors] - Impl std::error::Error for ErrorCode

### LINKED ISSUE #786: Update ErrorCode to follow Rust idiom
Currently ErrorCode is modeled based on Kafka protocol with simple number enums.  We should try to refactor into idiomatic Rust Enum

nicholastmosher:
Closes #786

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Release Notes
 
 ## Platform Version 0.9.9 - UNRELEASED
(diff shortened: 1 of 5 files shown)


sehz (approved):

tjtelan:
I think this change requires at least a patch bump to version for `fluvio-dataplane-protocol`

sehz:
use this as testing version check?

nicholastmosher:
> use this as testing version check?

What do you mean by this? Is that something we'd do before or after merging this?

Issue #1568, opened 2021-09-03
Feature:  Consuming multiple partitions


sehz:
Currently Consumer are able to read from a single partition.  We need ability to read from multiple partition concurrently.

* [Kafka](http://kafka.apache.org/21/javadoc/index.html?org/apache/kafka/clients/consumer/KafkaConsumer.html)
* [Pulsar](https://pulsar.apache.org/docs/v1.20.0-incubating/getting-started/ConceptsAndArchitecture)

By default, consume should retrieve records from all partitions in a topic.  But it also should support:
* list of partitions
* partitions matched by some wildcard.

Records retrieve among multiple partitions are not ordered

PR #1699, opened 2021-09-28
[Merged by Bors] - feature: Implement multiple partitions consumer

morenol:
Closes https://github.com/infinyon/fluvio/issues/1568

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,6 +109,16 @@ election-test: test-setup
 		${TEST_ARG_DEVELOP} \
 		${TEST_ARG_EXTRA}
 
+multiple-partition-test: TEST_ARG_EXTRA=--local $(EXTRA_ARG)
+multiple-partition-test: test-setup
+	$(TEST_BIN) multiple_partition --partition 10 \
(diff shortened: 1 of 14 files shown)


Makefile
```diff
@@ -109,6 +109,16 @@ election-test: test-setup
 		${TEST_ARG_DEVELOP} \
 		${TEST_ARG_EXTRA}
 
+multiple-partition-test: TEST_ARG_EXTRA=--local $(EXTRA_ARG)
```
sehz, line 112: is there reason why this only runs on local? this can be removed

sehz (changes requested):
Nicely done.!   A minor nit.   Also, add to changelog

sehz:
CI is failling.  Look at other test and make sure right args are passed

morenol:
> CI is failling. Look at other test and make sure right args are passed

The difference that I see is that the other tests are using `--local`. I added again this line:
```
 multiple-partition-test: TEST_ARG_EXTRA=--local $(EXTRA_ARG)
```

sehz (approved):

PR #1702, opened 2021-09-29
[Merged by Bors] - feat: Expose partition in fluvio consume --format command

### LINKED ISSUE #1701: Expose partition for `fluvio consume --format`
With #1568 complete, we have the ability to consume from all partitions at the same time.

We need to expose to the client the partition that a record belongs to. 

Such that we can use a command like: `fluvio consume -A -B --format "{{partition}} {{value}}`

The expected output would print the partition of the record value in column 1 and the value of the record starting at column 2.

morenol:
Closes #1701


sehz (approved):
LGTM

Issue #791, opened 2021-02-17
handle producer error correctly


sehz:
**What happened**
Error is not correctly propagated from SPU to client:
```
$ fluvio produce t1 -r $(ls /var/log/journal/**/*.journal)
Error: 
   0: Consumer Error
   1: Fluvio client error
   2: Fluvio socket error
   3: time out in serial: 0 request: 1
```

**Expected behavior**
record size exceed maximum message length of <xxx>

**Describe the setup**
* Set up any fluvio cluster
* Create topic t1: `flvd topic create t1`

**Environment (please complete the following information):**
 - OS: [Linux]
 - Fluvio Version [master]

PR #1709, opened 2021-10-01
[Merged by Bors] - Catch socket timeout and print nicely

nicholastmosher:
Closes https://github.com/infinyon/fluvio/issues/791.

Now when the client encounters a socket timeout, they get the following printed:

```
$ fluvio produce topic
...
...
Timed out while waiting on socket response
```

### DIFF
--- a/crates/fluvio-cli/src/error.rs
+++ b/crates/fluvio-cli/src/error.rs
@@ -112,6 +112,12 @@ impl CliError {
                 }
                 _ => Err(self),
             },
+            Self::ClientError(FluvioError::Socket(SocketError::Io(io)))
+                if io.kind() == ErrorKind::TimedOut =>
+            {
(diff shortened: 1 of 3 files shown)


crates/fluvio-cli/src/error.rs
```diff
@@ -112,6 +112,12 @@ impl CliError {
                 }
                 _ => Err(self),
             },
+            Self::ClientError(FluvioError::Socket(SocketError::Io(io)))
+                if io.kind() == ErrorKind::TimedOut =>
+            {
+                println!("Timed out while waiting on socket response");
```
sehz, line 118: This is bit technical.  Should be something like: "Network connection timed out while waiting response"

sehz (changes requested):

sehz (approved):

PR #1712, opened 2021-10-01
Fluvio test process

### LINKED ISSUE #627: fluvio-test - Spawn producers and consumers into processes
In the smoke test, consumers and producers created using separate connection.  If they are created from same Fluvio, it fails.

tjtelan:
Fixes #627 

This is only the longevity test that runs a producer and consumer in their own process. However, significant change was needed to decouple the test driver from the producer/consumer.

Our test strategy changed to defer using `async` as late as possible. The goal being that the first occurrence of `async` code will be in the test code itself. This enables us to avoid a problem I was seeing where our async code would block on `.await` in our child processes after process forking.


crates/fluvio-test/src/tests/election/mod.rs
```diff
@@ -52,137 +52,141 @@ impl TestOption for ElectionTestOption {
...
-    // this is hack now, because we don't have ack
-    sleep(Duration::from_secs(ACK_WAIT)).await;
-
-    let admin = test_driver.client().admin().await;
-
-    let partitions = admin
-        .list::<PartitionSpec, _>(vec![])
-        .await
-        .expect("partitions");
+    run_block_on(async {
```
sehz, line 58: Don't see why we couldn't keep existing signature.  Instead of requiring `run_block_on`, inject that from `fluvio_test` procedure macro so it will be pretty much same as before.  In fact: `test_driver
             .connect().await` can be part of generated code

crates/fluvio-test/src/tests/smoke/mod.rs
```diff
@@ -68,14 +77,90 @@ impl TestOption for SmokeTestOption {
 //}
 
 #[fluvio_test(topic = "test")]
-pub async fn smoke(mut test_driver: Arc<FluvioTestDriver>, mut test_case: TestCase) -> TestResult {
+pub fn smoke(mut test_driver: FluvioTestDriver, mut test_case: TestCase) -> TestResult {
```
sehz, line 80: So I think we either have two procedure macro or one procedure macro with different attribute.  If for example, attribute as in `#[fluvio_test(topic = 'test', async)]` this will inject async (in the election test), otherwise it will generate raw code

crates/fluvio-test/src/tests/smoke/mod.rs
```diff
@@ -68,14 +77,90 @@ impl TestOption for SmokeTestOption {
...
 
-    let start_offsets = produce::produce_message(test_driver.clone(), &smoke_test_case).await;
-    // println!("start sleeping");
-    // fluvio_future::timer::sleep(Duration::from_secs(40)).await;
-    // sleep(Duration::from_secs(40));
-    // println!("end sleeping");
-    consume::validate_consume_message(test_driver.clone(), &smoke_test_case, start_offsets).await;
+    // We're going to handle the `--consumer-wait` flag in this process
+    let producer_process = match fork() {
+        Ok(Fork::Parent(child_pid)) => child_pid,
```
sehz, line 86: this is not show stopper, but this could use regular macro to automate fork like this:
```
#[macro_export]
macro_rules! test_fork! {
    ($child:expr) => {{
         match fork() {
              Ok(Fork::Parent(child_pid)) => child_pid,
              Ok(Fork::Child) => {
                run_block_on($child)
              }
...
        }
    }};
}

```

So this can be written as:
```
  test_fork!(async {
            println!("Producer about to connect");
           let mut test_driver_consumer_wait = test_driver.clone();
          ...

 }
```

sehz (changes requested):
I think this is good start.  No really show stopper but couple of suggestions

crates/fluvio-test/src/tests/election/mod.rs
tjtelan: Both the smoke and longevity tests (and any future test where we're going to spawn processes) need to defer their async code until after fork(). I found that we can't async any earlier, just based on my small experimenting: https://github.com/infinyon/fluvio/issues/627#issuecomment-931547608

So that's one reason we wouldn't want to `test_driver.connect().await` in the generated code.

crates/fluvio-test/src/tests/smoke/mod.rs
tjtelan: I can play around with this pattern. The only part I don't want to be overconfident about is if it would be simple enough to join on the pid. Seems like it should be straightforward, and would make this pattern much easier to use.

sehz (approved):
Nice job!  This looks very clean with min changes.

PR #1716, opened 2021-10-02
[Merged by Bors] - fluvio-test - Reverse cluster start/stop cli logic

### LINKED ISSUE #1644: fluvio-test - Change cluster start/delete flags to use opt-in logic
After a few months of regularly using `fluvio-test`, I realized that I don't use the cluster start/stop features.

`fluvio-test -d -k <test-name>`  -- Typically, I'll always use this with the `-d` and `-k` flags.

---

Instead of:

* `--disable-install` or `-d`
  * Preventing the deployment of a cluster before a test.

And:
* `--keep-cluster` or `-k`
  * Preventing the deletion of a cluster after a test

We should have the opposite logic be the default.
* `--cluster-start`
  * Attempt to create a cluster at the start of a test. (Important detail: Don't fail if a cluster already exists. Be idempotent)
* `--cluster-delete`
  * Delete the cluster at the end of the test

---

We need short versions of these flags too. I want to use `-s` for start and `-d` for delete. 

However `-s` is the short flag for `--spu`. It is likely that I'm going to take this back. I feel like this won't be confusing as `fluvio cluster start` does not use `-s` as a short flag for `--spu`.

tjtelan:
Resolves #1644

By default, we should use whatever cluster is available, instead of attempting to cleanup and start new clusters. We opt-in to cluster management instead of opt-out.

---

Removing the following fluvio-test options:

`--disable-install` + `-d`
Which was used to opt-out of deleting a cluster, then restarting a new cluster.

`--keep-cluster` + `-k`
Which was used to opt-out of deleting your test cluster at the end of the test

and the short flag `-s` used for `--spu`

---

And adding the following options:

`--cluster-start` + `-s` (Repurposed short flag)
Which will try to start a cluster before the test starts (no harm if cluster is already started)

`--cluster-start-fresh`
Which will try to delete a cluster then re-start a new one before a test starts

`--cluster-delete` + `-d` (Repurposed short flag)
Which will try to delete a cluster at the end of the test

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -371,7 +371,7 @@ jobs:
         timeout-minutes: 2
         run: |
           date
-          make  FLUVIO_BIN=./fluvio TEST_BIN=./fluvio-test UNINSTALL=noclean EXTRA_ARG=--keep-cluster smoke-test-tls-root
+          make  FLUVIO_BIN=./fluvio TEST_BIN=./fluvio-test UNINSTALL=noclean EXTRA_ARG=--cluster-start smoke-test-tls-root
           date
           kubectl get partitions
           kubectl get partitions  -o=jsonpath='{.items[0].status.leader.leo}' | grep 100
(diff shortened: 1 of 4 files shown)


sehz (approved):
LGTM

PR #1718, opened 2021-10-03
Invalid smartstream error

### LINKED ISSUE #1462: SPU should report error when uploaded SmartStream has no `#[smartstream]` attribute
Currently, SmartStream code where the user forgets the `#[smartstream]` attribute is perfectly valid Rust that will compile with no errors, but will fail to process when it reaches the SPU because the filter/map/aggregate hooks are not present. This currently results in a Wasmtime error in the SPU such as "expected 1 export, found 0", which is unintuitive. We should detect this case and wrap it in a more user-friendly error reminding the user to add a `#[smartstream]` attribute.

nicholastmosher:
Closes #1462.

SmartStreams are written as Rust functions that are annotated with a `#[smartstream]` attribute, like so:

```rust
use fluvio_smartstream::{smartstream, Record, Result};

#[smartstream(filter)]
pub fn filter(record: &Record) -> Result<bool> {
    let string = std::str::from_utf8(record.value.as_ref())?;
    Ok(string.contains('a'))
}
```

However, if a user forgets the `#[smartstream]` attribute, the Rust will still successfully compile, but the SPU will be unable to use it, since it is missing generated code. Currently, if you use a SmartStream with a missing `#[smartstream]` attribute, the SPU will simply close the stream without any explanation to the client:

```
$ fluvio consume fruits --tail --filter=crates/fluvio-smartstream/examples/target/wasm32-unknown-unknown/debug/fluvio_wasm_filter.wasm
Consuming records starting 10 from the end of topic 'fruits'
Consumer stream has closed
```

With this PR, the client will now be sent an ErrorCode that alerts the user that the `#[smartstream]` attribute is missing. For example, when using WASM compiled from the following code:

```rust
use fluvio_smartstream::{Record, Result};

// Note the missing #[smartstream] attribute that should be here!
pub fn filter(record: &Record) -> Result<bool> {
    let string = std::str::from_utf8(record.value.as_ref())?;
    Ok(string.contains('a'))
}
```

The resulting message when using this SmartStream is now the following:

```
$ fluvio consume fruits --tail --filter=crates/fluvio-smartstream/examples/target/wasm32-unknown-unknown/debug/fluvio_wasm_filter.wasm
Consuming records starting 10 from the end of topic 'fruits'
Error:
   0: Fluvio client error
   1: SmartStream error
   2: WASM module is not a valid 'filter' SmartStream. Are you missing a #[smartstream(filter)] attribute?
```

When using `--map` or `--aggregate`, the error message updates accordingly:

```
   2: WASM module is not a valid 'map' SmartStream. Are you missing a #[smartstream(map)] attribute?
```

```
   2: WASM module is not a valid 'aggregate' SmartStream. Are you missing a #[smartstream(aggregate)] attribute?
```


morenol (approved):
lgtm

crates/fluvio-spu/src/services/public/stream_fetch.rs
```diff
@@ -536,6 +561,40 @@ impl StreamFetchHandler {
...
+        RequestMessage::<DefaultStreamFetchRequest>::response_with_header(header, stream_response);
+
+    {
+        let mut inner_sink = sink.lock().await;
+        inner_sink
+            .send_response(&response_msg, header.api_version())
+            .await?;
+    }
+
+    Err(SocketError::Io(IoError::new(ErrorKind::InvalidData, msg)))
```
sehz, line 595: Why is this needed? sending error should be sufficient

crates/fluvio-spu/src/services/public/stream_fetch.rs
```diff
@@ -138,74 +138,99 @@ impl StreamFetchHandler {
...
+                            send_back_error(
+                                &sink,
+                                &replica,
+                                &header,
+                                stream_id,
+                                error_code,
+                                "Invalid SmartStream Aggregate",
+                            )
+                            .await?;
+                            return Err(SocketError::Io(IoError::new(
```
sehz, line 228: this should not raise error

sehz (changes requested):

PR #1787, opened 2021-10-21
Remove wasm payload from smartmodule list by default

simlay:
After playing with this for a while, I concluded that if you're just doing a `list` with no filter parameters, by default there shouldn't be a wasm playoad in the response. However, if you do have a list with any filter at all (normally this is just an index), you almost always want the wasm payload.


sehz:
After thinking more, there should be better way.  Let's pause this bit

PR #1854, opened 2021-11-06
Consumer table output update rows

### LINKED ISSUE #1846: Consume Table rendering make effort to update rows instead
Right now, table rendering just appends new rows. This feature would be more usable if the table made an effort to keep updated data contained to a row.

Generalizing this may require some integration with the `fluvio table` workflow for hints about removing rows, so for now we want to make a best effort to render rows based on a client-side cache.

---

- We should keep `--output=table` and have this rendering available under `--output=full_table`

tjtelan:
Replace consumer output from using prettytable to tui. The row update is mandatory, and has no opt-out in this PR.

* Full screen table, exit with `q` or ESC (no ctrl+c in this PR)
* Evenly spaces columns based on number of keys
* Cell data clips (not wrap)
* First version alphabetizes the columns, and uses the 1st column as primary key for updating rows. No alternative ordering in this PR
* Scrolls with arrow up/down, and mouse wheel

Resolves #1846


tjtelan:
Closing in favor of #1860

PR #1860, opened 2021-11-07
[Merged by Bors] - Add consumer output type full_table

### LINKED ISSUE #1846: Consume Table rendering make effort to update rows instead

tjtelan:
Leaving `--output=table` intact and adding full screen table render at `--output=full_table`

Consideration for CI was made that caused some code duplication, because `full_table` requires a TTY to read events from keyboard + mouse. Without this guard, tests using consumer would panic in CI.

- Full screen table, exit with q or ESC (no ctrl+c in this PR)
- Evenly spaces columns based on number of keys
- Cell data clips (not wrap)
- First version alphabetizes the columns, and uses the 1st column as primary key for updating rows. No alternative ordering in this PR
- The row update behavior is mandatory, and has no opt-out in this PR.
- Scrolls with arrow up/down, and mouse wheel

Attempt 2 
Resolves #1846

Tested on Linux and MacOS in VSCode + Alacrtty terminals using live and still data to validate fullscreen drawing, keyboard + mouse scrolling, table clearing and table exiting.

### DIFF
--- a/crates/fluvio-cli/src/consume/record_format.rs
+++ b/crates/fluvio-cli/src/consume/record_format.rs
@@ -79,10 +79,12 @@ pub fn format_raw_record(record: &[u8]) -> String {
 }
 
 // -----------------------------------
-//  Table
+//  Table (basic table)
 // -----------------------------------
 
-/// Structure json data in table format
+/// Structure json data into table row
+/// Print table header if `print_header` is true
+/// Rows may not stay aligned with table header
 pub fn format_basic_table_record(record: &[u8], print_header: bool) -> String {
     use prettytable::{Row, cell, Cell, Slice};
     use prettytable::format::{self, FormatBuilder};
@@ -124,11 +126,6 @@ pub fn format_basic_table_record(record: &[u8], print_header: bool) -> String {
     let table_format = base_format;
     table.set_format(table_format.build());
 
-    // FIXME: Live display of table data easily misaligns column widths
-    // if there is a length diff between the header and the data
-    // The rows after the first (count == 0) don't line up with the header
-    // prettytable might not support the live display use-case we want
-
     let mut out = Vec::new();
     if print_header {
         table.print(&mut out).expect("Unable to print table");
@@ -140,6 +137,13 @@ pub fn format_basic_table_record(record: &[u8], print_header: bool) -> String {
     format!("{}", String::from_utf8_lossy(&out))
 }
 
+// -----------------------------------
+//  Full Table (fullscreen interactive table)
+// -----------------------------------
+
+/// Updates the TableModel used to render the TUI table during `TableModel::render()`
+/// Attempts to update relevant rows, but appends to table if the primary key doesn't exist
+/// Returned String is not intended to be used
 pub fn format_fancy_table_record(
     record: &[u8],
     terminal: &mut Terminal<CrosstermBackend<Stdout>>,
--- a/crates/fluvio-cli/src/consume/record_format.rs
+++ b/crates/fluvio-cli/src/consume/record_format.rs
@@ -5,15 +5,15 @@
 //!
 
 use fluvio_extension_common::{bytes_to_hex_dump, hex_dump_separator};
-use super::TableModel;
-
-use std::io::Stdout;
-use tui::{backend::CrosstermBackend, Terminal};
-use crossterm::{
-    event::DisableMouseCapture,
-    execute,
-    terminal::{disable_raw_mode, LeaveAlternateScreen},
-};
+//use super::TableModel;
+//
+//use std::io::Stdout;
+//use tui::{backend::CrosstermBackend, Terminal};
+//use crossterm::{
+//    event::DisableMouseCapture,
+//    execute,
+//    terminal::{disable_raw_mode, LeaveAlternateScreen},
+//};
 
 // -----------------------------------
 //  JSON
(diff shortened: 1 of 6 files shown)


crates/fluvio-cli/src/consume/record_format.rs
```diff
@@ -115,18 +126,74 @@ pub fn print_table_record(record: &[u8], count: i32) -> String {
...
 
-    // FIXME: Live display of table data easily misaligns column widths
-    // if there is a length diff between the header and the data
-    // The rows after the first (count == 0) don't line up with the header
-    // prettytable might not support the live display use-case we want
-    if count == 0 {
-        table.printstd();
+    let mut out = Vec::new();
+    if print_header {
+        table.print(&mut out).expect("Unable to print table");
```
sehz, line 131: remove expect

crates/fluvio-cli/src/consume/record_format.rs
```diff
@@ -115,18 +126,74 @@ pub fn print_table_record(record: &[u8], count: i32) -> String {
...
     } else {
         let slice = table.slice(1..);
-        slice.printstd();
+        slice.print(&mut out).expect("Unable to print row");
```
sehz, line 134: remove expect

crates/fluvio-cli/src/consume/record_format.rs
```diff
@@ -115,18 +126,74 @@ pub fn print_table_record(record: &[u8], count: i32) -> String {
...
+pub fn format_fancy_table_record(
+    record: &[u8],
+    terminal: &mut Terminal<CrosstermBackend<Stdout>>,
+    table_model: &mut TableModel,
+) -> String {
+    let maybe_json: serde_json::Value = match serde_json::from_slice(record) {
+        Ok(value) => value,
+        Err(_e) => {
+            // restore terminal
+            disable_raw_mode().expect("Disabling raw mode failed");
```
sehz, line 156: remove expect

crates/fluvio-cli/src/consume/record_format.rs
```diff
@@ -115,18 +126,74 @@ pub fn print_table_record(record: &[u8], count: i32) -> String {
...
+            execute!(
+                terminal.backend_mut(),
+                LeaveAlternateScreen,
+                DisableMouseCapture
+            )
+            .expect("Failed to leave alternate screen");
```
sehz, line 162: remove expect

crates/fluvio-cli/src/consume/record_format.rs
```diff
@@ -115,18 +126,74 @@ pub fn print_table_record(record: &[u8], count: i32) -> String {
...
+            terminal.show_cursor().expect("Show terminal cursor failed");
```
sehz, line 163: remove expect

crates/fluvio-cli/src/consume/record_format.rs
```diff
@@ -115,18 +126,74 @@ pub fn print_table_record(record: &[u8], count: i32) -> String {
...
+
+            panic!("Value not json")
```
sehz, line 165: convert to proper error

crates/fluvio-cli/src/consume/record_format.rs
```diff
@@ -115,18 +126,74 @@ pub fn print_table_record(record: &[u8], count: i32) -> String {
...
+        }
+    };
+
+    let obj = maybe_json.as_object().unwrap();
```
sehz, line 169: return error

crates/fluvio-cli/src/consume/record_format.rs
```diff
@@ -115,18 +126,74 @@ pub fn print_table_record(record: &[u8], count: i32) -> String {
...
+    let keys_str: Vec<String> = obj.keys().map(|k| k.to_string()).collect();
+
+    // serde_json's Value::String() gets wrapped in quotes if we use `to_string()`
+    let values_str: Vec<String> = obj
+        .values()
+        .map(|v| {
+            if v.is_string() {
+                v.as_str()
+                    .expect("Value not representable as str")
```
sehz, line 178: remove expect

crates/fluvio-cli/src/consume/record_format.rs
```diff
@@ -115,18 +126,74 @@ pub fn print_table_record(record: &[u8], count: i32) -> String {
...
+            } else {
+                v.to_string()
+            }
+        })
+        .collect();
+
+    let header = keys_str;
+    table_model
+        .update_header(header)
+        .expect("Unable to set table headers");
```
sehz, line 189: remove expect

crates/fluvio-cli/src/consume/record_format.rs
```diff
@@ -115,18 +126,74 @@ pub fn print_table_record(record: &[u8], count: i32) -> String {
...
+    table_model
+        .update_row(values_str)
+        .expect("Unable to update table row");
```
sehz, line 192: remove expect

crates/fluvio-cli/src/consume/table_format.rs
```diff
@@ -0,0 +1,254 @@
...
+    pub fn render(
+        &mut self,
+        terminal: &mut Terminal<CrosstermBackend<Stdout>>,
+        // primary_key: Option<String>,
+        // column_order: Option<Vec<String>,
+        //record: &[u8],
+    ) {
+        terminal
+            .draw(|frame| self.table_ui(frame))
+            .expect("Could not render table frame");
```
sehz, line 186: remove expect

sehz (changes requested):
Overall look good. Just need remove panic and expect

sehz:
invalid json leads to panic: ```flvd consume test2 -B --output table
Consuming records from the beginning of topic 'test2'
The application panicked (crashed).
Message:  Value not json
Location: crates/fluvio-cli/src/consume/record_format.rs:94```

this should be handle gracefully.

sehz:
Issues in display
* Terminal doesn't seems to take entire view
* Scrolling doesn't work

tjtelan:
> Issues in display
> 
>     * Terminal doesn't seems to take entire view
> 
>     * Scrolling doesn't work

Both of these issues are not present in Linux.

Testing on MacOS, I can repro these issues. Might need to play around with the different terminal backends, but I chose `crossterm` since it was supposed to support all the basic unix and Windows.

sehz:
Can you try different backends?

tjtelan:
Tried swapping backend from `crossterm` to `termion` and tested on my local macos box. It seems to clear the screen and draw the table correctly for the first frame. If you press any button it exits, and the table doesn't update.

On Linux, I don't even notice a change. It works with both crossterm and termion.

tjtelan:
I reverted the Termion commit after playing around with Crossterm on MacOS more. I get fullscreen tables that update, clear screen and exit cleanly. There seems to be more of an issue with the stability around redrawing frames if the data isn't updating.

Here's a screenshot from MacOS with the selection active, and immediately after clearing the screen. The output feels much more sluggish than on my Linux dev environment, but it does work with the keyboard and mouse inputs.

image

sehz:
Same problem on MacOS (VsCode, Terminal emulator, Alacrity).
- "C" command didn't work
- Scroll still doesn't work,

Code looks good so let's merge and fix these issues.

sehz (approved):
LGTM

Issue #1623, opened 2021-09-14
bug: SmartStream Aggregation doesn't work


sehz:
steps to reproduce:

```
make smoke-test-local DEFAULT_SPU=1 REPL=1  EXTRA_ARG=--keep-cluster^C
ubuntu@ip-172-31-20-82:~/fluvio$ flvd topic create one
topic "one" created
ubuntu@ip-172-31-20-82:~/fluvio$ flvd produce one
> 1
Ok!
> 2
Ok!
> 3
Ok!
> 4
Ok!
> 5
Ok!
> ^C
ubuntu@ip-172-31-20-82:~/fluvio$ fluvio consume one -B --aggregate crates/fluvio-smartstream/examples/target/wasm32-unknown-unknown/release/fluvio_wasm_aggregate.wasm --initial initial.txt
Consuming records from the beginning of topic 'one'
A
1
A
2
A
3
A
4
A
5
```

Expected outcome is: A1, A12,A123,A1234,A12345

sehz:
false alarm: same issue as; #1491.  Input needs to be trimmed: 
```
cat initial.txt | tr -d "\r" > t2.txt
```

PR #1869, opened 2021-11-09
[Merged by Bors] - Persist aggregate accumulator across file batches

nacardin:
Resolves https://github.com/infinyon/fluvio/issues/1623

Currently, the initial accumulator is passed into each file batch rather than using the accumulator from the previous batch. This is not desirable since we want to aggregate the entire specified input.

This fix will persist the accumulator across the entire stream.


sehz:
Can you more notes what was problem and how this PR fixes the issue?

crates/fluvio-spu/src/services/public/stream_fetch_test.rs
```diff
@@ -1274,17 +1274,9 @@ async fn test_stream_aggregate_fetch_multiple_batch(
...
     ctx.leaders_state().insert(test_id, replica.clone());
 
-    // Aggregate 5 records
-    // These records look like:
-    //
-    // 1
-    // 2
-    // 3
-    // 4
-    // 5
```
morenol, line 1284: could you adjust the comments instead of removing them?

morenol (changes requested):
lgtm, just a comment

crates/fluvio-spu/src/services/public/stream_fetch_test.rs
sehz: Yup

sehz (changes requested):
LGTM.  Can you add to changelog?

crates/fluvio-spu/src/services/public/stream_fetch_test.rs
nacardin: updated

nacardin:
> LGTM. Can you add to changelog?

done

sehz (approved):
LGTM.  Great work!

Issue #1843, opened 2021-11-03
SPU: Execute SmartStream Pipeline


sehz:
- [ ] Support Consume operation in CLI
- [ ] Process multiple smartstream module pipeline
- [ ] Support Leader to Leader Fetch
- [ ] Join operations

PR #1872, opened 2021-11-09
[Merged by Bors] - move join stream to lazy as possible

sehz:
Instead of getting records from right stream in the beginning of fetch cycle, this PR move them in side batch processing.  This decouple join from fetch and allow integration of SmartStream steps.

Bump up SmartStream crates and mark all example as unpublished.  Part of #1843

### DIFF
--- a/crates/fluvio-smartengine/src/smartstream/mod.rs
+++ b/crates/fluvio-smartengine/src/smartstream/mod.rs
@@ -2,6 +2,7 @@ use std::sync::{Arc, Mutex};
 use std::time::Instant;
 use std::fmt::{self, Debug};
 
+use dataplane::{ErrorCode};
 use dataplane::record::Record;
 use dataplane::smartstream::SmartStreamExtraParams;
 use tracing::{debug, trace};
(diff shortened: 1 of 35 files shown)


crates/fluvio-smartengine/src/smartstream/mod.rs
```diff
@@ -245,7 +248,18 @@ impl dyn SmartStream + '_ {
             let now = Instant::now();
 
             let mut join_record = vec![];
-            join_last_record.encode(&mut join_record, 0)?;
+            if let Some(right_join_stream) = right_join.as_mut() {
+                match right_join_stream.next().await {
```
morenol, line 252: I think that we could have issues here if we dont have new records in the right stream

sehz: No changes in semantics.  Only difference is that waiting for join is done at batch processing which allow decouple from stream fetch pipeline

morenol: but my concern here is that I think that this will require a new record per batch,

sehz: move back join update to inside fetch

Issue #1131, opened 2021-05-26
Remove dependency on `kubectl` binary


tjtelan:
Cluster start and cluster checks use the locally installed `kubectl`.

This is about using `kubectl` in the codebase. Not in scripts or for CI

* src/cluster/src/check/mod.rs
* src/cluster/src/start/k8.rs

---

Calls we need to convert to using `k8-api`

* `kubectl create secret` 
* `kubectl delete secret`
* `kubectl create service`
* `kubectl delete service`
* `kubectl auth`

pinkforest:
I came across this with minikube as the idiomatic way is to pipe alias kubectl via minikube kubectl -- foobar

And then naturally start was complaining about Kubectl not found which I guessed that it wants the executable in some $PATH

I first thought maybe it requires environment variable or maybe the intention was to port to using k8-API -

There is a kube-rs client which can use rustls as dep without rolling own with OpenAPI generators;

https://github.com/kube-rs/kube-rs

Too bad there isn't CLI mapping to API so have to translate the CLI calls to programmatic version.

PR #1873, opened 2021-11-09
[Merged by Bors] - Use a random available port for tests instead of a fixed one

qrilka:
While looking into #1131 I found `sink::test_sink_copy` failing because the port hardcoded in that test was used on my machine. This prevents such a failure.


sehz (approved):
Neat. We should start to use this pattern.

PR #1879, opened 2021-11-10
[Merged by Bors] - Update full_table rendering + input support

tjtelan:
Update consume CLI `--output=full_table` margins for table goes edge to edge

Remove mouse capture, so we can copy text from full_table

Adding page up/down key support


sehz (approved):
LGTM

PR #1882, opened 2021-11-10
Fix helm chart details

qrilka:
Found this because of `cargo test` failing at https://github.com/infinyon/fluvio/blob/master/crates/fluvio-cluster/src/charts/location.rs#L183

### DIFF
--- a/k8-util/helm/Makefile
+++ b/k8-util/helm/Makefile
@@ -3,6 +3,7 @@ VERSION := $(shell cat $(MAKE_DIR)/../../VERSION)
 APP_VERSION = $(VERSION)
 GIT_VERSION=$(shell git log -1 --pretty=format:"%H")
 REGISTRY=infinyon
+SHELL := bash
 CLOUD=minikube
 SC_LOG=info
 IMAGE_POLICY=IfNotPresent
@@ -23,8 +24,8 @@ dry-run-aws:
 	helm install --debug --dry-run goodly-guppy ./fluvio --set cloud=aws
 
 update_version:
-	yq w -i fluvio-core/Chart.yaml version $(VERSION)
-	yq w -i fluvio-core/values.yaml fluvioVersion $(VERSION)
+	yq w -i $(CHART_SRC)/Chart.yaml version $(VERSION)
+	yq w -i $(CHART_SRC)/values.yaml fluvioVersion $(VERSION)
 
 check_version:
 	if [[ $(CORE_VERSION) == $(SYS_VERSION) && $(CORE_VERSION) == $(VERSION)-latest ]]; \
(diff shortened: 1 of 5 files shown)


k8-util/helm/fluvio-app/Chart.yaml
```diff
@@ -2,4 +2,4 @@ apiVersion: v2
 name: fluvio-app
 description: A Helm chart for Fluvio
 type: application
-version: 0.9.0
+version: 0.9.13
```
sehz, line 5: why bump to 0.9.13?  Chart version is template structure version not tied to content.

k8-util/helm/fluvio-app/templates/sc-deployment.yaml
```diff
@@ -19,7 +19,7 @@ spec:
         {{- toYaml .Values.scPod.nodeSelector | nindent 8 }}
       containers:
         - name: fluvio-sc
-          image: {{ .Values.image.registry }}/fluvio:{{ .Values.image.tag | default .Chart.AppVersion }}
+          image: "{{ .Values.image.registry }}/fluvio:{{ .Values.image.tag | default .Chart.AppVersion }}"
```
sehz, line 22: why quote?

k8-util/helm/Makefile
```diff
@@ -23,8 +24,8 @@ dry-run-aws:
 	helm install --debug --dry-run goodly-guppy ./fluvio --set cloud=aws
 
 update_version:
```
sehz, line 26: Much of this stuff in this Makefile is obsolete with inline chart mechanism.   Chart is now compiled into installer rather than as separate independent install.

sehz (changes requested):

k8-util/helm/fluvio-app/templates/sc-deployment.yaml
qrilka: `helm` errors out without quotes:
```
Error: YAML parse error on fluvio-app/templates/sc-deployment.yaml: error converting YAML to JSON: yaml: line 22: mapping values are not allowed in this context
```
I was using `make install_minikube` in this case, tried version 3.3.4 and the latest 3.7.1 with the same result

k8-util/helm/fluvio-app/Chart.yaml
qrilka: that's what `make` have set up, could be reverted

k8-util/helm/Makefile
qrilka: It's not documented so I had doubts about something like that, let me know what could be a proper fix

sehz: Let me clean up to reflect current state of helm chart installation and add more notes.

qrilka: Thanks, let me know if I could help

sehz: This PR should fix the test issue:  https://github.com/infinyon/fluvio/actions/runs/1455781206

PR #1888, opened 2021-11-11
Send records when max_bytes reached

### LINKED ISSUE #1887: Send records when max_bytes reached for smartstreams

nacardin:
Resolves #1887


sehz (approved):
Changes Looks good.  Can you add integration test?

PR #1889, opened 2021-11-11
[Merged by Bors] - Ensure right record is updated in integration test

morenol:
Update comment in test and add sleep to integration test to ensure the update of the right record


sehz (approved):
LGTM

Issue #1835, opened 2021-11-01
[Bug]: Request size issues when using the smart producer


morenol:
If we use the smart producer with array_map and with map, the final request  size could be higher than the SPU configuration

PR #1892, opened 2021-11-12
[Merged by Bors] - Use one more than one batch in the producer when the batch is too large

morenol:
Fixes https://github.com/infinyon/fluvio/issues/1835

### DIFF
--- a/crates/fluvio/src/producer/mod.rs
+++ b/crates/fluvio/src/producer/mod.rs
@@ -1,6 +1,9 @@
 use std::sync::Arc;
 use std::collections::HashMap;
+use fluvio_protocol::Encoder;
+use tracing::debug;
 use tracing::instrument;
+use once_cell::sync::Lazy;
 
 mod partitioning;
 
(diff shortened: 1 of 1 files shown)


sehz (changes requested):
Looks good.  Can you add unit test?

crates/fluvio/src/producer/mod.rs
```diff
@@ -270,7 +273,10 @@ fn assemble_requests(
                 partition_index: partition,
                 ..Default::default()
             };
-            partition_request.records.batches.push(Batch::from(records));
+
+            let mut batches = create_batches(records);
+
+            partition_request.records.batches.append(&mut batches);
```
sehz, line 279: This just moves records around.  I don't think this will work since SPU will still enforce Max bytes.  Instead, we need automatic batch like here: https://github.com/infinyon/fluvio/pull/1631.

morenol: I tested this with the http connector and it worked. It looks like SPU enforce max_bytes per batch and not per request. We are getting the BatchTooLarge error in the http connector without this change

sehz: Ok

sehz (approved):
Need to revisit max req later on but good enough for now

Issue #1828, opened 2021-10-30
Fluvio protocol: Support for New types


sehz:
Make fluvio encoder and decoder work with new types. Example
```
#[derive(Debug, Default, Encoder, Decoder)]
pub struct MyList(Vec<String>);
```

Current this silently fails even thought it compiles.

PR #1897, opened 2021-11-15
[Merged by Bors] - Add support for tuple structs in fluvio protocol

morenol:
Fixes https://github.com/infinyon/fluvio/issues/1828

Allows use to successfully encode/decode tuple structs, eg:
```rust
pub struct MyTupleStruct(i32, HashMap<String, String>);
```

### DIFF
--- a/crates/fluvio-protocol/tests/tuple_struct.rs
+++ b/crates/fluvio-protocol/tests/tuple_struct.rs
@@ -8,9 +8,10 @@ pub struct TupleStruct(String, i32);
 
 #[test]
 fn test_encode_tuple_struct() -> Result<(), Error> {
-    let mut v1 = TupleStruct::default();
-    v1.0 = format!("Struct without named fields");
-    v1.1 = 42;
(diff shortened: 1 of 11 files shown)


crates/fluvio-protocol/tests/tuple_struct.rs
```diff
@@ -0,0 +1,25 @@
+use std::io::Cursor;
+use std::io::Error;
+
+use fluvio_protocol::{Decoder, Encoder};
+
+#[derive(Encoder, Default, Decoder, Debug)]
+pub struct TupleStruct(String, i32);
+
```
sehz, line 8: Can you add test for NewType? 
```
pub struct TestString(String)
```

sehz (changes requested):
Changes looks good. Just need couple of more tests:
* Test for New Type
* Add derive UI Test.
also changelog entries

sehz (approved):
Nice work!

PR #1902, opened 2021-11-15
test consume on topic before producing test data

### LINKED ISSUE #1819: Upgrade test - Update validation process
The upgrade test failed to catch the root cause of #1815, which released in `0.9.10` (for future context, current release is `0.9.11`)

The upgrade process currently does the following:

1. Create topic
2. Produce data to topic
3. Consume data from topic
4. Validate data integrity

What we want to change this to:

1. Create topic
2. Consume data from topic (as smoke test)
3. Produce data to topic
4. Consume data from topic
5. Validate data integrity from step 4

tjtelan:
fixes #1819


sehz (approved):
LGTM

PR #1913, opened 2021-11-17
[Merged by Bors] - Do not start different smartengines per send call

morenol:
With this change we can create a producer with something like:

```rust
    lat data = std::fs::read(name)?;
    let params = Default::default();
    let producer = fluvio.topic_producer(&self.fluvio_topic).await?.with_map(data, params)?
```

The smartengine will be initialized when `with_filter`, `with_map`... is called and not per `send()` call

### DIFF
--- a/crates/fluvio-smartengine/src/smartstream/mod.rs
+++ b/crates/fluvio-smartengine/src/smartstream/mod.rs
@@ -214,7 +214,7 @@ impl SmartStreamContext {
     }
 }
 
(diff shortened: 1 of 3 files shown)


sehz (approved):
LGTM

morenol:
@sehz  The flaky test that failed here should be solved with https://github.com/infinyon/fluvio/pull/1889/files 

A little sleep after producing to the right stream to make sure that it is updated in the stream fetch handler

PR #1915, opened 2021-11-17
[Merged by Bors] - Initial CLI consumer support of TableFormat

### LINKED ISSUE #1793: Consumer - Integrate Table spec into consume output
We need integration of Table and the consume step.

Prior work this builds on:
* #1759 handles creating the Table spec
   * The Table module defines a set of keys we want to select from our data as columns (and any renaming of these column names)
* #1642 handles printing json data in a table format during `fluvio consume`
   * And supports table printing w/ smartstream

The bulk of this scope relies on an existing Table for key selection that are not top-level. The consumer client expects that data transformation occurs prior to printing (as opposed to the original shape of the data). 

However, based on #1759, we intended to also have client-side column selection on post-transformation data. (no column renaming expected via CLI though).

---

The consumer maintains the table buffer in `TableModel` struct, so we need to add in the Table spec transformations against the buffer.

`TableModel` needs to be refactored a bit so we can modify attributes about the table display with minimum need to touch of all the buffer data for display order purposes. Tables currently are rendered in a Row-oriented manner with the cell values printed in the order of how the json parser ordered their keys into a list.

The ordering is always alphabetical and we need to support:
- column ordering
- column renaming
- formatting options. 
- key inclusion (keying on non-top-level keys, selection of subsets) -- The idea here is we trim data we aren't going to use for display.

The CRD should contribute to all these features on the SPU-side, but there are some features that the client-side should be able to configure as a final override.

---

Chores I can do along that are related: 

Rename the table CRD to something a lot more specific. Tableformat? Tableview? Open to suggestions.

---

Tasks:
- [x] #1906
- [ ] #1907
- [ ] #1908
- [ ] CLI: Add Primary keys other than 1st column
- [ ] SPU-side: Filter out keys
- [ ] CLI: Standard cell auto-format types (Dates, time, duration, currency,)

### LINKED ISSUE #1906: Rename table CRD to `tableformat`
`table` is not a descriptive enough name for what this provides. Rename to `tableformat`.

https://github.com/infinyon/fluvio/blob/master/k8-util/helm/fluvio-sys/templates/crd_table.yaml

tjtelan:
Add ability to rearrange columns, rename columns and declaring primary keys other than the 1st column default

fixes #1793 

This PR changes the existing behavior of the `fluvio consume <topic> --output full_table`

For row-level updates, create a `tableformat` with one of the columns defined as `primaryKey: true` to update rows with that key.

```yaml
name: "example-tableformat"
inputFormat: "JSON"
columns:
  - keyPath: "key1"
    headerLabel: "Column 1"
    primaryKey: true
  - keyPath: "key2"
    headerLabel: "Column 2"
  - keyPath: "key3"
    headerLabel: "Column 3"
```

### DIFF
--- a/crates/fluvio-cli/src/consume/mod.rs
+++ b/crates/fluvio-cli/src/consume/mod.rs
@@ -9,7 +9,7 @@ use std::io::{self, ErrorKind, Read, Stdout};
 use std::collections::{BTreeMap};
 use flate2::Compression;
 use flate2::bufread::GzEncoder;
-use fluvio_controlplane_metadata::tableformat::TableFormatSpec;
+use fluvio_controlplane_metadata::tableformat::{TableFormatSpec, TableColumn};
 use tracing::{debug, trace, instrument};
 use structopt::StructOpt;
 use structopt::clap::arg_enum;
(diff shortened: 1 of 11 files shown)


sehz (approved):
LGTM

Issue #1922, opened 2021-11-18
`tableformat` records do not serialize or deserialize with column info


tjtelan:
Related to #1793

`fluvio tableformat` is still in development. Currently `fluvio tableformat create` only results in a record that only includes the name.

(Note: The CRD definitions have been changing up slightly, so the fields might not be accurate to what is in `master` branch. This may be a contributing factor in the serialize/deserialize issue)

Example crd w/ columns

```yaml
# tableformat-example.yaml
apiVersion: fluvio.infinyon.com/v1
kind: TableFormat
metadata:
  name: "examplename"
spec:
  name: "exampleformat"
  inputFormat: "JSON"
  columns:
    - keyPath: "key2"
    - keyPath: "key1"
```

```
$ kubectl apply -f tableformat-example.yaml
```

```
$ fluvio topic create tablestuff

$ fluvio consume tablestuff -B --tableformat examplename
debug: Found tableformat: TableFormatSpec { name: "exampleformat", input_format: JSON, column: None, smartmodule: None }
Consuming records from the beginning of topic 'tablestuff'
```

---

However, when we manually add in column data, the create still only results in a record with name, with no column info.

Here's an example of hardcoding column data into the `fluvio tableformat create` command. (The intention is for users is to provide the TableFormat spec to `create` through a `--config` param which will be a file path to a yaml that maps to the CRD definition)

https://github.com/tjtelan/fluvio/blob/658b2ac1d01e511d2d217905de9d4a86ca0e64e3/crates/fluvio-cli/src/tableformat/create.rs#L28-L56

And this is what the record looks like

```
$ kubectl get tableformat nocolumninfo -o json
{
    "apiVersion": "fluvio.infinyon.com/v1",
    "kind": "TableFormat",
    "metadata": {
        "creationTimestamp": "2021-11-18T19:32:43Z",
        "generation": 1,
        "name": "nocolumninfo",
        "namespace": "default",
        "resourceVersion": "522475",
        "uid": "e9eda943-23b2-4869-bfe7-143fd35e9a61"
    },
    "spec": {
        "inputFormat": "JSON",
        "name": "nocolumninfo"
    }
}
```

---
Also, when we list the tableformats in the consumer, it fails to deserialize records that contain columns.

Here's an example record with the columns defined

```
$ kubectl get tableformat examplename -o json
{
    "apiVersion": "fluvio.infinyon.com/v1",
    "kind": "TableFormat",
    "metadata": {
        "annotations": {
[...]

PR #1923, opened 2021-11-18
[Merged by Bors] - Add tableformat create

morenol:
Fixes https://github.com/infinyon/fluvio/issues/1922

This adds the capability to tableformat create to use a config file and maps that to the CRD.

### DIFF
--- a/crates/fluvio-cli/src/tableformat/mod.rs
+++ b/crates/fluvio-cli/src/tableformat/mod.rs
@@ -7,7 +7,7 @@ use std::fs::File;
 use std::io::Read;
 
 use fluvio::Fluvio;
-use fluvio_controlplane_metadata::tableformat::TableFormatSpec;
+use fluvio_controlplane_metadata::tableformat::{InputFormat, TableFormatSpec, TableFormatColumnConfig};
 use fluvio_extension_common::Terminal;
 use fluvio_extension_common::COMMAND_TEMPLATE;
 mod create;
@@ -62,7 +62,9 @@ impl TableFormatCmd {
 
 #[derive(Debug, Deserialize, Clone)]
 pub struct TableFormatConfig {
-    spec: TableFormatSpec,
+    pub input_format: Option<InputFormat>,
+    pub columns: Vec<TableFormatColumnConfig>,
+    pub smartmodule: Option<String>,
 }
 
 impl TableFormatConfig {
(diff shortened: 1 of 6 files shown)


crates/fluvio-controlplane-metadata/src/tableformat/spec.rs
```diff
@@ -30,10 +30,56 @@ impl Default for InputFormat {
...
-//    pub path: String,
-//    pub format: String,
-//}
+#[derive(Encoder, Decoder, Default, Debug, PartialEq, Clone)]
+#[cfg_attr(
+    feature = "use_serde",
+    derive(serde::Serialize, serde::Deserialize),
+    serde(rename_all = "camelCase")
+)]
+pub struct TableFormatColumnConfig {
```
morenol, line 39: I think that one of these fields should be required, not sure which one. I marked them as option because in the CRD they are not required.

What do you think? @tjtelan

crates/fluvio-cli/src/tableformat/mod.rs
```diff
@@ -62,45 +60,30 @@ impl TableFormatCmd {
...
-//            name: config.name,
-//            type_: config.type_,
-//            topic: config.topic,
-//            parameters: config.parameters,
-//            secrets: config.secrets,
-//        }
-//    }
-//}
+#[derive(Debug, Deserialize, Clone)]
+pub struct TableFormatConfig {
```
morenol, line 64: Here I am using the exact same format that we use with `kubectl apply`. Is this what we want? or do we want to have something more simpler for the user? @tjtelan

tjtelan: We want just the `spec` part of the yaml for `kubectl apply`.

So for a k8 config

```yaml
apiVersion: fluvio.infinyon.com/v1
kind: TableFormat
metadata:
  name: "example-column-ordered"
spec:
  name: "example-column-ordered"
  inputFormat: "JSON"
  columns:
    - keyPath: "key2"
    - keyPath: "key1"
```

We want the user to give a config like:

```yaml
inputFormat: "JSON"
columns:
  - keyPath: "key2"
  - keyPath: "key1"
```

The name is provided as the CLI positional arg

morenol: got it

crates/fluvio-controlplane-metadata/src/tableformat/spec.rs
tjtelan: The `key_path` is the only field that needs to be required

tjtelan (approved):
Tested locally. Works for me, and the `fluvio tableformat list` output returns all the data as expected

Issue #1894, opened 2021-11-14
SmartStream aggregate  for version 0.3.0 doesn't work for current version 


sehz:
Steps to reproduce.

Deploy latest dev release of fluvio.
```
$ git checkout tags/v0.9.12 -b v9.12
$ make build_smartstreams
$ fluvio Topic create
```
Consume fails
```
$ fluvio consume test -B --aggregate  crates/fluvio-smartstream/examples/target/wasm32-unknown-unknown/release/fluvio_wasm_aggregate.wasm 

Consuming records from the beginning of topic 'test'
Consumer stream has closed
```

morenol:
So, the issue is that now we changed the output of the aggregate function to return SmartStreamAggregateOutput instead of SmartStreamOutput (it includes now the accumulator, in the new version of fluvio we expect to have the accumulator in the output so it fails decoding, I can change the smartengine to be backward compatible but if I do that the old error from aggregate(not be persistent across batches) is maintained for old versions

nacardin:
Closed by https://github.com/infinyon/fluvio/pull/1921

PR #1924, opened 2021-11-19
[Merged by Bors] - Use version from the client in SmartEngine to encode/decode input/output

morenol:
Closes https://github.com/infinyon/fluvio/issues/1894

Backward compatibility with old wasm modules is keep due to we now check for both 
```rust
 TypedFunc<(i32, i32), i32>;
 TypedFunc<(i32, i32, u32), i32>;
```


sehz (approved):
Nice job!

PR #1935, opened 2021-11-22
Remove payload from smartmodule

### LINKED ISSUE #1781: Add Remove wasmmodule from response to list request for smartmodules by default
Currently when you list the smartmodules, the API returns all the bits for a smart module. This isn't scalable and will break soon.

simlay:
This PR ads a `SmartModuleMetadatSpec` and a `SmartModuleMetadata` object to the public API. This object is simply derived from `SmartModule`/`SmartModuleSpec`.

Closes #1781.


sehz:
CI failure

crates/fluvio-controlplane-metadata/src/lib.rs
```diff
@@ -31,6 +31,7 @@ pub mod extended {
         Partition,
         ManagedConnector,
         SmartModule,
+        SmartModuleMetadata,
```
sehz, line 34: Don't think this should be in here since this is not storage Spec

crates/fluvio-sc-schema/src/objects/create.rs
```diff
@@ -44,6 +45,7 @@ pub enum ObjectCreateRequest {
     Topic(TopicSpec),
     CustomSpu(CustomSpuSpec),
     SmartModule(SmartModuleSpec),
+    SmartModuleMetadata(SmartModuleMetadataSpec),
```
sehz, line 48: should not be in create request

crates/fluvio-sc-schema/src/smartmodule/mod.rs
```diff
@@ -34,4 +34,21 @@ mod convert {
 
     ObjectTryFrom!(WatchResponse, SmartModule);
     ObjectTryFrom!(ListResponse, SmartModule);
+
+    use super::SmartModuleMetadataSpec;
+
+    impl AdminSpec for SmartModuleMetadataSpec {
```
sehz, line 40: No need to implement AdminSpec.  It should be associated type of SmartModuleSpec

crates/fluvio-sc/src/services/public_api/create.rs
```diff
@@ -43,6 +43,9 @@ pub async fn handle_create_request<AC: AuthContext>(
             super::smartmodule::handle_create_smartmodule_request(common, create, auth_context)
                 .await?
         }
+        ObjectCreateRequest::SmartModuleMetadata(_) => {
```
sehz, line 46: should not implement this

crates/fluvio-sc/src/services/public_api/smartmodule/list.rs
```diff
@@ -0,0 +1,63 @@
...
+use fluvio_sc_schema::objects::{ListResponse, Metadata};
+use fluvio_sc_schema::smartmodule::SmartModuleMetadataSpec;
+use fluvio_controlplane_metadata::store::KeyFilter;
+use fluvio_controlplane_metadata::extended::SpecExt;
+use fluvio_auth::{AuthContext, TypeAction};
+
+use crate::services::auth::AuthServiceContext;
+
+#[instrument(skip(filters, auth_ctx))]
+pub async fn handle_metadata_fetch_request<AC: AuthContext>(
```
sehz, line 14: rather than re-implementing this, can use common fetch list

sehz (changes requested):

crates/fluvio-sc-schema/src/smartmodule/mod.rs
simlay: I think it's needed to used for [`ListResponse`](https://github.com/infinyon/fluvio/blob/master/crates/fluvio-sc-schema/src/objects/list.rs#L19-L26)

sehz: that's ListRequest

simlay: I mean it's required for both though `ListRequest` and `ListResponse`.

sehz:
```
 pub trait AdminSpec: Spec + Encoder + Decoder {
        type ListFilter: Encoder + Decoder + Sized + Debug;
        type ListType: Encoder + Decoder + Debug;
        type WatchResponseType: Spec + Encoder + Decoder;
    }
```

The only requirement for ListType is that it can be serialized/deserialized thru Fluvio Encoder

sehz:
obsolete. superseded by summary mode

Issue #1841, opened 2021-11-03
Audit connector secrets in logs


simlay:
maybe figure out how to skip the `secrets` field in the debug field?

nacardin:
Maybe wrap String in struct which doesn't have debug output. This can deref to String.

morenol:
Closed by #1952

PR #1952, opened 2021-11-30
[Merged by Bors] - Wrap secrets in a struct that does not leak internal value in display…

morenol:
… and debug implementation

Solves #1841


sehz (approved):
LGTM

PR #1964, opened 2021-12-01
[Merged by Bors] - Bump fluvio-future

### LINKED ISSUE #1629: HandshakeError does not expose underlying error
We only see the following for any connection error:
```
❯ fluvio topic list                                                                          
Error: 
   0: Fluvio client error
   1: Fluvio socket error
   2: HandshakeError
```

This may be something that needs to be fixed in our futures crate.

morenol:
Closes  #1629

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1729,9 +1729,9 @@ dependencies = [
 
 [[package]]
 name = "fluvio-future"
-version = "0.3.11"
+version = "0.3.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d2645caef3993b4bcb8a4aa90da3546d993a9ce8a710acebb5fed588d4c2f1e7"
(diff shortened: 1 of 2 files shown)


sehz:
Is there other crates that need to updated their version of fluvio future?

morenol:
We may enforce all the crates to use fluvio-future>=0.3.12 but I dont think that it is necessary now. Should we do that?

 I am just updating Cargo.lock with that so the binaries and docker images created use that.

sehz:
Only for crates need to depend for some explicit behavior.  This is mainly for external crates which may not pick up behavior changes

morenol:
ok, will add that to fluvio client, not sure if we need that elsewhere, this fluvio-future update includes the display representation update for its error type

sehz (approved):
LGTM

PR #1965, opened 2021-12-01
[Merged by Bors] - Delete unused files from crates

morenol:
There are some files that are not used(they were used probably when these fcrates were in their own repository).

 For instance, there are some empty `.gitignore`, a .github/workflow/ not in the root of the repo and a couple of unused Cargo.lock.
 
 This PR removes some of them


sehz (approved):
Nice clean up

PR #1971, opened 2021-12-01
[Merged by Bors] - Add json array support for full_table

### LINKED ISSUE #1967: Enable feeding an Aggregate to a Tableformat
Right now, there is no way to feed an aggregated stream into a Tableformat to get a pretty display. This is problematic because we would like to be able to view aggregates as a live-updating table, which is what tableformat is for.

The root problem is that Aggregates typically represent a collection of data points in _each_ Record in a stream. For example, if we have Records in an aggregate stream that look like this, they cannot be used in a tableformat:

```
[
  {"repo": "infinyon/fluvio", "stars": 790},
  {"repo": "infinyon/node-bindgen", "stars": 236}
]
```

For this, we would expect a tableformat with columns namd `repo` and `stars`, with two rows.

One way to make this happen is to enable `array-map`s in a derived stream. The array-map would flatten the collection of data points and create a stream of the individual data points, which the tableformat can understand. Currently, however, array-map is not implemented/supported for derived streams, so this solution is unavailable.

tjtelan:
Table row updates now check for the json type. Handles objects and arrays of objects (for multiple updates)

Fixes #1967

Manual test:

Input
```
{"rand_int":80,"rand float":14.106}
[{"rand_int":40,"rand float":7.106}]
[{"rand_int":20,"rand float":4.106}, {"rand_int":10,"rand float":2.106}]
```

Output:
`flvd consume test -B --output full_table`

```
┌('c' to clear table | 'q' or ESC to exit) | Items: 4───┐
│rand float                  rand_int                   │
│14.106                      80                         │
│7.106                       40                         │
│4.106                       20                         │
│2.106                       10                         │
└───────────────────────────────────────────────────────┘
```

---

### DIFF
--- a/crates/fluvio-cli/src/consume/record_format.rs
+++ b/crates/fluvio-cli/src/consume/record_format.rs
@@ -8,6 +8,7 @@ use fluvio_controlplane_metadata::tableformat::TableFormatColumnConfig;
 use fluvio_extension_common::{bytes_to_hex_dump, hex_dump_separator};
 use super::TableModel;
 use std::collections::BTreeMap;
+use crate::{CliError, Result};
 
 // -----------------------------------
 //  JSON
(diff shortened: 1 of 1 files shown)


sehz (approved):
LGTM

PR #1974, opened 2021-12-02
Do not display spinner in CI

### LINKED ISSUE #1968: Hide spinner and progress indicator in CI
We can either detect that we are on CI, or add a flag to hide the spinner and progress indicator if we are in CI

morenol:
This PR adds the change to not display the spinner in the CI. It can also be hidden from the cli with the --hide-spinner flag

Closes #1968

### DIFF
--- a/crates/fluvio-cluster/src/render.rs
+++ b/crates/fluvio-cluster/src/render.rs
@@ -1,4 +1,48 @@
+use indicatif::ProgressBar;
+
 pub trait ProgressRenderedText {
     /// Rendered text of the last finished step in the progress
     fn msg(&self) -> String;
+}
+
+#[derive(Debug)]
+pub enum ProgressRenderer {
+    /// Render the progress using eprintln macro
+    Std,
+    /// Render the progress using Indicatiff
+    Indicatiff(ProgressBar),
+}
+
+impl ProgressRenderer {
+    pub fn println(&self, msg: String) {
+        match self {
+            ProgressRenderer::Std => eprintln!("{}", msg),
+            ProgressRenderer::Indicatiff(pb) => pb.println(&msg),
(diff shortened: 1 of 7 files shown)


crates/fluvio-cluster/src/render.rs
```diff
@@ -1,4 +1,48 @@
...
+#[derive(Debug)]
+pub enum ProgressRenderer {
+    /// Render the progress using eprintln macro
+    Std,
+    /// Render the progress using Indicatiff
+    Indicatiff(ProgressBar),
+}
+
+impl ProgressRenderer {
+    pub fn println(&self, msg: String) {
```
sehz, line 17: if this is change to: `pub fn println(&self, msg: &str)` then don't need `into()`

crates/fluvio-cluster/src/render.rs
```diff
@@ -1,4 +1,48 @@
...
+        match self {
+            ProgressRenderer::Std => eprintln!("{}", msg),
+            ProgressRenderer::Indicatiff(pb) => pb.println(&msg),
+        }
+    }
+
+    pub fn set_message(&self, msg: String) {
```
sehz, line 24: same here. `set_message(&self, msg: &str)` would be better

sehz (changes requested):
Looks great.  Couple of minor suggestions.

crates/fluvio-cluster/src/render.rs
morenol: I was having issues with the fact that set_message in progress bar has as parameter a  `impl Into<Cow<'static, str>>);`.  I will change this to also use this

sehz (approved):
LGTM

Issue #1940, opened 2021-11-24
Consume with end


sehz:
In order to fulfill use case for [bulk read and write](https://github.com/infinyon/fluvio/issues/1898), we need to limit consume when it reaches certain criteria instead of waiting for further inputs.  In order to that we should provide 'end` options:

```
$ fluvio consume my_topic -B --end <offset>
```

<end_option> is optional.  if it is not specified, it would read to current offset at time of reading.

sehz:
@bohlmannc, interested in this issue?

bohlmannc:
Yes, I will look into it.

tjtelan:
We have the `disable_continuous` option which sort of implements this by closing the stream when we've reached the end of a topic. We would need to pass in an optional ending offset to the stream handling loop so we can exit early

https://github.com/infinyon/fluvio/blob/e9807a82eb9fdffc7391a7f39cfa8f2a75d6ccb7/crates/fluvio-cli/src/consume/mod.rs#L353-L358

bohlmannc:
Hey tj just to clarify, when reading records from a stream, the offset is not returned as part of the result, so would it suffice to just count how many records are successfully read from the stream and then exit early when that count equals the optional ending offset? I think I was overcomplicating things before I saw your comment above.

tjtelan:
@bohlmannc The [`Record`](https://docs.rs/fluvio/0.10.0/fluvio/consumer/struct.Record.html) type we receive from `stream_with_config` (which we call within `consumer_records_stream`) does include the offset returned alongside the data content.

We should add an optional command-line flag (to `ConsumeOpt`) to take in this end offset, and use it when building the consumer's config. I was thinking this end offset would be added to `ConsumerConfig` as an `Option<i64>` (to match the numerical type we use inside `Record`). Then we can compare offsets in the stream fetch loop after printing the record.

https://github.com/infinyon/fluvio/blob/8e77bda1b0b7315c2c038965ebb3ba25fcc57b1e/crates/fluvio/src/consumer.rs#L711-L727

After we close the stream, we should print a message stating that we've exited the consumer stream bc we've reached the offset threshold.

bohlmannc:
Okay that makes way more sense, thanks for explaining. Finding the record struct in the dataplane crate also helped.

nacardin:
Closed by https://github.com/infinyon/fluvio/pull/2023

Issue #1898, opened 2021-11-15
Bulk Save and Restore


sehz:
For development and testing purposes, we need way to move data between clusters.   This utility should work on both local and k8 clusters.   It should be designed only for bulk copying not replacement for consumer/producer.  

Something like this:
```
$ fluvio-bulk export /tmp/
<this will create zip file for existing data set>
```
fluvio-data.zip
```
$ fluvio cluster start --import /tmp/fluvio-data.zip
```

tjtelan:
Related: #1672 

We need some assurance that no writes to the cluster are possible during the save or restore actions

sehz:
related issue: https://github.com/infinyon/fluvio/issues/1504.

```
kubectl cp fluvio-spg-main-0:/var/lib/fluvio/data data1
```

This works.   Can probably use similar technique to reload them.  Just probably need to restart SPU after load them.

sehz:
Copying 784M using `kubectl cp` took 1 min 10 seconds
```
time kubectl cp fluvio-spg-main-0:/var/lib/fluvio/data data3
tar: removing leading '/' from member names
error: unexpected EOF

real	1m10.510s
user	0m3.215s
sys	0m4.167s
```

Consuming using fluvio CLI took: ~1m 15 seconds:
```
fluvio consume producer-test -B > producer3.data
```

sehz:
Similar for produce:
``
time fluvio produce test -f producer3.data

real	0m53.774s
user	0m13.656s
sys	0m0.907s
``

Closing this issue since there are no benefit of having alternative path to copy data from/to from pod.

PR #1984, opened 2021-12-02
Consume with end

bohlmannc:
# What does this PR do?

This PR adds a new option for consuming called `end-offset` in order to cease reading from the stream when the ending offset is found in the stream's current record, instead of waiting for new inputs.
 
# Motivation

Helping out on good first issues.

# Related issues

A list of issues closed by or relevant to this PR.

- Supports [#1898] (https://github.com/infinyon/fluvio/issues/1898)
- Closes [#1940] (https://github.com/infinyon/fluvio/issues/1940).

# Additional Notes

None

# Checklist

- [ ] Created or updated all relevant documentation
- [ ] Added unit tests to relevant added or changed code
- [ ] Bumped required crate version numbers
- [ ] Updated CHANGELOG.md

### DIFF
--- a/crates/fluvio-cli/src/consume/mod.rs
+++ b/crates/fluvio-cli/src/consume/mod.rs
@@ -639,6 +639,19 @@ impl ConsumeOpt {
                 )
                 .bold()
             );
+<<<<<<< HEAD
+=======
+        // If --end-offset=X
+        } else if let Some(end) = self.end_offset {
+            eprintln!(
+                "{}",
+                format!(
+                    "Consuming records starting from the end record {} in topic '{}'",
+                    end, &self.topic
+                )
+                .bold()
+            );
+>>>>>>> 1acd71b6... rephrasing end offset message
         // If no offset config is given, read from the end
         } else {
             eprintln!(
--- a/crates/fluvio-cli/src/consume/mod.rs
+++ b/crates/fluvio-cli/src/consume/mod.rs
@@ -100,6 +100,10 @@ pub struct ConsumeOpt {
     #[structopt(short, long, value_name = "integer", conflicts_with_all = &["from_beginning", "tail"])]
     pub offset: Option<u32>,
 
+    /// The offset of the last record to consume 
+    #[structopt(long, value_name = "integer")]
+    pub end_offset: Option<i64>,
+
     /// Consume records starting X from the end of the log (default: 10)
     #[structopt(short = "T", long, value_name = "integer", conflicts_with_all = &["from_beginning", "offset"])]
     pub tail: Option<Option<u32>>,
@@ -347,6 +351,10 @@ impl ConsumeOpt {
             builder.disable_continuous(true);
         }
 
+        if let Some(end_offset) = self.end_offset {
+            builder.end_offset(Some(end_offset));
+        }
+
         let consume_config = builder.build()?;
         debug!("consume config: {:#?}", consume_config);
 
@@ -369,6 +377,7 @@ impl ConsumeOpt {
         tableformat: Option<TableFormatSpec>,
     ) -> Result<()> {
         self.print_status();
+        let potential_offset: i64 = config.end_offset.unwrap_or(-1);
         let mut stream = consumer.stream_with_config(offset, config).await?;
 
         let templates = match self.format.as_deref() {
@@ -482,6 +491,11 @@ impl ConsumeOpt {
                     &mut None,
                     &mut None,
                 );
+
+                if potential_offset == record.offset {
+                    println!("Ending offset has been reached; exiting consumer stream.");
+                    break;
+                }
             }
         }
 
--- a/crates/fluvio-cli/src/consume/mod.rs
+++ b/crates/fluvio-cli/src/consume/mod.rs
@@ -639,8 +639,6 @@ impl ConsumeOpt {
                 )
                 .bold()
             );
-<<<<<<< HEAD
-=======
         // If --end-offset=X
         } else if let Some(end) = self.end_offset {
             eprintln!(
@@ -651,7 +649,6 @@ impl ConsumeOpt {
                 )
                 .bold()
             );
->>>>>>> 1acd71b6... rephrasing end offset message
         // If no offset config is given, read from the end
         } else {
             eprintln!(
--- a/crates/fluvio/src/consumer.rs
+++ b/crates/fluvio/src/consumer.rs
@@ -724,6 +724,8 @@ pub struct ConsumerConfig {
     pub(crate) smartmodule: Option<SmartModuleInvocation>,
     #[builder(default)]
     pub(crate) derivedstream: Option<DerivedStreamInvocation>,
+    #[builder(default)]
+    pub end_offset: Option<i64>,
 }
 
 impl ConsumerConfig {


sehz:
@morenol Can you test and review the code?

sehz:
add changelog

crates/fluvio-cli/src/consume/mod.rs
```diff
@@ -438,6 +447,11 @@ impl ConsumeOpt {
                                 &mut maybe_terminal_stdout,
                                 &mut maybe_table_model,
                             );
+
+                            if potential_offset == record.offset {
+                                println!("Ending offset has been reached; exiting consumer stream.");
```
morenol, line 452: I think this should go to stderr
```suggestion
                                eprintln!("Ending offset has been reached; exiting consumer stream.");
```

crates/fluvio-cli/src/consume/mod.rs
```diff
@@ -482,6 +496,11 @@ impl ConsumeOpt {
                     &mut None,
                     &mut None,
                 );
+
+                if potential_offset == record.offset {
+                    println!("Ending offset has been reached; exiting consumer stream.");
```
morenol, line 501: Also this should go to stderr

crates/fluvio-cli/src/consume/mod.rs
```diff
@@ -438,6 +447,11 @@ impl ConsumeOpt {
                                 &mut maybe_terminal_stdout,
                                 &mut maybe_table_model,
                             );
+
+                            if potential_offset == record.offset {
```
morenol, line 451: I think that we could have issues here if we run:

`fluvio consume topic --end-offset 10` in a topic that has more than 10 records. It will never break in that case. Maybe we should do `<=` instead but then we may need to use something different than -1 as default value

morenol (commented):
It is working as expected, I added a couple of suggestions in the code

I also think that we need to validate that given --end-offset  is positive

crates/fluvio-cli/src/consume/mod.rs
bohlmannc: I see what you're saying here. What if we set disable_continuous to be true if a positive end-offset is specified? That way, if someone accidentally types 55 instead of 5 in a topic with 10 records, at least the loop will break (although they will still get more records than they originally wanted)

morenol: The problem that I see with that is that disable_continuous will close the stream even if we are not reach the `end offset` that we want.

I think that we can avoid most of these issues with something like:
```rust
if let Some(potencial_offset) = maybe_potencial_offset {
       if  record.offset >= potencial_offset {
          eprintln!("Ending offset has been reached; exiting consumer stream.");
          break;
       }
}
```

bohlmannc:
```
I think that we could have issues here if we run:
fluvio consume topic --end-offset 10 in a topic that has more than 10 records.
```
I'm a bit confused. Did you mean to say less than 10 records here?

For example, say i send 10 records (indexed as 0-9 inclusive) to a new topic (foo, bar, baz, blah, more blah, foo, bar, baz, blah, more blah)

With the code change you provided above (which I completely agree with and have been using), a user inputting `flvd consume example -B --end-offset 9` will see this in the output:
```
foo
bar
baz
blah
more blah
foo
bar
baz
blah
more blah
Ending offset has been reached; exiting
Consumer stream has closed
```

So far, so good. However, if the user inputs `flvd consume example -B --end-offset 10` (ie an offset that is greater than the highest current offset in the topic), then the stream will not be closed and will be waiting on potential new inputs and would require the user to ctrl-c to end consuming. 

Shouldn't we want to close the stream if we've reached the end of the topic even if we inadvertently specify an end-offset that is too high? This is the behavior I was interpreting based on Sehyo and TJ's original comments on the issue. Sehyo mentioned we don't want to wait on further inputs, in which case we'd want to exit early at a specified end-offset or exit when we reach the end of the topic. 

Note that inputting `flvd consume example -B --end-offset 10 -d` would close the stream without waiting for new inputs. In which case, I think it does make sense to force disable_continuous to be true if end-offset is specified, on the off chance someone adds an extra 0 and specifies 100 as an end-offset when they meant 10. There is still the problem of them getting extra values that they didn't want in certain scenarios (for example if they accidentally input 100, really only wanted 10, but got 50 records as the -d flag read until the end of the records on that topic, which happened to be more than 10 but less than 100).

```
The problem that I see with that is that disable_continuous will close the stream even if we havent reached the end offset that we want.
```
I didn't see that behavior on my local cluster. Am I missing something/doing something wrong?

 `flvd consume example -B --end-offset 4 -d` yielded:
```
foo
bar
baz
blah
more blah
```

morenol: > I'm a bit confused. Did you mean to say less than 10 records here?

I mean, when you use `fluvio consume` without -B, it will fetch the records from the end of the stream. If we run `flvd consume example --end-offset 9` in that topic that you mentioned, it will never break (assuming that we are not using disable_continuous behavior) for the `potential_offset == record.offset` case because the first record that it will get is the record with offset 10.

> So far, so good. However, if the user inputs flvd consume example -B --end-offset 10 (ie an offset that is greater than the highest current offset in the topic), then the stream will not be closed and will be waiting on potential new inputs and would require the user to ctrl-c to end consuming.

I think that this ise the expected behavior, I think that we shouldn't exit early unless we set the disable_continuous flag. Could you confirm which is the expected behavior for this case @sehz?

> I didn't see that behavior on my local cluster. Am I missing something/doing something wrong?

I meant the case that you described above: it will exit earlier if we use as offset a number greater than the highest current offset in the topic.

> (for example if they accidentally input 100, really only wanted 10, but got 50 records as the -d flag read until the end of the records on that topic)

Something that we can do, is t update  the print_status function to include information about --end-offset so the user could get that information just in case that they accidentally put a large input

bohlmannc: @morenol Luis thanks for your comments. I made some change to force a read from beginning, even if the user forgot the -B flag when specifying an end-offset by changing some logic in the calculate_offset fn. Let me know if you think that makes sense. I am still checking if -B is specified with a particular offset before defaulting to 0.

I also added some logic to ensure a positive end-offset, but noticed that `flvd consume topic -B  --end-offset "-9"` or `flvd consume topic -B  --end-offset -9` already yields `error: Found argument '-9' which wasn't expected, or isn't valid in this context`. So the code I added may be redundant?

I did update print_status as well to specify the end-offset we expect to consume until.

sehz: start and end offset should be thought of as independent parameters similar to range(min,max).  Of course, basic validation must be enforced (min <= max) and both must be positive offset. 

So we can have the following scenarios.  LEO (Log End Offset) is the last offset in the topic.
(1) LEO = 4,  start_offset = 0 (-B), end_offset =  2,  fetch 0..2
(2) LEO = 4,  start_offset = 4 (no argument),  end_offset = 4,  fetch 4..4
(3) LEO = 4,  start_offset = 4 (no argument),  end_offset = 6,  wait and fetch 4..6
(4) LEO = 4,  start_offset = 2, end_offset = 10,   fetch 2..4, wait, fetch 4.10

crates/fluvio-cli/src/consume/mod.rs
```diff
@@ -644,6 +652,16 @@ impl ConsumeOpt {
                 )
                 .bold()
             );
+        // If --end-offset=X
+        } else if let Some(end) = self.end_offset {
+            eprintln!(
+                "{}",
+                format!(
+                    "Consuming records until record {} in topic '{}'",
```
morenol, line 660: I think that in this case, the message should still mention that it will start from the end

crates/fluvio-cli/src/consume/mod.rs
```diff
@@ -684,6 +702,12 @@ impl ConsumeOpt {
         } else if let Some(maybe_tail) = self.tail {
             let tail = maybe_tail.unwrap_or(DEFAULT_TAIL);
             Offset::from_end(tail)
+        } else if let Some(maybe_end_offset) = self.end_offset {
```
morenol, line 705: I think that this block is not needed

crates/fluvio-cli/src/consume/mod.rs
morenol: Thanks @bohlmannc I added some comments to the changes.

crates/fluvio-cli/src/consume/mod.rs
bohlmannc: Yeah makes sense. I think we were on the right track initially, based on sehyo's updated comments above. I'll fix this.

sehz:
@morenol Can you review changes?

crates/fluvio-cli/src/consume/mod.rs
```diff
@@ -347,6 +351,16 @@ impl ConsumeOpt {
             builder.disable_continuous(true);
         }
 
+        if let Some(end_offset) = self.end_offset {
+            if end_offset <= 0 {
+                eprintln!("Argument end-offset must be greater than zero")
+            } else if end_offset > offset {
```
morenol, line 357: I think that we should fail if end_offset is less than initial offset 
```suggestion
            } else if end_offset < offset {
```

morenol (changes requested):
Thanks again, I added some comments, please also update the changelog in the root of the repo and fix the lint issues

sehz:
CI failures: https://github.com/infinyon/fluvio/actions/runs/1580141181

bohlmannc:
Had to close this because I lost necessary commits when force-pushing. Updated here: https://github.com/infinyon/fluvio/pull/2022

PR #1985, opened 2021-12-03
[Merged by Bors] - Adding connector test to smoke tests

### LINKED ISSUE #1982: Test managed connector as part of smoke test
Create a managed connector as part of the smoke test.

Default no connectors. Must opt in.

tjtelan:
Fixes #1982 

Pass a connector config to the smoke test to include in consume testing. Checks that connector topic actively receives data.

- Made changes to test runner to catch test panics in CI so errors and exit codes would report accurately as failures
- Added CLI tests to publish requirements
- Left a note in the `k8_cluster_test` for how to change the way the k8 smoke tests w/ tls run when the support comes in.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -787,7 +787,7 @@ jobs:
       - local_cluster_test
       - k8_cluster_test
       - k8_upgrade_test
-     # - k8_cli_smoke
+      - k8_cli_smoke
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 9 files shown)


tjtelan:
I thought there was an issue w/ connector pods not starting in the smoke test (`smoke-test-k8-tls-root`), but turns out that the policy being used in the test is working properly. But we should run the version of the tls test w/o the policy.

sehz (approved):
lgtm

PR #1989, opened 2021-12-03
[Merged by Bors] - Store and use client TLS for connectors

nacardin:
Resolves https://github.com/infinyon/fluvio/issues/1970


nacardin:
The `fs_group` is needed for `fluvio` user to read mounted secret (client TLS certs).

crates/fluvio/src/config/tls.rs
```diff
@@ -122,18 +123,38 @@ pub struct TlsCerts {
...
+}
+
 impl TlsCerts {
     /// Attempts to write the inline TLS certs into temporary files
     ///
     /// Returns a `TlsPaths` populated with the paths where the
     /// temporary files were written.
     pub fn try_into_temp_files(&self) -> Result<TlsPaths, IoError> {
         use std::fs::write;
-        let tmp = std::env::temp_dir();
```
sehz, line 132: Is there a reason for writing a cert in the client rather than installer?

sehz (changes requested):
Looks good. Just have one issue

sehz:
CI failures. I think it is better to move certs generation to the installer

crates/fluvio/src/config/tls.rs
nacardin: It was probably done to keep it in the impl of TlsCerts, but it makes more sense to have this is installer. Will fix

sehz (approved):
LGTM

PR #2003, opened 2021-12-09
[Merged by Bors] - Add CLI test - confirm cluster delete

### LINKED ISSUE #1844: CLI test - Confirm cluster delete
Cluster deletion should also delete connectors and other relevant objects in Kubernetes. We need to confirm that is the case in the CLI smoke test

tjtelan:
Deletes cluster and verifies that all CRD resources were also deleted
Added to Makefile so it will be run in CI

Fixes #1844


sehz (approved):
LGTM

PR #2013, opened 2021-12-10
[Merged by Bors] - Batching producer

### LINKED ISSUE #2000: automatic batching
With automatic batching, the producer takes care of batch management which should result in better throughput and lower latency. 

For example, suppose we have a topic with 3 partitions and we select RoundRobin mapping. We create a producer that has a maximum of 1000 bytes for batch size and 100 ms for linger time.
```
let producer = ProducerBuilder("temperature")
    .partition_map(SiphashRoundRobinPartitioner::default())
   . max_buffer_size(1000)
   .max_linger(Duration::from_millis(100))
  .build();
```

This producer can send data to 3 partitions simultaneously.   Suppose we have 10 cities we use as key ["SFO","LAX","NYC","MIA","SEA","ATL","SAN","CHI","PARIS","LON"].

Partition 0 has keys ["ATL","CHI","LAX"]
Partition 1 has keys ["LON","MIA","NYC"]
Partition 2 has keys ["PARIS","SAN","SFO"]

```
producer.send("SF","{ time: "2AM", temperature: 11.2}").await?;    // send to buffer for partition 2.
producer.send("LON","{ time: "1AM", temperature: 9.2}".await?:    // send to buffer for partition 1
.....other data
```

Now suppose the buffer for partition 2 has 988 bytes in the buffer and partition 0 has 977 bytes in the buffer. Then next write to partition 2 and 0 should flush.  Flush should be done concurrently.
```
producer.send("SF","{ time: "10AM", temperature: 2}").await?;      // no more room, this should flush existing buffer for partition 2 and this data should be in buffer.
producer.send("SF","{ time: "9AM", temperature: 3}").await?;   // put in buffer 
producer.send("ATL","{ time: "9AM", temperature: 10}").await?;   // no more room, trigger flush
```

if a partition producer is not finished sending but the buffer is full then `send` should wait til buffer is available.
```
producer.send("SF","{ time: "9AM", temperature: 3}").await?;   // put in buffer, but buffer is full capacity now and producer is still busy
producer.send("SF","{ time: "9AM", temperature: 4}").await?;  // wait til buffer has room
```

At any time, flushing can be manually accomplished or when lingering time has expired for each partition.
```
producer.flush().await?;        // flush all partitions
```

Error handling:

For some reason, produce is not able to recover from failure then, it should trigger an error
```
if let Err(ProducerFailure(err) = producer.send("SF","{ time: "10AM", temperature: 2}").await? {
    ......
}
```

Future:

We can extend the API to get the progress of the individual records
```
let progress = producer.send("SF","{ time: "10AM", temperature: 2}").await?;
assert_eq!(progress.is_buffered());        // is waiting
... sometime later
assert_eq!(progress.is_sending());         // transmission
assert_eq!(progress.is_done());             // done   

// or error
if let Some(err) = progress.error() {
}
```

Or we can wait for records to finish if you have a dependency between data
```
let progress = producer.send("SF","{ time: "10AM", temperature: 2}").await?;
progress.wait()?;
let progress = producer.send("NYC","{ time: "9AM", temperature: 3}").await?;
```

morenol:
Closes #2000

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -121,6 +121,28 @@ multiple-partition-test: test-setup
                 ${TEST_ARG_DEVELOP} \
                 ${TEST_ARG_EXTRA}
 
+batch-failure-test: TEST_ARG_EXTRA=--local $(EXTRA_ARG)
+batch-failure-test: FLV_SOCKET_WAIT=25
+batch-failure-test: DEFAULT_SPU=1
+batch-failure-test: REPL=1
+batch-failure-test: test-setup
+	$(TEST_BIN) produce_batch  \
+                ${TEST_ARG_SPU} \
+                ${TEST_ARG_LOG} \
+                ${TEST_ARG_REPLICATION} \
+                ${TEST_ARG_DEVELOP} \
+                ${TEST_ARG_EXTRA}
+
+batching-test: TEST_ARG_EXTRA=--local $(EXTRA_ARG)
+batching-test: DEFAULT_SPU=1
+batching-test: REPL=1
+batching-test: test-setup
+	$(TEST_BIN) batching  \
+                ${TEST_ARG_SPU} \
+                ${TEST_ARG_LOG} \
+                ${TEST_ARG_REPLICATION} \
+                ${TEST_ARG_DEVELOP} \
+                ${TEST_ARG_EXTRA}
 
 reconnection-test: TEST_ARG_EXTRA=--local $(EXTRA_ARG)
 reconnection-test: DEFAULT_SPU=1
--- a/crates/fluvio-cli/src/produce/mod.rs
+++ b/crates/fluvio-cli/src/produce/mod.rs
@@ -4,7 +4,7 @@ use std::path::PathBuf;
 use structopt::StructOpt;
 use tracing::error;
 
-use fluvio::{Fluvio, TopicProducer, RecordKey};
+use fluvio::{Fluvio, FluvioError, TopicProducer, TopicProducerConfigBuilder, RecordKey};
 use fluvio_types::print_cli_ok;
 use crate::common::FluvioExtensionMetadata;
 use crate::Result;
@@ -56,7 +56,17 @@ fn validate_key_separator(separator: String) -> std::result::Result<(), String> 
 
 impl ProduceOpt {
     pub async fn process(self, fluvio: &Fluvio) -> Result<()> {
-        let producer = fluvio.topic_producer(&self.topic).await?;
+        let config = if self.interactive_mode() {
+            TopicProducerConfigBuilder::default()
+                .linger(std::time::Duration::from_millis(10))
+                .build()
+                .map_err(FluvioError::from)?
+        } else {
+            Default::default()
+        };
+        let producer = fluvio
+            .topic_producer_with_config(&self.topic, config)
+            .await?;
 
         if self.raw {
             // Read all input and send as one record
--- a/crates/fluvio-cli/src/produce/mod.rs
+++ b/crates/fluvio-cli/src/produce/mod.rs
@@ -2,14 +2,12 @@ use std::fs::File;
 use std::io::{BufReader, BufRead};
 use std::path::PathBuf;
 use structopt::StructOpt;
-use tracing::{error, debug};
+use tracing::error;
 
 use fluvio::{Fluvio, TopicProducer, RecordKey};
 use fluvio_types::print_cli_ok;
 use crate::common::FluvioExtensionMetadata;
-use crate::{Result, CliError};
-
-const DEFAULT_BATCH_SIZE: usize = 50;
+use crate::Result;
 
 // -----------------------------------
 // CLI Options
@@ -76,6 +74,7 @@ impl ProduceOpt {
             self.produce_lines(&producer).await?;
         }
 
+        producer.flush().await?;
         if self.interactive_mode() {
             print_cli_ok!();
         }
@@ -87,11 +86,8 @@ impl ProduceOpt {
         match &self.file {
             Some(path) => {
                 let reader = BufReader::new(File::open(path)?);
-                let lines: Vec<_> = reader.lines().filter_map(|it| it.ok()).collect();
-                let batches: Vec<_> = lines.chunks(DEFAULT_BATCH_SIZE).collect();
-                for batch in batches {
-                    let batch: Vec<_> = batch.iter().map(|it| &**it).collect();
-                    self.produce_strs(producer, &batch).await?;
+                for line in reader.lines().filter_map(|it| it.ok()) {
+                    self.produce_line(producer, &line).await?;
                 }
             }
             None => {
@@ -100,7 +96,7 @@ impl ProduceOpt {
                     eprint!("> ");
                 }
                 while let Some(Ok(line)) = lines.next() {
-                    self.produce_strs(producer, &[&line]).await?;
+                    self.produce_line(producer, &line).await?;
                     if self.interactive_mode() {
                         print_cli_ok!();
                         eprint!("> ");
@@ -112,69 +108,44 @@ impl ProduceOpt {
         Ok(())
     }
 
-    async fn produce_strs(&self, producer: &TopicProducer, strings: &[&str]) -> Result<()> {
-        if self.kv_mode() {
-            self.produce_key_values(producer, strings).await?;
+    async fn produce_line(&self, producer: &TopicProducer, line: &str) -> Result<()> {
+        if let Some(separator) = &self.key_separator {
+            self.produce_key_value(producer, line, separator).await?;
         } else {
-            let batch = strings.iter().map(|&s| (RecordKey::NULL, s));
-            producer.send_all(batch).await?;
-            if self.verbose {
-                for string in strings {
-                    println!("[null] {}", string);
-                }
-            }
+            producer.send(RecordKey::NULL, line).await?;
         }
+
         Ok(())
     }
 
-    fn kv_mode(&self) -> bool {
-        self.key_separator.is_some()
+    async fn produce_key_value(
+        &self,
+        producer: &TopicProducer,
+        line: &str,
+        separator: &str,
+    ) -> Result<()> {
+        let maybe_kv = line.split_once(separator);
+        let (key, value) = match maybe_kv {
+            Some(kv) => kv,
+            None => {
+                error!(
+                    "Failed to find separator '{}' in record, skipping: '{}'",
+                    separator, line
+                );
+                return Ok(());
+            }
+        };
+
+        if self.verbose {
+            println!("[{}] {}", key, value);
+        }
+
+        producer.send(key, value).await?;
+        Ok(())
     }
 
     fn interactive_mode(&self) -> bool {
         self.file.is_none() && atty::is(atty::Stream::Stdin)
-    }
-
-    async fn produce_key_values(&self, producer: &TopicProducer, strings: &[&str]) -> Result<()> {
-        if let Some(separator) = &self.key_separator {
-            self.produce_key_value_via_separator(producer, strings, separator)
-                .await?;
-            return Ok(());
-        }
-
-        Err(CliError::Other(
-            "Failed to send key-value record".to_string(),
-        ))
-    }
-
-    async fn produce_key_value_via_separator(
-        &self,
-        producer: &TopicProducer,
-        strings: &[&str],
-        separator: &str,
-    ) -> Result<()> {
-        debug!(?separator, "Producing Key/Value:");
-
-        let pairs: Vec<_> = strings
-            .iter()
-            .filter_map(|s| {
-                let pieces: Vec<_> = s.split(separator).collect();
-                if pieces.len() < 2 {
-                    error!("Failed to find separator '{}' in record '{}'", separator, s);
-                    return None;
-                }
-
-                let key: &str = pieces[0];
-                let value: String = (&pieces[1..]).join(&*separator);
-                if self.verbose {
-                    println!("[{}] {}", key, value);
-                }
-                Some((key, value))
-            })
-            .collect();
-
-        producer.send_all(pairs).await?;
-        Ok(())
     }
 
     pub fn metadata() -> FluvioExtensionMetadata {
--- a/crates/fluvio-test/src/tests/producer_fail/mod.rs
+++ b/crates/fluvio-test/src/tests/producer_fail/mod.rs
@@ -73,11 +73,11 @@ pub async fn produce_batch(
 
     let result: Result<_, FluvioError> = (|| async move {
         let mut results = Vec::new();
-        for i in 0..5000 {
+        for i in 0..1000 {
             let result = producer.send(RecordKey::NULL, i.to_string()).await?;
             results.push(result);
         }
-        println!("Send 5000");
+        println!("Send 1000");
         let mut i = 0;
         for result in results.into_iter() {
             let record = result.wait().await.expect("result is not ok");
@@ -88,20 +88,14 @@ pub async fn produce_batch(
             i += 1;
         }
 
-        assert_eq!(i, 5000);
+        assert_eq!(i, 1000);
 
         cluster_manager.terminate_spu(leader).expect("terminate");
         println!("Terminate SPU");
 
-        let mut results = Vec::with_capacity(16000);
-        for i in 0..16_000 {
-            let result = producer.send(RecordKey::NULL, i.to_string()).await?;
-            results.push(result);
-        }
-        println!("Send 1000");
-
+        let result = producer.send(RecordKey::NULL, i.to_string()).await?;
         // This should fail because the SPU is terminated.
-        results.remove(0).wait().await?;
+        result.wait().await?;
         Ok(())
     })()
     .await;
--- a/crates/fluvio-test/src/tests/producer_fail/mod.rs
+++ b/crates/fluvio-test/src/tests/producer_fail/mod.rs
@@ -90,14 +90,23 @@ pub async fn produce_batch(
 
         assert_eq!(i, 1000);
 
-        fluvio_future::timer::sleep(std::time::Duration::from_secs(5)).await;
+        fluvio_future::timer::sleep(std::time::Duration::from_secs(3)).await;
 
         cluster_manager.terminate_spu(leader).expect("terminate");
         println!("Terminate SPU");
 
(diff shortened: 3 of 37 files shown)


crates/fluvio/src/producer/mod.rs
```diff
@@ -1,44 +1,331 @@
...
+            .spu_pool
+            .metadata
+            .partitions()
+            .lookup_by_key(&replica_key)
+            .await?
+            .ok_or_else(|| FluvioError::PartitionNotFound(self.topic.to_string(), self.partition))?
+            .spec;
+        let leader = partition_spec.leader;
+
+        let spu_client = self
```
sehz, line 163: why need to create a client at this point?

crates/fluvio/src/producer/mod.rs
```diff
@@ -196,354 +486,69 @@ impl TopicProducer {
...
-            assert_eq!(partition_1_request.records.batches.len(), 1);
-            let batch = partition_1_request.records.batches.get(0).unwrap();
-            assert_eq!(batch.records().len(), 2);
-            let record_1_0 = batch.records().get(0).unwrap();
-            assert_eq!(record_1_0.value.as_ref(), b"C");
-            let record_1_1 = batch.records().get(1).unwrap();
-            assert_eq!(record_1_1.value.as_ref(), b"D");
-        }
+    /// Wait for all record metadata of all records sent using smartmodule
+    #[cfg(feature = "smartengine")]
```
sehz, line 533: why this needs to be under smartengine`?

crates/fluvio/src/producer/record.rs
```diff
@@ -0,0 +1,33 @@
+use dataplane::Offset;
+use fluvio_types::PartitionId;
+use async_channel::Receiver;
+
+use crate::error::Result;
+
+use super::error::ProducerError;
+
+#[derive(Clone, Debug, Default)]
+pub struct RecordMetadata {
```
sehz, line 10: I think this is too heavy, it should be light as possible.  Probably some kind of a simple token or hash

sehz (changes requested):

Makefile
```diff
@@ -121,6 +121,18 @@ multiple-partition-test: test-setup
                 ${TEST_ARG_DEVELOP} \
                 ${TEST_ARG_EXTRA}
 
+batch-test: TEST_ARG_EXTRA=--local $(EXTRA_ARG)
+batch-test: FLV_SOCKET_WAIT=25
```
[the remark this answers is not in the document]
morenol: In the batch test, I killed the spu, and with the default value of FLV_SOCKET_WAIT it takes 5 minutes to raise a failure (which is giving timeout errors in CI)

crates/fluvio/src/producer/accumulator.rs
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    fn push_record(&mut self, record: Record) -> Result<FutureRecordMetadata> {
+        let relative_offset = self.records.len() as i64;
+        let (sender, receiver) = async_channel::bounded(1);
+        let record_size = record.write_size(0);
+
+        if self.current_size + record_size > self.write_limit {
+            self.is_full = true;
+            return Err(FluvioError::Other("cannot be added to this batch".into()));
```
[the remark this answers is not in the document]
morenol: We cannot add that to this batch because it doesnt have enough space, when this error happens, we are creating a new batch and adding the records there, see line 52

Makefile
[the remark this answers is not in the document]
sehz: so that's fair.  maybe then rename so it's clear this is more than a simple batch test

crates/fluvio/src/producer/accumulator.rs
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    pub(crate) fn create_time(&self) -> &Instant {
+        &self.create_time
+    }
+
+    fn push_record(&mut self, record: Record) -> Result<FutureRecordMetadata> {
+        let relative_offset = self.records.len() as i64;
+        let (sender, receiver) = async_channel::bounded(1);
+        let record_size = record.write_size(0);
```
sehz, line 116: use version from server instead of hard coding

crates/fluvio/src/producer/accumulator.rs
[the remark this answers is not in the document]
sehz: Then why need to return error here?

Makefile
```diff
@@ -121,6 +121,18 @@ multiple-partition-test: test-setup
                 ${TEST_ARG_DEVELOP} \
                 ${TEST_ARG_EXTRA}
 
+batch-test: TEST_ARG_EXTRA=--local $(EXTRA_ARG)
```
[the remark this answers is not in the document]
morenol: All the other tests are running with local, not sure why they fail without that

crates/fluvio/src/producer/mod.rs
morenol: Only when we are using smartengine we can have multiple records from a single send call

crates/fluvio/src/producer/accumulator.rs
morenol: I think that in that case, we may have issues if there is a change in the SPU leader and it has a different version, maybe we could use a fixed version from the client side, since all of this computation is done before we know to which SPU this will be sent?

crates/fluvio/src/producer/mod.rs
morenol: We need to get the socket before sending the request, this under the hood does not create a new socket unless the previous one was staled

crates/fluvio-test/src/tests/producer_fail/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+            let result = producer.send(RecordKey::NULL, i.to_string()).await?;
+            results.push(result);
+        }
+        println!("Send 5000");
+        let mut i = 0;
+        for result in results.into_iter() {
+            let record = result.wait().await.expect("result is not ok");
+
+            // Check that offset is correct for each record sent.
+            assert_eq!(record.offset(), i);
```
morenol, line 86: This is what I mentioned that is getting the right offset

sehz (changes requested):
There should be an integration test to check for scenarios descirbed in the issue: https://github.com/infinyon/fluvio/issues/2000.

crates/fluvio/src/producer/accumulator.rs
```diff
@@ -0,0 +1,161 @@
...
+        }
+
+        trace!(
+            partition_id,
+            "Batch is full. Creating a new batch for partition"
+        );
+
+        let mut batch = ProducerBatch::new(self.batch_size);
+
+        match batch.push_record(record) {
```
sehz, line 68: add test for this

crates/fluvio/src/producer/accumulator.rs
sehz: sure

crates/fluvio/src/producer/accumulator.rs
```diff
@@ -0,0 +1,161 @@
...
+
+        self.current_size += record_size;
+
+        self.records.push(record);
+
+        Some(PartialFutureRecordMetadata::new(
+            relative_offset,
+            self.batch_metadata.clone(),
+        ))
+    }
```
sehz, line 156: need to add test for this method

crates/fluvio/src/producer/mod.rs
```diff
@@ -33,12 +45,224 @@ use crate::producer::partitioning::{Partitioner, SiphashRoundRobinPartitioner, P
...
-    pool: Arc<SpuPool>,
-    partitioner: Box<dyn Partitioner + Send + Sync>,
+    inner: Arc<InnerTopicProducer>,
     #[cfg(feature = "smartengine")]
     pub(crate) smartmodule_instance: Option<Arc<RwLock<Box<dyn SmartModuleInstance>>>>,
 }
+
+/// Struct that is responsible for sending produce requests to the SPU in a given partition.
+struct PartitionProducer {
+    topic: String,
```
sehz, line 55: there already `ReplicaKey` that to represents replica (topic, partition)

crates/fluvio/src/producer/mod.rs
```diff
@@ -33,12 +45,224 @@ use crate::producer::partitioning::{Partitioner, SiphashRoundRobinPartitioner, P
...
+    partition: PartitionId,
+    spu_pool: Arc<SpuPool>,
+    batches_lock: Arc<Mutex<VecDeque<ProducerBatch>>>,
+    linger: Duration,
+}
+
+impl PartitionProducer {
+    fn new(
+        topic: String,
```
sehz, line 64: use ReplicaKey

crates/fluvio/src/producer/mod.rs
```diff
@@ -33,12 +45,224 @@ use crate::producer::partitioning::{Partitioner, SiphashRoundRobinPartitioner, P
...
+            .partitions()
+            .lookup_by_key(&replica_key)
+            .await?
+            .ok_or_else(|| FluvioError::PartitionNotFound(self.topic.to_string(), self.partition))?
+            .spec;
+        let leader = partition_spec.leader;
+
+        let spu_socket = self
+            .spu_pool
+            .create_serial_socket_from_leader(leader)
```
sehz, line 107: why need to create a socket when flushing? Shouldn't connect already exist?

crates/fluvio/src/producer/mod.rs
```diff
@@ -33,12 +45,224 @@ use crate::producer::partitioning::{Partitioner, SiphashRoundRobinPartitioner, P
...
+            .await?;
+
+        let mut batches_ready = vec![];
```
sehz, line 110: shouldn't batch be already calculated at this time?

crates/fluvio/src/producer/mod.rs
```diff
@@ -33,12 +45,224 @@ use crate::producer::partitioning::{Partitioner, SiphashRoundRobinPartitioner, P
...
+            if ready {
+                if let Some(batch) = batches.pop_front() {
+                    batches_ready.push(batch);
+                }
+            } else {
+                break;
+            }
+        }
+
+        // Send each batch and notify base offset
```
sehz, line 126: Shouldn't be only a single batch?

crates/fluvio/src/producer/mod.rs
```diff
@@ -33,12 +45,224 @@ use crate::producer::partitioning::{Partitioner, SiphashRoundRobinPartitioner, P
...
+        let value = record.value.as_ref();
+        let partition = self.partitioner.partition(&partition_config, key, value);
+
+        let push_record = self
+            .record_accumulator
+            .push_record(record, partition)
+            .await?;
+        let self_clone = self.clone();
+        if push_record.is_full {
+            fluvio_future::task::spawn(async move {
```
sehz, line 245: why spawn instead of just wait here?

crates/fluvio/src/producer/mod.rs
```diff
@@ -33,12 +45,224 @@ use crate::producer::partitioning::{Partitioner, SiphashRoundRobinPartitioner, P
...
+                if let Err(e) = self_clone.flush(false).await {
+                    error!("Failed to flush producer: {:?}", e);
+                }
+            });
+        }
+
+        if push_record.new_batch {
```
sehz, line 252: what does mean new batch?

crates/fluvio/src/producer/mod.rs
```diff
@@ -33,12 +45,224 @@ use crate::producer::partitioning::{Partitioner, SiphashRoundRobinPartitioner, P
...
+            let linger = self.linger;
+            fluvio_future::task::spawn(async move {
+                fluvio_future::timer::sleep(linger).await;
+                if let Err(e) = self.flush(false).await {
```
sehz, line 256: flush doesn't need to be spawn

sehz (changes requested):
more feedback

crates/fluvio/src/producer/mod.rs
morenol: Then how could I run the flush after the linger time is elapsed?

crates/fluvio/src/producer/mod.rs
morenol: So, in the RecordAccumulator, we have a VecDeq of batches per partition, each time a batch is full (due to max size constraints, we create a new batch in the VecDeq and add the record there in the new batch. Every time it is created a new batch we need to spawn a flush once the linger time is reached.

This implementation is pretty much based on this 

https://github.com/apache/kafka/blob/394175688a76d52d1895c4419711edb4b47bfd47/clients/src/main/java/org/apache/kafka/clients/producer/internals/RecordAccumulator.java#L180

crates/fluvio/src/producer/mod.rs
morenol: There could be batches in the VecDeq that are not ready to be flushed (they are not full and their creation time is smaller than the linger time

crates/fluvio/src/producer/mod.rs
morenol: There could be more than one batch in the VecDeq that is ready, but this is only true because I spawn  the flush()  call when I have a batch full, if I do that without spawn, then there could be only one batch

crates/fluvio/src/producer/mod.rs
morenol: This is reusing the same socket, this is being done  before the send in case that the SPU leader have changed

crates/fluvio/src/producer/mod.rs
morenol: I just tried to do this similar to the kafka implementation, there when the batch is full they wakeup the sender thread and then return
https://github.com/apache/kafka/blob/394175688a76d52d1895c4419711edb4b47bfd47/clients/src/main/java/org/apache/kafka/clients/producer/KafkaProducer.java#L989

crates/fluvio/src/producer/mod.rs
sehz: there are 3 ways to flush:
* after duration
* when min size threshold has been reached
* manual flushing

so if the other 2 has been reached, lingering should be reset which implies use of `select`

crates/fluvio/src/producer/mod.rs
sehz: see other comments about lingering.  Kafka uses traditional thread-based approach while we use async approach which requires different implementation strategy

crates/fluvio/src/producer/mod.rs
morenol: It is not clear to me how can I use select! here without creating a sender dispatcher

crates/fluvio/src/producer/config.rs
```diff
@@ -0,0 +1,67 @@
+use std::time::Duration;
+
+use crate::producer::partitioning::{Partitioner, SiphashRoundRobinPartitioner};
+
+const DEFAULT_LINGER_MS: u64 = 250;
+const DEFAULT_BATCH_SIZE_BYTES: usize = 16_000;
+
+/// Builder for `TopicProducerConfig`
+#[derive(Default)]
```
sehz, line 9: Can you use DeriveBuilder?

sehz (changes requested):
Generally looks good after first pass.  Minor comment about using DeriveBuilder.

crates/fluvio/src/producer/config.rs
morenol: I think that in order to use that we need to make ` partitioner: Box<dyn Partitioner + Send + Sync>` clonable

sehz: Can use `#[builder(pattern = "owned")] ` which doesn't require cloning

crates/fluvio/src/producer/partition_producer.rs
```diff
@@ -0,0 +1,238 @@
...
+
+    #[instrument(skip(self, end_event, flush_event))]
+    async fn run(
+        &self,
+        end_event: Arc<StickyEvent>,
+        flush_event: (Arc<EventHandler>, Arc<EventHandler>),
+    ) {
+        use tokio::select;
+
+        let mut linger_sleep = sleep(std::time::Duration::from_secs(1800));
```
sehz, line 94: lingering time should be getting from config value

crates/fluvio/src/producer/partition_producer.rs
```diff
@@ -0,0 +1,238 @@
...
+                    debug!("flush event received");
+                    if let Err(e) = self.flush(true).await {
+                        error!("Failed to flush producer: {}", e);
+                        self.set_error(e).await;
+                        flush_event.1.notify().await;
+
+                        continue;
+                    }
+                    flush_event.1.notify().await;
+                    linger_sleep = sleep(std::time::Duration::from_secs(1800));
```
sehz, line 114: get value from config

sehz (changes requested):

crates/fluvio/src/producer/partition_producer.rs
morenol: I am just adding a large value as linger time by default, and this is only changed to the config value once we get a new batch event

crates/fluvio/src/producer/partition_producer.rs
morenol: every time that there is a manual flush, there is not need to wait for the linger time because all is supposed to be removed from the accumulator. This is just setting the linger time as something large. This is only changed to the config value once we get a new batch event

crates/fluvio/src/producer/partition_producer.rs
morenol: I will change this to None, so it is more clear

crates/fluvio-cli/src/produce/mod.rs
```diff
@@ -112,71 +108,46 @@ impl ProduceOpt {
...
+            println!("[{}] {}", key, value);
+        }
 
-        producer.send_all(pairs).await?;
+        producer.send(key, value).await?;
         Ok(())
     }
 
+    fn interactive_mode(&self) -> bool {
+        self.file.is_none() && atty::is(atty::Stream::Stdin)
```
sehz, line 148: for interactive use, set lingering time to low (like 10ms) otherwise, it may appear to be slow

sehz (changes requested):
I think this is good enough for now.  One minor changes is need for CLI interactive use case

sehz (approved):
Nice work!

PR #2023, opened 2021-12-15
[Merged by Bors] - Consume w end

bohlmannc:
# What does this PR do?

This PR adds a new option for consuming called `end-offset` in order to cease reading from the stream when the ending offset is found in the stream's current record, instead of waiting for new inputs.
 
# Motivation

Helping out on good first issues.

# Related issues

A list of issues closed by or relevant to this PR.

- Supports [#1898] (https://github.com/infinyon/fluvio/issues/1898)
- Closes [#1940] (https://github.com/infinyon/fluvio/issues/1940).

# Additional Notes

None

# Checklist

- [ ] Created or updated all relevant documentation
- [ ] Added unit tests to relevant added or changed code
- [ ] Bumped required crate version numbers
- [x] Updated CHANGELOG.md


bohlmannc:
@morenol 
I'm stuck again. I tried reforking the repo so it would have the latest 0.11.0 version of fluvio and match crates.io. So the version in `crates/fluvio/Cargo.toml` looks better now. But obviously, i'm still getting the same CI error. How would rebasing help here? 

Is the issue because I am making a change in `crates/fluvio/src/consumer.rs` in addition to the changes being made in `crates/fluvio-cli/src/consume/mod.rs`?

tjtelan:
@bohlmannc The CI error you have is from a check on our public crate versions.

In this case, you just want to update the version for the `fluvio` crate's minor version in the `Cargo.toml`.

`version = "0.11.0"` -> `version = "0.12.0"`

Since our crate publish is automatic, CI enforces any changes to the public crates also get a version increase. We follow a semantic versioning model. Our convention are for changes to public interfaces (like structs, enums or function signatures) get a minor version bump, but otherwise patch updates.

morenol:
> @bohlmannc The CI error you have is from a check on our public crate versions.
> 
> In this case, you just want to update the version for the `fluvio` crate's minor version in the `Cargo.toml`.
> 
> `version = "0.11.0"` -> `version = "0.12.0"`
> 
> Since our crate publish is automatic, CI enforces any changes to the public crates also get a version increase. We follow a semantic versioning model. Our convention are for changes to public interfaces (like structs, enums or function signatures) get a minor version bump, but otherwise patch updates.

I think that the bump is not needed in this PR since we are not adding any changes to that crate here. I think that the problem is that we need to rebase this on top of master so this branch includes all the changes that are already on master.

I think that you could do that with 

`git pull origin master --rebase` and solve the conflicts that you may have. Before doing that, it would help to squash all your commits into a single commit. A merge commit with everything that is in the current master also should solve the problem

crates/fluvio/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio"
-version = "0.11.0"
+version = "0.12.0"
```
morenol, line 3: I think that a patch update is ok for this change, (0.11.1)

crates/fluvio-extension-common/Cargo.toml
```diff
@@ -27,5 +27,5 @@ futures-lite = { version = "1.7.0" }
 thiserror = "1.0.20"
 semver = { version = "1.0.0", features = ["serde"] }
 
-fluvio = { version = "0.11.0", path = "../fluvio", optional = true }
+fluvio = { version = "0.11.1", path = "../fluvio", optional = true }
```
morenol, line 30: Right now the CI, is failing because of this, I think that this change is needed if we are updating fluvio to 0.12.0, but since fluvio is still 0.11.X, we dont need to update the dependency here

bohlmannc: After talking with Nick C, the changes to the ConsumerConfig were undone, making a version change unnecessary as no changes are made to fluvio crate now.

crates/fluvio/Cargo.toml
bohlmannc: After talking with Nick C, the changes to the ConsumerConfig were undone, making a version change unnecessary as no changes are made to fluvio crate now.

bohlmannc:
After discussing with Nick C, simplifying the code and not making and changes to the ConsumerConfig in the fluvio crate made this PR much simpler. The logic in mod.rs didn't change that much as compared to prior discussions.

morenol (approved):
lgtm, and it is working as expected. 

@bohlmannc Thank you

sehz:
Great.  Seems like there are some issues with GH actions

PR #2028, opened 2021-12-17
[Merged by Bors] - Return base offset in partition produce response

### LINKED ISSUE #2025: Partition response does not incluse the right base offset
Currently partition producer handler does not set the base offset in partition response so that is getting the default value (0)

 https://github.com/infinyon/fluvio/blob/5e7c0f73f37734ad582d3e1a36bc8413ee3ad216/crates/fluvio-spu/src/services/public/produce_handler.rs#L73

morenol:
Get current LEO and return that as base offset in write_record_set function, so we can use that value in partition produce response

Fixes #2025


crates/fluvio-spu/src/storage/mod.rs
```diff
@@ -145,6 +145,9 @@ where
         );
 
         let mut writer = self.write().await;
+
+        let base_offset = writer.get_leo();
```
sehz, line 149: is this correct? `leo` is not base offset

sehz (changes requested):

PR #2030, opened 2021-12-21
[Merged by Bors] - Remove duplicated version of trybuild. Fixes #2029

### LINKED ISSUE #2029: [Bug]: found duplicate version of package `trybuild v1.0.42` vendored from two sources
**What happened**
trying to vendor the source does not work because of a duplicate dependency.

**Expected behavior**
Trying to package this for Nixos (a linux distribution), the most common way to package rust executables is by vendoring depedencies. It would make things much easier if dependencies could be vendored.

**Describe the setup**
- Just trying to build fluvio-cli from source

1. Run `cargo vendor` at the root of the directory to reproduce the failure

 - OS: Mac
 - Fluvio Version: 0.9.15

**Additional context**
Like I was saying, this is in the context of trying to package this for a linux distro (albeit a minor one). If something can be done to vendor the sources, it makes it a lot easier. I would perfectly understand if nothing can be done.

morenol:
Fixes #2029

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ VERSION
 .scripts
 .env
 .vscode/
+vendor/
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1823,7 +1823,7 @@ dependencies = [
  "quote",
  "syn",
  "tracing",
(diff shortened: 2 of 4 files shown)


sehz (approved):
Lgtm

PR #2079, opened 2022-01-07
[Merged by Bors] - CI - Handle test panics without SIGTERM

### LINKED ISSUE #2042: CI says "Canceled" rather than "Failed" due to Panic Handler
CI failed but said it was actually canceled for unclear reasons.
This was the main part of the stacktrace:
```
Creating the topic: test
topic "test" created
Test panicked: PanicInfo {
About to start producer test
    payload: Any { .. },
    message: Some(
        called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" },
    ),
    location: Location {
        file: "crates/fluvio-test/src/tests/smoke/mod.rs",
        line: 261,
        col: 48,
    },
}
+--------------+--------------+
| Test Results |              |
+--------------+--------------+
| Pass?        | false        |
+--------------+--------------+
| Duration     | 197.453274ms |
+--------------+--------------+

found topic: test offset: 0
found topic: test offset: 0
make: *** [Makefile:74: smoke-test] Terminated
```

image

With my particular bug, it was because I moved a file and forgot to update the path. In any case, this should have said "Failed".

tjtelan:
* Added a new `fluvio-test` `self_check` for self-diagnostic usage
* Log collections additionally occur when workflow is cancelled instead of just on failure.
* Handle `fluvio-test` panic with file-based flags to compensate for Github runner signal handling changes
* Minor fix to `action-slack` to hopefully fix missing fields in Slack messages

Resolves #2042


sehz (approved):
LGTM.

later, we should consolidate slack flags into common fields.

sehz:
Good job figuring out complex interactions between panic, processes, and GitHub environment!

PR #2136, opened 2022-01-24
[Merged by Bors] - Add check in `CD_Release` for `Release` success

### LINKED ISSUE #1681: `CD_Release` job should check on the success of `Release`
Just like in #1613, we need to add a check in the `CD_Release` job to look at the last `Release` run. We only need the workflow to run when the last Release is successful.

tjtelan:
Only run tests if the most recent Release job passed

Resolves #1681


sehz (approved):
LGTM

PR #2150, opened 2022-01-31
[Merged by Bors] - feat: translate _ in connector parameters keys to -

### LINKED ISSUE #2149: translate _ to - in connector parameters
Currently, in the connector controller we call the parameters as they come from the config file https://github.com/infinyon/fluvio/blob/e655710bcede6933d021b697b048dd8025d1c47b/crates/fluvio-sc/src/k8/controllers/managed_connector_deployment.rs#L228

So, if we have as parameters:
```yaml
parameters:
  some_parameter: foo
```
it is called as `--some_parameter` instead of `--some-parameter`.
To fix that, we should translate all _ to - in the parameter key.

morenol:
Also remove the `.clone()` calls in that line

Fixes #2149


sehz:
@simlay please review

simlay (approved):
lgtm

PR #2156, opened 2022-02-03
[Merged by Bors] - Ignore records larger than requested offset

### LINKED ISSUE #2002: [Bug]: Incorrect consumer offset when producer uses `send_all`
**What happened**
After producing records to a topic using `producer.send_all`, a consumer reading from `Offset::from_end(1)` receives the record at the beginning of the latest batch, rather than the latest record.

**Expected behavior**
I expect consuming from `Offset::from_end(1)` should give me the latest record from the latest batch.

**Describe the setup**
- Are you using a local Fluvio install? Minikube? Fluvio Cloud?
  - minikube
- What version of Fluvio are you using? `fluvio version`

```
Fluvio CLI           : 0.9.14
Fluvio CLI SHA256    : c4b27f5f5747932169852461272b009472fece31a3cf4cfdfacaaac33176d21f
Fluvio Platform      : 0.9.14 (minikube)
Git Commit           : 5d8e9e306f75a66393209d5a5076182564cd9ca3
OS Details           : Darwin 11.4 (kernel 20.5.0)
=== Plugin Versions ===
Fluvio Runner (fluvio-run)     : 0.0.0
Infinyon Cloud CLI (fluvio-cloud) : 0.1.6
```

- Create a crate with the following `Cargo.toml` and `src/main.rs`

```toml
[package]
name = "offset-bug"
version = "0.1.0"
edition = "2021"

[dependencies]
fluvio = "0.10"
tokio = { version = "1", features = ["full" ] }
futures-util = "*"
```

```rust
use fluvio::{Fluvio, RecordKey, Offset};
use futures_util::StreamExt;

type Result<T, E = Box<dyn std::error::Error + Send + Sync + 'static>> = core::result::Result<T, E>;

#[tokio::main]
async fn main() -> Result<()> {
    let fluvio = Fluvio::connect().await?;
    let producer = fluvio.topic_producer("test").await?;
    let consumer = fluvio.partition_consumer("test", 0).await?;

    // Write five batches of ten records
    for b in 0..5 {
        let numbers = (0..10).map(|i| (RecordKey::NULL, ((b * 10) + i).to_string()));
        producer.send_all(numbers).await?;
    }

    let mut stream = consumer.stream(Offset::from_end(1)).await?;
    let result: Option<core::result::Result<_, _>> = stream.next().await;
    let record = result.unwrap().unwrap();
    let string = String::from_utf8_lossy(record.value()).to_string();

    println!("Record at Offset::from_end(1) is {}", string);

    Ok(())
}
```

Then run:

- `fluvio topic create test`
- `cargo run`
  - Output is `Record at Offset::from_end(1) is 40`
- Check contents of topic: `fluvio consume test -B`
  - See records numbered 0-49 in order, therefore latest record should be 49

[diagnostics-2021-12-08-16-29-24.tar.gz](https://github.com/infinyon/fluvio/files/7679080/diagnostics-2021-12-08-16-29-24.tar.gz)

morenol:
Fixes #2002 

Currently, when we do a stream request with a given offset to the SPU, in the SPU response we get the batch where the record with that particular offset is. Since SPU does not do any batch manipulation and send the raw bytes of the batch that it gets from the file, we need to handle this on the client side, by skipping the records of the first batch that are greater than the offset that we want.

This PRs do that, it does not change any public API, just change the internal request_stream function to also return the offset of the expected starting offset so we can use that to filter all the records with an offset greater than that.

Added a CLI smoke test for --tail 1

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1491,7 +1491,7 @@ dependencies = [
 
 [[package]]
 name = "fluvio"
-version = "0.12.3"
+version = "0.12.4"
 dependencies = [
  "async-channel",
  "async-lock",
--- a/crates/fluvio/Cargo.toml
+++ b/crates/fluvio/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio"
-version = "0.12.3"
+version = "0.12.4"
 edition = "2021"
 license = "Apache-2.0"
 authors = ["Fluvio Contributors <team@fluvio.io>"]
--- a/crates/fluvio/src/consumer.rs
+++ b/crates/fluvio/src/consumer.rs
@@ -151,25 +151,31 @@ where
         offset: Offset,
         config: ConsumerConfig,
     ) -> Result<impl Stream<Item = Result<Record, ErrorCode>>, FluvioError> {
-        let stream = self.stream_batches_with_config(offset, config).await?;
+        let (stream, start_offset) = self
+            .inner_stream_batches_with_config(offset, config)
+            .await?;
         let partition = self.partition;
-        let flattened =
-            stream.flat_map(move |result: Result<Batch, _>| match result {
-                Err(e) => Either::Right(once(err(e))),
-                Ok(batch) => {
-                    let base_offset = batch.base_offset;
-                    let records = batch.own_records().into_iter().enumerate().map(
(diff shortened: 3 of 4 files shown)


sehz (approved):
LGTM

PR #2160, opened 2022-02-06
[Merged by Bors] - Add `wasi` support to smartengine

### LINKED ISSUE #1874: Debugging: Support logging in the WASM module
Either using WASI or some other API to provide logging.

galibey:
Added `WASI` support to `smartengine` that allows user's code to have access to `stdout` and `stderr` for debug purposes. 
It's guarded by feature `wasi` and disabled by default.
User's code has to be compiled against `wasm32-wasi` target.

Fixes #1874

### DIFF
--- a/makefiles/check.mk
+++ b/makefiles/check.mk
@@ -44,4 +44,4 @@ run-client-doc-test: install_rustup_target
 
 
 fluvio_run_bin: install_rustup_target
-	cargo build --bin fluvio-run $(RELEASE_FLAG) --target $(TARGET)
+	cargo build --bin fluvio-run --features wasi $(RELEASE_FLAG) --target $(TARGET)
--- a/makefiles/check.mk
+++ b/makefiles/check.mk
@@ -44,4 +44,4 @@ run-client-doc-test: install_rustup_target
 
 
 fluvio_run_bin: install_rustup_target
-	cargo build --bin fluvio-run --features wasi $(RELEASE_FLAG) --target $(TARGET)
+	cargo build --bin fluvio-run $(RELEASE_FLAG) --target $(TARGET) $(DEBUG_SMARTMODULE_FLAG)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Release Notes
 
 ## Platform Version 0.9.19 - UNRELEASED
+* Add WASI support to SmartEngine ([#1874](https://github.com/infinyon/fluvio/issues/1874))
 
 ## Platform Version 0.9.18 - 2022-01-31
 * Show Platform version for fluvio run ([#2104](https://github.com/infinyon/fluvio/issues/2104))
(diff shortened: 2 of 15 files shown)


sehz:
Thanks for PR!  There is one Clippy error:  https://github.com/infinyon/fluvio/runs/5085332885?check_suite_focus=true

sehz:
@morenol can you review this as well?

sehz:
This is a step toward adding debugging the WASM module but not the complete solution.   In order to have a complete solution (which can be done as another PR or issue)
* Build Fluvio with WASI feature flag outputting stderr to log (for a temporary workaround)
* Implement custom WASI to output logging to fluvio topic (this is probably a better long-term solution).

morenol (changes requested):
nice work! 

It is working as expected. 

Please fix the clippy issue and add an entry into the changelog.

Also, I think that we should update the makefiles to optionally use the wasi feature while we run:

https://github.com/infinyon/fluvio/blob/bcda4fe105c7bca5ed9bb01fd91c8d9f54acfde5/makefiles/check.mk#L53 and https://github.com/infinyon/fluvio/blob/bcda4fe105c7bca5ed9bb01fd91c8d9f54acfde5/makefiles/build.mk#L11

galibey:
> * Implement custom WASI to output logging to fluvio topic (this is probably a better long-term solution).
This is sounds like it deserves own API so it would be more declarative for users. Many people would be surprised that data will be sent over network when they do simple `println` or `dbg` and a lot of things might get wrong during that process. So, in my opinion, some function-like proc macro or function which explicitly says about it looks nice here. Also, there will be room for tweaking and documenting behaviour whereas if WASI method will be overridden to handle it, it would feel like magic.

makefiles/check.mk
```diff
@@ -44,4 +44,4 @@ run-client-doc-test: install_rustup_target
 
 
 fluvio_run_bin: install_rustup_target
-	cargo build --bin fluvio-run $(RELEASE_FLAG) --target $(TARGET)
+	cargo build --bin fluvio-run --features wasi $(RELEASE_FLAG) --target $(TARGET)
```
morenol, line 47: We don't want to build with this feature by default. I think that we can add here $(DEBUG_SMARTMODULE_FLAG) just like we do with `RELEASE_FLAG`.

We can define `DEBUG_SMARTMODULE_FLAG` in the Makefile in the root of the repo with something like 

```bash
DEBUG_SMARTMODULE_FLAG=$(if $(DEBUG_SMARTMODULE),--features wasi,)
```

The same applies to `makefiles/build.mk`

morenol (changes requested):
just one last detail

makefiles/check.mk
galibey: Sorry, never dealt with Makefiles. Done now.

morenol (approved):
LGTM!

sehz (approved):
LGTM.

Nice Job!

galibey:
Thanks for review, @morenol @sehz!

Issue #2168, opened 2022-02-08
[CLI]: channel config populated with invalid version when using version channel


LanceEa:
**What happened**
When using the command `fluvio version create 123` an error will be thrown because the version cannot be resolved.  Despite not being a valid version the `channel` config file is being updated with the channel key as seen here:

```toml
current_channel = "stable"
[channel.123]
binary_location = "/Users/lance/.fluvio/bin/fluvio-123"
extensions = "/Users/lance/.fluvio/bin/extensions-123"
image_tag_strategy = "Version"

[channel.stable]
binary_location = "/Users/lance/.fluvio/bin/fluvio-stable"
extensions = "/Users/lance/.fluvio/extensions"
image_tag_strategy = "Version"

[channel."0.9.18"]
binary_location = "/Users/lance/.fluvio/bin/fluvio-0.9.18"
extensions = "/Users/lance/.fluvio/bin/extensions-0.9.18"
```

Also, running `fluvio version list` shows the invalid version:

```bash
fluvio version list
  0.9.18
  123
* stable
```

**Expected behavior**

I would expect the resolution of the version to happen prior to making any changes to the channel config so as not to pollute the available versions.

**Describe the setup**
- local fluvio cli installed
- `fluvio version`
Release Channel      : stable
Fluvio CLI           : 0.9.18
Fluvio CLI SHA256    : f97262eec5fcb1f0b05c3cd6d1a7cdf40cfe83d1c5b702be49b82aa4575f409e
Fluvio Platform      : 0.9.18 (kind-fluvio)
Git Commit           : 2b6397f56e3935807ebb9e84c60517f815426d51
OS Details           : Darwin 12.1 (kernel 21.2.0)
=== Plugin Versions ===
Fluvio Runner (fluvio-run)     : 0.0.0
Infinyon Cloud CLI (fluvio-cloud) : 0.1.6

**How to reproduce it (as minimally and precisely as possible)**
Steps to reproduce the behavior on Mac:
1. Run the command `fluvio version create 1234`
2. An unresolved version error message will be shown
3. cat ~/.fluvio/channel
4. run the command `fluvio version list`

You will notice the bad version has been written to the config file

**Environment (please complete the following information):**
 - OS: Mac M1
 - Fluvio Version:  0.9.18

rrrodzilla:
[...]

PR #2170, opened 2022-02-09
[Merged by Bors] - Resolve issue_2168 with regression test

rrrodzilla:
- add: dev dependencies predicates, assert_cmd, rand
- fix: issue_2168

### DIFF
--- a/crates/fluvio-channel-cli/tests/issue_2168_regression.rs
+++ b/crates/fluvio-channel-cli/tests/issue_2168_regression.rs
@@ -15,7 +15,7 @@
 // just in case the version was indeed written to the file.
 // This should never succeed and we assert that.
 #[test]
-fn issue_2168_regression_test() -> Result<(), Box<dyn std::error::Error>> {
+fn issue_2168_regression_test() {
     use assert_cmd::prelude::*; // Add methods on commands
     use predicates::prelude::*; // Used for writing assertions
     use std::process::Command; // Run programs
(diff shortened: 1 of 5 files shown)


sehz:
@tjtelan Can you review this PR?

tjtelan (changes requested):
I needed to change one line in `fluvio-channel-cli`'s `Cargo.toml` to get the test to compile. Could you make that change please?

```diff
- fluvio-cli-common = { path = "../fluvio-cli-common", default-features = false }
+ fluvio-cli-common = { path = "../fluvio-cli-common"}
```

Also, thanks for writing a test! Could you add this test to [this makefile, in the `run-all-unit-test` target](https://github.com/rrrodzilla/fluvio/blob/7137d22309b50e998e5b2a153f6fccf8911818b3/makefiles/check.mk#L28) so we can run it during CI?

```diff
+ cargo test -p fluvio-channel-cli $(BUILD_FLAGS)
```

Otherwise I manually tested and it fixes issue #2168 ✅

LGTM!

rrrodzilla:
Hey thanks for having a review.  I've made the requested changes with this last commit.  Thanks again!

crates/fluvio-channel-cli/tests/issue_2168_regression.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    //verify our test channel name is not currently in the channel config
+    let mut cmd = Command::cargo_bin("fluvio-channel")?;
+    cmd.arg("version").arg("list");
+    cmd.assert()
+        .success()
+        .stdout(predicate::str::contains(&random_channel_name).count(0));
+
+    //if we get this far, we can attempt to create the channel version
+    let mut cmd = Command::cargo_bin("fluvio-channel")?;
```
sehz, line 48: same here

sehz (changes requested):

rrrodzilla:
Updates to the test with the requested changes pushed!

rrrodzilla (commented):
I've updated the test to handle errors internally with `expect`

sehz (approved):

sehz:
There is test failure: https://github.com/infinyon/fluvio/runs/5133172297?check_suite_focus=true

rrrodzilla:
@tjtelan The test running from the makefile is expecting a channel config file in the CI environment and it's not finding one.  Is there not one or should I expect it in a different location?

tjtelan:
> @tjtelan The test running from the makefile is expecting a channel config file in the CI environment and it's not finding one. Is there not one or should I expect it in a different location?

Yeah, the CI environment wouldn't have a channel file because we are only using the standard github runners. You'll need to create the config file when in the CI environment before running the test.

You can do a CI check by looking for the [`CI` environment variable](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables), which Github Actions will always set to `true`.

If that env var exists and it is set to `true`, you can run `fluvio-channel version switch stable`. I just tested locally, and it will create the `~/.fluvio` directory and the `channel` config for the rest of the test to use.

tjtelan (approved):

sehz:
Good job.  Thanks for contribution!

PR #2175, opened 2022-02-09
[Merged by Bors] -  Reimport metadata for tableformat in fluvio client 

### LINKED ISSUE #2176: Add tableformat metadata to fluvio::metadata
Currently, fluvio client does not export metadata for TableFormat

morenol:
Fixes #2176

Required by https://github.com/infinyon/fluvio-client-wasm/issues/246

### DIFF
--- a/crates/fluvio/src/lib.rs
+++ b/crates/fluvio/src/lib.rs
@@ -240,6 +240,10 @@ pub mod metadata {
         pub use fluvio_sc_schema::objects::*;
     }
 
+    pub mod tableformat {
+        pub use fluvio_sc_schema::tableformat::*;
+    }
+
     pub mod core {
         pub use fluvio_sc_schema::core::*;
     }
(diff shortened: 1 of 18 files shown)


crates/fluvio/src/lib.rs
```diff
@@ -240,6 +240,10 @@ pub mod metadata {
         pub use fluvio_sc_schema::objects::*;
     }
 
+    pub mod tableformat {
+        pub use fluvio_sc_schema::tableformat::*;
```
sehz, line 244: need to bump up fluvio version.  also add to changelog

sehz (changes requested):

crates/fluvio/src/lib.rs
morenol: The version as already bumped in a previous PR, I will add to the changelog

morenol: do you mean minor bump?

sehz: Yes

crates/fluvio/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio"
-version = "0.12.4"
+version = "0.13.0"
```
sehz, line 3: is there need to bump to full version rather than patch

sehz (changes requested):

crates/fluvio/Cargo.toml
morenol: I think that no, but I will get back to the previous version, but we dont need to patch more because 0.12.4 is not released yet

sehz (approved):
LGTM

PR #2187, opened 2022-02-11
[Merged by Bors] - fluvio-test data generator

### LINKED ISSUE #2129: fluvio-test - Data generator
We need to be able to configure:
- \# producers
- batching options
- test payload size
- run duration
  - This is mostly for automation
  - But the default duration being infinite. Otherwise end when the user exits test.
  - Test should include user input for the purpose of exiting cleanly. Press ESC or something
- (stretch goal) Rate limiting
- (stretch goal) Multiple data output characteristics (e.g., continuous, periodic, producer ramp-up, size ramp-up, random)

This will build out some foundation I need for performance testing, but we will not be doing any performance testing here. The producers will generate data prior to running, and we want all the producers to start at the same time. We will wait for the producers to signal when they are ready to start. The synchronization of the producers will be using Fluvio topics as an event bus.

Also needs to support the possibility of multiple instances of this test running against the same cluster.

tjtelan:
Resolves #2129 

Add test `generator`. Uses a topic as event bus to synchronize all producers (as foundation for perf testing)

```
FLAGS:
    -h, --help         Prints help information
        --multi-run    When set, unique topics will be used so multiple instances of generator against cluster won't
                       collide
    -V, --version      Prints version information
    -v, --verbose      Opt-in to detailed output printed to stdout

OPTIONS:
        --linger <batch-linger-ms>             Producer linger (ms)
        --batch-size <batch-size>              Producer batch size (bytes)
        --producers <producers>                Number of producers to create [default: 1]
        --record-size <record-size>            Record payload size used by test (bytes) [default: 1000]
        --runtime-seconds <runtime-seconds>    Max time we want the producer to run, in seconds - Default: forever
        --topics <topics>                      Number of topics for producers to send traffic to [default: 1]
```

Additional flag `--multi-run` provided for the case of running multiple instances of generator against the same cluster w/o collision.


sehz (approved):
LGTM

sehz:
clippy

tjtelan:
bors ping

morenol:
bors retry

Issue #2154, opened 2022-02-03
Connector:  Update config command


sehz:
The connector should restart to pick up the new configuration.

The new command should be in form of:
```
$ fluvio connector update -c <config_file_name>
```

This should update the existing connector configuration and re-deploy connector pods.

PR #2188, opened 2022-02-11
[Merged by Bors] - feat(connector): update

pinkforest:
Closing #2154 (see caveats)

Adds fluvio connector update -c config_file

### DIFF
--- a/crates/fluvio-cli/src/connector/update.rs
+++ b/crates/fluvio-cli/src/connector/update.rs
@@ -66,10 +66,6 @@ impl UpdateManagedConnectorOpt {
                 .await
             {
                 Err(FluvioError::AdminApi(ApiError::Code(ErrorCode::TopicAlreadyExists, _))) => {
-                    debug!(
-                        ?replica_spec,
-                        "UpdateManagedConnectorOpt topic spec - Re-Using"
-                    );
                     Ok(())
                 }
                 Ok(_) => {
--- a/crates/fluvio-cli/src/connector/update.rs
+++ b/crates/fluvio-cli/src/connector/update.rs
@@ -1,0 +1,69 @@
+//!
+//! # Update a Managed Connector
+//!
+//! CLI tree to generate Update a Managed Connector
+//!
+
+use fluvio::metadata::topic::ReplicaSpec;
+use structopt::StructOpt;
+use tracing::debug;
+
+use fluvio::{Fluvio, FluvioError};
+use fluvio::metadata::{
+    topic::{TopicSpec, TopicReplicaParam},
+    connector::ManagedConnectorSpec,
+};
+
+use fluvio_sc_schema::ApiError;
+use fluvio_sc_schema::errors::ErrorCode;
+
+use crate::CliError;
+use super::ConnectorConfig;
+
+// -----------------------------------
+// CLI Options
+// -----------------------------------
+
+#[derive(Debug, StructOpt, Default)]
+pub struct UpdateManagedConnectorOpt {
+    /// The name for the update Managed Connector
+    #[structopt(short = "c", long = "config", value_name = "config")]
+    pub config: String,
+}
+
+impl UpdateManagedConnectorOpt {
+    pub async fn process(self, fluvio: &Fluvio) -> Result<(), CliError> {
+        let config = ConnectorConfig::from_file(&self.config)?;
+        let spec: ManagedConnectorSpec = config.clone().into();
+        let name = spec.name.clone();
+
+        debug!("updating managed_connector: {}, spec: {:#?}", name, spec);
+
+        let admin = fluvio.admin().await;
+
+/* We are going to handle it better than below crude delete/create restart :)
+
+       admin.delete::<ManagedConnectorSpec, _>(&self.name).await?;
+       if config.create_topic {
+            let replica_spec = ReplicaSpec::Computed(TopicReplicaParam::new(1, 1, false));
+            debug!("topic spec: {:?}", replica_spec);
+            match admin
+                .create::<TopicSpec>(config.topic, false, replica_spec.into())
+                .await
+            {
+                Err(FluvioError::AdminApi(ApiError::Code(ErrorCode::TopicAlreadyExists, _))) => {
+                    //println!("Topic already exists");
+                    Ok(())
+                }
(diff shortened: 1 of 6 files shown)


crates/fluvio-cli/src/connector/update.rs
```diff
@@ -0,0 +1,87 @@
...
+        };
+
+        debug!("Found connector: {:?}", existing_config);
+
+        // create_topic will succeed even if the topic already exists. PR: 1823
+        // A New Topic might have been defined - We will not delete the old one.
+        if config.create_topic {
+            // Check if the topic already exists by trying to create it
+            let replica_spec = ReplicaSpec::Computed(TopicReplicaParam::new(1, 1, false));
+            debug!("UpdateManagedConnectorOpt topic spec: {:?}", replica_spec);
```
morenol, line 63:
```suggestion
            debug!(?replica_spec, "UpdateManagedConnectorOpt topic spec");
```

crates/fluvio-cli/src/connector/update.rs
```diff
@@ -0,0 +1,87 @@
...
+use crate::CliError;
+use super::ConnectorConfig;
+
+// -----------------------------------
+// CLI Options
+// -----------------------------------
+
+#[derive(Debug, StructOpt, Default)]
+pub struct UpdateManagedConnectorOpt {
+    /// The name for the update Managed Connector
```
morenol, line 29: I think that we should update this and also for the create option
```suggestion
    /// The filename with the configurations to update Managed Connector
```

morenol (commented):
just some small nits

crates/fluvio-cli/src/connector/update.rs
```diff
@@ -0,0 +1,87 @@
...
+                    Ok(())
+                }
+                Err(e) => Err(e),
+            }?;
+        }
+
+        admin
+            .delete::<ManagedConnectorSpec, _>(&config.name)
+            .await?;
+        admin.create(config.name, false, spec).await?;
```
simlay, line 83: Hmm. So, this takes a new config, delete's the one with the same name and then recreates it. While technically this does update the `ManagedConnectorSpec`, if we want resume-from-last-offset using persistent storage, this won't work as the pod will be deleted.

I think we might need an [`UpdateRequest` in the public API](https://github.com/infinyon/fluvio/blob/0c5560a3fa1f6e024896dd0801878d16fa5150d1/crates/fluvio-sc/src/services/public_api/public_server.rs#L80-L119). @sehz What do you think?

sehz: Or we can rename, `CreateRequest` as `CreateOrUpdateRequest`.   This can be mapped to Kubernetes `apply`.

simlay: I'm okay with that. The only thing there is that all our `CreateRequests` would then support updates where as now if you try and create the same topic, it errors saying "topic already exists".

crates/fluvio-cli/src/connector/update.rs
```diff
@@ -0,0 +1,87 @@
...
+            None => {
+                return Err(CliError::InvalidConnector(format!(
+                    "{} - does not exist.",
+                    config.name
+                )))
+            }
+            Some(c) => c,
+        };
+
+        debug!("Found connector: {:?}", existing_config);
```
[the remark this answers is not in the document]
pinkforest: I don't like using ?pattern in as it makes it harder to read.

crates/fluvio-cli/src/connector/update.rs
pinkforest: We decided to add update request into API

crates/fluvio-cli/src/connector/update.rs
```diff
@@ -0,0 +1,87 @@
...
+    pub config: String,
+}
+
+impl UpdateManagedConnectorOpt {
+    pub async fn process(self, fluvio: &Fluvio) -> Result<(), CliError> {
+        let config = ConnectorConfig::from_file(&self.config)?;
+        let spec: ManagedConnectorSpec = config.clone().into();
+        let name = spec.name.clone();
+
+        debug!(connector_name = %name, connector_spec = >spec "updating managed_connector");
```
morenol, line 40: I think that I had a typo here, it should be
```suggestion
        debug!(connector_name = %name, connector_spec = ?spec "updating managed_connector");
```

simlay (approved):
bors r+

morenol:
bors r-

I just added a new comment because I think that there is a typo

crates/fluvio-cli/src/connector/update.rs
```diff
@@ -0,0 +1,87 @@
...
+    pub config: String,
+}
+
+impl UpdateManagedConnectorOpt {
+    pub async fn process(self, fluvio: &Fluvio) -> Result<(), CliError> {
+        let config = ConnectorConfig::from_file(&self.config)?;
+        let spec: ManagedConnectorSpec = config.clone().into();
+        let name = spec.name.clone();
+
+        debug!(connector_name = %name, connector_spec = ?spec "updating managed_connector");
```
morenol, line 40:
```suggestion
        debug!(connector_name = %name, connector_spec = ?spec, "updating managed_connector");
```

crates/fluvio-cli/src/connector/update.rs
```diff
@@ -0,0 +1,87 @@
...
+        if config.create_topic {
+            // Check if the topic already exists by trying to create it
+            let replica_spec = ReplicaSpec::Computed(TopicReplicaParam::new(1, 1, false));
+            debug!(?replica_spec, "UpdateManagedConnectorOpt topic spec");
+            match admin
+                .create::<TopicSpec>(config.topic, false, replica_spec.into())
+                .await
+            {
+                Err(FluvioError::AdminApi(ApiError::Code(ErrorCode::TopicAlreadyExists, _))) => {
+                    debug!(?replica_spec, "UpdateManagedConnectorOpt topic spec - Re-Using")
```
morenol, line 69:
```suggestion
                    debug!(?replica_spec, "UpdateManagedConnectorOpt topic spec - Re-Using");
```

morenol:
bors retry

PR #2189, opened 2022-02-12
Add error message on 'fluvio connector logs <name>' when request connector is undefined

### LINKED ISSUE #2161: Connector log from an invalid connector should return an error
No connectors configured:

```
$ fluvio connector list
no managed connectors
```

Use log command:

```
$ fluvio connector logs foo
```

The command just returns, it should say:
`Error: Connector foo does not exist`

Afourcat:
- An error message was missing when requesting logs from an undefined connector.

```bash
$ fluvio connector logs test-undefined
Error: 
   0: Invalid argument: Connector test-undefined does not exist
```
- Fix bug where we could see logs of all pods not only connectors
- Fix bug where we could match with the wrong pod

Fixes #2161

### DIFF
--- a/crates/fluvio-cli/src/connector/logs.rs
+++ b/crates/fluvio-cli/src/connector/logs.rs
@@ -36,7 +36,9 @@ impl LogsManagedConnectorOpt {
             .stdout;
 
         let pods = String::from_utf8_lossy(&pods);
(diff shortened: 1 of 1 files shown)


sehz:
@simlay Please review this change and confirm this works

simlay (approved):
Thanks for the PR! I've confirmed this works as expected.

PR #2192, opened 2022-02-14
[Merged by Bors] - docs: clarify that `send()` does not `flush()` automatically in producer documentation

### LINKED ISSUE #2191: awaiting `TopicProducer.send` in the rust client does not automatically flush message to server
**What happened**
When following the Hello-World Rust example it suggests creating a TopicProducer, sending and awaiting within an async message. However, it appears that the message is not flushed to the topic.

Screen Shot 2022-02-13 at 3 41 58 PM

As you can see from the screenshot the consumer in the right terminal is just waiting to consume messages and running the sample application doesn't produce a message on the topic.

After some debugging and playing around I noticed that if I sleep on the thread for a second, wait for the RecordMetadata or make an explicit call to the `TopicProducer.flush` then it will successfully send the messages. I attempted to send 3 messages but only two were consumed.

Screen Shot 2022-02-13 at 3 55 59 PM

Here is the sample code based on the HelloWorld Rust docs:

```rust
use fluvio::FluvioError;

#[async_std::main]
async fn main() {
    let result = produce("Hello", "ok is it working now").await;
    println!("{:?}", result);

    let result = produce_metadata_wait("RecordMetadata", "i waited for you").await;
    println!("{:?}", result);

    let result = produce_flush("Flush", "bye bye message").await;
    println!("{:?}", result);

    println!("That is all folks");
}

async fn produce(key: &str, value: &str) -> Result<(), FluvioError> {
    let producer = fluvio::producer("hello-fluvio").await?;
    producer.send(key, value).await?;

    Ok(())
}

async fn produce_metadata_wait(key: &str, value: &str) -> Result<(), FluvioError> {
    let producer = fluvio::producer("hello-fluvio").await?;
    let _record_metadata = producer.send(key, value).await?.wait().await?;

    Ok(())
}

async fn produce_flush(key: &str, value: &str) -> Result<(), FluvioError> {
    let producer = fluvio::producer("hello-fluvio").await?;
    producer.send(key, value).await?.wait().await?;
    producer.flush().await?;

    Ok(())
}
``` 

**Expected behavior**
First, is this expected behavior or a bug? If it is not a bug and expected behavior then below would be my expected behavior.

As new comer to both Rust and Fluvio, the fact that  awaiting `TopicProducer.send` does not explicitly flush the record tripped me up. The TopicProducer created in the `fn produce` is being dropped when it returns the result which seems to be occurring prior to being flushed to the topic. 

1. I would have expected the HelloWorld Rust docs to just work without needing to know about flush or else it would have called it out.
2. In preparing this issue, I stumbled upon the Send docs found here so https://www.fluvio.io/api/fluvio/rust/#send. I have seen this same pattern with Ubers Zap Golang logging library where we ensure that it is flushed on graceful shutdown so its not out of ordinary. However, I guess I would have thought it would auto-flush when Drop is called on the TopicProducer(???).
3. To improve the developer experience it would be good to document the need for calling `flush` in the rust comment docs for the function. This way as a developer I wouldn't have to go searching for the api docs on the fluvio website and could see directly in my IDE.

Screen Shot 2022-02-13 at 4 09 18 PM

**Describe the setup**
- local Fluvio install against a local Kinds cluster

```bash
Release Channel      : stable
Fluvio CLI           : 0.9.19
Fluvio CLI SHA256    : fb4ff5949eb8883f65bb3fdf2994ce49652d9fbddebe2146f6ece49994d0a231
Fluvio Platform      : 0.9.19 (kind-fluvio)
Git Commit           : db2cbffabb657821180173dfaecedc906897baf2
OS Details           : Darwin 12.1 (kernel 21.2.0)
=== Plugin Versions ===
Fluvio Runner (fluvio-run)     : 0.0.0
Infinyon Cloud CLI (fluvio-cloud) : 0.1.6
```

1. Following the hello-world rust docs as seen here: https://www.infinyon.com/tutorials/rust/hello-world/
2. Include the additional functions provided in the description above

 - OS: Mac M1
 - Fluvio 0.9.19
 - kind version 0.11.1
 - kubectl version

```Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.2", GitCommit:"9d142434e3af351a628bffee3939e64c681afa4d", GitTreeState:"clean", BuildDate:"2022-01-19T17:27:51Z", GoVersion:"go1.17.6", Compiler:"gc", Platform:"darwin/arm64"}
Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.1", GitCommit:"5e58841cce77d4bc13713ad2b91fa0d961e69192", GitTreeState:"clean", BuildDate:"2021-05-21T23:06:30Z", GoVersion:"go1.16.4", Compiler:"gc", **Platform:"linux/arm64"}**
```

morenol:
Fixes #2191

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1597,7 +1597,7 @@ dependencies = [
 
 [[package]]
 name = "fluvio"
-version = "0.12.4"
+version = "0.12.5"
 dependencies = [
  "async-channel",
  "async-lock",
--- a/crates/fluvio/Cargo.toml
+++ b/crates/fluvio/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio"
-version = "0.12.4"
+version = "0.12.5"
 edition = "2021"
 license = "Apache-2.0"
 authors = ["Fluvio Contributors <team@fluvio.io>"]
--- a/crates/fluvio/src/lib.rs
+++ b/crates/fluvio/src/lib.rs
@@ -96,7 +96,10 @@ pub mod config;
 use tracing::instrument;
 pub use error::FluvioError;
 pub use config::FluvioConfig;
-pub use producer::{TopicProducerConfigBuilder, TopicProducerConfig, TopicProducer, RecordKey};
+pub use producer::{
+    TopicProducerConfigBuilder, TopicProducerConfig, TopicProducer, RecordKey, ProduceOutput,
+    FutureRecordMetadata, RecordMetadata,
+};
 
 pub use consumer::{
     PartitionConsumer, ConsumerConfig, MultiplePartitionConsumer, PartitionSelectionStrategy,
--- a/crates/fluvio/src/producer/mod.rs
+++ b/crates/fluvio/src/producer/mod.rs
@@ -37,7 +37,7 @@ pub use self::error::ProducerError;
 use self::event::EventHandler;
 pub use self::output::ProduceOutput;
 use self::partition_producer::PartitionProducer;
-pub use self::record::FutureRecordMetadata;
+pub use self::record::{FutureRecordMetadata, RecordMetadata};
 
 use crate::error::Result;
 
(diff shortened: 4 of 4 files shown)


sehz (approved):
LGTM

PR #2198, opened 2022-02-16
[Merged by Bors] - feat(connector): deprecate meta flag create_topic

### LINKED ISSUE #2200: Connectors: Redundant topic_create flag
Since the connectors require topic and the topic create is non-destructive action I think it would be safe to deprecate the now redundant create_topic flag?

pinkforest:
Closes #2200
- Deprecate a redundant flag since the topic needs to exist for connectors and it can be safely checked that it exists with create #1823

### DIFF
--- a/crates/fluvio-cli/src/connector/create.rs
+++ b/crates/fluvio-cli/src/connector/create.rs
@@ -51,7 +51,7 @@ impl CreateManagedConnectorOpt {
                 Ok(())
             }
             Ok(_) => Ok(()),
-                Err(e) => Err(e),
+            Err(e) => Err(e),
         }?;
 
         admin.create(name.to_string(), false, spec).await?;
--- a/crates/fluvio-cli/src/connector/create.rs
+++ b/crates/fluvio-cli/src/connector/create.rs
@@ -39,21 +39,21 @@ impl CreateManagedConnectorOpt {
         debug!("creating managed_connector: {}, spec: {:#?}", name, spec);
 
         let admin = fluvio.admin().await;
-        if config.create_topic {
-            let replica_spec = ReplicaSpec::Computed(TopicReplicaParam::new(1, 1, false));
-            debug!("topic spec: {:?}", replica_spec);
-            match admin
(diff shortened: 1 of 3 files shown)


crates/fluvio-cli/src/connector/mod.rs
```diff
@@ -94,8 +94,6 @@ pub struct ConnectorConfig {
     pub(crate) topic: String,
     pub(crate) version: Option<String>,
     #[serde(default)]
-    pub(crate) create_topic: bool,
```
simlay, line 97: Serde just ignores items when they're in the thing still, right?

simlay (approved):
LGTM.

crates/fluvio-cli/src/connector/mod.rs
pinkforest: yep - it's non-exhaustive like structopt is.

If we want to enforce unknown fields deny then we need the below:

```
#[serde(deny_unknown_fields)]

Always error during deserialization when encountering unknown fields. When this attribute is not present, by default unknown fields are ignored for self-describing formats like JSON.

Note: this attribute is not supported in combination with [flatten](https://serde.rs/field-attrs.html#flatten), neither on the outer struct nor on the flattened field.
```

crates/fluvio-cli/src/connector/create.rs
```diff
@@ -39,21 +39,21 @@ impl CreateManagedConnectorOpt {
...
-                Err(FluvioError::AdminApi(ApiError::Code(ErrorCode::TopicAlreadyExists, _))) => {
-                    //println!("Topic already exists");
-                    Ok(())
-                }
-                Ok(_) => Ok(()),
-                Err(e) => Err(e),
-            }?;
-        }
+
+        let replica_spec = ReplicaSpec::Computed(TopicReplicaParam::new(1, 1, false));
```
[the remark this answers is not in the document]
pinkforest: new() ?

wendyOn:
bors retry

wendyOn:
bors retry

Issue #2148, opened 2022-01-31
feature: Show partition size


sehz:
in the CLI, show the size of the data for each partition:
```
$ fluvio partition list
 TOPIC      PARTITION  LEADER  REPLICAS  RESOLUTION  SIZE HW     LEO    LRS  FOLLOWER OFFSETS 
 longevity      0       5001    []        Online      4G  88455  88455  0    []       
```

In order to do this, we need to add size to the status reported by SPU to SC

Afourcat:
Hello,
I was trying to solve this issue, but, I feel a little bit confused, could someone tell me if I am on the right track and/or maybe give me a hint?

First of all, is the **"size"** the space taken, in bytes, by the stored records in the SPU?

## CLI Side
On the CLI, data shown in the `fluvio partition list` output are fetched from here:
https://github.com/infinyon/fluvio/blob/0b991889205213c3a043c9ce7a9d0864b7eb6737/crates/fluvio-cli/src/partition/list.rs#L32 
In the metadata I need to add:
```rust
Metadata {
    name: "cat-facts-3-partionned-4",
    spec: PartitionSpec {
        leader: 0,
        replicas: [
            0,
        ],
        cleanup_policy: None,
        storage: None
    },
    status: PartitionStatus {
        resolution: Online,
        leader: ReplicaStatus {
            spu: 0,
            hw: 5,
            leo: 5,
        },
        lsr: 0,
        size: 240000, // size of the partition in bytes ?
        replicas: [],
        is_being_deleted: false,
    },
},

```

## SPU Side
If I understand well enough, the following function sends back the status from the SPU's Leader to the SC.
https://github.com/infinyon/fluvio/blob/0b991889205213c3a043c9ce7a9d0864b7eb6737/crates/fluvio-spu/src/control_plane/dispatcher.rs#L189-L208

It seems that I have to add a field somewhere, in this `UpdateLrsRequest`, but where exactly and how to fetch partitions' size is unclear.

If I'm right, partitions can be spread across multiple SPUs. So, do I need to gather partitions' size from followers and leaders, and send them to the sc through the `UpdateLrsRequest`?

Or am I completely wrong?

Thank you for your time,

sehz:
Yes, this requires adding sizing API to ReplicaStorage in: `https://github.com/infinyon/fluvio/blob/master/crates/fluvio-storage/src/lib.rs`.

Afourcat:
While implementing the feature, I met a compilation error.
[...]

PR #2214, opened 2022-02-18
[Merged by Bors] - Feature/fluvio cli partition list show size

Afourcat:
Add a human-readable partition size field to the `fluvio partition list` command.

```bash
$ fluvio partition list
 TOPIC      PARTITION  LEADER  REPLICAS  RESOLUTION  SIZE    HW  LEO  LRS  FOLLOWER OFFSETS 
 cat-facts  0          5001    []        Online      6 KB    12  12   0    [] 
 cat-facts  1          5001    []        Online      2.1 KB  10  10   0    [] 
 cat-facts  2          5001    []        Online      4.7 KB  10  10   0    [] 
 cat-facts  3          5001    []        Online      4.6 KB  9   9    0    []
 ```
 Prints `ERROR` when the leader is not able to fetch log size.
 ```bash
 TOPIC      PARTITION  LEADER  REPLICAS  RESOLUTION  SIZE   HW  LEO  LRS  FOLLOWER OFFSETS 
 cat-facts  0          5001    []        Online      ERROR  12  12   0    [] 
 cat-facts  1          5001    []        Online      ERROR  10  10   0    [] 
 cat-facts  2          5001    []        Online      ERROR  10  10   0    [] 
 cat-facts  3          5001    []        Online      ERROR  9   9    0    []
 ```

 
 Force bash shell in the `k8-util/docker/build.sh` to make the script work in a non-bash-compatible environment.
 
 Solves #2148

### CODE REVIEW

**Comment by sehz** [comment]
rebase to resolve conflict

**Comment by Afourcat** [comment]
Hello,
should I bump the version of `fluvio-controlplane-metadata` from
```
version = "0.13.2"
```
to
```
version = "0.14.0"
```
It seems like a breaking change to me, I added a field to the `PartitionStatus` struct.
Thank you,

**Comment by Afourcat** [comment]
Hello again,
Same question for the `fluvio-sc-schema` crate,
Should I bump PATCH or MINOR?

Thanks again,

**Comment by sehz** [comment]
I it's ok to only bump PATCH version since public API has not changed.

**Comment by sehz** [comment]
Upgrade test fails.  To fix this:
* Bump up API version for ObjectApiListRequest in `fluvio-sc-schema`
* Add ` #[fluvio(min_version = <new_version>)] to new field in the `SpuStatus`

**Comment by Afourcat** [comment]
Hello,
Do I need to change the default version, here?

```rust
impl Request for ObjectApiListRequest {
    const API_KEY: u16 = AdminPublicApiKey::List as u16;
    // From: const DEFAULT_API_VERSION: i16 = 4;
    const DEFAULT_API_VERSION: i16 = 5;
    type Response = ObjectApiListResponse;
}
```

And I didn't change `SpuStatus`, you probably mean `PartitionStatus`?

```rust
#[derive(Decoder, Encoder, Default, Debug, Clone, PartialEq)]
#[cfg_attr(
    feature = "use_serde",
    derive(serde::Serialize, serde::Deserialize),
    serde(rename_all = "camelCase")
)]
pub struct PartitionStatus {
    pub resolution: PartitionResolution,
    pub leader: ReplicaStatus,
    // TODO: Next time we make a breaking protocol change, rename this to `lrs`
    // TODO: There is no such thing as `lsr`, it is a typo
    pub lsr: u32,
    pub replicas: Vec<ReplicaStatus>,
    #[fluvio(min_version = 5)] // <-------- HERE
    pub size: i64,
    pub is_being_deleted: bool,
}
```
does this seem right to you?

There is this comment asking someone to change `lsr` to `lrs`. It may be the time to do it?

sehz:
Thats seems right. Changing LRS should be done in separate PR

Afourcat:
Hey,
I have no clue on how to make the CI go green.

Afourcat:
I'm sorry,
I have no clue on how to make the CI go green.
Can someone lend a hand?

Thank you,

sehz:
Can you try run failed CI job locally? You should be able to run them in your machine by just following the steps.  If you are able to run successfully, this is probably a flaky CI run and re-running action should fix the problem

sehz:
also rebase, CI step should no longer hang

Afourcat:
Hello,

> Can you try run failed CI job locally? You should be able to run them in your machine by just following the steps. If you are able to run successfully, this is probably a flaky CI run, and re-running action should fix the problem

Locally these tests do not pass. I'm struggling a bit to understand why.

> also rebase, CI step should no longer hang

I rebased, Indeed it fixed a failing test.

I'm currently reading the sources of failing tests.

So,
[For this test](https://github.com/infinyon/fluvio/runs/5272283645?check_suite_focus=true), if I understand it well.
It should fail because the stable cluster is not compatible with the latest CLI.

However,
[this test should pass](https://github.com/infinyon/fluvio/runs/5272284278?check_suite_focus=true,),
Upgrading from one version to another should work.

Do I need to mark the CLI and the stable cluster as incompatible, somehow?
I'm sorry, I make you lose some time :sweat_smile: 

Thank you,

sehz:
both tests need to pass in order to get this PR merged since we want to ensure compatibility.    For the failed run, there should be a log artifact in the Github Action run.  You can look at the log file and find errors.  

In your local cluster, you can also see it in the SC and SPU log files.    We need to add more doc on how to debug fluvio in the developer README.

Afourcat:
Hello,
> For the failed run, there should be a log artifact in the Github Action run. You can look at the log file and find errors.

Thank you, I didn't know about artifacts.

As expected, reading the SC logs uploaded by the CI, I found:
```
JsonError(Error(\"missing field `size`\", line: 1, column: 1230))
```
I hopped that `#[fluvio(min_version = 5)]` would fix it.

I'm currently trying with something like `#[fluvio(min_version = 5, ignorable)]` that I found while searching in the codebase.
I will investigate more tonight.

Afourcat:
I finally found out how to make the `upgrade-test` go green.
But, I don't know if it is what is expected.
```rust
#[derive(Decoder, Encoder, Default, Debug, Clone, PartialEq)]
#[cfg_attr(
    feature = "use_serde",
    derive(serde::Serialize, serde::Deserialize),
    serde(rename_all = "camelCase")
)]
pub struct PartitionStatus {
    pub resolution: PartitionResolution,
    pub leader: ReplicaStatus,
    // TODO: Next time we make a breaking protocol change, rename this to `lrs`
    // TODO: There is no such thing as `lsr`, it is a typo
    pub lsr: u32,
    pub replicas: Vec<ReplicaStatus>,
    #[cfg_attr(feature = "use_serde", serde(default))] // <-- HERE
    #[fluvio(min_version = 5)]
    pub size: i64,
    pub is_being_deleted: bool,
}
```

Also, do I need to add the following commit?
in `fluvio-controlplane/src/requests/update_lrs.rs`
```rust
 impl Request for UpdateLrsRequest {
     const API_KEY: u16 = InternalScKey::UpdateLrs as u16;
+    const DEFAULT_API_VERSION: i16 = 1;
     type Response = UpdateLrsResponse;
 }
 
@@ -63,6 +64,7 @@ pub struct LrsRequest {
     pub id: ReplicaKey,
     pub leader: ReplicaStatus,
     pub replicas: Vec<ReplicaStatus>,
+    #[fluvio(min_version = 1)]
     pub size: i64,
 }
 ```
 
Thank you,

**Comment by sehz** [comment]
`update_lrs.rs` is between SC and SPU which assume to be using same version for now, thus there is no need add version yet.

Can you perform following manual tests?:
* Stable CLI and new cluster
* New CLI and stable CLI

These ensure we have have full backward compatibility.  We probably need to move `cd` tests to CI later.

Great job btw.

**Comment by sehz** [comment]
@morenol  can you do code review?  Also, confirm backward compatibility tests.

File: crates/fluvio-storage/src/replica.rs [hunk 1]
```diff
@@ -97,6 +99,39 @@ impl ReplicaStorage for FileReplica {
...
+            let os_file_name = entry.file_name();
+            if os_file_name.to_string_lossy().ends_with(".log") {
+                let metadata = entry.metadata().await.map_err(|err| {
+                    error!(
+                        "fetching metadata for file \"{}\" failed: {err}",
+                        os_file_name.to_string_lossy(),
+                    );
+                    ErrorCode::StorageError
+                })?;
+                return Ok(metadata.len());
```
morenol, line 125: This is only returning the size of the first log file that it gets. We can have multiple segments in the topic, so I think that we should return the sum of the size of all of them. 

You can test this by using a topic with a low segment-size: `fluvio topic create topicname --segment-size 1024`, once the first log file reaches 1024, it creates a new .log file.

Not sure if we also should count the size of the replication.chk and .index files, what do you think @sehz ?

connect.yaml
```diff
@@ -0,0 +1,9 @@
+api_version: v1
```
morenol, line 1: Please remove this file from the PR

morenol (changes requested):
Nice work @Afourcat,

I added a couple of comments. 

Other than that,  It seems to be working nicely using:

new cli  - new spu
stable cli - new spu

When I use new cli and stable spu, running `fluvio partition list` always return 0 as size. I think that we need to find a way to hide that field if the SPU does not support the size field.

crates/fluvio-storage/src/replica.rs
sehz: replication and index are internal file so let's not expose them yet (and smaller too)

Afourcat:
> When I use new cli and stable spu, running fluvio partition list always return 0 as size. I think that we need to find a way to hide that field or have a special message if the SPU does not support the size field.

Right now, if the size is `-1`. The CLI shows an `ERROR`.
```
TOPIC      PARTITION  LEADER  REPLICAS  RESOLUTION  SIZE   HW  LEO  LRS  FOLLOWER OFFSETS 
cat-facts  0          5001    []        Online      ERROR  12  12   0    [] 
cat-facts  1          5001    []        Online      ERROR  10  10   0    [] 
cat-facts  2          5001    []        Online      ERROR  10  10   0    [] 
cat-facts  3          5001    []        Online      ERROR  9   9    0    []
```
We may default the `size` field to `-2` instead of `0`.
And hide the field in that case?

morenol:
> We may default the `size` field to `-2` instead of `0`. And hide the field in that case?

Yes, we can do that, please define constants and use them for that. Maybe we shouldn't hide, but have a different message

Afourcat:
```
 TOPIC  PARTITION  LEADER  REPLICAS  RESOLUTION  SIZE           HW  LEO  LRS  FOLLOWER OFFSETS 
 lyon   0          0       []        Online      NOT SUPPORTED  11  11   0    []
 ```
As you can see, I set the text to `NOT SUPPORTED`.

**Comment by sehz** [comment]
'NA' is simpler

**Comment by ajhunyady** [comment]
It seems like SIZE is something that should always be available. Why is there no data, and the need for NA?

**Comment by sehz** [comment]
Thats for running against old version

**Comment by ajhunyady** [comment]
> Thats for running against old version

got it.

**Comment by sehz** [comment]
@morenol all good?

File: crates/fluvio-spu/src/replication/leader/replica_state.rs [hunk 3]
```diff
@@ -291,14 +291,20 @@ where
                 (*follower_id, follower_info.hw, follower_info.leo).into()
             })
             .collect();
+        let storage_reader = self.storage.read().await;
+        let size = storage_reader
+            .get_partition_size()
+            .await
+            .map(|e| e as i64)
+            .unwrap_or(-1);
```
morenol, line 299: Please use here `PartitionStatus::SIZE_ERROR`

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Release Notes
 
 ## Platform Version 0.9.21 - UNRELEASED
+* Add a `size` field in the `fluvio partition list` output. This field represents the size of logs in the partition. ([#2214](https://github.com/infinyon/fluvio/pull/2214))
```
morenol, line 4: Please point to the #2148

Afourcat: Should I point to both the issue and the pull request?

morenol (approved):
LGTM, Nice work!

Afourcat:
Thank you

sehz (approved):

sehz:
rebase

morenol:
bors retry

PR #2219, opened 2022-02-22
[Merged by Bors] - refactor: use async RwLock in SPU instead of sync version

morenol:
While reading the code due to #2220, I found that we are using the sync RwLock version in the Spu crate. This changes that to the async_lock version.


sehz (approved):
LGTM

morenol:
bors retry

PR #2226, opened 2022-02-24
[Merged by Bors] - Fix panic in 'fluvio connector logs'

dervus:
If cluster has no installed connectors, `fluvio connector logs <name>` will panic on line 63 with:

* Attempt to subtract with overflow in debug build, i.e. `0usize - 17usize` = overflow panic.
* Byte index 18446744073709551599 is out of bounds of '' in release build, i.e. overflow check is skipped, but slice boundaries check is still there.
 
This happens because `pod_name` will contain an empty string in case of missing connectors. So, this PR:

* Makes `is_matching_connector_name` safe to work with any input string.
* Skips search if `kubectl` returned empty string to begin with.

### DIFF
--- a/crates/fluvio-cli/src/connector/logs.rs
+++ b/crates/fluvio-cli/src/connector/logs.rs
@@ -36,9 +36,11 @@ impl LogsManagedConnectorOpt {
             .stdout;
 
         let pods = String::from_utf8_lossy(&pods);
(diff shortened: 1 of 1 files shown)


simlay (approved):
Ah. Good catch. I should have reviewed better on the last PR for this bit.

Thanks!

PR #2229, opened 2022-02-25
Report Error codes in RecordMetadata

morenol:


crates/fluvio/src/producer/partition_producer.rs
```diff
@@ -216,10 +216,9 @@ impl PartitionProducer {
         for (batch_notifier, response) in batch_notifiers.into_iter().zip(response.responses.iter())
         {
             let base_offset = response.partitions[0].base_offset;
-            if let Err(_e) = batch_notifier.send(base_offset).await {
-                trace!(
-                    "Failed to notify producer of successful produce because receiver was dropped"
-                );
+            let fluvio_error = response.partitions[0].error_code.clone();
+            if let Err(_e) = batch_notifier.send((base_offset, fluvio_error)).await {
+                trace!("Failed to notify produce result because receiver was dropped");
```
sehz, line 221: Shouldn't this be error or propogate?

sehz (changes requested):

crates/fluvio/src/producer/partition_producer.rs
morenol: No, this error happens when we dont keep the receiver,
```rust
This will log that
let _ = producer.send(key, value).await?;
// This won't log that
let produce_output = producer.send(key, value).await?;
```

sehz:
rebase

sehz (approved):
LGTM

Issue #2230, opened 2022-02-25
Optimize Partition Size computation


sehz:
With #2212, we have ability get size of the partitions.  However, performance of the size computation can be improved.  Instead of querying metadata each time, it can query segments which already have size cached.

sehz:
@Afourcat interested?

Afourcat:
Hello,
The issue is happening without the commented code. In fact, I also have this issue on stable cluster and CLI.
```
Release Channel      : stable
Fluvio CLI           : 0.9.20
Fluvio CLI SHA256    : 01e4d42525292d6ca924ca6f34443f7533c269c974981634bcca1f676c80fad1
Fluvio Platform      : 0.9.20 (minikube)
Git Commit           : 50446a0ec086b9d87f7800f143a8f4f4f66285cd
OS Details           : Gentoo 2.8 (kernel 5.15.5-gentoo-dist)
=== Plugin Versions ===
Fluvio Runner (fluvio-run)     : 0.0.0
Infinyon Cloud CLI (fluvio-cloud) : 0.1.6
```

I think that this is this line that throws the "22: Invalid argument" error.
in `fluvio/crates/fluvio-storage/src/index.rs:119`.
```rust
        // make sure it is log file
        let (m_file, file) =
            MemoryMappedFile::open(&index_file_path, INDEX_ENTRY_SIZE as u64).await?;
```

```
2022-02-27T19:51:58.305053Z DEBUG handle_request{host="0.0.0.0:9010"}:respond{socket=Socket(1317) _connection=peer("127.0.0.1:39330")}:handle_produce_request{id=1 client=fluvio}:handle_produce_topic{topic=lyon-seg}:handle_produce_partition{replica_id=lyon-seg-0}:write_record_set:write_record_set:write_recordset:write_batch: fluvio_storage::index: opening index index_file_path="/home/alex/fluvio-logs/spu-logs-5001/lyon-seg-0/00000000000000000001.index"
2022-02-27T18:53:28.725918Z ERROR handle_request{host="0.0.0.0:9005"}:respond{socket=Socket(25) _connection=peer("172.17.0.1:22939")}:handle_produce_request{id=6 client=fluvio}:handle_produce_topic{topic=stable}:handle_produce_partition{replica_id=stable-0}: fluvio_spu::services::public::produce_handler: Error writing to replica: Io(
    Os {
        code: 22,
        kind: InvalidInput,
        message: "Invalid argument",
    },
) replica_id=stable-0
```

[...]

PR #2231, opened 2022-02-28
Add a test for segmented topics' log file rollout.

Afourcat:
Requested in #2230

### DIFF
--- a/crates/fluvio-storage/src/replica.rs
+++ b/crates/fluvio-storage/src/replica.rs
@@ -414,7 +414,7 @@ mod tests {
     use dataplane::core::{Decoder, Encoder};
     use dataplane::record::{Record, RecordSet};
     use dataplane::batch::MemoryRecords;
-    use dataplane::fixture::{BatchProducer, create_batch};
+    use dataplane::fixture::{BatchProducer, create_batch, create_batch_with_producer};
     use dataplane::fixture::read_bytes_from_file;
     use flv_util::fixture::ensure_clean_dir;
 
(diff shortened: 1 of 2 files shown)


sehz:
Can you reproduce issue by adding or update tests in: https://github.com/infinyon/fluvio/blob/master/crates/fluvio-storage/src/replica.rs?

Afourcat:
Hello,
I find out where this is coming from.
# Why
When rolling out a new segment, we are using the `as_segment` function for the old segment which ends up using:
```rust
let (m_file, file) =
    MemoryMappedFile::open(&index_file_path, INDEX_ENTRY_SIZE as u64).await?;
```
Which calls the `fluvio-future/src/fs/mmap.rs -> MemoryMappedFile::open` function.
```rust
/// open memory file, specify minimum size
pub async fn open<P>(path: P, min_len: u64) -> Result<(Self, File), IoError>
where
    P: AsRef<Path>,
{
    let m_path = path.as_ref().to_owned();
    let (m_map, mfile, _) = spawn_blocking(move || {
        let mfile = OpenOptions::new().read(true).open(&m_path).unwrap();
        let meta = mfile.metadata().unwrap();
        if meta.len() == 0 {
            mfile.set_len(min_len)?;
        }
                                                                                      
        unsafe { Mmap::map(&mfile) }.map(|mm_file| (mm_file, mfile, m_path))
    })
    .await?;
                                                                                      
    Ok((MemoryMappedFile::from_mmap(m_map), mfile.into()))
}
```
If the length of the file is not zero, we are using `mfile.set_len(min_len)` [which throws an `Error`](https://doc.rust-lang.org/std/fs/struct.File.html#method.set_len) if the file is not opened with write capacity.
# Fix
A simple fix is to open the file with write capacity.
```rust
                                          vvvvvvvvvvv
let mfile = OpenOptions::new().read(true).write(true).open(&m_path).unwrap();
```
Even if `mmap` creates a read-only memory map, do we specifically want the file to be opened read-only?
If yes, should we open the file in write mode to set its length and then re-open it in read-only mode?
# Conclusion

Do you want me to open a pull request on the [`fluvio-future`](https://github.com/infinyon/future-aio) project?

Afourcat:
Thanks,
[I made the PR](https://github.com/infinyon/future-aio/pull/137).
I will bump the version of `fluvio-future` when available.
I should now be able to make this issue 🎉

PR #2238, opened 2022-03-09
[Merged by Bors] - fix(cli): #2207 connector delete error handling

### LINKED ISSUE #2207: Connector: Deleting connector should give more readable error
If trying to delete an unknown connector, CLI gives an unfriendly error

 ```$ fluvio connector delete jjj
Error: 
   0: Fluvio client error
   1: Fluvio controlplane error: Received error code: ManagedConnectorNotFound (Some("not found"))
   2: Received error code: ManagedConnectorNotFound (Some("not found"))
```

### PR DESCRIPTION
**Pull request by pinkforest**
Fixes #2207
Proposes a precedent for #2240

By default CLIError::Unhandled intent is a "I'm a BUG - I should be fixed" in a predictable easily automation-parseable form -

All CLI is mapped out in contextual presentable form whilst library will still do #[from]

Any Unhandled errors can be quickly fixed by identifying straight from the report which one needs to be matched instead of go with status quo bad presentation. e.g. if CI screams suddenly 100 unhandled errors they can be mapped much faster.

This way when someone adds CLI test cases for any negative tests they can identify all errors that need proper user visible contextual presentation vs library presentation.

e.g. when this would be now Unhandled - 
```
$ ./target/debug/fluvio connector delete foobar
Error: 
   0: BUG! Report me - Unhandled error: AdminApi(Code(ManagedConnectorNotFound, Some("not found")))
```

But as a mapped error now for connector not found:
```
$ ./target/debug/fluvio connector delete fffd
Error:
   0: Invalid connector: fffd not found
```

### DIFF
--- a/crates/fluvio-cli/src/connector/delete.rs
+++ b/crates/fluvio-cli/src/connector/delete.rs
@@ -21,10 +21,24 @@ pub struct DeleteManagedConnectorOpt {
     name: String,
 }
 
+use fluvio::FluvioError;
+use fluvio_sc_schema::ApiError;
+use fluvio_sc_schema::errors::ErrorCode;
+
 impl DeleteManagedConnectorOpt {
     pub async fn process(self, fluvio: &Fluvio) -> Result<(), CliError> {
         let admin = fluvio.admin().await;
-        admin.delete::<ManagedConnectorSpec, _>(&self.name).await?;
+
+        admin
+            .delete::<ManagedConnectorSpec, _>(&self.name)
+            .await
+            .map_err(|e| match e {
+                FluvioError::AdminApi(ApiError::Code(ErrorCode::ManagedConnectorNotFound, _)) => {
+                    CliError::InvalidConnector(format!("{} not found", &self.name))
+                }
+                _ => CliError::Other(format!("{:?}", e)),
+            })?;
+
         Ok(())
     }
 }
--- a/crates/fluvio-cli/src/connector/delete.rs
+++ b/crates/fluvio-cli/src/connector/delete.rs
@@ -36,7 +36,8 @@ impl DeleteManagedConnectorOpt {
                 FluvioError::AdminApi(ApiError::Code(ErrorCode::ManagedConnectorNotFound, _)) => {
                     CliError::InvalidConnector(format!("{} not found", &self.name))
                 }
-                _ => CliError::Other(format!("{:?}", e)),
+                // Raise a bug for any Unhandled error.
+                _ => CliError::Unhandled(format!("{:?}", e)),
             })?;
 
         Ok(())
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: crates/fluvio-cli/src/connector/delete.rs [hunk 1]
```diff
@@ -21,10 +21,24 @@ pub struct DeleteManagedConnectorOpt {
...
-        admin.delete::<ManagedConnectorSpec, _>(&self.name).await?;
+
+        admin
+            .delete::<ManagedConnectorSpec, _>(&self.name)
+            .await
+            .map_err(|e| match e {
+                FluvioError::AdminApi(ApiError::Code(ErrorCode::ManagedConnectorNotFound, _)) => {
+                    CliError::InvalidConnector(format!("{} not found", &self.name))
+                }
+                _ => CliError::Other(format!("{:?}", e)),
```
morenol, line 39: I think that display representation is better for cli

morenol:
Just a little nit

morenol (approved):

crates/fluvio-cli/src/connector/delete.rs
wendyOn: Decision was to leave the Debug to leave for identified bug purposes that will enable velocity for mapping out CLI errors into presentable form.

wendyOn: Further we decided that we will add Unhandled that will inform that BUG has been found.

wendyOn: Done. I've updated the description per intent.

nacardin:
Nice! I like this

nacardin (approved):

simlay (approved):
bors r+

simlay:
bors retry

crates/fluvio-cli/src/connector/delete.rs
```diff
@@ -21,10 +21,25 @@ pub struct DeleteManagedConnectorOpt {
...
-        admin.delete::<ManagedConnectorSpec, _>(&self.name).await?;
+
+        admin
+            .delete::<ManagedConnectorSpec, _>(&self.name)
+            .await
+            .map_err(|e| match e {
+                FluvioError::AdminApi(ApiError::Code(ErrorCode::ManagedConnectorNotFound, _)) => {
+                    CliError::InvalidConnector(format!("{} not found", &self.name))
+                }
+                // Raise a bug for any Unhandled error.
```
sehz, line 39: I agree that we need to properly handle all error messages correctly but this is probably not the preferred approach.
* First, BUG is a serious allegation that may or may not be correct for any unhandled. 
* We need to fix systematic not in a single isolated code.

sehz (changes requested):

PR #2243, opened 2022-03-10
[Merged by Bors] - Rework/optimize partition size computation

Afourcat:
This PR Implements #2230.

- Add smoke-test
- Bumps `fluvio-future` to `0.3.15`.
- Replace File-based computation of logs' length by the sum of active and previous `Segment`s `msg_log` size.
Which removes the need for file system IO.
- Change tests

### DIFF
--- a/crates/fluvio-storage/src/replica.rs
+++ b/crates/fluvio-storage/src/replica.rs
@@ -103,12 +103,15 @@ impl ReplicaStorage for FileReplica {
     async fn get_partition_size(&self) -> Result<u64, ErrorCode> {
         let active_len = self.active_segment.get_msg_log().get_pos();
 
-        debug!("Active segment length: {active_len}.");
+        debug!(active_len, "Active segment length");
         let total_prev_segments_len = {
             let reader = self.prev_segments.read().await;
             reader.get_total_logs_len() as u32
         };
-        debug!("Cumulated previous segments length: {total_prev_segments_len}.");
+        debug!(
+            total_prev_segments_len,
+            "Cumulated previous segments length"
+        );
         let total_len = active_len + total_prev_segments_len;
 
         return Ok(total_len.into());
--- a/crates/fluvio-storage/src/replica.rs
+++ b/crates/fluvio-storage/src/replica.rs
@@ -103,40 +103,17 @@ impl ReplicaStorage for FileReplica {
     /// return the size in byte
     #[instrument(skip(self))]
     async fn get_partition_size(&self) -> Result<u64, ErrorCode> {
-        let mut entries = read_dir(&self.option.base_dir).await.map_err(|err| {
-            error!(
-                "failed to open the log base dir \"{}\": {}",
-                self.option.base_dir.to_string_lossy(),
-                err
-            );
-            ErrorCode::StorageError
-        })?;
+        let active_len = self.active_segment.get_msg_log().get_pos();
 
-        let mut total_size = None;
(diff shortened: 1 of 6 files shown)


crates/fluvio-storage/src/replica.rs
```diff
@@ -721,11 +698,11 @@ mod tests {
...
             .expect("delete base dir");
 
-        let error = replica
+        let size = replica
             .get_partition_size()
             .await
-            .expect_err("error partition size");
-        assert_eq!(error, dataplane::ErrorCode::StorageError)
+            .expect("error partition size");
+        assert_eq!(size, 79);
```
Afourcat, line 705: I had to change this test, removing the logs' directory does not lead to the `get_partition_size` method to throw an error anymore.
Instead, it returns the memory-stored size of active and previous segments. Even if these segments don't exist anymore.
What kind of behavior are we expecting here?

crates/fluvio-storage/src/replica.rs
```diff
@@ -103,40 +103,17 @@ impl ReplicaStorage for FileReplica {
...
-                })?;
-                let file_len = metadata.len();
-                debug!("Log segment found. Add {file_len} byte to the total_size.");
-                *total_size.get_or_insert(0) += file_len;
-            }
-        }
+        debug!("Active segment length: {active_len}.");
+        let total_prev_segments_len = {
+            let reader = self.prev_segments.read().await;
+            reader.get_total_logs_len() as u32
```
Afourcat, line 111: I didn't use atomic here, as advised, since the bump of `fluvio-future` did solve my issue.
Do you want me to change the implementation?

crates/fluvio-storage/src/replica.rs
```diff
@@ -103,40 +101,17 @@ impl ReplicaStorage for FileReplica {
...
-                        os_file_name.to_string_lossy(),
-                    );
-                    ErrorCode::StorageError
-                })?;
-                let file_len = metadata.len();
-                debug!("Log segment found. Add {file_len} byte to the total_size.");
-                *total_size.get_or_insert(0) += file_len;
-            }
-        }
+        debug!("Active segment length: {active_len}.");
```
sehz, line 106: should change to `debug!(active_len, "Active segment length")`

crates/fluvio-storage/src/replica.rs
sehz: No this is fine.

crates/fluvio-storage/src/replica.rs
```diff
@@ -103,40 +101,17 @@ impl ReplicaStorage for FileReplica {
...
+        let total_prev_segments_len = {
+            let reader = self.prev_segments.read().await;
+            reader.get_total_logs_len() as u32
+        };
+        debug!("Cumulated previous segments length: {total_prev_segments_len}.");
```
sehz, line 111: `debug!(total_prev_segments_len,....)`

crates/fluvio-storage/src/replica.rs
sehz: I think this is correct.  total size = active_segment_size + prev_segment_size.  if that doesn't reflect deletion then it is separate bug.

sehz (changes requested):
Looks pretty good.  Couple of minor nits.  Can you also add to changelog?

Thanks!

sehz (approved):
LGTM.  Great work!

PR #2247, opened 2022-03-12
[Merged by Bors] - add compression in data generator

morenol:
This updates data generator with the capability to also set the compression


tjtelan (approved):
looks good to me

PR #2264, opened 2022-03-24
[Merged by Bors] - Topic level compression fix

### LINKED ISSUE #2265: SPU does not return error code if produce request is using a compression algorithm not accepted by topic configuration
They way to reproduce is to use old client without support to topic level compression and send uncompressed data to a topic with an specific compression configuration

morenol:
Closes #2265

Adding #[fluvio(min_version) in Replica struct was provoking using the default for compression type. 

This changes fixes that


sehz (changes requested):
(1) Can you create issue for this?
(2) Add unit test or integration test to verify this is fixed.

morenol:
I had an integration test for this https://github.com/infinyon/fluvio/blob/master/crates/fluvio-spu/src/services/public/produce_handler.rs, seems to be working as expected but the issue is that there, the ReplicaLocalStore is mocked.

Not sure how to test this in another way since I cannot create producer with invalid compression algorithm

sehz:
https://github.com/infinyon/fluvio/issues/2265 should have steps to reproduce.  That should be put into integration test.

morenol:
I added in the description how to replicate that, but still it requires the usage of old client version, not sure how to add a test with that constraint

sehz:
I think that scenario (old client) to new cluster is already described in current CI tests @tjtelan?

morenol:
> I think that scenario (old client) to new cluster is already described in current CI tests @tjtelan?

Yes, but that is something that we will only be able to test when old client is 0.9.21

sehz (approved):
LGTM

PR #2267, opened 2022-03-24
[Merged by Bors] - Report error code in CLI `fluvio produce`

### LINKED ISSUE #2266: `fluvio produce` does not fail if SPU sends an error code.
It should return and error and report the error.

for instance BatchTooBig and CompressionError spu error codes are not being reported

morenol:
Fixes #2266

Added a test to the CLI smoke tests, that should fail due to batchToBig spu error code


sehz (approved):
LGTM

PR #2269, opened 2022-03-25
Simple example about using your https://infinyon.cloud/ credentials

devcdcc:
This MR aims to provide a simple example about producing/consuming messages with your https://infinyon.cloud/ cluster.


examples/05-produce-cloud/src/main.rs
```diff
@@ -0,0 +1,47 @@
+//! A minimal example showing how to produce messages on Fluvio
```
morenol, line 1: Maybe we can rephrase this to make it clear that it is using a profile different than the default. Maybe instead of produce-cloud we could name this as `produce-custom-profile` and then we say that we will use the `cloud` profile

examples/05-produce-cloud/src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+#[async_std::main]
+async fn main() {
+    if let Err(e) = produce().await {
+        println!("Produce error: {:?}", e);
+    }
+}
+
+async fn produce() -> Result<(), fluvio::FluvioError> {
+    let config = ConfigFile::load(None).unwrap();
+    let fluvio_config = config.config().cluster.get("cloud").unwrap();
```
morenol, line 37: I think that we should mention in some place that we need to make sure that this is created, by calling first `fluvio cloud login` with the infinyon.cloud credentials

examples/06-consume-cloud/src/main.rs
```diff
@@ -0,0 +1,60 @@
+//! A minimal example showing how to produce messages on Fluvio
```
morenol, line 1: Same here, I think that we can change this to say custom-profile

examples/05-produce-cloud/src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+
+#[async_std::main]
+async fn main() {
+    if let Err(e) = produce().await {
+        println!("Produce error: {:?}", e);
+    }
+}
+
+async fn produce() -> Result<(), fluvio::FluvioError> {
+    let config = ConfigFile::load(None).unwrap();
```
morenol, line 36: Please use, ? instead of `unwrap()`

examples/06-consume-cloud/src/main.rs
```diff
@@ -0,0 +1,60 @@
...
+        }
+        // The consumer should run forever except for the timeout above
+        _ => unreachable!("Consumer should last forever"),
+    }
+}
+
+async fn consume() -> Result<(), fluvio::FluvioError> {
+    use futures_lite::StreamExt;
+
+    let config = ConfigFile::load(None).unwrap();
```
morenol, line 48: Please use, ? instead of unwrap()

examples/06-consume-cloud/src/main.rs
```diff
@@ -0,0 +1,60 @@
...
+    let fluvio_config = config.config().cluster.get("cloud").unwrap();
```
morenol, line 49: Please use, ? instead of unwrap()

morenol (changes requested):
Thanks for the PR, I added some comments

PR #2276, opened 2022-03-29
[Merged by Bors] - Change default max_bytes in consumer

### LINKED ISSUE #2259: `fluvio consume` stalls if produced batch size is too large
This came out of testing https://github.com/infinyon/fluvio/pull/2258

I produced 1 million JSON records (~1GB) with a batch size of 1MB, no compression.

When I tried consume from this topic (with `fluvio consume -B -d my-topic`) the consumer just sat there and never returned any data.

This worked fine when I used the `-T` option (instead of `-B`) to pull the last 10 records.

I expect there to be some output (even if I have to wait a while) but even after a few minutes I never received data.

---
I deleted the topic and produced the records again, but with a batch of 500K, no compression. And `fluvio consumer -B -d my-topic` worked

morenol:
This changes `MAX_FETCH_BYTES` to be the same value of `STORAGE_MAX_BATCH_SIZE` + the size of the other felds in a `FetchableTopicResponse` in order to be able to receive responses with at least a batch of that size.

Fixes #2259

### DIFF
--- a/crates/fluvio/src/consumer.rs
+++ b/crates/fluvio/src/consumer.rs
@@ -576,10 +576,19 @@ mod publish_stream {
 /// MAX FETCH BYTES
 static MAX_FETCH_BYTES: Lazy<i32> = Lazy::new(|| {
     use std::env;
+    use fluvio_protocol::Encoder;
+    use crate::dataplane::fetch::FetchResponse;
+    use crate::dataplane::fetch::FetchableTopicResponse;
+    use crate::dataplane::fetch::FetchablePartitionResponse;
+    use crate::dataplane::batch::MemoryRecords;
     let var_value = env::var("FLV_CLIENT_MAX_FETCH_BYTES").unwrap_or_default();
-    let max_bytes: i32 = var_value
-        .parse()
-        .unwrap_or_else(|_| STORAGE_MAX_BATCH_SIZE.try_into().unwrap_or(1048588));
+    let max_bytes: i32 = var_value.parse().unwrap_or_else(|_| {
+        FetchResponse::<MemoryRecords>::default().write_size(0) as i32
+            + FetchableTopicResponse::<MemoryRecords>::default().write_size(0) as i32
+            + FetchablePartitionResponse::<MemoryRecords>::default().write_size(0) as i32
+            + 64 // using max size of topic name
+            + STORAGE_MAX_BATCH_SIZE.try_into().unwrap_or(1048588)
+    });
     max_bytes
 });
 
--- a/crates/fluvio/src/consumer.rs
+++ b/crates/fluvio/src/consumer.rs
@@ -7,6 +7,7 @@ use futures_util::future::{Either, err, join_all};
 use futures_util::stream::{StreamExt, once, iter};
 use futures_util::FutureExt;
 
+use fluvio_types::defaults::STORAGE_MAX_BATCH_SIZE;
 use fluvio_spu_schema::server::stream_fetch::{
     DefaultStreamFetchRequest, DefaultStreamFetchResponse, GZIP_WASM_API, SMART_MODULE_API,
     LegacySmartModulePayload, SmartModuleWasmCompressed, WASM_MODULE_API, WASM_MODULE_V2_API,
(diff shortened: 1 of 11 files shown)


sehz:
Can you turn off draft mode?

crates/fluvio/src/consumer.rs
```diff
@@ -575,8 +576,19 @@ mod publish_stream {
...
+    use crate::dataplane::fetch::FetchableTopicResponse;
+    use crate::dataplane::fetch::FetchablePartitionResponse;
+    use crate::dataplane::batch::MemoryRecords;
     let var_value = env::var("FLV_CLIENT_MAX_FETCH_BYTES").unwrap_or_default();
-    let max_bytes: i32 = var_value.parse().unwrap_or(1000000);
+    let max_bytes: i32 = var_value.parse().unwrap_or_else(|_| {
+        FetchResponse::<MemoryRecords>::default().write_size(0) as i32
+            + FetchableTopicResponse::<MemoryRecords>::default().write_size(0) as i32
+            + FetchablePartitionResponse::<MemoryRecords>::default().write_size(0) as i32
+            + 64 // using max size of topic name
```
sehz, line 589: Where is this come from?

morenol: I am using https://github.com/infinyon/fluvio/blob/5d1ca946e0be25c866068b3d165e3891a3522a6e/crates/fluvio-dataplane-protocol/src/fetch/response.rs#L54 which has a `name` attribute which is a String with the topic name, I thought that topic name size is limited to the kubernetes max size of a label which is 63, not sure if that is the actual limit of a topic name

sehz (changes requested):
This seems highly dependent on specific fields which could be fragile.  Can you add test(s)?

crates/fluvio/src/producer/accumulator.rs
```diff
@@ -188,6 +189,7 @@ impl ProducerBatch {
                 Compression::None => 1.0,
                 Compression::Gzip | Compression::Snappy | Compression::Lz4 => 0.5,
             }) as usize
+            + Batch::<RawRecords>::default().write_size(0)
```
morenol, line 192: Previosly, this computation only considered record size, but when we use a batch_size large enough like 1048588, when we added other Batch fields, the size was higher than that, therefore SPU was rejecting that with BatchTooBig, that is the reason for this change

tests/cli/smoke_tests/consume-batch-size.bats
```diff
@@ -0,0 +1,41 @@
...
+
+# Create topic
+@test "Create topics for test" {
+    debug_msg "topic: $TOPIC_NAME"
+    run timeout 15s "$FLUVIO_BIN" topic create "$TOPIC_NAME"
+}
+
+@test "Produce message with batch large" {
+    run bash -c "yes abc |head -c 1500000 > $TOPIC_NAME.txt"
+    run bash -c 'timeout 35s "$FLUVIO_BIN" produce "$TOPIC_NAME" --batch-size 1048588 --file $TOPIC_NAME.txt --linger 5s'
```
morenol, line 30: without the changes in `accumulator.rs` this produce was failing due to BatchTooBig

tests/cli/smoke_tests/consume-batch-size.bats
```diff
@@ -0,0 +1,41 @@
...
+    assert_success
+}
+
+# Consume message and compare message
+@test "Consume message" {
+    run timeout 15s "$FLUVIO_BIN" consume "$TOPIC_NAME" -B -d
```
morenol, line 36: This is currently failing when the first Batch is larger than 1000000 bytes.

crates/fluvio/src/consumer.rs
sehz, line 589: can you make this constant and put in with rest of defaults?

sehz (changes requested):
Looks very good.  One minor nitpick

crates/fluvio/src/consumer.rs
morenol: I found that, actually the max length is 252 (so it is 255 in bytes (similar to kafka)) I am adding a const for that and changing that here

morenol: Sure, I also noticed that actually the max size is 255

sehz (approved):
Sounds good.

PR #2296, opened 2022-04-13
[Merged by Bors] - Installer script update

tjtelan:
Update installer to install the same version as Fluvio CLI, if specified instead of always using `stable`


tjtelan:
Looks like this change starts the cluster. Example: https://github.com/infinyon/fluvio/runs/6000966168?check_suite_focus=true

sehz (approved):
LGTM

Issue #2288, opened 2022-04-08
Add record timestamps


morenol:
Ensure every batch has correct timestamp

morenol:
Currently all timestamp related fields in Records and Batches are set to 0. We need to update the producer to populate the timestamp fields

morenol:
Regarding CLI:

> Can we leverage https://docs.rs/chrono/0.4.19/chrono/format/strftime/index.html#specifiers?


@sehz since we are using Handlebars, we need to pass {{ name }}, for every specifier that we want. Do we want to be able to specify them with something like `{{ Y }}`, `{{ C }}`? I think that for CLI it is enough to follow a specific RFC format

PR #2322, opened 2022-04-18
[Merged by Bors] - Add timestamp to records

morenol:
This PR updates producer to set the fields in batch header: `first_timestamp`, with the timestamp of the first record of the batch and `max_time_stamp`, with the timestamp of the last record of the batch. And also updates the records in the batches to populate the `timestamp_delta` field in the header.

Part of #2288

### DIFF
--- a/crates/fluvio-cli/src/consume/mod.rs
+++ b/crates/fluvio-cli/src/consume/mod.rs
@@ -4,13 +4,11 @@
 //! CLI command for Consume operation
 //!
 
-use std::time::UNIX_EPOCH;
 use std::{io::Error as IoError, path::PathBuf};
 use std::io::{self, ErrorKind, Read, Stdout};
 use std::collections::{BTreeMap};
 use flate2::Compression;
 use flate2::bufread::GzEncoder;
-use fluvio::dataplane::batch::NO_TIMESTAMP;
 use fluvio::metadata::tableformat::{TableFormatSpec};
 use tracing::{debug, trace, instrument};
 use clap::{Parser, ArgEnum};
@@ -82,10 +80,8 @@ pub struct ConsumeOpt {
     /// Provide a template string to print records with a custom format.
     /// See --help for details.
     ///
-    /// Template strings may include the variables {{key}}, {{value}}, {{offset}}, {{partition}}, {{timestamp_humanize}} and {{timestamp}}
+    /// Template strings may include the variables {{key}}, {{value}}, {{offset}} and {{partition}}
     /// which will have each record's contents substituted in their place.
-    /// Note that timestamp is displayed using RFC3339, is always UTC and ignores system timezone.
-    ///
     /// For example, the following template string:
     ///
     /// Offset {{offset}} has key {{key}} and value {{value}}
@@ -588,48 +584,11 @@ impl ConsumeOpt {
             }
             (_, Some(templates)) => {
                 let value = String::from_utf8_lossy(record.value()).to_string();
-                let timestamp_rfc3339 = if record.timestamp() == NO_TIMESTAMP {
-                    "NA".to_string()
-                } else {
-                    format!(
-                        "{}",
-                        humantime::format_rfc3339_millis(
-                            UNIX_EPOCH
-                                + std::time::Duration::from_millis(
-                                    record.timestamp().try_into().unwrap_or_default()
-                                )
-                        )
-                    )
-                };
-
-                let human_readable_timestamp: String = if record.timestamp() == NO_TIMESTAMP {
-                    "NA".to_string()
-                } else {
-                    format!(
-                        "{}",
-                        humantime::format_duration(
-                            (UNIX_EPOCH
-                                + std::time::Duration::from_millis(
-                                    record.timestamp().try_into().unwrap_or_default()
-                                ))
-                            .elapsed()
-                            .unwrap_or_default()
-                        )
-                    )
-                    .split_whitespace()
-                    .take(3)
-                    .take_while(|s| !s.contains("us"))
-                    .collect::<Vec<&str>>()
-                    .join(" ")
-                };
-
                 let object = serde_json::json!({
                     "key": formatted_key,
                     "value": value,
                     "offset": record.offset(),
                     "partition": record.partition(),
-                    "timestamp": timestamp_rfc3339,
-                    "timestamp_humanize": human_readable_timestamp,
                 });
                 templates.render(USER_TEMPLATE, &object).ok()
             }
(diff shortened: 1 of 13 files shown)


crates/fluvio-dataplane-protocol/src/batch.rs
```diff
@@ -356,6 +357,14 @@ impl BatchHeader {
         let compression_bits = compression as i16 & COMPRESSION_CODEC_MASK;
         self.attributes = (self.attributes & !COMPRESSION_CODEC_MASK) | compression_bits;
     }
+
+    pub fn set_first_timestamp(&mut self, timestamp: i64) {
```
sehz, line 361: this should probably not make it public since it's internal

crates/fluvio-dataplane-protocol/src/batch.rs
```diff
@@ -356,6 +357,14 @@ impl BatchHeader {
...
+        self.first_timestamp = timestamp;
+    }
+
+    pub fn set_max_time_stamp(&mut self, timestamp: i64) {
```
sehz, line 365: same.  should be only crate wide

crates/fluvio-dataplane-protocol/src/record.rs
```diff
@@ -358,6 +358,14 @@ impl RecordHeader {
     pub fn offset_delta(&self) -> Offset {
         self.offset_delta
     }
+
+    pub fn set_timestamp_delta(&mut self, delta: i64) {
```
sehz, line 362: make it private or crate wide

crates/fluvio-dataplane-protocol/src/record.rs
```diff
@@ -537,6 +547,10 @@ impl<B> ConsumerRecord<B> {
     pub fn inner(&self) -> &B {
         &self.record
     }
+
+    pub fn timestamp(&self) -> i64 {
```
sehz, line 551: also, let's use type alias for timestamp

crates/fluvio/src/consumer.rs
```diff
@@ -160,14 +160,21 @@ where
             Err(e) => Either::Right(once(err(e))),
             Ok(batch) => {
                 let base_offset = batch.base_offset;
+                let batch_first_timestamp = batch.header.first_timestamp;
                 let records = batch.own_records().into_iter().enumerate().filter_map(
                     move |(relative, record)| {
                         let record_offset = base_offset + relative as i64;
+                        let timestamp = if batch_first_timestamp <= 0 {
```
sehz, line 167: this logic should be moved into protocol

crates/fluvio/src/producer/accumulator.rs
```diff
@@ -158,6 +159,14 @@ impl ProducerBatch {
     fn push_record(&mut self, mut record: Record) -> Option<PartialFutureRecordMetadata> {
         let relative_offset = self.records.len() as i64;
         record.preamble.set_offset_delta(relative_offset as Offset);
+        record.preamble.set_timestamp_delta(
```
sehz, line 162: should move to internal to batch

crates/fluvio/src/producer/accumulator.rs
```diff
@@ -207,6 +216,23 @@ impl TryFrom<ProducerBatch> for Batch<RawRecords> {
 
         header.set_compression(compression);
 
+        let first_timestamp = match p_batch.create_time.duration_since(SystemTime::UNIX_EPOCH) {
```
sehz, line 219: move to batch as internal algorithm.

crates/fluvio/src/producer/accumulator.rs
```diff
@@ -264,7 +290,7 @@ mod test {
 
         // Producer batch that can store three instances of Record::from(("key", "value"))
         let mut pb = ProducerBatch::new(
-            size * 3 + 1 + Batch::<RawRecords>::default().write_size(0),
+            size * 3 + 1 + Batch::<RawRecords>::default().write_size(0) + 4,
```
sehz, line 293: instead of hard coding 4, let's derived from data type

crates/fluvio-cli/src/consume/mod.rs
```diff
@@ -584,11 +588,48 @@ impl ConsumeOpt {
             }
             (_, Some(templates)) => {
                 let value = String::from_utf8_lossy(record.value()).to_string();
+                let timestamp_rfc3339 = if record.timestamp() == NO_TIMESTAMP {
+                    "NA".to_string()
+                } else {
+                    format!(
+                        "{}",
```
sehz, line 595: Instead of hard coding time values, can we use formatted time like here? https://rust-lang-nursery.github.io/rust-cookbook/datetime/parse.html.  We can also treat human time similarly

sehz (changes requested):
Looks great.   2 feedback:
1) move timestamp logic to batch so timestamp attribute can be hidden
2) more flexible way to format timestamp

crates/fluvio-cli/src/consume/mod.rs
morenol: We can, I just didn't want to add chronos as dependency, but if that is not a problem, I can proceed with that library

crates/fluvio-dataplane-protocol/src/batch.rs
morenol: Not sure how to set timestamp_delta without making this public. That is being set by accumulator, and it needs to know the batch creation timestamp

sehz: that can be a time of inserting the record.  first time stamp if the batch is empty otherwise it will be the max time stamp?

morenol: But all these things are done by the `BatchingProducer` struct that is in the fluvio crate.  It needs to set the timestamp_delta before inserting the record because timestamp_delta is varint and the size of the record (which is needed before inserting) could change because of that

crates/fluvio-cli/src/consume/mod.rs
morenol: Is that better than the actual RFC format?

I also can extract year, month, day, hour, minute and second, and then we could do something like:

`"{{year}}-{{month}}-{{day}}-{{hour}}-{{min}}-{{sec}}"`

sehz: Can we leverage https://docs.rs/chrono/0.4.19/chrono/format/strftime/index.html#specifiers?

sehz: Also, it would probably be better to split this PR since formatting make needs additional refinement.  We can merge API-related changes first.

sehz (changes requested):
Let's take the formatting part as separate PR so we can merge core stuff first.

sehz (approved):
LGTM

PR #2324, opened 2022-04-19
[Merged by Bors] - Improve error reporting in `SmartModule` runtime errors

### LINKED ISSUE #2286: Improve error reporting in smartmodule runtime errors
When there is an error in the smartmodule execution, we get an error like this one:

```
 Error:
   0: Dataplane error: Other error: SmartModule err wasm trap: wasm `unreachable` instruction executed
   1: Other error: SmartModule err wasm trap: wasm `unreachable` instruction executed
```

We may want to have better error reporting, so it is more clear the cause of the failure

galibey:
Currently, error reporting from user-defined `SmartModule` has two issues:
1. Conversion from any error to `eyre::Report` compiles but panics at runtime.
2. Panics lead to non-informative `wasm trap: wasm unreachable instruction executed`.

This PR fixes the first issue by enabling the feature `auto-install` on `eyre`. This feature enables support for the conversion of the most commonly used errors. Example of the error output after the fix:
```
Error:
   0: Dataplane error: SmartModule runtime error invalid type: integer `1`, expected a string at line 1 column 8

      SmartModule Info:
          Type: Filter
          Offset: 2
          Key: NULL
          Value: {"key":1}
   1: SmartModule runtime error invalid type: integer `1`, expected a string at line 1 column 8

      SmartModule Info:
          Type: Filter
          Offset: 2
          Key: NULL
          Value: {"key":1}
```

The second issue can't be fixed at this moment due to `wasm32-unknown-unknown` and `wasm32-wasi` targets do not support `panic=unwind`, only `panic=abort`. We can't catch panic. When it occurs, the execution just stopped and `wasmtime` engine just returns `Trap` with the message `wasm `unreachable` instruction executed`. So, there is no way to somehow enrich the current error reporting in case of panics.

Closes #2286

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2080,7 +2080,7 @@ dependencies = [
 
 [[package]]
 name = "fluvio-smartmodule"
-version = "0.2.3"
+version = "0.2.4"
 dependencies = [
  "eyre",
  "fluvio-dataplane-protocol",
--- a/crates/fluvio-smartmodule/Cargo.toml
+++ b/crates/fluvio-smartmodule/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio-smartmodule"
-version = "0.2.3"
+version = "0.2.4"
 edition = "2021"
 license = "Apache-2.0"
 authors = ["Fluvio Contributors <team@fluvio.io>"]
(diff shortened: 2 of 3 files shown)


sehz (approved):
LGTM

morenol (approved):

PR #2331, opened 2022-04-21
[Merged by Bors] - Add `std`, `derive` features of `clap`

galibey:
This change fixes the compilation error for `fluvio-extension-common` crate as it requires now these features.


morenol (approved):
Lgtm

PR #2334, opened 2022-04-25
[Merged by Bors] - Move to dataplane-protocol convertion from MemoryRecords to RawRecords

morenol:
This moves the conversion from RawRecords to MemoryRecords to the fluvio-dataplane-protocol.

Also creates a type alias for Timestamp and set default to NO_TIMESTAMP in batch headers

This is part of #2288


sehz (approved):
LGTM

Issue #2302, opened 2022-04-14
Producer must respect ReadCommitted isolation


sehz:
We need to support following isolation level for producer
* ReadUncommitted - Just wait for leader to write message (only wait for LEO update)
* ReadCommitted - Wait for messages to be committed (wait for HW)

This was implemented in previous version of Fluvio but somehow it got lost.  The isolation level must be exposed in the Producer Config.

galibey:
This is the current state of this feature:
- Isolation is the property of the Consumer. Specified in ConsumerConfig. It is used by the Stream handler and defines when records should be sent back to the consumer. 
- As stated in the description: 
     `ReadUncommitted` - Just wait for leader to write message (only wait for LEO update)
     `ReadCommitted` - Wait for messages to be committed (wait for HW)
- `fluvio-cli` doesn't support specifying the isolation, only Rust API client does. 

I did local testing with 3 spu and depending on isolation the latency of read/write operation differs significantly indicating that the feature works. Do we need some deep verification of the isolation guarantees?

sehz:
Yes. We should allow CLI to specify isolation level for Consumer and producer.

PR #2336, opened 2022-04-28
[Merged by Bors] - Support `ReadCommitted` isolation in SPU for Produce requests

galibey:
The produce handler uses the `acks` field from `ProduceRequest` to determine which isolation will be taken. 
In case of `ReadCommitted` isolation, it waits until the replica's `hw` includes records from the requests before returning the response. Awaiting is limited by the `timeout_ms` field of the request. For now, the timeout is not shared between partitions, it is counted separately for each partition. 

For `ReadUncommitted` there is no waiting.

Related #2302

### DIFF
--- a/crates/fluvio-dataplane-protocol/src/error_code.rs
+++ b/crates/fluvio-dataplane-protocol/src/error_code.rs
@@ -35,6 +35,9 @@ pub enum ErrorCode {
     #[fluvio(tag = 6)]
     #[error("the given SPU is not the leader for the partition")]
     NotLeaderForPartition,
+    #[fluvio(tag = 7)]
+    #[error("the request timed out.")]
+    RequestTimedOut,
     #[fluvio(tag = 10)]
     #[error("the message is too large to send")]
     MessageTooLarge,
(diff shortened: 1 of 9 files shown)


sehz (changes requested):
The first pass looks good.  

Meantime, can you add integration tests similar to fetch tests?

crates/fluvio-dataplane-protocol/src/error_code.rs
```diff
@@ -35,6 +35,9 @@ pub enum ErrorCode {
     #[fluvio(tag = 6)]
     #[error("the given SPU is not the leader for the partition")]
     NotLeaderForPartition,
+    #[fluvio(tag = 7)]
+    #[error("the request timed out.")]
+    RequestTimedOut,
```
sehz, line 40: Can we add timeout period and what kind?

crates/fluvio-spu/src/services/public/tests/produce.rs
```diff
@@ -197,3 +197,190 @@ async fn test_produce_invalid_compression() {
...
+        .send_and_receive(RequestMessage::new_request(produce_request))
+        .await
+        .expect("send offset");
+
+    // Check error code
+    assert_eq!(produce_response.responses.len(), 1);
+    assert_eq!(produce_response.responses[0].partitions.len(), 1);
+    assert_eq!(
+        produce_response.responses[0].partitions[0].error_code,
+        ErrorCode::None
```
sehz, line 380: Should check hw of the response to ensure hw is updated

sehz (changes requested):
Looks very good. Couple of minor comments

crates/fluvio-dataplane-protocol/src/error_code.rs
galibey: Added

crates/fluvio-spu/src/services/public/tests/produce.rs
galibey: There is no hw in the response, only `base_offset`. I added a check from the leader replica state.

sehz: sounds good

PR #2340, opened 2022-05-02
[Merged by Bors] - Add isolation to producer and consumer clis

### LINKED ISSUE #2302: Producer must respect ReadCommitted isolation

galibey:
Closes #2302


crates/fluvio-dataplane-protocol/src/common.rs
```diff
@@ -21,6 +22,25 @@ impl Default for Isolation {
     }
 }
 
+impl FromStr for Isolation {
```
sehz, line 25: Why put conversion in this crate?  Opinionated conversion (user interface) should be put in their respective area (CLI?)

crates/fluvio/src/producer/partition_producer.rs
```diff
@@ -210,8 +210,11 @@ impl PartitionProducer {
         }
 
         topic_request.partitions.push(partition_request);
-        request.acks = 1;
-        request.timeout_ms = 1500;
+        request.acks = match self.config.isolation {
```
sehz, line 213: Maybe we should change acks to Rust enum instead of doing conversion here.

sehz (changes requested):
Overall looks good.  Couple of comments

crates/fluvio/src/producer/partition_producer.rs
galibey: It will break Kafka wire protocol. Do we need to follow it? If no, then enum looks better here, I agree.

crates/fluvio-dataplane-protocol/src/common.rs
galibey: Due to Orphan rules, we can't impl `FromStr` anywhere else. We might put some custom conversion method in `fluvio-cli` but we may have needed it in other places in the future. Anyways, I don't have a strong opinion here, maybe just a method inside `fluvio-cli` crate is better for now. What would you recommend?

sehz: For now, just a method in the `fluvio-cli` since that's only used there.  If need to make it more usable, you can use https://doc.rust-lang.org/rust-by-example/generics/new_types.html to get around orphan rules.

crates/fluvio/src/producer/partition_producer.rs
sehz: We don't have to follow Kafka wire protocol.  You can still keep protocol compatible but use Rust Enum by just conforming to the same encoding and decoding behavior.  The only downside is bumping up the minor version number but I think it's worthwhile.  We should migrate legacy C or Kafka-style API to Rust native

galibey: Done

crates/fluvio-dataplane-protocol/src/common.rs
galibey: Added parse function to `fluvio-cli` crate.

crates/fluvio-dataplane-protocol/src/produce/request.rs
```diff
@@ -91,6 +79,99 @@ where
     pub records: R,
 }
 
+impl<R> Encoder for ProduceRequest<R>
```
sehz, line 82: Instead of writing custom encoder, you should directly implement encoder and decoder for duration in fluvio protocol crate

crates/fluvio-dataplane-protocol/src/produce/request.rs
```diff
@@ -91,6 +79,99 @@ where
...
+        self.isolation = Isolation::from(IsolationData::decode_from(src, version)?);
+        self.timeout = Duration::try_from(TimeoutData::decode_from(src, version)?)?;
+        self.topics = Decoder::decode_from(src, version)?;
+        Ok(())
+    }
+}
+
+/// Isolation is represented in binary format as i16 value (field `acks` in Kafka wire protocol).
+#[derive(Encoder, Decoder, FluvioDefault, Debug)]
+struct IsolationData(i16);
```
sehz, line 123: Same for isolation

sehz (changes requested):

crates/fluvio-dataplane-protocol/src/produce/request.rs
galibey: This encoding of duration and timeout is specific for produce requests only. For other places, it fairly might be done differently.

sehz: For simplicity, I think we should provide a default implementation.  Could always provide custom implementation later if needs arise.

sehz:
rebase

sehz (approved):
LGTM.  Nice job!

PR #2341, opened 2022-05-02
[Merged by Bors] - Add timestamp to `ConsumerRecord`

morenol:
This PRs adds timestamp() to ConsumerRecord API. This is the last protocol-related change needed to add timestamp formatting to CLI.

Part of #2288

### DIFF
--- a/crates/fluvio-dataplane-protocol/src/batch.rs
+++ b/crates/fluvio-dataplane-protocol/src/batch.rs
@@ -539,7 +539,6 @@ mod test {
     use crate::batch::Batch;
     use super::BatchHeader;
     use super::BATCH_HEADER_SIZE;
-    use super::memory::MemoryBatch;
 
     #[test]
     fn test_batch_size() {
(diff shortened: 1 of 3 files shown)


morenol:
> Can we leverage https://docs.rs/chrono/0.4.19/chrono/format/strftime/index.html#specifiers?
@sehz since we are using Handlebars, we need to pass {{ name }}, for every specifier that we want. Do we want to be able to specify them with something like `{{ Y }}`, `{{ C }}`? I think that for CLI it is enough to follow a specific RFC format

crates/fluvio-dataplane-protocol/src/batch.rs
```diff
@@ -226,6 +229,24 @@ impl Batch {
         }
         self.header.last_offset_delta = self.records().len() as i32 - 1;
     }
+
+    pub fn into_consumer_records_iter(
```
sehz, line 233: can you add unit tests?

crates/fluvio-dataplane-protocol/src/record.rs
```diff
@@ -560,6 +563,14 @@ impl ConsumerRecord<DefaultRecord> {
     pub fn value(&self) -> &[u8] {
         self.record.value().as_ref()
     }
+    /// Return the timestamp of the Record
```
sehz, line 566: add tests

sehz (changes requested):

crates/fluvio-dataplane-protocol/src/batch.rs
```diff
@@ -740,8 +760,35 @@ mod test {
...
 
+    #[test]
+    fn test_into_consumer_records_iter() {
+        let mut batch = Batch::from(vec![
+            Record::default(),
+            Record::default(),
+            Record::default(),
+        ]);
+
+        batch.header.first_timestamp = 1_500_000_000;
```
sehz, line 771: shouldn't this be computed instead of hardcoded based on initialization?  Or this should be put under different tests for it?

crates/fluvio-dataplane-protocol/src/batch.rs
```diff
@@ -740,8 +760,35 @@ mod test {
...
+        assert_eq!(consumer_records.len(), 3);
+        assert_eq!(consumer_records[0].offset(), 0);
+        assert_eq!(consumer_records[1].offset(), 1);
+        assert_eq!(consumer_records[2].offset(), 2);
+
+        consumer_records.iter().for_each(|record| {
+            assert_eq!(record.timestamp(), 1_500_000_000);
+            assert_eq!(record.partition, partition_id);
+        });
+    }
```
sehz, line 786: need to test if timestamp changes after putting new another batch

sehz (changes requested):

crates/fluvio-dataplane-protocol/src/batch.rs
morenol: That's part of MemoryBatch that has all the clock related stuff

crates/fluvio-dataplane-protocol/src/batch.rs
morenol: Timestamp creation is done in MemoryBatch tests, this is just testing that this function works as expected

sehz (approved):
Ok, LGTM.

PR #2344, opened 2022-05-03
[Merged by Bors] - Add timestamp to `fluvio consume --format`

morenol:
This PRs adds `{{time}}` to --format that will display the timestamp with RFC3339 format, and NA in case of no timestamp in record

``` fluvio consume topic --format "{{timestamp}}"
2022-05-03T19:12:56.980Z
2022-05-03T19:13:00.481Z
```

### DIFF
--- a/crates/fluvio-cli/src/consume/mod.rs
+++ b/crates/fluvio-cli/src/consume/mod.rs
@@ -82,7 +82,7 @@ pub struct ConsumeOpt {
     /// Provide a template string to print records with a custom format.
     /// See --help for details.
     ///
-    /// Template strings may include the variables {{key}}, {{value}}, {{offset}}, {{partition}} and {{timestamp}}
+    /// Template strings may include the variables {{key}}, {{value}}, {{offset}}, {{partition}} and {{time}}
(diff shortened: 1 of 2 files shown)


crates/fluvio-cli/src/consume/mod.rs
```diff
@@ -584,11 +588,26 @@ impl ConsumeOpt {
...
+                        )
+                    )
+                };
+
                 let object = serde_json::json!({
                     "key": formatted_key,
                     "value": value,
                     "offset": record.offset(),
                     "partition": record.partition(),
+                    "timestamp": timestamp_rfc3339,
```
sehz, line 610: how about `time` rather than `timestamp`?

sehz (changes requested):

crates/fluvio-cli/src/consume/mod.rs
morenol: I am ok with both options

sehz: Yes, its less typing 😉

sehz (approved):
LGTM

PR #2349, opened 2022-05-05
Support configuring compression level

### LINKED ISSUE #2248: Add support for compression levels
Some compression algorithms have compression levels to define the compression quality , we can add that support to the producer configuration

mtopolnik:
Fixes #2248.

Of the currently used compression libraries, I found only Gzip supports setting a compression level. It allows levels 1..9. Snappy doesn't support any levels, and LZ4 generally does support them, but the Rust wrapper doesn't. This may be for a good reason, because almost all the levels LZ4 supports use the HC variant, which is much slower than the main variant.

The proposed code supports levels 1..9, plus "0" which means the undelying algorithm's default level.

Since Fluvio encodes the compression format + level in a bitfield of the batch header, I'm not sure if supporting this many levels is necessary, since it uses additional 4 bits. Maybe we could support only two or three levels (max ratio, max compression, balanced).

### DIFF
--- a/crates/fluvio-compression/src/lib.rs
+++ b/crates/fluvio-compression/src/lib.rs
@@ -14,10 +14,10 @@ use serde::{Serialize, Deserialize};
 #[serde(rename_all = "lowercase")]
 #[repr(i8)]
 pub enum Compression {
-    None = 0,
-    Gzip = 1,
-    Snappy = 2,
-    Lz4 = 3,
+    None,
+    Gzip(GzipLevel),
+    Snappy,
+    Lz4,
 }
 
 impl Default for Compression {
@@ -31,13 +31,25 @@ impl TryFrom<i8> for Compression {
     fn try_from(v: i8) -> Result<Self, CompressionError> {
         match v {
             0 => Ok(Compression::None),
-            1 => Ok(Compression::Gzip),
+            1 => Ok(Compression::Gzip(GzipLevel::default())),
             2 => Ok(Compression::Snappy),
             3 => Ok(Compression::Lz4),
             _ => Err(CompressionError::UnknownCompressionFormat(format!(
-                "i8 representation: {}",
+                "u8 representation: {}",
                 v
             ))),
+        }
+    }
+}
+
+impl From<Compression> for u8 {
+    fn from(compression: Compression) -> Self {
+        use Compression::*;
+        match compression {
+            None => 0,
+            Gzip(_) => 1,
+            Snappy => 2,
+            Lz4 => 3,
         }
     }
 }
@@ -48,7 +60,7 @@ impl FromStr for Compression {
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match s.to_lowercase().as_str() {
             "none" => Ok(Compression::None),
-            "gzip" => Ok(Compression::Gzip),
+            "gzip" => Ok(Compression::Gzip(GzipLevel::default())),
             "snappy" => Ok(Compression::Snappy),
             "lz4" => Ok(Compression::Lz4),
             _ => Err(CompressionError::UnknownCompressionFormat(s.into())),
@@ -58,14 +70,10 @@ impl FromStr for Compression {
 
 impl Compression {
     /// Compress the given data, returning the compressed data
-    pub fn compress(
-        &self,
-        src: &[u8],
-        level: CompressionLevel,
-    ) -> Result<Vec<u8>, CompressionError> {
+    pub fn compress(&self, src: &[u8]) -> Result<Vec<u8>, CompressionError> {
         match *self {
             Compression::None => Ok(src.to_vec()),
-            Compression::Gzip => gzip::compress(src, level),
+            Compression::Gzip(level) => gzip::compress(src, level),
             Compression::Snappy => snappy::compress(src),
             Compression::Lz4 => lz4::compress(src),
         }
@@ -75,7 +83,7 @@ impl Compression {
     pub fn uncompress(&self, src: &[u8]) -> Result<Option<Vec<u8>>, CompressionError> {
         match *self {
             Compression::None => Ok(None),
-            Compression::Gzip => {
+            Compression::Gzip(_) => {
                 let output = gzip::uncompress(src)?;
                 Ok(Some(output))
             }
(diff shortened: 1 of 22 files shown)


crates/fluvio-compression/src/lib.rs
```diff
@@ -58,10 +58,14 @@ impl FromStr for Compression {
 
 impl Compression {
     /// Compress the given data, returning the compressed data
-    pub fn compress(&self, src: &[u8]) -> Result<Vec<u8>, CompressionError> {
+    pub fn compress(
```
morenol, line 61: I think that if we wrap CompressionLevel internally in Compression type, we would not need to add level to `compress()`.

Maybe:
```rust
pub struct Compression {
   kind: CompressionAlgorithm,
   level: Option<CompressionLevel>
 }
// or:
 pub enum Compression {
   None,
   Gzip(CompressionLevel),
   Snappy,
    Lz4
 }
 ```
 
 With that we won't need to pass compression_lvel to Batch structs. We would have to change the encoder/decoder implementation to only consider the  compression algorithm

File: crates/fluvio-compression/src/lib.rs [hunk 2]
```diff
@@ -58,10 +58,14 @@ impl FromStr for Compression {
 
 impl Compression {
```
morenol, line 59: I think that we need to bump this crate to `0.2.0`

crates/fluvio-compression/src/lib.rs
mtopolnik: With the second approach we can specialize `CompressionLevel` to correspond exactly to what Gzip offers, and we can also generalize the parameter to more than just compression level. LZ4, for example, has some other compression options, like choosing the max block size and whether the compression dictionary is carried over to the next block.

mtopolnik: A problem I see with non-unit enum variants is that you can't use the explicit discriminant with them.

crates/fluvio-compression/src/lib.rs
```diff
@@ -90,9 +102,95 @@ impl std::fmt::Display for Compression {
...
+            9 => Ok(Level9),
+            _ => Err(CompressionError::UnknownCompressionFormat(format!(
+                "i8 representation: {}",
+                v
+            ))),
+        }
+    }
+}
+
+impl TryFrom<GzipLevel> for flate2::Compression {
```
galibey, line 175: There is no need to have fallible conversion here. `GzipLevel` is local type here and can always be convertible to `flate2::Compression`. If we implement `From` instead of `TryFrom`, it will be more convenient to use.

crates/fluvio-compression/src/lib.rs
```diff
@@ -90,9 +102,88 @@ impl std::fmt::Display for Compression {
...
+            ))),
+        }
+    }
+}
+
+impl From<GzipLevel> for flate2::Compression {
+    fn from(level: GzipLevel) -> Self {
+        let int_level = level as u32;
+        if int_level < 1 || int_level > 9 {
+            panic!("Invalid GzipLevel discriminant: {int_level}")
```
galibey, line 179: It is not safe to panic here. If someone forgets and adds `level10` it can kill the process in runtime. Better to cover all variants in `match`, and the correspondence will be controlled during compilation time.

galibey (approved):
LGTM

sehz:
@morenol ?

crates/fluvio/src/producer/mod.rs
```diff
@@ -297,28 +297,32 @@ impl TopicProducer {
             match topic_spec.get_compression_type() {
                 CompressionAlgorithm::Any => config.compression.unwrap_or_default(),
                 CompressionAlgorithm::Gzip => match config.compression {
-                    Some(Compression::Gzip) | None => Compression::Gzip,
+                    None => Compression::Gzip(GzipLevel::default()),
```
morenol, line 300: I think that here you can use `Default::default()`, which won't require code changes if change the inner elements of this enum variant

crates/fluvio-spu/src/services/public/produce_handler.rs
```diff
@@ -154,7 +154,9 @@ fn validate_records<R: BatchRecords>(
         match compression {
             CompressionAlgorithm::Any => true,
             CompressionAlgorithm::None => batch_compression == Compression::None,
-            CompressionAlgorithm::Gzip => batch_compression == Compression::Gzip,
+            CompressionAlgorithm::Gzip => {
+                batch_compression == Compression::Gzip(GzipLevel::default())
```
morenol, line 158: probably here, we could use `matches!( batch_compression, Compression::Gzip(_))`

crates/fluvio-dataplane-protocol/src/batch.rs
```diff
@@ -360,7 +360,7 @@ pub struct BatchHeader {
     pub partition_leader_epoch: i32,
     pub magic: i8,
     pub crc: u32,
-    pub attributes: i16,
+    pub attributes: u16,
```
morenol, line 363: Is there an specific reason to make it `u16`. I think that this change is good to have and probably harmless but not sure if this should be in this PR

morenol (changes requested):
LGTM.

I added a couple of comments, not sure about the change from `i16` to `u16`.

Alse we need to bump `fluvio` and `fluvio-dataplane-protocol` versions

crates/fluvio-dataplane-protocol/src/batch.rs
sehz: I agree, let's reduce changes

mtopolnik: Yeah, this change wasn't a must, it just made more sense to deal with unsigned values in combination with shifting and masking operations. The choice of signed int also propagates to conversions between `Compression` and primitive values.

I'll revert it to keep the PR focused.

morenol (changes requested):
Please rebase and add to Changelog

morenol (commented):
LGTM, 

do you know an input data that I can use where I can see different sizes by using different gzip levels? So far all the inputs that I have tried provides similar results in terms of size after compression.

Also, with the latest Cargo.toml, these crates need a bump 
```
* fluvio-controlplane-metadata
* fluvio-extension-common
* fluvio-sc-schema
* fluvio-smartengine
* fluvio-smartmodule
* fluvio-spu-schema
```

PR #2353, opened 2022-05-06
[Merged by Bors] - Increase storage max batch size

### LINKED ISSUE #2342: Need the ability to produce/consume messages beyond 1MB in size.
Currently, when a client producer a record of 5881006 bytes, the producer panics:

```
I have a rust producer and pass a  JSON string. It results in an error: Record too large. What's wrong?
thread 'main' panicked at 'Failed to send message.: Producer(RecordTooLarge(16384))', src/main.rs:39:9
```

morenol:
Fixes #2342

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Release Notes
 
 ## Platform Version 0.9.26 - UNRELEASED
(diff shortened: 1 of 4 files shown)


sehz (approved):
LGTM

PR #2354, opened 2022-05-07
Add cross compat test to CI

tjtelan:
From CD_Dev, but use CI artifacts. Using channels to manage cluster and CLI versions


sehz (approved):
LGTM

PR #2383, opened 2022-05-16
Force 1 fetch stream per consumer and replica

### LINKED ISSUE #2367: Starting a Consumer Strream with the same id should fail if used by another stream
This task also includes knowing when a stream is active and when a stream ends, denoting that the old stream is inactive.

This is a subtask of https://github.com/infinyon/fluvio/issues/2366.

galibey:
This PR adds extra information to stream publishers, allowing validation if there are no duplicated streams per consumer and replica.
Also, deleting of finished streams moved up to prevent dangled streams publishers. A stream dangled due to early returns if errors occur.

This is a draft version to validate the idea.

Fixes #2367

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -725,9 +725,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "3.1.17"
+version = "3.1.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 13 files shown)


galibey:
I have added removing spu-side structures of consumer streams when streams are dropped on the client-side.

PR #2389, opened 2022-05-18
[Merged by Bors] - Add fish support to install.sh

joshchngs:
I recently installed Fluvio from the online installer, but I felt left out when it came time to adding the bindir to my path.

The first commit is the important one, the others just clean the output up slightly, now there are 3 variants.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -141,12 +141,12 @@ remind_path() {
     say "💡 You'll need to add '~/.fluvio/bin/' to your PATH variable"
     say "    You can run the following to set your PATH on shell startup:"
     
-    # shellcheck disable=SC2016
-    local bash_hint='echo '\''export PATH="${HOME}/.fluvio/bin:${PATH}"'\'' >> ~/.bashrc'
-    # shellcheck disable=SC2016
-    local zsh_hint='echo '\''export PATH="${HOME}/.fluvio/bin:${PATH}"'\'' >> ~/.zshrc'
-    # shellcheck disable=SC2016
-    local fish_hint='fish_add_path "$HOME/.fluvio/bin"'
+    # shellcheck disable=SC2016,SC2155
+    local bash_hint="$(tput bold)"'echo '\''export PATH="${HOME}/.fluvio/bin:${PATH}"'\'' >> ~/.bashrc'"$(tput sgr0)"
+    # shellcheck disable=SC2016,SC2155
+    local zsh_hint="$(tput bold)"'echo '\''export PATH="${HOME}/.fluvio/bin:${PATH}"'\'' >> ~/.zshrc'"$(tput sgr0)"
+    # shellcheck disable=SC2016,SC2155
+    local fish_hint="$(tput bold)"'fish_add_path "$HOME/.fluvio/bin"'"$(tput sgr0)"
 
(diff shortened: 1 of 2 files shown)


sehz:
Thanks for contribution

install.sh
```diff
@@ -140,14 +140,30 @@ verify_checksum() {
...
+
+    case "$(basename "${SHELL}")" in
+        bash)
+            say "      ${bash_hint}"
+            ;;
+        zsh)
+            say "      ${zsh_hint}"
+            ;;
+        fish)
+            say "      ${bash_hint}"
```
tjtelan, line 159: Typo: `fish_hint`

tjtelan (changes requested):
This works as expected using `bash` and `zsh`, but it doesn't seem to work at all with `fish`.

In each of my examples, the user has the variant shell set as their login, to ensure the `$SHELL` var is set correctly. This should always return the same value even if you were to pipe the script to a different shell. 

bash
```
bashy@2ad9c674bbea:~$ echo $SHELL
/bin/bash
bashy@2ad9c674bbea:~$ cat install.sh | bash
[...]
fluvio:       echo 'export PATH="${HOME}/.fluvio/bin:${PATH}"' >> ~/.bashrc
```

zsh
```
2ad9c674bbea% echo $SHELL
/usr/bin/zsh
2ad9c674bbea% cat install.sh|zsh
[...]
fluvio:       echo 'export PATH="${HOME}/.fluvio/bin:${PATH}"' >> ~/.zshrc
```

For me, the script has an error if I try to execute w/ `fish`
```
fishy@2ad9c674bbea ~> echo $SHELL
/usr/bin/fish

fishy@2ad9c674bbea ~> cat install.sh | fish
fish: Unexpected ')' found, expecting '}'
        x86_64-unknown-linux-gnu)
                                ^
```

This would print `fish_hint`, but there's a little typo.

```
fishy@2ad9c674bbea ~ [0|127]> cat install.sh | bash
[...]
fluvio:       echo 'export PATH="${HOME}/.fluvio/bin:${PATH}"' >> ~/.bashrc

fishy@2ad9c674bbea ~> cat install.sh | zsh
[...]
fluvio:       echo 'export PATH="${HOME}/.fluvio/bin:${PATH}"' >> ~/.bashrc
```

When I changed L159 to `fish_hint`, this does print correctly, however. 

```
fishy@2ad9c674bbea ~> cat install.sh | zsh
[...]
fluvio:       fish_add_path "$HOME/.fluvio/bin"
```

install.sh
joshchngs: Oops! That's from testing - fixed.

joshchngs:
This isn't intended to execute under `fish` - that wouldn't work. The install flow should stay the same - `curl | bash`.

The change is just that the PATH-setting instructions are emitted differently. Here's how it looks for me (on fish):
Screenshot 2022-05-19 at 09 04 12
:

tjtelan (approved):

sehz:
can you add to changelog?

joshchngs:
Sure. I didn't make an issue like the other entries - just link to this PR?

sehz (approved):
Great.  thanks

PR #2392, opened 2022-05-24
chore(typo): Rename partition's `lsr` -> `lrs`

tjtelan:
Found old TODO for a typo fix on next minor version bump
Basically this is redoing #1298

### DIFF
--- a/crates/fluvio-controlplane-metadata/src/partition/status.rs
+++ b/crates/fluvio-controlplane-metadata/src/partition/status.rs
@@ -29,10 +29,8 @@ use super::ElectionScoring;
 pub struct PartitionStatus {
     pub resolution: PartitionResolution,
     pub leader: ReplicaStatus,
-    // TODO: Next time we make a breaking protocol change, rename this to `lrs`
-    // TODO: There is no such thing as `lsr`, it is a typo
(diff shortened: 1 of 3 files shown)


crates/fluvio-controlplane-metadata/src/partition/status.rs
```diff
@@ -114,9 +113,8 @@ impl PartitionStatus {
         self.resolution != PartitionResolution::Online
     }
 
-    // TODO: At next breaking change, deprecate this and replace with `fn lrs` to fix typo
-    pub fn lsr(&self) -> u32 {
```
morenol, line 118: Can we keep this with the deprecated attribute?
```rust
         #[deprecated = "Replace by lrs"]
```

crates/fluvio-controlplane-metadata/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "fluvio-controlplane-metadata"
 edition = "2021"
-version = "0.15.4"
+version = "0.16.0"
```
morenol, line 4: I think that with the `deprecated` attribute, we don't need to do major bump

Issue #2361, opened 2022-05-10
Producer Retries


galibey:
We need to implement initial retry logic for the Producer. We will distinguish all errors into two categories: `retriable` and `nonretriable`.
There can be four groups of errors:
1. Transport errors (including IO) - `retriable`.
2. Producer or system configuration issues - `nonretriable`.
3. SPU responded with the error - `retriable` (not always, depending on the error).
4. SPU or topic configuration issue - `nonretriable`.

Two parameters can configure the retry logic:
1. Integer parameter `retries` - how many times to re-send the record to SPU in case of errors. 0-no retries.
2. `delivery.timeout.ms` - If retries are enabled, it defines the upper time bound for all retries.

galibey:
Will be implemented in https://github.com/infinyon/fluvio/issues/2481

PR #2397, opened 2022-05-26
Add generic trait impl for everything that dereferences to `Request`

galibey:
Added generic impl of `Request` trait for everything that dereferences to `Request`.

Also, impl `Decoder` and `Encoder` for Arc<T: Decoder + Encoder>.

These changes allow using `Arc<any Request>` in retries without cloning the data.

Related to #2361

### DIFF
--- a/crates/fluvio-protocol/src/core/decoder.rs
+++ b/crates/fluvio-protocol/src/core/decoder.rs
@@ -4,6 +4,7 @@ use std::io::Error;
 use std::io::ErrorKind;
 use std::io::Read;
 use std::marker::PhantomData;
+use std::sync::Arc;
 
 use bytes::Buf;
 use bytes::BufMut;
@@ -400,6 +401,21 @@ impl DecoderVarInt for Option<Vec<u8>> {
     }
 }
 
+impl<M> Decoder for Arc<M>
+where
+    M: Decoder,
+{
+    fn decode<T>(&mut self, src: &mut T, version: Version) -> Result<(), Error>
+    where
+        T: Buf,
+    {
+        let mut_ref = Arc::get_mut(self).ok_or_else(|| {
+            Error::new(ErrorKind::Other, "not exclusive reference is being decoded")
+        })?;
+        M::decode(mut_ref, src, version)
+    }
+}
+
 #[cfg(test)]
 mod test {
 
@@ -409,6 +425,7 @@ mod test {
     use bytes::Buf;
     use std::io::Cursor;
     use std::io::Error;
+    use std::sync::Arc;
 
     #[test]
     fn test_decode_i18_not_enough() {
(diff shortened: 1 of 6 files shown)


morenol (approved):
LGTM!

crates/fluvio-protocol/src/core/decoder.rs
```diff
@@ -400,6 +401,21 @@ impl DecoderVarInt for Option<Vec<u8>> {
     }
 }
 
+impl<M> Decoder for Arc<M>
```
sehz, line 404: Do we need this? This should be done transparently by Arc since it implements Deref.

sehz (changes requested):
Adding Deref sounds reasonable, not sure about adding a special case for Arc.  Can you test without it?

crates/fluvio-protocol/src/core/decoder.rs
galibey: It is not necessary for Arc<R:Encoder> variable because it will be coerced to R that impl Encode, that's correct. But trait bound `Arc<R>: Encoder`needs also for Arc<R:Request> to impl Request.

sehz: Problem is that this assumes there is one owner since it performs `get_mut` which doesn't seem to make sense for decoding.   Instead, in this case, it's better to have the caller have an explicit conversion, so this can be written as:
```
   let data = Arc::get_mut(self).
   let some = data.decode(....)
```

sehz (changes requested):

crates/fluvio-protocol/src/core/decoder.rs
galibey: This is done also for `Arc<R:Request>` to impl `Request` because `Request` requires `Encoder` and `Decoder`. 
It is very unlikely that Arc will be shared at the time of decoding. Usually, decoding happens right after instantiation. But if it happens, we provide error, so it can be identified and fixed.

sehz: What's the use case for decoding directly from Arc? This seems to be an unsafe operation.  Either we should wrap with a Lock or another kind of atomic operation such that it's safe to access underlying data.

galibey: For example, we have a method with the following signature:
```
async fn send_receive<R>(&self, request: R) where R: Request {...}
```
and we have some request:
```
struct ProduceRequest<R> where R: Encoder + Decoder + Default + Debug
```
Now with that changes, we can do:
```
let request = Arc::new(ProduceRequest{...});
socker.send_receive(request.clone())
```
It becomes useful in context with retries. We don't need to clone request, just clone arc reference:
```
let request = Arc::new(ProduceRequest{...});
retry(..., || socker.send_receive(request.clone())).await;
```

galibey: It is not unsafe. If the error occurs, it is just IO error that we return, nothing special. But, if it happens it is the developer's mistake, nothing external can incur it. So, we will just fix it.

galibey:
Implemented another approach in https://github.com/infinyon/fluvio/pull/2501

PR #2415, opened 2022-06-13
POC: SmartModule with jolt template support

sehz:


crates/fluvio-cli/src/smartmodule/generate.rs
```diff
@@ -0,0 +1,115 @@
...
+    Ok(template_path)
+}
+
+#[derive(ArgEnum, Debug, Clone, PartialEq)]
+#[allow(non_camel_case_types)]
+pub enum DataType {
+    json,
+}
+
+impl Default for DataType {
```
tomuxmon, line 35: Starting with [rust 1.62](https://blog.rust-lang.org/2022/06/30/Rust-1.62.0.html) you can use derive default on enum variants. A bit nicer to read IMHO.
 ```rust
#[derive(ArgEnum, Default, Debug, Clone, PartialEq)]
#[allow(non_camel_case_types)]
pub enum DataType {
    #[default]
    json,
}
```

sehz:
obsolete by `smdk`

PR #2424, opened 2022-06-18
[Merged by Bors] - Add perf reporting to client Producer

### LINKED ISSUE #2423: Add performance-related probes in Producer client
In the `fluvio` crate, the producer should collect session statistics for:

* throughput (latency x time)
* bandwidth (MB/s)
* resource usage (cpu, memory)

Additionally, provide it to the CLI with a sticky progress bar, like in the consumer CLI.

tjtelan:
Resolves #2423 

Added `ClientStats` and `ClientStatsUpdate` to `fluvio` crate and integrated into `TopicProducer` with latency, throughput, memory and cpu stats. This behavior is configurable through `TopicProducerConfig` using  `ClientStatsDataCollect` enum.

Added user facing `ClientStatsDataPoint` for display of stats collection. Used in fluvio-cli.

Exposed functionality in the CLI with `--stats`, `--stats-plus`, `--stats-path` options

e.g. `fluvio produce <topic> --stats`

* `--stats`
  * This configures the producer for collecting data for throughput and latency reporting
* `--stats-plus`
  * This spawns a separate thread to poll the system for CPU and memory usage.
* `--stats-path`
  * When combined with `--stats`- or `--stats-plus`, this will write the datapoints to file in csv format for loading into spreadsheet
* `--no-stats-bar`
  * Prevent display and update of stats bar. Useful when in combination with `--stats-path`
* `--stats-summary`
  * Implies `--stats`. Don't print live stats. Only the summary.


crates/fluvio/src/producer/partition_producer.rs
```diff
@@ -25,6 +38,7 @@ pub(crate) struct PartitionProducer {
     batches_lock: Arc<Mutex<VecDeque<ProducerBatch>>>,
     batch_events: Arc<BatchEvents>,
     last_error: Arc<RwLock<Option<ProducerError>>>,
+    client_stats: Option<Arc<RwLock<ClientStats>>>,
```
sehz, line 41: The internal stat counter should use Atomic to be highly performant, Lock will be too slow.

sehz (changes requested):

sehz:
rebase

crates/fluvio/src/producer/partition_producer.rs
```diff
@@ -198,11 +224,17 @@ impl PartitionProducer {
...
+
         for p_batch in batches_ready {
             let notify = p_batch.notify.clone();
             let batch = p_batch.batch();
 
-            let raw_batch = batch.try_into()?;
+            let raw_batch: Batch<RawRecords> = batch.try_into()?;
+
+            if self.client_stats.is_some() {
+                flush_total_bytes += BATCH_FILE_HEADER_SIZE + raw_batch.records().0.len();
```
sehz, line 236: this seems like magic.  raw_batch should return len instead of computing here

crates/fluvio/src/producer/partition_producer.rs
```diff
@@ -213,23 +245,83 @@ impl PartitionProducer {
         request.isolation = self.config.isolation;
         request.timeout = self.config.timeout;
         request.topics.push(topic_request);
-        let response = spu_socket.send_receive(request).await?;
+
+        let (response, stats_update) = if self.client_stats.is_some() {
+            Self::measure_send_receive(&spu_socket, request).await?
```
sehz, line 250: This would be more cleaner like this:
```
if let Some(stat) = self.client.stats {
     stat.measure_send_receive(&spu_socket,request).await?
}
```

crates/fluvio/src/producer/partition_producer.rs
```diff
@@ -198,11 +224,17 @@ impl PartitionProducer {
...
         for p_batch in batches_ready {
             let notify = p_batch.notify.clone();
             let batch = p_batch.batch();
 
-            let raw_batch = batch.try_into()?;
+            let raw_batch: Batch<RawRecords> = batch.try_into()?;
+
+            if self.client_stats.is_some() {
+                flush_total_bytes += BATCH_FILE_HEADER_SIZE + raw_batch.records().0.len();
+            }
```
sehz, line 237: It would be more clear like this:
```
if let Some(stat) = self.client_stat. {
   state.update_flush_bytes(raw.batch_len());
}
```

crates/fluvio/src/producer/partition_producer.rs
```diff
@@ -213,23 +245,83 @@ impl PartitionProducer {
...
             {
                 trace!("Failed to notify produce result because receiver was dropped");
             }
+
             if fluvio_error.is_error() {
                 return Err(FluvioError::from(ProducerError::from(fluvio_error)));
             }
+
+            // Update stats
+            if let Some(client_stats) = &self.client_stats {
```
sehz, line 275: see comment above

crates/fluvio/src/producer/mod.rs
```diff
@@ -190,6 +201,20 @@ impl InnerTopicProducer {
             .push_record(record, partition)
             .await?;
 
+        // Increase record count stat
+        if let Some(client_stats) = &self.client_stats {
```
sehz, line 205: not sure what this does?

sehz (changes requested):
few more comments

crates/fluvio/src/producer/mod.rs
tjtelan: If we've configured the producer to collect stats, we count each record as we push them into the batch.

Atomically updating the struct through the `ArcSwap` takes a little bit of digging. I can bring this into the `ClientStats` struct clean this up.

crates/fluvio/src/producer/partition_producer.rs
tjtelan: I was calculating `batch_len` because it was getting lost in the type conversion. This is fixed in `Batch` now.

crates/fluvio/src/producer/partition_producer.rs
tjtelan: This is fixed. There were bugs in the batch type conversions where we were losing the batch length.

crates/fluvio/src/producer/partition_producer.rs
```diff
@@ -213,23 +248,44 @@ impl PartitionProducer {
...
                 return Err(FluvioError::from(ProducerError::from(fluvio_error)));
             }
         }
 
+        // Commit stats update
+        if let Some(client_stats) = self.client_stats.as_ref() {
+            ClientStats::shared_update(
+                client_stats,
+                client_stats_update.offset(Some(base_offset as i32)),
+            );
```
sehz, line 286: can it be written as:
```
client.stats.update(base_offset))
```

sehz (changes requested):

crates/fluvio-dataplane-protocol/src/batch.rs
```diff
@@ -156,6 +156,11 @@ impl<R> Batch<R> {
         self.header.decode(src, version)?;
         Ok(())
     }
+
+    /// Return the size of the batch header + records
+    pub fn batch_len(&self) -> i32 {
```
sehz, line 161: add unit test

crates/fluvio-dataplane-protocol/src/batch.rs
```diff
@@ -499,7 +505,10 @@ pub mod memory {
 
     impl From<MemoryBatch> for Batch<MemoryRecords> {
         fn from(p_batch: MemoryBatch) -> Self {
-            let mut batch = Self::default();
+            let mut batch = Self {
+                batch_len: p_batch.current_size_uncompressed as i32,
+                ..Default::default()
```
sehz, line 510: need unit test here

sehz (changes requested):
Overall looks very good.  Just a couple of nits left.

sehz:
Also, need CLI Test to see if the stat is correct.

sehz:
rebase

crates/fluvio/src/producer/partition_producer.rs
```diff
@@ -212,27 +239,42 @@ impl PartitionProducer {
...
-                .flat_map(|response| response.partitions),
-        ) {
-            let base_offset = partition_response.base_offset;
-            let fluvio_error = partition_response.error_code.clone();
+        } else {
+            spu_socket.send_receive(request).await?
+        };
+
+        let mut base_offset = 0;
+        for (batch_notifier, response) in batch_notifiers.into_iter().zip(response.responses.iter())
```
morenol, line 255: could you change this to
```suggestion
         for (batch_notifier, partition_response) in batch_notifiers.into_iter().zip(responses.into_iter().flat_map(|response| response.partitions))
```

crates/fluvio/src/producer/partition_producer.rs
```diff
@@ -212,27 +239,42 @@ impl PartitionProducer {
...
+        {
+            base_offset = response.partitions[0].base_offset;
```
morenol, line 257:
```suggestion
            base_offset = partition_response.base_offset;
```

crates/fluvio/src/producer/partition_producer.rs
```diff
@@ -212,27 +239,42 @@ impl PartitionProducer {
...
+            let fluvio_error = response.partitions[0].error_code.clone();
```
morenol, line 258:
```suggestion
            let fluvio_error = partition_response.error_code.clone();
```

crates/fluvio/src/stats/data_point.rs
```diff
@@ -0,0 +1,316 @@
...
+        }
+
+        if (self.stats_collect() == ClientStatsDataCollect::All)
+            || (self.stats_collect() == ClientStatsDataCollect::System)
+        {
+            header.extend(system_related);
+        }
+
+        // Render header as CSV w/ trailing newline
+        header.push("\n".to_string());
```
digikata, line 200: These two lines end up creating the header line with a trailing comma.  Appending the "\n" to the joined string would avoid that.

crates/fluvio/src/stats/data_point.rs
```diff
@@ -0,0 +1,316 @@
...
+
+        if (self.stats_collect() == ClientStatsDataCollect::All)
+            || (self.stats_collect() == ClientStatsDataCollect::System)
+        {
+            data_point.extend(system_related);
+        }
+
+        // Render header as CSV w/ trailing newline
+        data_point.push("\n".to_string());
+        Ok(data_point.join(","))
```
digikata, line 246: These two lines end up creating the data line with a trailing comma.  Appending the "\n" to the joined string would avoid that.

crates/fluvio-dataplane-protocol/src/batch.rs
```diff
@@ -150,12 +150,17 @@ impl<R> Batch<R> {
     where
         T: Buf,
     {
```
sehz, line 152: need to bump up version

crates/fluvio/Cargo.toml
```diff
@@ -40,7 +40,11 @@ siphasher = "0.3.5"
 cfg-if = "1.0.0"
 derive_builder = "0.10"
 async-trait = "0.1.51"
-
+chrono = "0.4"
```
sehz, line 43: These dependencies should be under separate feature like `stat`?

sehz (changes requested):
Looks couple.  Couple of comments.

crates/fluvio/src/producer/partition_producer.rs
```diff
@@ -212,27 +244,46 @@ impl PartitionProducer {
...
             }
+
             if fluvio_error.is_error() {
                 return Err(FluvioError::from(ProducerError::from(fluvio_error)));
             }
         }
 
+        // Commit stats update
+        if self.client_stats.is_collect(ClientStatsDataCollect::Data) {
+            client_stats_update.push(ClientStatsMetricRaw::Offset(base_offset as i32));
```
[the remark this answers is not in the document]
tjtelan: This is true, the last offset update is what is what is recorded. The offset in this case is purely for reporting purposes, and not for calculations. I might ask what is your expected behavior for this case? My intent was to reflect the status of the producer after flush is complete.

I haven't tested with multiple partitions yet, so my concern is about whether we accurately capture the volume of traffic sending to all partitions.

crates/fluvio/Cargo.toml
tjtelan: I ended up getting rid of chrono. I was only using it in one place for timestamps, so I replaced it

crates/fluvio/src/producer/partition_producer.rs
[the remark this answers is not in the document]
tjtelan: I'm testing on a 1 node cluster, but I wasn't able to verify that the client sends records to all partitions in a single request using `iftop`

One partition
```
$ fluvio topic create one -p 1
topic "one" created
$ fluvio produce one -f /usr/share/dict/american-english
```

```
TX:             cum:   1.65MB   peak:   6.59Mb      rates:      0b   1.32Mb   422Kb
RX:                    20.4KB           81.7Kb                  0b   16.3Kb  5.10Kb
TOTAL:                 1.67MB           6.67Mb                  0b   1.33Mb   427Kb
```

Ten partitions
```
$ fluvio topic create ten -p 10
topic "ten" created
$ fluvio produce ten -f /usr/share/dict/american-english
```


```
TX:             cum:   1.67MB   peak:   6.68Mb      rates:      0b   1.34Mb  1.34Mb
RX:                    27.9KB            112Kb                  0b   22.3Kb  22.3Kb
TOTAL:                 1.70MB           6.79Mb                  0b   1.36Mb  1.36Mb
```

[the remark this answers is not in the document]
sehz: I think we can create a separate issue for multiple partitions so it's not a show stopper for this PR

crates/fluvio/Cargo.toml
```diff
@@ -40,7 +40,11 @@ siphasher = "0.3.5"
 cfg-if = "1.0.0"
 derive_builder = "0.10"
 async-trait = "0.1.51"
-
+sysinfo = "0.24"
```
sehz, line 43: In the previous conversation, I requested that these additional dependencies should be under the feature flag and ensure they work in the web assembly.

sehz (changes requested):

crates/fluvio/src/producer/partition_producer.rs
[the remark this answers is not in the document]
galibey: I looked more carefully. There is always one partition in this place because we create one PartitionProducer per partition. So, using the latest offset here is totally valid.

sehz (approved):
LGTM.  Nice work

sehz:
armv7 build failed:  https://github.com/infinyon/fluvio/runs/7265516240?check_suite_focus=true

sehz:
disable stats for armv7?

tjtelan:
> disable stats for armv7? This probably can be done thru feature flag

I fixed this by having 32 bit arm use the same types as wasm32. Builds for me now.

```
$ cross build --bin fluvio --release --target armv7-unknown-linux-gnueabihf
   Compiling fluvio v0.13.0 (/project/crates/fluvio)
   Compiling fluvio-extension-common v0.8.0 (/project/crates/fluvio-extension-common)
   Compiling fluvio-cluster v0.0.0 (/project/crates/fluvio-cluster)
   Compiling fluvio-cli v0.0.0 (/project/crates/fluvio-cli)
    Finished release [optimized] target(s) in 1m 01s
```

PR #2425, opened 2022-06-20
[Merged by Bors] - Add specific log_dir creation error

joshchngs:
Previous implementation simply prints:
	0: Fluvio cluster error
	1: Permission denied (os error 13)

...which gives no information regarding which operation caused the
error, or even to which path the error relates.

This commit adds an interstitial error to the chain, which prints:
	1: An error occurred creating the cluster log directory "/usr/local/var/log/fluvio"

### DIFF
--- a/.github/workflows/ci_mac.yaml
+++ b/.github/workflows/ci_mac.yaml
@@ -236,7 +236,7 @@ jobs:
           log_dir="/obviously/not/a/writable/path"
           stderr="$(mktemp)"
           echo "Starting cluster with '--log-dir=${log_dir}'"
-          fluvio cluster start --local --develop --log-dir "${log_dir}" 2>$stderr
+          fluvio cluster start --local --develop --skip-checks --log-dir "${log_dir}" 2>$stderr
           [ $? -eq 0 ] && { echo "Expected nonzero exit code, got $?"; exit 2; }
           grep -qF "${log_dir}" $stderr || { echo "Expected to find '${log_dir}' in stderr"; exit 3; }
 
--- a/.github/workflows/ci_mac.yaml
+++ b/.github/workflows/ci_mac.yaml
@@ -233,6 +233,7 @@ jobs:
       - name: Start local fluvio cluster with bad --log-dir
         timeout-minutes: 1
         run: |
+          set +e
           log_dir="/obviously/not/a/writable/path"
           stderr="$(mktemp)"
           echo "Starting cluster with '--log-dir=${log_dir}'"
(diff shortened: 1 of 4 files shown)


joshchngs:
@nicholastmosher I hit this issue because my user doesn't have any write permissions inside `/usr/local`.
I'm on Apple Silicon, MacOS 12.4.

I don't have a quick way to confirm with a fresh install, but I *think* this is the new default permission set on arm64.
Question: what's the reason for the different default log_dir under macos, and is it still valid?

tjtelan (changes requested):
> I don't have a quick way to confirm with a fresh install, but I _think_ this is the new default permission set on arm64.

@joshchngs Is there any way you can write a simple test that uses the CLI to throw the error and validates your error message matches what you set? 

Please add it to [the Mac CI](https://github.com/infinyon/fluvio/blob/master/.github/workflows/ci_mac.yaml). This is a filesystem permissions issue, not specifically a MacOS related issue. But it will be the easiest area for you to add CLI tests against a local cluster.

Copy this job as a base for your local CLI test

https://github.com/infinyon/fluvio/blob/477b2609a68cfa728a2fab661665e98efefe132b/.github/workflows/ci_mac.yaml#L121

Replace the job name, modify the job to set your test conditions and run your test
https://github.com/infinyon/fluvio/blob/477b2609a68cfa728a2fab661665e98efefe132b/.github/workflows/ci_mac.yaml#L175-L179

joshchngs:
That's done. It's a fairly trivial test which comes with 10 minutes of boilerplate docker/k8s setup. I'd suggest rolling it into the  existing local cluster setup as an additional step if the overhead becomes a problem.

joshchngs:
Alright, that's ready to merge from my end. Pushed a CI test for the default case, which seems OK on the x86_64 macos-12 runner, so I unpushed that commit. So, either something else changed `/usr` permissions on my setup or it's actually different on arm64. At least if somebody else hits it there is a more informative error.

tjtelan (changes requested):
One last ask before merge: Can you add a reference to this fix in [`CHANGELOG.md`](https://github.com/infinyon/fluvio/blob/master/CHANGELOG.md)?

Otherwise, LGTM, thanks!

joshchngs:
Done  - against 0.9.30 (rebased on master)

tjtelan (approved):

sehz:
see https://stackoverflow.com/questions/16432071/how-to-fix-homebrew-permissions

sehz:
For homebrew users, we didn't encounter this issue since we probably fix the permission problem a long time ago and totally forgotten about it.

PR #2426, opened 2022-06-20
[Merged by Bors] - Add custom deserializer for connector yaml

### LINKED ISSUE #2410: Allow keys and dictionaries to parameters section of connector yaml
This is a subtask of https://github.com/infinyon/fluvio-connectors/issues/260

Our connectors yaml should support values and dictionaries in the `parameters` section. Example:
```yaml
version: dev
name: kafka-output
type: kafka-source
topic: fluvio-output
create_topic: true
parameters:
  kafka-topic: fluvio
  kafka-url: "192.168.1.12:9092"
  header:
    - "foo: bar"
    - "bar: baz"
```

In this example, the SC controller will `headers` dictionary will add duplicate commandline arguments to the connector at https://github.com/infinyon/fluvio/blob/32433325185edab66b8bffed46af0f7730887af9/crates/fluvio-sc/src/k8/controllers/managed_connector_deployment.rs#L224-L229

### LINKED ISSUE #2421: Expose producer options in connector config
Intent is to expose options that affect the performance of a connector

For example:
```yaml
version: 0.2.0
name: source-connector-draft
type: http
topic: cat-facts
direction: source
producer:
  linger: 1ms
  batch_size: 44
  compression: gzip
parameters:
  endpoint: https://catfact.ninja/fact
  interval: 10
  map: "catfact-map"
```

simlay:
* Closes #2421
* Closes #2410.

This enables the connectors yamls to be thusly:
```yaml
version: 0.1.0
name: my-test-mqtt
type: mqtt
topic: my-mqtt
create_topic: false
direction: source
parameters:
  param_1: "mqtt.hsl.fi"
  param_2:
    foo: bar
    bar: foo
  param_3:
    - "baz"
secrets:
  foo: bar
producer:
  linger: 1ms
  batch_size: 44
  compression: gzip
consumer:
  partition: 10
```

### DIFF
--- a/crates/fluvio-cli/src/connector/mod.rs
+++ b/crates/fluvio-cli/src/connector/mod.rs
@@ -193,7 +193,7 @@ impl ConnectorConfig {
             if let Some(batch_size_string) = &producer.batch_size_string {
                 let batch_size = batch_size_string
                     .parse::<ByteSize>()
-                    .map_err(|e| CliError::Other(e))?;
+                    .map_err(CliError::Other)?;
                 producer.batch_size = Some(batch_size);
             }
         }
--- a/crates/fluvio-cli/src/connector/mod.rs
+++ b/crates/fluvio-cli/src/connector/mod.rs
@@ -116,16 +116,34 @@ fn config_test() {
     let connector_cfg = ConnectorConfig::from_file("test-data/test-config.yaml")
         .expect("Failed to load test config");
     let expected_params = BTreeMap::from([
-        ("param_1".to_string(), YamlParameter{ context: vec!["mqtt.hsl.fi".to_string()] }),
-        ("param_2".to_string(), YamlParameter{ context: vec!["foo:bar".to_string(), "bar:foo".to_string()] }),
-        ("param_3".to_string(), YamlParameter{ context: vec!["baz".to_string()] })
+        (
+            "param_1".to_string(),
+            YamlParameter {
+                context: vec!["mqtt.hsl.fi".to_string()],
+            },
+        ),
+        (
+            "param_2".to_string(),
+            YamlParameter {
+                context: vec!["foo:bar".to_string(), "bar:foo".to_string()],
+            },
+        ),
+        (
+            "param_3".to_string(),
+            YamlParameter {
+                context: vec!["baz".to_string()],
+            },
+        ),
     ]);
     assert_eq!(connector_cfg.parameters, expected_params);
     let out: ManagedConnectorSpec = connector_cfg.into();
     let expected_params = BTreeMap::from([
         ("consumer-partition".to_string(), vec!["10".to_string()]),
         ("param_1".to_string(), vec!["mqtt.hsl.fi".to_string()]),
-        ("param_2".to_string(), vec!["foo:bar".to_string(), "bar:foo".to_string()]),
+        (
+            "param_2".to_string(),
+            vec!["foo:bar".to_string(), "bar:foo".to_string()],
+        ),
         ("param_3".to_string(), vec!["baz".to_string()]),
         ("producer-batch-size".to_string(), vec!["44".to_string()]),
         ("producer-compression".to_string(), vec!["Gzip".to_string()]),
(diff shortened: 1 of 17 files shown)


crates/fluvio-cli/src/connector/mod.rs
```diff
@@ -92,12 +94,47 @@ pub struct ConnectorConfig {
...
+    consumer: Option<ConsumerParameters>,
+}
+
+#[test]
+fn config_test() {
+    let connector_cfg = ConnectorConfig::from_file("test-data/test-config.yaml")
+        .expect("Failed to load test config");
+    println!("{:#?}", connector_cfg);
+    let out: ManagedConnectorSpec = connector_cfg.into();
+    println!("{:#?}", out);
```
sehz, line 120: There should be assertion

sehz: There should be fail cases as well

crates/fluvio-controlplane-metadata/src/connector/k8/spec.rs
```diff
@@ -42,7 +42,7 @@ pub struct K8ManagedConnectorSpec {
     #[cfg_attr(feature = "use_serde", serde(rename = "type"))]
     pub type_: String, // syslog, github star, slack
     pub topic: String,
-    pub parameters: BTreeMap<String, String>,
+    pub parameters: BTreeMap<String, Vec<String>>,
```
sehz, line 45: This should bump up crate version

sehz (changes requested):

simlay:
One thing that might be worth adding to this PR would be adding a top level key of `smart_modules`. Thoughts?

crates/fluvio-cli/src/connector/mod.rs
```diff
@@ -112,20 +149,95 @@ impl ConnectorConfig {
...
+            if let Some(linger) = producer.linger {
+                let linger = humantime::format_duration(linger).to_string();
+                parameters.insert("producer-linger".to_string(), vec![linger]);
+            }
+            if let Some(compression) = producer.compression {
+                let compression = format!("{:?}", compression);
+                parameters.insert("producer-compression".to_string(), vec![compression]);
+            }
+            if let Some(batch_size) = producer.batch_size {
+                let batch_size = format!("{}", batch_size);
```
tjtelan, line 168: Since we're adding units to linger, can we do the same w/ batch size? In the fluvio CLI (maybe currently only in `topic`), this is done w/ [ByteSize](https://crates.io/crates/bytesize).

It is not obvious from the key name what units this is in, and it would be reasonable for a user to assume bytes or kilobytes.

tjtelan (changes requested):
I think this otherwise looks like a good change

tjtelan:
> One thing that might be worth adding to this PR would be adding a top level key of `smart_modules`. Thoughts?

My only apprehension is that we don't have a simple interface for SmartModules. We currently need to specify the type of module (`map`, `filter`, etc...)

My ideal representation of a `smart_modules` top-level key implies chaining behavior, and  looks like this:

```yaml
smart_modules:
  - my_filter
  - my_map
  - my_agg
```
Where the SmartModule transforms in the order it is specified, and does not require passing extra info such as what type of SmartModule. 

---

Considering the current behavior, I would be ok with `smart_module` (no plural), at the top level. Something like this?

```yaml
smart_module:
  name: my_filter
  type: filter
```

crates/fluvio-cli/src/connector/mod.rs
simlay: Hmm. The `ByteSize` docs don't really show how one can use `1GB` is a param in json/serde.

tjtelan: Hm. Yeah, docs aren't as clear as they could be.

I had to look at the [repo's code](https://github.com/hyunsik/bytesize/blob/master/src/parse.rs), but it seems like you can use `parse` with a type.

```
      fn parse(s: &str) -> u64 {
          s.parse::<ByteSize>().unwrap().0
      }
```

And at least the way we using in in `fluvio topic create`,  the  [`.as_u64()`](https://docs.rs/bytesize/1.1.0/bytesize/struct.ByteSize.html#method.as_u64) method returns the `ByteSize` value in bytes.

crates/fluvio-cli/test-data/test-config.yaml
```diff
@@ -5,8 +5,17 @@ topic: my-mqtt
...
+    foo: bar
+    bar: foo
+  param_3:
+    - "baz"
 secrets:
   foo: bar
+producer:
+  linger: 1ms
+  batch_size: '44 MB'
+  compression: gzip
```
sehz, line 19: Is there failure cases?

sehz (changes requested):

crates/fluvio-cli/test-data/test-config.yaml
simlay: I've added tests for the error cases.

sehz: great

sehz (approved):
LGTM

PR #2428, opened 2022-06-22
[Merged by Bors] - Fix `wasi` functions binding relying on order

galibey:
The binding of host function `copy_record` for `wasi` expects that the function always goes first in the list of imports. This is not always correct. Instead of lookup by order, we will do a lookup by name.
Also, the PR adds integration tests that will check the smart module instantiation works for both `wasm` and `wasi` cases.


sehz (approved):
LGTM

morenol (approved):

sehz:
clippy strikes...

PR #2435, opened 2022-06-27
Move CRDs from templates/ to crds/

tothandras:
Is there any reason not to move CRDs to the crds/ folder in `fluvio-sys`?
See: https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-1-let-helm-do-it-for-you


sehz (approved):
LGTM

sehz (changes requested):
Looks like CI test fail: https://github.com/infinyon/fluvio/runs/7095527168?check_suite_focus=true.

I think you need to bump chart version

sehz:
CI still fails. Can you find out why it is failing?

sehz (changes requested):
There is no need to bump up major version.  So this should be 0.9.10.

sehz (approved):
LGTM

sehz:
CI still failed.  Can you check if inline helm still works?  https://github.com/infinyon/fluvio/blob/master/k8-util/helm/Makefile

sehz:
Ensure upgrade helm upgrade works.

tothandras:
@sehz I'm running helm install / upgrade without issues, I'll check the test case.

PR #2442, opened 2022-06-28
[Merged by Bors] - Fix issue with producer when sending more than one batch per request

### LINKED ISSUE #2443: [Bug]: Producer failed to retrieve record metadata when sending more than one batch in a request
**What happened**
When the request of producer has more than one batch.
 The producer only sends back the base offset of the first batch. Therefore batch metadata is not resolved for the batches that are after the first one.

**Expected behavior**
record metadata in:
```rust
let produce_output = producer.send(RecordKey::NULL, value).await?;
let record_metadata = produce_output.wait().await?;
```
should not fail with : `receiving from an empty and closed channel`

1. Run the command:
fluvio topic create topicname
fluvio produce topicname -f input.txt

with this file:
[input.txt](https://github.com/infinyon/fluvio/files/9003188/input.txt)

morenol:
Fixes #2443 

 The producer only sends back the base offset of the first batch. Therefore batch metadata is not resolved for the batches that are after the first one.

With this change, we put each batch in its own PartitionProduceResponse, so SPU returns back the base offset of each batch

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Platform Version 0.9.30 - UNRELEASED
 * Improve CLI error output when log_dir isn't writable ([#2425](https://github.com/infinyon/fluvio/pull/2425))
+* Fix issue in producer when sending more than one batch in a request ([#2443](//github.com/infinyon/fluvio/issues/2443))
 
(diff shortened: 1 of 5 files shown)


sehz (approved):
LGTM

PR #2444, opened 2022-06-28
[Merged by Bors] - Re-added producer config to connector yaml

### LINKED ISSUE #2421: Expose producer options in connector config

simlay:
This is the non breaking changes changes from #2426.

Closes #2421.

### DIFF
--- a/crates/fluvio-cli/src/connector/mod.rs
+++ b/crates/fluvio-cli/src/connector/mod.rs
@@ -6,8 +6,10 @@ use std::collections::BTreeMap;
 use std::path::PathBuf;
 use std::fs::File;
 use std::io::Read;
+use std::time::Duration;
+use bytesize::ByteSize;
 
-use fluvio::Fluvio;
+use fluvio::{Fluvio, Compression};
 use fluvio::metadata::connector::{ManagedConnectorSpec, SecretString};
 use fluvio_extension_common::Terminal;
 use fluvio_extension_common::COMMAND_TEMPLATE;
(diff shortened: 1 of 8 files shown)


crates/fluvio-cli/src/connector/mod.rs
```diff
@@ -92,40 +94,146 @@ pub struct ConnectorConfig {
...
     parameters: BTreeMap<String, String>,
+
     #[serde(default)]
     secrets: BTreeMap<String, SecretString>,
+
+    #[serde(default)]
+    producer: Option<ProducerParameters>,
+
+    #[serde(default)]
+    consumer: Option<ConsumerParameters>,
```
simlay, line 111: @morenol should I add a `smartmodule` section to this?

crates/fluvio-cli/src/connector/mod.rs
```diff
@@ -92,40 +94,146 @@ pub struct ConnectorConfig {
...
+}
+
+#[derive(Debug, Clone, Deserialize)]
+pub struct ConsumerParameters {
+    #[serde(default)]
+    partition: Option<i32>,
```
simlay, line 117: I'm tempted to add a `offset` field here but it's unclear how it should be specified. Thoughts?

sehz: should be the same as CLI but can be add later

crates/fluvio-cli/src/connector/mod.rs
morenol, line 111: Do the feature flag for source and sink work as we would expect? @simlay

sehz (approved):
LGTM. We can add other changes later

crates/fluvio-cli/src/connector/mod.rs
morenol: My bad, that is a fluvio-connectors thing

PR #2445, opened 2022-06-28
Add SmartModule config to connector top level configuration file

### LINKED ISSUE #2455: Add support for user defined parameters in smartmodules used by connectors
Related to https://github.com/infinyon/fluvio-connectors/issues/285

### LINKED ISSUE #2471: Move SmartModule configuration in connectors to top level configuration in Connectors config:
We should configure smart modules in their own section `smartmodule` of the connectors config.
```yaml
...
smartmodule: # optional
  name: string # required
  type: string # (valid values: filter, map, filter_map, aggregate, array_map) -> Optional once we have: https://github.com/infinyon/fluvio/issues/2457
  aggregate_initial_value: 1 # optional
  parameters: # optional
    key: value
    foo: bar
...
```

morenol:
Adds a `smartmodule` top level config:

```yaml
...
smartmodule: # optional
  name: some-name # required
  type: aggregate # optional (valid values: filter, map, filter_map, aggregate, array_map)
  aggregate_initial_value: 1 # optional
  parameters: # optional
    key: value
    foo: bar
...
```

Also changed struct name from `ManageConnectorParameterValue` to `ManagedConnectorParameterValue`

Closes #2471
Closes #2455

### DIFF
--- a/crates/fluvio-cli/src/connector/mod.rs
+++ b/crates/fluvio-cli/src/connector/mod.rs
@@ -350,7 +350,7 @@ impl From<ManagedConnectorSpec> for ConnectorConfig {
             ManagedConnectorParameterValueInner::String(smartmodule_name),
         )) = parameters
             .remove("filter-map")
-            .or(parameters.remove("filter_map"))
+            .or_else(|| parameters.remove("filter_map"))
         {
             (Some(smartmodule_name), Some(SmartModuleType::FilterMap))
         } else {
@@ -362,7 +362,7 @@ impl From<ManagedConnectorSpec> for ConnectorConfig {
                 ManagedConnectorParameterValueInner::String(agrgate_initial_value),
             )) = parameters
                 .remove("aggregate-initial-value")
-                .or(parameters.remove("aggregate_initial_value"))
+                .or_else(|| parameters.remove("aggregate_initial_value"))
             {
                 Some(agrgate_initial_value)
             } else {
--- a/crates/fluvio-cli/src/connector/mod.rs
+++ b/crates/fluvio-cli/src/connector/mod.rs
@@ -125,6 +125,9 @@ pub struct ConnectorConfig {
 
     #[serde(default, skip_serializing_if = "Option::is_none")]
     consumer: Option<ConsumerParameters>,
+
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    smartmodule: Option<SmartModuleParameters>,
 }
 
 #[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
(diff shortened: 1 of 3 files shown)


crates/fluvio-cli/src/connector/mod.rs
```diff
@@ -152,6 +155,41 @@ pub struct ProducerParameters {
...
+    Aggregate,
+    FilterMap,
+}
+
+impl ToString for SmartModuleType {
+    fn to_string(&self) -> String {
+        match self {
+            SmartModuleType::Filter => "filter",
+            SmartModuleType::Map => "map",
+            SmartModuleType::Arraymap => "arraymap",
```
simlay, line 173: `arraymap` or `array-map`?

crates/fluvio-cli/src/connector/mod.rs
```diff
@@ -318,6 +447,20 @@ fn full_yaml_test() {
             "param_6".to_string(),
             vec!["-10".to_string(), "-10.0".to_string()].into(),
         ),
+        ("aggregate".to_string(), "myaggregate".to_string().into()),
```
simlay, line 450: Part of me thinks that we should have the parameters going to the connectors as `smartmodule-aggregate`, `smartmodule-map`, etc. similar to `producer-compression`, `producer-linger`, etc.

crates/fluvio-cli/src/consume/mod.rs
```diff
@@ -158,29 +158,40 @@ pub struct ConsumeOpt {
...
     /// They should be passed using key=value format
     /// Eg. fluvio consume topic-name --filter filter.wasm -e foo=bar -e key=value -e one=1
-    #[clap(short = 'e', long= "extra-params", parse(try_from_str = parse_key_val), number_of_values = 1)]
+    #[clap(
+        short = 'e',
+        requires = "smartmodule_group",
+        long= "extra-params",
+        parse(try_from_str = parse_key_val),
+        number_of_values = 1
+    )]
```
simlay, line 194: Are the changes to the `fluvio consume` cli required for this change?

morenol: no, moved to https://github.com/infinyon/fluvio/pull/2476

crates/fluvio-cli/src/connector/mod.rs
morenol: I think that in the future we should just use `--smartmodule`. I can change with `--smartmodule-name` if needed but I use --smartmodule in order to be similar to `fluvio consume` flag.

I just keep the old --map, --filter, --aggregate... for backward compatibility but maybe in the future we can just remove those options.

crates/fluvio-cli/src/connector/mod.rs
morenol: It looks like in `fluvio-connector` we are using arraymap: See https://github.com/infinyon/fluvio-connectors/blob/4b894d7bacd3c761712da681bab9c63dd847144a/rust-connectors/common/src/opt.rs#L83

simlay: Sounds good.

crates/fluvio-cli/src/connector/mod.rs
simlay: ah, yes. If `--smartmodule` is just the name of the smartmodule, that works. I agree that we should keep the `map`, `filter`, `aggregate` for now to maintain backward compatability. Ideally we could have both but if we add a `--smartmodule-map`, etc. it'll break the connector start as it'll be an "unexpected commandline option". It's burdensome that we have to update two repo's at the sametime to change the name. :/

PR #2452, opened 2022-06-30
[Merged by Bors] - Move stream publishers to connection-level context

galibey:
This is the first PR of implementation of dropping a stream's server-side counterpart when a stream on the client-side is dropped.  This is all part of the activities for managing offsets per consumer on the SPU side.

Currently, the lifetime of `StreamPublisher` is equal to the lifetime of Tcp connection between SPU and client. We do not clean up `StreamPublishers` until the connection is dropped. Therefore, there is no need to keep them inside `GlobalContext` and have it multi-threaded, because it is always touched by only one thread (the one that handles connection). 

This PR doesn't add or remove any functional behavior. The `StreamPublishers` will be dropped after disconnect as it does now. The actual drop mechanism will go in another PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Release Notes
 
 ## Platform Version 0.9.31 - UNRELEASED
+* Move stream publishers to connection-level context ([#2452](https://github.com/infinyon/fluvio/pull/2452))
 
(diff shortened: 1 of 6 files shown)


sehz (approved):
LGTM

PR #2463, opened 2022-07-04
[Merged by Bors] - Detect smart module type on SmartEngine

### LINKED ISSUE #2457: Identify type of SmartModule by signature
Right now in order to use a SmartModule, the user has to identify the type of SmartModule 

Example: Fluvio CLI consume
```
$ fluvio consume --filter my-filter example-topic
$ fluvio consume --map my-map example-topic
```

But this can be a little confusing, and you can easily make mistakes.

Example of potential error
```
$ fluvio consume --array-map my-mystery-smartmodule example-topic
```

It seems like name should suffice, and the responsibility for type checking should go to the cluster.

Example w/ more generic flag
```
$ fluvio consume --smartmodule my-mystery-smartmodule example-topic
```

morenol:
Fixes #2457 

This PR adds a way to defer to the SPU the smarmodule type resolution.

We can use now in CLI and API a generic --smartmodule type and the SPU will try to detect the smart module type by trying to initialize the smart module engine with each one of them until there is success

example:

`fluvio consume numbers --smartmodule double -B` Where double is https://github.com/infinyon/fluvio/tree/master/crates/fluvio-smartmodule/examples/map_double. Should return a stream with the numbers of `number` topic multiplied by 2

### DIFF
--- a/crates/fluvio-smartengine/Cargo.toml
+++ b/crates/fluvio-smartengine/Cargo.toml
@@ -33,7 +33,7 @@ dataplane = { version = "0.11.0", path = "../fluvio-dataplane-protocol", package
 ] }
 fluvio-controlplane-metadata = { path = "../fluvio-controlplane-metadata", version = "0.17.0", optional = true }
 flate2 = { version = "1.0", optional = true }
-fluvio-spu-schema = { version = "0.9.0", path = "../fluvio-spu-schema" }
+fluvio-spu-schema = { version = "0.10.0", path = "../fluvio-spu-schema" }
 
 [dev-dependencies]
 fluvio-types = { path = "../fluvio-types" }
--- a/crates/fluvio-smartengine/Cargo.toml
+++ b/crates/fluvio-smartengine/Cargo.toml
@@ -34,6 +34,7 @@ dataplane = { version = "0.11.0", path = "../fluvio-dataplane-protocol", package
 fluvio-controlplane-metadata = { path = "../fluvio-controlplane-metadata", version = "0.17.0", optional = true }
 flate2 = { version = "1.0", optional = true }
 fluvio-spu-schema = { version = "0.10.0", path = "../fluvio-spu-schema" }
+thiserror = "1.0.31"
 
 [dev-dependencies]
 fluvio-types = { path = "../fluvio-types" }
--- a/crates/fluvio-smartengine/Cargo.toml
+++ b/crates/fluvio-smartengine/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio-smartengine"
-version = "0.2.11"
+version = "0.3.0"
 edition = "2021"
 license = "Apache-2.0"
 authors = ["Fluvio Contributors <team@fluvio.io>"]
--- a/crates/fluvio-smartengine/src/smartmodule/join.rs
+++ b/crates/fluvio-smartengine/src/smartmodule/join.rs
@@ -46,14 +46,14 @@ impl SmartModuleJoin {
         let join_fn = if let Ok(join_fn) = base
             .instance
             .get_typed_func(&mut base.store, JOIN_FN_NAME)
-            .map_err(|_err| Error::NamedExport(JOIN_FN_NAME))
+            .map_err(|_err| Error::NotNamedExport(JOIN_FN_NAME))
         {
             JoinFnKind::New(join_fn)
         } else {
             let join_fn = base
                 .instance
                 .get_typed_func(&mut base.store, JOIN_FN_NAME)
-                .map_err(|_err| Error::NamedExport(JOIN_FN_NAME))?;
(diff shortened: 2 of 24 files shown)


tjtelan (approved):
LGTM

sehz:
still marked as draft

crates/fluvio-smartengine/Cargo.toml
```diff
@@ -33,7 +33,7 @@ dataplane = { version = "0.11.0", path = "../fluvio-dataplane-protocol", package
 ] }
 fluvio-controlplane-metadata = { path = "../fluvio-controlplane-metadata", version = "0.17.0", optional = true }
```
sehz, line 34: I think this requires major bump with enum changes

crates/fluvio-smartengine/src/smartmodule/mod.rs
```diff
@@ -233,6 +238,45 @@ impl SmartModuleWithEngine {
         let aggregate = SmartModuleAggregate::new(self, params, accumulator, version)?;
         Ok(aggregate)
     }
+
+    fn create_smartmodule(
```
sehz, line 242: change this to more descriptive? `create_generic_smartmodule` and add doc

crates/fluvio-smartengine/src/smartmodule/mod.rs
```diff
@@ -233,6 +238,45 @@ impl SmartModuleWithEngine {
...
+        &self,
+        params: SmartModuleExtraParams,
+        version: i16,
+        context: &SmartModuleContextData,
+    ) -> Result<Box<dyn SmartModuleInstance>> {
+        if let Ok(filter) = self.create_filter(params.clone(), version) {
+            return Ok(Box::new(filter));
+        }
```
sehz, line 250: what about error cases? It is possible that create_filter could fail for other reasons.  Should distinguish between true error and not matching fn signature

crates/fluvio-smartengine/src/smartmodule/mod.rs
```diff
@@ -233,6 +238,45 @@ impl SmartModuleWithEngine {
...
+        if let Ok(map) = self.create_map(params.clone(), version) {
+            return Ok(Box::new(map));
```
sehz, line 252: same as previously

crates/fluvio-smartengine/Cargo.toml
sehz: if we bump up the major version, maybe we can deprecate other manual smart module types

crates/fluvio-smartengine/src/smartmodule/mod.rs
```diff
@@ -233,6 +238,45 @@ impl SmartModuleWithEngine {
...
+            return Ok(Box::new(join));
+        }
+        if let Ok(join_stream) = self.create_join_stream(params, version) {
+            return Ok(Box::new(join_stream));
+        }
+
+        Err(Error::msg(
+            "Unable to initialize smartmodule. Make sure that the wasm module includes the correct imports.",
+        ))
+    }
```
sehz, line 279: need unit test here as well

sehz (changes requested):

crates/fluvio-smartengine/src/smartmodule/mod.rs
morenol: I create a internal error type: that represent the two types of error:
1) Not matching signature
2) Failed to instantiate


For the first one it will try to continue with the next smartmodule type. With the second one it will return that error so it doesn't try the next smartmodule types.

crates/fluvio-smartengine/Cargo.toml
morenol: I did the bump, not sure what is the expected approach to deprecate those manual smart modules

sehz: k. let's keep as is

crates/fluvio-smartengine/src/smartmodule/join.rs
```diff
@@ -40,17 +41,21 @@ impl SmartModuleJoin {
...
-            } else {
-                let join_fn = base
-                    .instance
-                    .get_typed_func(&mut base.store, JOIN_FN_NAME)?;
-                JoinFnKind::Old(join_fn)
-            };
+        let join_fn = if let Ok(join_fn) = base
+            .instance
+            .get_typed_func(&mut base.store, JOIN_FN_NAME)
+            .map_err(|_err| Error::NotNamedExport(JOIN_FN_NAME))
```
sehz, line 49: can you include error?

sehz (changes requested):

crates/fluvio-smartengine/src/smartmodule/join.rs
morenol: Added, unfortunately it is anyhow::Error

PR #2464, opened 2022-07-05
[Merged by Bors] - Add fluvio connector config to cli to get connector configs.

### LINKED ISSUE #2458: Add `connector describe` command
Users need the ability to see what are the connector parameters used to provision the connectors.

```
fluvio connector describe github-stars-input
name: github-stars-input
type: http
topic: github-stars
create_topic: true
direction: source
parameters:
  endpoint: 'https://api.github.com/repos/infinyon/fluvio'
  method: GET
  interval: 30
  header: 'Authorization: token ****'
  aggregate: star-selector
  aggregate-initial-value: "use-last"
```

simlay:
Closes #2458.

~Notes from discussion with @ajhunyady: add json output type option.~

We chose to move to using `fluvio connector config <name> [-o output-file]`


crates/fluvio-cli/src/connector/describe.rs
```diff
@@ -0,0 +1,55 @@
...
+        let admin = fluvio.admin().await;
+        let connector_name = self.connector_name;
+        let connector_list = admin
+            .list::<ManagedConnectorSpec, _>(vec![connector_name.clone()])
+            .await?;
+        let connector = connector_list.first();
+        let connector = if let Some(connector) = connector {
+            connector.spec.clone()
+        } else {
+            return Err(CliError::ConnectorNotFound(connector_name));
```
sehz, line 41: no need semicolon?

crates/fluvio-cli/src/connector/describe.rs
```diff
@@ -0,0 +1,55 @@
...
+        };
+        let connector: ConnectorConfig = connector.into();
+        let connector = serde_yaml::to_string(&connector)?;
```
sehz, line 44: We should support all output types right? (YAML,JSON, Text) We should probably have trait for it so can standardize it.

sehz (changes requested):
Generally looks good.  Unfortunately, I think we don't seem to have an existing framework for `describe` which we should start

crates/fluvio-cli/src/connector/describe.rs
simlay: We do actually kinda already have this. If you were to do `fluvio connector list -O [json|yaml]` you'll get the connectors spec as `json/yaml`. The issue is that it's pretty much just the CRD as json or yaml. Not the connector yaml which the user will want to copy and paste for a `fluvio connector create` or `fluvio connector update` command.

crates/fluvio-cli/src/connector/describe.rs
simlay: I removed it and `cargo fmt` re-added it.

crates/fluvio-cli/src/connector/describe.rs
simlay: Should I try to make use of this it? I'm sure we'll want this for stuff like for `fluvio table-format`.

tjtelan (commented):
Can we have a CLI test that starts a connector and just runs the describe command and tests that it exits appropriately?

sehz (changes requested):
Looks good.  a very minor nit.

sehz (approved):
LGTM.  Good work!

PR #2472, opened 2022-07-12
[Merged by Bors] - Make version required in connector yaml

### LINKED ISSUE #2456: Connector: Enforce version field as mandatory
[Right now in the connector yaml if the `version` field isn't specified, the connector defaults to `latest`](https://github.com/infinyon/fluvio/blob/749df76b6a277bbb3a8c966b511c2b9ad7002ca8/crates/fluvio-sc/src/k8/controllers/managed_connector_deployment.rs#L267-L274) which means that if the connector is re-created (or if the cluster is upgraded), the connector will try to get the `latest` image which may have breaking changes.

If the `version` field is unspecified, the connector should pin the version of the latest image and this image version/tag should be pinned to the version it was created with.

simlay:
Closes #2456.

### DIFF
--- a/crates/fluvio-controlplane-metadata/src/connector/spec.rs
+++ b/crates/fluvio-controlplane-metadata/src/connector/spec.rs
@@ -55,7 +55,7 @@ impl ToString for ConnectorVersionInner {
     fn to_string(&self) -> String {
         match self {
             ConnectorVersionInner::String(inner) => inner.to_string(),
-            ConnectorVersionInner::Option(inner) => inner.clone().unwrap_or_else(|| "".to_string()),
+            ConnectorVersionInner::Option(inner) => inner.clone().unwrap_or_else(|| "latest".to_string()),
         }
     }
 }
--- a/crates/fluvio-controlplane-metadata/src/connector/spec.rs
+++ b/crates/fluvio-controlplane-metadata/src/connector/spec.rs
@@ -252,6 +252,7 @@ parameters:
 secrets: {}
 topic: poc1
 type: kafka-sink
+version: latest
 "#;
         use super::ManagedConnectorSpec;
         let connector_spec: ManagedConnectorSpec =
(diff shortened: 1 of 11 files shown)


ajhunyady:
Is this change backward compatible?

simlay:
> Is this change backward compatible?

Yes.

sehz (changes requested):
is there a negative test? if required version is missing, it should fail

sehz (approved):
LGTM

crates/fluvio-controlplane-metadata/src/connector/spec.rs
```diff
@@ -30,6 +30,74 @@ pub struct ManagedConnectorSpec {
...
+impl From<String> for ConnectorVersionInner {
+    fn from(inner: String) -> Self {
+        Self::String(inner)
+    }
+}
+impl ToString for ConnectorVersionInner {
+    fn to_string(&self) -> String {
+        match self {
+            ConnectorVersionInner::String(inner) => inner.to_string(),
+            ConnectorVersionInner::Option(inner) => inner.clone().unwrap_or_else(|| "".to_string()),
```
morenol, line 58: this is for the old connectors right? shouldn't it be `latest` by default?

Not sure what is the expected response to `fluvio connector config old-connector` when it was created with the previos version.

crates/fluvio-sc-schema/src/objects/list.rs
```diff
@@ -27,7 +27,7 @@ where
 
 impl Request for ObjectApiListRequest {
     const API_KEY: u16 = AdminPublicApiKey::List as u16;
-    const DEFAULT_API_VERSION: i16 = 8;
+    const DEFAULT_API_VERSION: i16 = 9;
```
morenol, line 30: Probably we want to update this also for `watch.rs` so we don't have problems when we do the Update controller

morenol (approved):
looks good to me!

simlay:
bors retry

tjtelan:
bors retry

tjtelan:
@simlay

The reason Bors can't merge because fluvio check test is failing.
https://github.com/infinyon/fluvio/runs/7312057421?check_suite_focus=true

simlay:
bors retry

PR #2474, opened 2022-07-13
Add `AtMostOnce` and `AtLeastOnce` delivery semantics

galibey:
Added two delivery semantics to Producer:
1. `AtMostOnce` - we send records and do not wait for a response. To wait for the response, one could use `ProduceOutput::wait` after sending the record.
2. `AtLeastOnce` - we send records, wait for the response and retry if a network error occurs.

The default retry policy:
retries: 4,
max timeout: 300s
delays sequence: 20ms, 400ms, 8s, 160s (Exponential Backoff)

### DIFF
--- a/crates/fluvio-cli/src/produce/mod.rs
+++ b/crates/fluvio-cli/src/produce/mod.rs
@@ -230,8 +230,7 @@ impl ProduceOpt {
 
             #[cfg(feature = "stats")]
             if self.is_stats_collect() && self.is_print_live_stats() {
-                self.update_stats_bar(maybe_stats_bar.as_ref(), &producer, "")
-                    .await;
+                self.update_stats_bar(maybe_stats_bar.as_ref(), &producer, "");
             }
         } else {
             // Read input line-by-line and send as individual records
@@ -242,12 +241,13 @@ impl ProduceOpt {
             self.produce_lines(producer.clone()).await?;
         };
 
+        producer.flush().await?;
+
         #[cfg(feature = "stats")]
         if self.is_stats_collect() {
             producer_summary(&producer, maybe_stats_bar.as_ref(), self.stats_summary).await;
         }
 
-        producer.flush().await?;
         if self.interactive_mode() {
             print_cli_ok!();
         }
@@ -291,8 +291,7 @@ impl ProduceOpt {
                     #[cfg(feature = "stats")]
                     if self.is_stats_collect() {
                         if self.is_print_live_stats() {
-                            self.update_stats_bar(maybe_stats_bar, &producer, &line)
-                                .await;
+                            self.update_stats_bar(maybe_stats_bar, &producer, &line);
                         }
 
                         stats_dataframe_check = write_csv_dataframe(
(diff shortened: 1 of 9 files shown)


crates/fluvio-cli/src/produce/mod.rs
```diff
@@ -104,6 +104,12 @@ pub struct ProduceOpt {
     /// Experimental: Only print the stats summary. Implies `--stats` and `--no-stats-bar`
     #[clap(long)]
     pub stats_summary: bool,
+
+    /// Delivery guarantees that producer must respect. Supported values:
+    /// at_most_once (AtMostOnce) - send records without waiting from response,
+    /// at_least_once (AtLeastOnce) - send records and retry if error occurred.
+    #[clap(long, parse(try_from_str = parse_delivery_semantic))]
+    pub delivery_semantic: Option<DeliverySemantic>,
```
sehz, line 112: I can you can use default flag in `clap` so doesn't requires Optional

crates/fluvio-cli/src/produce/mod.rs
```diff
@@ -161,6 +167,16 @@ impl ProduceOpt {
             _ => config_builder,
         };
 
+        // Delivery Semantic
+        let config_builder = if let Some(delivery_semantic) = self.delivery_semantic {
+            if delivery_semantic == DeliverySemantic::AtMostOnce && self.isolation.is_some() {
+                warn!("Isolation is ignored for AtMostOnce delivery semantic");
```
sehz, line 173: @morenol I think there is way to enforce at clap using conditional flag?

crates/fluvio-cli/src/produce/mod.rs
```diff
@@ -455,3 +473,11 @@ async fn init_ctrlc(
     }
     Ok(())
 }
+
+pub(crate) fn parse_delivery_semantic(s: &str) -> Result<DeliverySemantic, String> {
```
sehz, line 477: I think you can map enum in the clap

crates/fluvio-cli/src/produce/stats/report.rs
```diff
@@ -25,7 +25,7 @@ pub async fn start_csv_report(
             .truncate(true)
```
sehz, line 25: Looks like this can be separated into small PR to merge quickly.

sehz (changes requested):
The first pass looks good.  
Couple of minor nits in the CLI.
I will do a more in-depth review. 

It will be easier to review if this can be broken into multiple PR:
   - stats 
   - protocols
   - producer
   - ?

crates/fluvio-cli/src/produce/mod.rs
morenol: I think that we can enforce at clap level with `requires` when we want to enable/disable a parameter depending on the presence of other, but since this one is a enum type, I think that we cannot use that.

https://github.com/clap-rs/clap/blob/ef3121cbdf938fef9a6b2dc199bc15998043939a/examples/tutorial_derive/04_03_relations.rs#L35

Maybe an option is to use `--at-most-once-delivery` and `--at-least-once-delivery` instead of the enum type, but not sure if that is better for the CLI users

sehz:
I assume this is obsolete?

galibey:
> I assume this is obsolete?

Yes, all parts have been merged within https://github.com/infinyon/fluvio/issues/2481

PR #2476, opened 2022-07-13
[Merged by Bors] - Do not allow --initial when we are not using aggregate smartmodule

morenol:
I noticed that in `fluvio consume` we can use `--initial` and `--join-topic` when we are not using their respective smart modules. This fixes that, now if we use `--initial` without using --aggregate or --smartmodule the CLI will fail with the right description.

The same applies to `--extra-params`


sehz (approved):
LGTM

Issue #2481, opened 2022-07-14
Support for `at least once` and `at most once` in the Producer Client


galibey:
## Acceptance Criteria:

- Implement a new API per `Producer with delivery semantic Flag`.  We should implement `at-most-once` and `at-least-once`.
- CLI or integration tests.
- Update Doc with supported delivery semantics

PR #2483, opened 2022-07-14
[Merged by Bors] - Measure latency for stats using macro

galibey:
Added macro for measuring latency for stats instead of using the function. It will decouple it from unnecessary details and make it more generic.

Also, I changed some async functions into sync where they aren't needed.

Related #2481


sehz (approved):
LGTM

PR #2484, opened 2022-07-14
[Merged by Bors] - Keep serving incoming requests even if socket closed to write

galibey:
This change is needed for `at most once` delivery semantic. The client can send a batch and close the socket without waiting for a response. If the server can't write response it should continue to handle incoming requests that are received and waiting to be served.

Related #2481


sehz (approved):
LGTM

galibey:
bors retry

PR #2488, opened 2022-07-15
[Merged by Bors] - Support async response in socket

galibey:
Added an ability to receive a response asynchronously. We re-use the current functionality of `AsyncResponse` and streaming, just saying that `AsyncResponse` is now also a Future that can resolve to one result instead of a sequence of results for the streaming case.

Related #2481

### DIFF
--- a/crates/fluvio-socket/Cargo.toml
+++ b/crates/fluvio-socket/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio-socket"
-version = "0.13.0"
+version = "0.12.1"
 edition = "2021"
 authors = ["Fluvio Contributors <team@fluvio.io>"]
 description = "Provide TCP socket wrapper for fluvio protocol"
--- a/crates/fluvio-socket/Cargo.toml
+++ b/crates/fluvio-socket/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio-socket"
-version = "0.12.0"
+version = "0.13.0"
 edition = "2021"
 authors = ["Fluvio Contributors <team@fluvio.io>"]
 description = "Provide TCP socket wrapper for fluvio protocol"
--- a/crates/fluvio-socket/src/multiplexing.rs
+++ b/crates/fluvio-socket/src/multiplexing.rs
@@ -11,6 +11,7 @@ use std::sync::atomic::Ordering::SeqCst;
 use std::sync::atomic::AtomicI32;
 use std::time::Duration;
 use std::fmt;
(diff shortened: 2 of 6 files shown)


crates/fluvio-socket/src/multiplexing.rs
```diff
@@ -459,27 +482,31 @@ impl MultiPlexingResponseDispatcher {
...
-                            format!(
-                                "problem sending to queue socket: {}, err: {}",
-                                correlation_id, err
-                            ),
-                        )
-                        .into()
-                    })
+                    // sender was dropped before response arrives
+                    if queue_sender.is_closed() {
+                        debug!("attempt to send data to closed socket: {}", correlation_id);
```
sehz, line 487: Can you use trace format? 
`debug!(correlation_id,"attemp....")`

crates/fluvio-socket/src/multiplexing.rs
```diff
@@ -459,27 +482,31 @@ impl MultiPlexingResponseDispatcher {
...
-            Err(IoError::new(
-                ErrorKind::BrokenPipe,
-                format!(
-                    "no socket receiver founded for id: {}, abandoning sending",
-                    correlation_id
-                ),
-            )
-            .into())
+            // sender was dropped and unregistered before response arrives
+            debug!(
```
sehz, line 505: use trace lib format. see comment above

crates/fluvio-socket/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio-socket"
-version = "0.12.0"
+version = "0.13.0"
```
sehz, line 3: Is there API incompatibility? If we are just adding a new API, then just need to bump up patch version

sehz (changes requested):
Overall looks good.  Just few minor nits

crates/fluvio-socket/src/multiplexing.rs
galibey: Done

crates/fluvio-socket/src/multiplexing.rs
galibey: Done

crates/fluvio-socket/Cargo.toml
galibey: I thought that implementing `Future`  on `AsyncResponse` potentially could be a breaking change, but `orphan rules` would not allow it, so you're right. Changed to 0.12.1.

sehz (approved):
LGTM

PR #2498, opened 2022-07-25
[Merged by Bors] - Add back pressure to longevity test producers

galibey:
Currently, in `fluvio-test` producers send records to the fluvio cluster asynchronously and without any throughput control. The cluster runs in k8s env and has CPU and memory limits, and it can't handle the generated load. Therefore, each producer constantly accumulates local records, leading to increased memory usage. For the test with 25 producers and 10 topics it is 250 partition producers that generate data without pauses. They eat all memory, and the cluster stuck during segments rotation holding the lock, and it stops serving all requests.

This PR adds periodic waiting for all batches to be flushed before generating the next records. This behaves like a back pressure mechanism and prevents the test producer from generating the data to local heap memory.

Also, linger and batch size from passed cli args are taken into account in topic producer configuration.

### DIFF
--- a/crates/fluvio-test/src/tests/longevity/producer.rs
+++ b/crates/fluvio-test/src/tests/longevity/producer.rs
@@ -1,7 +1,7 @@
-use fluvio::RecordKey;
+use fluvio::{RecordKey, TopicProducerConfig, TopicProducerConfigBuilder};
 use fluvio_test_util::test_runner::test_driver::TestDriver;
 use fluvio_test_util::test_meta::environment::EnvDetail;
-use std::time::SystemTime;
+use std::time::{Duration, Instant, SystemTime};
 use tracing::debug;
 
 use super::LongevityTestCase;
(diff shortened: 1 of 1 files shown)


crates/fluvio-test/src/tests/longevity/producer.rs
```diff
@@ -62,6 +75,13 @@ pub async fn producer(test_driver: TestDriver, option: LongevityTestCase, produc
                 .expect("Producer Send failed");
         }
 
+        if last_flush_time.elapsed() > flush_period {
+            for p in &producers {
+                let _ = p.flush().await;
```
sehz, line 80: Why is this is necessarily? Shouldn't linger perform flush?

galibey: To hold the current thread until all records are sent reducing the overall pressure.

sehz: I see.  Since there are too many producers, internal accumulation might be too large until flushing.

sehz (approved):
LGTM

tjtelan (approved):

PR #2501, opened 2022-07-25
[Merged by Bors] - Make produce request cloneable

galibey:
This is needed for being able to retry produce requests if an error occurs.

Related #2481


sehz (approved):
LGTM.  Hopefully, this passes hourly...

PR #2504, opened 2022-07-26
[Merged by Bors] - Try to deprecate LSR field typo

morenol:
This is a PR that we need before fully changing from LSR to LRS https://github.com/infinyon/fluvio/pull/2392

Without actually renaming the field but adding support in advance to using both `lrs` and `lsr` in the deserialization of the status


sehz (approved):
LGTM

morenol:
bors retry

Issue #2007, opened 2021-12-09
[Bug]: Confirmation message when creating `smart-module`


ajhunyady:
``` 
% fluvio smart-module create credit-transactions-stream --wasm-file=./target/wasm32-unknown-unknown/release/credit_transactions.wasm

```
should return

```
 % fluvio smart-module create credit-transactions-stream --wasm-file=./target/wasm32-unknown-unknown/release/credit_transactions.wasm
smart-module `credit-stream` has been created.
```

nacardin:
Let's do this for create API for all objects

sehz:
done

PR #2513, opened 2022-07-28
[Merged by Bors] - Adds feedback and debug info to 'smart-module create'

JacobBoling:
Fixes issue #2007

### DIFF
--- a/crates/fluvio-cli/src/smartmodule/create.rs
+++ b/crates/fluvio-cli/src/smartmodule/create.rs
@@ -35,7 +35,7 @@ impl CreateSmartModuleOpt {
             wasm: SmartModuleWasm::from_binary_payload(buffer),
             ..Default::default()
         };
-        
+
         debug!(name = self.name, "creating smart-module");
         let admin = fluvio.admin().await;
         admin.create(self.name.to_string(), false, spec).await?;
--- a/crates/fluvio-cli/src/smartmodule/create.rs
+++ b/crates/fluvio-cli/src/smartmodule/create.rs
@@ -1,4 +1,5 @@
 use std::path::PathBuf;
+use tracing::debug;
 use clap::Parser;
 use crate::Result;
 use fluvio::Fluvio;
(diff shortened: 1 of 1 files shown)


crates/fluvio-cli/src/smartmodule/create.rs
```diff
@@ -34,8 +35,12 @@ impl CreateSmartModuleOpt {
             wasm: SmartModuleWasm::from_binary_payload(buffer),
             ..Default::default()
         };
+        
+        debug!("creating smart-module: {}", self.name);
```
sehz, line 39: can you use structure logging? https://docs.rs/tracing/0.1.35/tracing/macro.debug.html

sehz (changes requested):

crates/fluvio-cli/src/smartmodule/create.rs
JacobBoling: Did you put in lines 38 and 39? Sorry, I thought I did that.

sehz: something like this:
```
debug!(name = self.name, "creating smart-module");
```

crates/fluvio-cli/src/smartmodule/create.rs
```diff
@@ -34,8 +35,12 @@ impl CreateSmartModuleOpt {
...
         let admin = fluvio.admin().await;
         admin.create(self.name.to_string(), false, spec).await?;
+        println!("smart-module \"{}\" has been created.", self.name);
```
sehz, line 42: why blackslash?

JacobBoling: That is how the message is structured in topic/create

`println!("topic \"{}\" created", name);`

sehz: k

crates/fluvio-cli/src/smartmodule/create.rs
JacobBoling: Ok I changed the line to that. Also letting you know how it is done in topic/create:

`debug!("creating topic: {} spec: {:#?}", name, topic_spec);`

sehz: Yup.  There's lots of code using legacy format.

sehz (approved):
LGTM

sehz:
`cargo fmt` check failure

Issue #2510, opened 2022-07-27
Add Minimum supported Rust version (MSRV) check to CI


tjtelan:
This one caught me while trying to fix the Fluvio website CI.

The root cause of the failure was the Rust Docker image we were using as a reference environment.

It was using `rust:1.61` but new updates to the repo use new standard traits, so the minimum version the `fluvio` crate will build on is `1.62.1`.

Maybe we can use this?
https://github.com/marketplace/actions/minimum-supported-rust-version

aaronaaeng:
I'll take this.  Do you think this should be handled on merges to `master` or should it run as a weekly cron job?  I've noticed that running `msrv` takes 15-20 minutes for the repository.  I personally prefer the idea of running it on each new merge so that it's updated with the code changes.

sehz:
Yes, let's try to run on each commit.

PR #2519, opened 2022-07-30
2510 - Add Minimum Supported Rust Version Badge

aaronaaeng:
Addresses #2510 by adding an msrv badge to the `README`.  The action runs on every push to `master`.  If a different msrv is needed, the runner commits and pushes the new badge.


tjtelan (approved):
LGTM

morenol:
I think that `master` branch is a protected branch so not sure if that push will work without changing the protection rules

tjtelan:
> I think that `master` branch is a protected branch so not sure if that push will work without changing the protection rules

I think you might be right.

I found this [Github Action](https://github.com/marketplace/actions/push-to-status-check-protected-branches) that says it can handle push to protected branches, but we need to provide an access token with the permissions

tjtelan:
https://github.com/infinyon/fluvio/pull/2526 also needs a personal access token created to push to master, so this PR will be unblocked once that occurs

Issue #1820, opened 2021-10-27
CHANGELOG CI check


tjtelan:
We need a CI check to enforce updates to `CHANGELOG.md` when they are appropriate. Our convention for the changelog thus far has been source code changes that are non-test.

I think this check *may* only be needed for changes within the `crates` dir that aren't test related? So excluding the `fluvio-test-*` crates and any changes in a crate that isn't in the `src` dir, (Needs confirmation)

tjtelan:
> We could use something like:
> 
> https://github.com/rusty-celery/rusty-celery/blob/41efda696132bff7451b191d185a48129ca1b2e3/.github/workflows/ci.yml#L19

This is a step in the direction we want. The way we use our changelog is more fine grained, as we make note of changes that have a public facing impact. The way `rusty-celery` does it is to enforce a changelog entry to any updates within `src/`.

Maybe if we can agree that this is an approach we'd like to try for the crates we publish, then this would be an easy-ish task to implement.

PR #2526, opened 2022-08-02
fix: Generate CHANGELOG.md on push to `master`

tjtelan:
Generates a changelog using Git log using [`git-cliff`](https://github.com/orhun/git-cliff) + a config file that I tuned to Fluvio's repo.
- We still follow [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) format
  - The Unreleased block no longer has the version number from `VERSION`, for templating purposes. The final file has this rendered.
- We are limited to commits in the Git history, and not so much Github PRs.
  - We categorize commits based on keywords in the commit message, and display commits by their subject line.
    - This means we need to keep commit titles useful.
    - Commits that meet [Conventional Commit standards](https://www.conventionalcommits.org/en/v1.0.0/) will be categorized in changelog by those standards.
- A sample (unsquashed) changelog is generated in CI and kept as a CI artifact
- A final changelog is generated and committed after push by Bors to `master`
  - The reason we make this change directly to `master` is because the commits of the PR aren't yet squashed.

Modified the release so it will still extract the latest changelog for the release notes.

Added `pr_merge.yml` workflow that triggers on push to `master`
- Changelog is generated, merged into the most recent PR commit, and force pushed to the `master` branch.
- We need a personal access token w/ `public_repo` in order to push to `master` through the branch protection.
  - `${{ secrets.CI_COMMIT_PUSH_PAT }}`

Resolves https://github.com/infinyon/fluvio/issues/1633
Closes https://github.com/infinyon/fluvio/issues/1820


sehz:
Can you link to issue?

cliff.toml
```diff
@@ -0,0 +1,120 @@
...
+    { pattern = "\\(#([0-9]+)\\)", replace = "([#${1}](https://github.com/infinyon/fluvio/issues/${1}))"}
+]
+
+# regex for parsing and grouping commits
+commit_parsers = [
+
+    { message = ".*[Bb]ump", group = "Noise", skip = true},
+    { message = ".*[Rr]evert", group = "Noise", skip = true},
+    { message = ".*[Cl]ippy", group = "Noise", skip = true},
+    { message = "^Merge pull request", group = "Noise", skip = true},
```
morenol, line 50: Could you add a "[Ii]gnore.*" or "[Ss]kip.*", so we can skip a commit with that too?

.github/workflows/ci.yml
```diff
@@ -50,6 +50,40 @@ jobs:
       ##  if: github.event_name == 'push' && github.ref == 'refs/heads/staging'
       #  run: echo "TEST_RUNS='[r1,r2,r3,r4,r5]'" | tee -a $GITHUB_ENV
 
+  generate_changelog:
```
sehz, line 53: Looks like this was for testing.  Is this still needed?

sehz (changes requested):

.github/workflows/ci.yml
tjtelan: I suppose not, but this is an early check for the changelog rendering, even if it isn't final. I don't see the harm in having it run but we can disconnect it from any requirements to pass CI?

sehz: Gotcha.  The only issue is that the workflow is duplicated.  Can you use a re-usable workflow?  https://docs.github.com/en/actions/using-workflows/reusing-workflows

PR #2529, opened 2022-08-02
[Merged by Bors] - refactor(fluvio-test): Add testcase derive macro

tjtelan:
And delete a lot of code from the tests


sehz (approved):
LGMT.  Nice work!  less is better :-)

PR #2539, opened 2022-08-04
[Merged by Bors] - fix(stream_model): prevent collisions between namespaces

nacardin:


crates/fluvio-stream-model/src/store/k8.rs
```diff
@@ -163,11 +163,13 @@ pub trait K8ExtendedSpec: Spec {
 
     fn convert_from_k8(
         k8_obj: K8Obj<Self::K8Spec>,
+        multi_namespace_context: bool,
     ) -> Result<MetadataStoreObject<Self, K8MetaItem>, K8ConvertError<Self::K8Spec>>;
 }
 
 pub fn default_convert_from_k8<S>(
```
morenol, line 170: I think that we should add some references to the way that we should use `fluvio_stream_dispatcher::store::StoreContext` when using multiple namespaces for all the methods that have `key: S::IndexKey` as a parameter. Maybe as docstrings?

morenol (approved):

crates/fluvio-stream-model/src/store/k8.rs
nacardin: added docstring to `default_convert_from_k8`

sehz:
Can you add more description to this PR?

sehz:
This is probably not fix since this is not bug fix

PR #2557, opened 2022-08-14
refactor(SPU) Remove global contexet

sehz:
`Global Context` in the SPU is carry over from very earlier day of fluvio.  
`Once_cell` is better way to define global variables.


sehz:
Closing this for now.  Moving to static variable make it harder to test

PR #2562, opened 2022-08-16
[Merged by Bors] - feat(producer): added blocking if the batch queue is full

### LINKED ISSUE #2512: Throughput Control
We need to find a way to control the throughput that we produce during integration testing. Currently, it is limited only by available system resources and is unstable due to that fact. 

We need to figure out on which level we want to do this. Possible options: `fluvio-test` where we generate the test data or `fluvio` client where we can add a back pressure mechanism to control the throughput that goes to the Producer.

galibey:
Added the batch queue size control using `Condvar`.
If the configured queue size is reached, we block pushing new records until there is enough space for a new batch. During the flush, we notify all waiting threads that they can progress. 

Fixed #2512

### DIFF
--- a/crates/fluvio/Cargo.toml
+++ b/crates/fluvio/Cargo.toml
@@ -46,14 +46,14 @@ quantities = { version = "0.10.0", features = ["datavolume", "datathroughput", "
 hdrhistogram = { version = "7.5.0", optional = true }
 strum = { version = "0.24", features = ["derive"], optional = true }
 num-traits = { version = "0.2", optional = true }
-async-std = { version = "1.12", default-features = false, features=["std"] }
 
 # Fluvio dependencies
-fluvio-future = { version = "0.4.1", features = [
+fluvio-future = { version = "0.4.2", features = [
     "task",
     "openssl_tls",
     "task_unstable",
-    "retry"
+    "retry",
+    "sync"
 ] }
 fluvio-types = { version = "0.3.7", features = [
     "events",
@@ -81,6 +81,7 @@ fluvio-spu-schema = { version = "0.10.0", path = "../fluvio-spu-schema" }
 fluvio-spu-schema = { version = "0.10.0", path = "../fluvio-spu-schema" }
 
 [dev-dependencies]
+async-std = { version = "1.6.4", default-features = false }
 fluvio-future = { version = "0.4.0", features = ["io", "fixture"] }
 rand = "0.8.5"
 
--- a/crates/fluvio/Cargo.toml
+++ b/crates/fluvio/Cargo.toml
@@ -46,6 +46,7 @@ quantities = { version = "0.10.0", features = ["datavolume", "datathroughput", "
 hdrhistogram = { version = "7.5.0", optional = true }
 strum = { version = "0.24", features = ["derive"], optional = true }
 num-traits = { version = "0.2", optional = true }
+async-std = { version = "1.12", default-features = false, features=["std"] }
 
 # Fluvio dependencies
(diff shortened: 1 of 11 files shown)


crates/fluvio/Cargo.toml
```diff
@@ -46,6 +46,7 @@ quantities = { version = "0.10.0", features = ["datavolume", "datathroughput", "
 hdrhistogram = { version = "7.5.0", optional = true }
 strum = { version = "0.24", features = ["derive"], optional = true }
 num-traits = { version = "0.2", optional = true }
+async-std = { version = "1.12", default-features = false, features=["std"] }
```
sehz, line 49: Unfortunately, `async-std` bring large sets of dependencies.  Can you try to use https://docs.rs/event-listener?  `CondVar` uses internally.

crates/fluvio/src/producer/accumulator.rs
```diff
@@ -25,30 +27,45 @@ use crate::error::Result;
 
 use super::event::EventHandler;
 
-pub(crate) type BatchHandler = (Arc<BatchEvents>, Arc<Mutex<VecDeque<ProducerBatch>>>);
+const RECORD_ENQUEUE_TIMEOUT: Duration = Duration::from_secs(30);
+
+pub(crate) type BatchHandler = (
+    Arc<BatchEvents>,
+    Arc<(Mutex<VecDeque<ProducerBatch>>, Condvar)>,
```
sehz, line 34: Probably should abstract out: 
```
Arc<(Mutex<VecDeque<ProducerBatch>>, Condvar)>
```

as `NewType` so can independently unit test and easy to reason about

sehz (changes requested):
This approach seems to be reasonable approach with minimum complexity.

Couple of changes I would like to see:
1) Replace  `CondVar` with `EventListener` which we use lot in the code base.  This should remove need to depend on entire `async-std` 
2) Unit Test queue which I believe should be achievable by abstracting out `Mutex<VedDeque...`

crates/fluvio/Cargo.toml
galibey: The key feature of `Condvar` that `Event Listener` doesn't have is that you can `wait` under the lock: it drops the lock, goes to sleep, and then regains the lock after awaking `automatically` unnoticeable for our code. With `Event Listener` all these manipulations need to be done manually.

sehz: ok. make sense.  However, we should not bring rest of dependency from async-std which will make it harder for us to port to WASM or more strict env.  Can we just port `CondVar`?

sehz: Looks like we are including `std` due to `task` feature flag in the `fluvio_future`.   Can we re-export `CondVar` in the `fluvio-future` under `sync` flag?  That way, we don't explicit depend on `async_std` directly from Fluvio?

tjtelan (approved):
LGTM

crates/fluvio/Cargo.toml
morenol: Done

crates/fluvio/src/producer/accumulator.rs
morenol: new internal type created for that

morenol:
Feedback addressed, I also tried hourly tests with these changes to ensure that it is working as expected

sehz (approved):
LGTM.  Nice work @galibey @morenol

sehz:
hopefully this will improve hourly test stability

PR #2566, opened 2022-08-17
[Merged by Bors] - feat(cluster): add wait for SC availability

nacardin:
Added functionality to check if at least one replicas of the SC deployment is available before attempting to port forward.

Removes assumption that pod will be ready in 3 seconds

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -619,7 +619,7 @@ jobs:
         os: [ubuntu-latest]
         rust-target: [x86_64-unknown-linux-musl]
         run: [r1]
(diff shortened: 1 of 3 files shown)


sehz:
Where is `MAX_SC_DEPLOYMENT_AVAILABLE_WAIT` value defined?

nacardin:
> Where is `MAX_SC_DEPLOYMENT_AVAILABLE_WAIT` value defined?

It's not. Just like `FLV_CLUSTER_MAX_SC_SERVICE_WAIT`, its an env variable that can be set if needed.

sehz:
Got it.  thanks

sehz (approved):
LGTM

PR #2588, opened 2022-08-24
[Merged by Bors] - chore(dependencies): Fix unused dependencies check

morenol:
It failed in last push to master branch:

https://github.com/infinyon/fluvio/actions/runs/2921183973/attempts/1

### DIFF
--- a/crates/fluvio-storage/src/segments.rs
+++ b/crates/fluvio-storage/src/segments.rs
@@ -178,14 +178,6 @@ impl SegmentList {
         self.segments.len()
     }
 
-    #[instrument(skip(self))]
-    pub fn get_total_logs_len(&self) -> usize {
-        self.segments
-            .values()
-            .map(|segment| segment.get_msg_log().get_len() as usize)
-            .sum()
-    }
-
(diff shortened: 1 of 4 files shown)


crates/fluvio-storage/src/segments.rs
```diff
@@ -178,14 +178,6 @@ impl SegmentList {
         self.segments.len()
     }
 
-    #[instrument(skip(self))]
```
sehz, line 181: why was this deleted?

morenol: I saw that in the output of the failure:
```rust
warning: associated function `get_total_logs_len` is never used
   --> crates/fluvio-storage/src/segments.rs:182:12
    |
182 |     pub fn get_total_logs_len(&self) -> usize {
    |            ^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(dead_code)]` on by default
```

https://github.com/infinyon/fluvio/actions/runs/2921183973/attempts/1

sehz: got it.

sehz (approved):
LGTM

Issue #2550, opened 2022-08-11
[CI] Fix crate publish readiness check


tjtelan:
This CI run had a false positive:
https://github.com/infinyon/fluvio/runs/7687135285?check_suite_focus=true

This test should have failed: 
https://github.com/infinyon/fluvio/blob/master/.github/workflows/ci.yml#L297-L298

Which eventually resulted in a cargo publish job failing:
https://github.com/infinyon/fluvio/runs/7778435436?check_suite_focus=true

---

The cause was due to a [path dependency update](https://github.com/infinyon/fluvio/pull/2535/commits/514db8d9d59d720f28dfccbec5b2d0ab8a39e6ed#diff-90bf77bf380b339556f28c39666719c5b25074ee1667cf6879b5d3646147b773L30) where the version number of the crate was removed. 

This fine for development, since the build is unaffected. But the steps during the publish build is different. The Cargo book says that [crates with path dependencies can't be published](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-path-dependencies). The paths are ignored and cargo falls back on the public crate version.

So it seems we're exposed to this only when we are updating crates, after Bors has merged. It isn't an obvious thing to spot in a PR.

tjtelan:
Closed by https://github.com/infinyon/fluvio/pull/2589

PR #2589, opened 2022-08-24
[Merged by Bors] - test: add check for pinned versions

xenago:
Validate that crates with [multiple locations](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations) defined have a version specified/pinned.

When replicating the problem described in #2550 (removing the pinned version for `fluvio-package-index` in the `Cargo.toml` for `fluvio-extension-common`), the test now fails with a relevant error message:
image

### DIFF
--- a/release-scripts/check-publish-crates.sh
+++ b/release-scripts/check-publish-crates.sh
@@ -16,19 +16,40 @@ function cargo_publish_dry_run_all() {
         
         # Validate that dependencies with multiple locations defined have
         # a specific version pinned, to avoid failures at publish time (#2550)
+        bad_dep=""
         deps_header=0
         while read -r line; do
-            if [[ $line =~ ^\[dependencies\]$ ]]; then
-                # In the dependencies section of Cargo.toml
+            if [[ $deps_header -eq 0 ]] && \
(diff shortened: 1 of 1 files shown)


release-scripts/check-publish-crates.sh
```diff
@@ -13,6 +13,24 @@ function cargo_publish_dry_run_all() {
     for crate in "${PUBLISH_CRATES[@]}" ; do
         echo "$crate";
         pushd crates/"$crate";
+        
+        # Validate that dependencies with multiple locations defined have
+        # a specific version pinned, to avoid failures at publish time (#2550)
+        deps_header=0
+        while read -r line; do
+            if [[ $line =~ ^\[dependencies\]$ ]]; then
```
morenol, line 21: I think that we also need to handle sections like `[target.'cfg(not(target_arch = "wasm32"))'.dependencies]`

See:

https://github.com/infinyon/fluvio/blob/040655069aabb8ae358c6f5517a1cf0bbb25403f/crates/fluvio/Cargo.toml#L75

xenago: Good point. I will adjust the regex to include those headers as well.

tjtelan (approved):
LGTM. Thanks for the contribution

Issue #2590, opened 2022-08-25
[CI] Convert crate publish readiness checks to Rust


tjtelan:
These tests together are pretty useful for ensuring release doesn't surprise use with crate version number related noise, but doesn't give enough feedback for new or external contributors.

* https://github.com/infinyon/fluvio/blob/master/release-scripts/test-crate-version.sh
* https://github.com/infinyon/fluvio/blob/master/release-scripts/check-publish-crates.sh

---

For context, my ideal solution is to use something like https://github.com/obi1kenobi/cargo-semver-check, since it provides feedback about what kind of version bump to perform.

After a (very) quick check, I did not find a single tool that handles the edge cases that we look for before during publishing.

Examples of changes
- Valid, but trivial file changes, such as control flow logic, comments, whitespace, variable renames and no version change
- Valid, but Cargo.toml only changes, like for updating dependencies and no version change

This conversion will help close the gap for the ability to contribute to a community-supported crate that will handle our edge cases.

mfdorst:
So, to be clear, this issue is only proposing to translate the functionality that currently exists in `test-crate-version.sh` and `check-publish-crates.sh` into Rust, not to add any additional functionality?

tjtelan:
I suppose this is a little open ended, as long as it improves the overall reliability of the automated release process. However, the minimum is to cover the example cases I mentioned, which the scripts currently addresses during CI.

These scripts are admittedly naive, so I'm open to solutions that aren't exact ports, as long as there's justification. If you have other ideas, I'm open to hearing you out. 

Ideally, no one should need to think too hard about resolving these issues if they arise. I think this will be addressed by giving clear feedback to contributors.

mfdorst:
From what I can tell, `test-crate-version.sh` downloads `https://static.crates.io/db-dump.tar.gz` (a fairly large file) for the sole purpose of checking if the crate is published. There is a much faster way - just request `https://crates.io/api/v1/crates/{crate_name}/versions` and if it returns 404, the crate is not published. I'm using that in my rewrite unless there's some other reason to download the whole `db-dump` archive.

PR #2595, opened 2022-08-27
[Merged by Bors] - test: Rewrite test-crate-version in Rust

mfdorst:
**Notable changes**
+ Does not download `https://static.crates.io/db-dump.tar.gz` to check if crates are published (very slow)
   + Instead requests `https://crates.io/api/v1/crates/{crate_name}/versions` - 404 indicates that the crate is not published

**TODO**
+ [x] Test source diffing
+ [x] Display source changes
+ [x] Test manifest diffing
  + [x] Decide whether deleted manifest keys should be ignored or flagged
+ [x] Display manifest changes
+ [x] Download crates without an external tool (currently using `cargo-download`)
+ [ ] ~~Diff source without an external tool (currently using `diff`) - maybe? Might not be worth the effort.~~
+ [x] Add a `--verbose` mode
+ [x] Download crates concurrently
+ [x] Colorize diff

Related: #2590


mfdorst:
Thinking of using [treediff] in conjunction with [cargo-manifest] for showing changes in manifest files.
[treediff]: https://docs.rs/treediff/4.0.2/treediff/
[cargo-manifest]: https://docs.rs/cargo-manifest/0.3.0/cargo_manifest/

mfdorst:
So an interesting thing about the current state of the manifest diffing:
+ Reordering keys shows as no change (good!)
+ Deleting a key also shows not change (bad! or maybe potentially okay?)
I honestly don't know if this is a problem. Removing an unused dependency should not require a version bump (I think?) but there are probably other keys that ought to require a version bump if removed. But maybe this is a rare enough thing that we don't care? Feedback would be appreciated.

mfdorst:
I couldn't find a good crate for diffing toml files, so I'm making my own. [`toml-diff`]

[`toml-diff`]: https://github.com/mfdorst/toml-diff

sehz:
https://github.com/infinyon/k8-api/tree/master/src/k8-diff

tjtelan:
> So an interesting thing about the current state of the manifest diffing:
> 
>     * Reordering keys shows as no change (good!)
> 
>     * Deleting a key also shows no change (bad! or maybe potentially okay?)
> 
> 
> I honestly don't know if this is a problem. It may even be desirable. Removing an unused dependency should not require a version bump (I think?) but there are probably other keys that ought to require a version bump if removed. But maybe this is a rare enough thing that we don't care? Feedback would be appreciated.

The desire to have a version bump is primarily driven by our release process. Since we [use path dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-path-dependencies) for the public crates in the repo during development and testing, we have expectations that the public version is compatible after release.

So if we could be objective in order to decide what changes require or don't require a new version bump, and cover our internal dependency tree, then that sounds ok to me. As long as we're consistent between our internal dependencies.

mfdorst:
> https://github.com/infinyon/k8-api/tree/master/src/k8-diff

There might be a way to get that working for toml, but I wasn't able to easily figure it out. Good news is my `toml-diff` crate is coming along nicely - it's working for basic toml, though it could use some improvement on nested arrays/tables.

mfdorst:
I wasn't sure if I should include the crate in the source, or if I should publish it myself and have it pull from crates.io.

mfdorst:
Yes it is.

sehz:
Great.  Thanks for the nice work!
@tjtelan Please provide feedback

release-tools/check-crate-version/src/main.rs
```diff
@@ -0,0 +1,177 @@
...
+use serde::Deserialize;
+use toml_diff::TomlDiff;
+
+mod download;
+
+use download::download_crate;
+
+const PUBLISH_LIST_PATH: &str = "./publish-list.toml";
+const CRATES_DIR: &str = "../../crates";
+const CRATES_IO_DIR: &str = "./crates_io";
```
tjtelan, line 13: These paths can keep those default values, but they should be overridable via CLI flags and environment variables. I believe Clap will handle both.

release-tools/check-crate-version/src/main.rs
```diff
@@ -0,0 +1,177 @@
...
+
+    let list_toml = fs::read_to_string(PUBLISH_LIST_PATH).unwrap();
+    let list: PublishList = toml::from_str(&list_toml).unwrap();
+    list.publish_list
+}
+
+/// Checks crates.io to see if the crate has been published. Returns `true` if it has.
+async fn check_crate_published(crate_name: &str) -> bool {
+    // We have to identify ourselves, or crates.io will ignore our request
+    // TODO: Add an email to the user_agent filed - this will make us less likely to have our
```
tjtelan, line 117: you can use `team@infinyon.com` for crates.io's crawler policy

release-tools/check-crate-version/toml-diff/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "toml-diff"
+version = "0.0.0"
```
tjtelan, line 3: Can you also add `publish = false`? I don't think we intend to publish this crate at this time. At least not while it is living in this repo

release-tools/check-crate-version/toml-diff/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+
+#[derive(Debug)]
+pub enum TomlChange<'a> {
+    Added(Vec<&'a str>, &'a TomlValue),
+    Deleted(Vec<&'a str>, &'a TomlValue),
+}
+
+impl<'a> TomlDiff<'a> {
+    /// Return a list of differences between `a` and `b`. A is considered "new" and `b` is
+    /// considered "old", so items missing from `a` are considdered "deletions", while items
```
tjtelan, line 22: Small nitpick, a few typos of the same word this this doc comment.

`considdered` -> `considered`

release-tools/check-crate-version/toml-diff/src/test.rs
```diff
@@ -0,0 +1,245 @@
+use super::{TomlChange, TomlDiff};
```
tjtelan, line 1: These tests look good, but can you add a control case where there is no diff

release-tools/check-crate-version/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "check-crate-version"
+version = "0.1.0"
```
tjtelan, line 3: Can you update the version of this crate to `0.0.0` and add `publish = false`? If decide to publish this, it most likely won't be from this repo.

release-tools/check-crate-version/src/main.rs
```diff
@@ -0,0 +1,177 @@
...
+    let padding = publish_list
+        .iter()
+        .fold(0, |len, name| max(len, name.len()));
+
+    let status_checks: Vec<_> = publish_list
+        .into_iter()
+        .map(|crate_name| {
+            (
+                crate_name.clone(),
+                tokio::spawn(check_crate_status(crate_name)),
```
tjtelan, line 44: Can we add a rate limit for these checks so we can abide by crates.io's crawler policy of `1 request per second or less`?

https://crates.io/policies#crawlers

.github/workflows/ci.yml
```diff
@@ -297,14 +297,14 @@ jobs:
         run: make run-all-unit-test
       - name: Crate version check
         if: matrix.check == 'crate-version'
-        run: ./release-scripts/test-crate-version.sh
+        run: ./release-tools/test-crate-version.sh
```
tjtelan, line 300: You should run your tool to perform this check

tjtelan (commented):
This looks like it took a lot of effort and will be a lot easier to maintain. Good job with `toml-diff`!

release-tools/check-crate-version/toml-diff/src/lib.rs
mfdorst: I don't think I copied those lines, I think my fingers think it's spelled like that lol

release-tools/check-crate-version/Cargo.toml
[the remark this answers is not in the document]
mfdorst: Fixed

release-tools/check-crate-version/src/main.rs
mfdorst: Implemented

release-tools/check-crate-version/src/main.rs
mfdorst: Implemented

release-tools/check-crate-version/toml-diff/Cargo.toml
mfdorst: Fixed

release-tools/check-crate-version/toml-diff/src/lib.rs
mfdorst: Fixed

release-tools/check-crate-version/toml-diff/src/test.rs
mfdorst: Added

release-tools/check-crate-version/Cargo.toml
mfdorst: Fixed

release-tools/check-crate-version/src/main.rs
mfdorst: Bummer, cause it means there was no point in making this async, but you're right, that is what the crawler policy says. Implemented.

.github/workflows/ci.yml
mfdorst: Implemented

tjtelan (approved):
LGTM, I really appreciate the hard work that went into this

release-tools/check-crate-version/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "check-crate-version"
+version = "0.0.0"
+publish = false
+edition = "2021"
+
```
sehz, line 6: can you add a license?  this should be apache2

release-tools/check-crate-version/toml-diff/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+publish = false
+edition = "2021"
+
```
sehz, line 6: please add license

sehz (changes requested):
Overall looks reasonable.  Just a couple of minor nits regarding the license.

release-tools/check-crate-version/Cargo.toml
mfdorst: Done

release-tools/check-crate-version/toml-diff/Cargo.toml
mfdorst: Done

sehz (approved):
LGTM.   This is a fairly good effort.  Thanks for working on this.

PR #2599, opened 2022-08-30
feat(cli): added support of `transform` section for connector config

galibey:
Added `transform` section to Connector config.
The yaml config example:
```yaml
transform:
  - uses: infinyon/json-sql
    invoke: insert
    with:
      table: "topic_message"
      map-columns:
        "device_id":
          json-key: "device.device_id"
          value:
            type: "int"
            default: 0
            required: true
        "record":
          json-key: "$"
          value:
            type: "jsonb"
            required: true
```
Everything in `with` section has a free format depending on the smartmodule receiver, which defined in `uses` section.

### DIFF
--- a/crates/fluvio-cli/src/connector/mod.rs
+++ b/crates/fluvio-cli/src/connector/mod.rs
@@ -341,9 +341,9 @@ where
         {
             let json: serde_json::Value =
                 Deserialize::deserialize(serde::de::value::MapAccessDeserializer::new(map))?;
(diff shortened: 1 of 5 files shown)


galibey:
Closed as this functionality is about to be deprecated. The PR to the new one is coming instead.

ajhunyady:
what does this mean? will the new PR have the same requirements?

sehz:
Yes, this will be moved to connector repo

Issue #2597, opened 2022-08-28
Remove dependency on POSIX coreutils


BatmanAoD:
This is a suggested companion ticket to #1131.

The following `Command`s invoke coreutils:

```
Command::new("pgrep")
Command::new("pkill")
Command::new("sync")
```

These are not available by default on Windows, and the functionality they provide should be available either in the standard library or in Rust crates.

`pgrep` and `pkill` can be replaced using the SysInfo crate ([process search](https://docs.rs/sysinfo/latest/sysinfo/trait.SystemExt.html#method.processes_by_exact_name), [signalling](https://docs.rs/sysinfo/latest/sysinfo/trait.ProcessExt.html#tymethod.kill_with)).

The POSIX `sync` doesn't appear to have an exact equivalent in the Windows API; it appears that Windows only permits syncing [specific filehandles](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-flushfilebuffers). But it seems preferable to only flush the files pertinent to Fluvio anyway, rather than all system files.

sehz:
fixed

PR #2604, opened 2022-09-03
[Merged by Bors] - fix: Remove dependency on POSIX coreutils #2597 

mckaymatt:
https://github.com/infinyon/fluvio/issues/2597 There are two commits. The first addresses.

> The POSIX `sync` doesn't appear to have an exact equivalent in the Windows API; it appears that Windows only permits syncing [specific filehandles](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-flushfilebuffers). But it seems preferable to only flush the files pertinent to Fluvio anyway, rather than all system files.

I looked into it and it seems like on Windows `flush()` is noop, but `sync_all()` calls `FlushFileBuffers`.  
I looked around for places where it seems like config files are being written and I added a `sync_all`.  It is expensive though, so you may want to cross-reference and make sure I didn't place it anywhere it doesn't belong.

We can see what is being called on the Windows by tracing back starting with the all-platform `std::fs:sync_all` 
https://github.com/rust-lang/rust/blob/1.63.0/library/std/src/fs.rs#L424

The Windows implementation of `fsync` calls `FlushFileBuffers`  https://github.com/rust-lang/rust/blob/1.63.0/library/std/src/sys/windows/fs.rs#L307

On windows (and unix) `flush()` is a noop. https://github.com/rust-lang/rust/blob/1.63.0/library/std/src/sys/windows/fs.rs#L447

The second commit replaces calls to `pgrep` and `pkill` with the `Sysinfo` library. 
I was able to run the cluster locally using `kind`, start the `--local` deploy, and use `ps` to see the the target procs are both created and killed.
```
$ flvd cluster start --local --develop

# See the local procs
$ ps x |grep fluvio 
1663086 pts/0    Sl     0:00 /home/user/workspace/fluvio/target/debug/fluvio run sc --local
1663100 pts/0    Sl     0:02 /home/user/workspace/fluvio/target/debug/fluvio-run sc --local
1663131 pts/0    Sl     0:00 /home/user/workspace/fluvio/target/debug/fluvio run spu -i 5001 -p 0.0.0.0:9010 -v 0.0.0.0:9011 --log-base-dir /home/user/.fluvio/data
1663145 pts/0    Sl     0:01 /home/user/workspace/fluvio/target/debug/fluvio-run spu -i 5001 -p 0.0.0.0:9010 -v 0.0.0.0:9011 --log-base-dir /home/user/.fluvio/data

# Run cluster delete
$ flvd cluster delete
Uninstalled fluvio kubernetes components
Uninstalled fluvio local components
Objects and secrets have been cleaned up

# See that the procs are killed.
$ ps  |grep -c fluvio
0
```

### DIFF
--- a/crates/fluvio-channel/src/lib.rs
+++ b/crates/fluvio-channel/src/lib.rs
@@ -345,7 +345,8 @@ impl ChannelConfig {
             toml::to_vec(self).map_err(|err| IoError::new(ErrorKind::Other, format!("{}", err)))?;
 
         let mut file = File::create(path_ref)?;
-        file.write_all(&toml)
+        file.write_all(&toml)?;
+        file.sync_all()
     }
 }
 
--- a/crates/fluvio-cluster/src/delete.rs
+++ b/crates/fluvio-cluster/src/delete.rs
@@ -3,6 +3,7 @@ use std::fs::remove_dir_all;
 
 use derive_builder::Builder;
 use tracing::{info, warn, debug, instrument};
+use sysinfo::{ProcessExt, System, SystemExt};
 
 use fluvio_command::CommandExt;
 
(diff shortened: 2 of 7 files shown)


sehz:
Thanks for PR!

sehz:
@morenol Please review

mckaymatt:
The tests are failing because it's trying to kill a proc owned by root. 
The reason this wasn't an issue before is because the `Command` returns `Ok` even if the command returns a non-zero status. [example here](https://play.rust-lang.org/?code=%23!%5Ballow(unused)%5D%0Afn%20main()%20%7B%0Ause%20std%3A%3Aprocess%3A%3ACommand%3B%0Ause%20std%3A%3Aio%3A%3A%7Bself%2C%20Write%7D%3B%0Alet%20output%20%3D%20Command%3A%3Anew(%22%2Fbin%2Fcat%22)%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20.arg(%22file.txt%22)%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20.output()%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20.expect(%22failed%20to%20execute%20process%22)%3B%0A%0Aprintln!(%22status%3A%20%7B%7D%22%2C%20output.status)%3B%0Aio%3A%3Astdout().write_all(%26output.stdout).unwrap()%3B%0Aio%3A%3Astderr().write_all(%26output.stderr).unwrap()%3B%0A%0Aassert!(output.status.success())%3B%0A%7D&edition=2021)

I'll change it to only debug log, not error, when the kill command fails. This will preserve the old behavior. 
It seems like this should only run for a `--local` install, where the procs don't run as root, but it isn't for some reason.

crates/fluvio-channel/src/lib.rs
```diff
@@ -345,7 +345,8 @@ impl ChannelConfig {
             toml::to_vec(self).map_err(|err| IoError::new(ErrorKind::Other, format!("{}", err)))?;
 
         let mut file = File::create(path_ref)?;
-        file.write_all(&toml)
+        file.write_all(&toml)?;
+        file.sync_all()
```
morenol, line 349: Since this is only important in windows, I thin that we could use a feature flag here:
```#[cfg(windows)]```

We could use https://github.com/alexcrichton/cfg-if

The same applies to all the changes in the first commit

File: crates/fluvio-cluster/src/delete.rs [hunk 2]
```diff
@@ -131,23 +132,36 @@ impl ClusterUninstaller {
...
-            .arg("fluvio run")
-            .output()
-            .map_err(UninstallError::IoError)?;
-        Command::new("pkill")
-            .arg("-f")
-            .arg("fluvio-run")
-            .output()
-            .map_err(UninstallError::IoError)?;
+
+        let kill_proc = |name: &str, commands: Option<&[String]>| {
```
morenol, line 137: nit: I see that sysinfo api uses `cmd` but I think that maybe is more appropriate to use something like `command_args` instead of `commands`

morenol (changes requested):
Nice work! Thanks for your PR @mckaymatt!

LGTM!

I think that we could use feature flag to only run `sync_all` in windows

morenol (approved):
LGTM! thanks!

sehz:
Rust failing to install on Mac....  let's try again

PR #2605, opened 2022-09-04
fix: #1131 - remove `kubectl` for `check` and `delete` operations

BatmanAoD:
This is a partial implementation of #1131. It relies on https://github.com/infinyon/k8-api/pull/191. It only addresses two uses of `kubectl`, one for checking the k8s server version and one for deleting collections of objects.

I couldn't find valid `Spec` type definitions corresponding to SPU groups, persistent volume claims, or managed connectors. Each of these object types does have a `...Spec` type defined, but they implement `fluvio_stream_model::core::core_model::Spec` rather than the required `k8_types::spec_def::Spec`. I'm guessing that these _should_ implement both `Spec` traits, but I don't feel like I understand this part of the code well enough to dive in yet. Notably, it looks to me like SPU groups _do_ have a valid `Spec`-implementing type, `spg::k8::spec::K8SpuGroupSpec`, but `spg::k8` is a private module and it doesn't look like `K8SpuGroupSpec` is publicly re-exported anywhere.

I also found `TableFormatSpec` rather than `TableSpec`, so I'm not sure that's the right `Spec` to use, either.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1735,7 +1735,7 @@ dependencies = [
  "humantime-serde",
  "indicatif",
  "k8-client",
- "k8-config 2.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "k8-config 2.0.0",
  "k8-types",
  "pretty_assertions",
  "semver 1.0.13",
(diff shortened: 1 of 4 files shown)


sehz:
nice progress

PR #2609, opened 2022-09-08
[Merged by Bors] - fix: use `wait_time` by default in `StoreContext`

morenol:
`fn wait_action` should use `self.wait_time by default instead of `MAX_WAIT_TIME` const.

That way we can use `set_wait_time` and the `StoreContext` will use that as it would be expected. Note that default value of `wait_time` is `MAX_WAIT_TIME`.


sehz (approved):
LGTM

PR #2620, opened 2022-09-25
feat: `SmartModuleSourceCode` fields public

EstebanBorai:
Makes `SmartModuleSourceCode` fields public in order to be visible for API consumers.

---

Currently on [fluvio-client-node/#262][1] an instance of `Metadata<SmartModuleSpec>` must be returned as a JavaScript object for user consumption. The inner `SmartModuleSpec` holds a `SmartModuleSourceCode` instance, which fields are not accesible to external crates.

[1]: https://github.com/infinyon/fluvio-client-node/pull/262
[2]: https://github.com/infinyon/fluvio-client-node/blob/285faa60452d611569e45c9e06a60f5c42f5637a/src/admin.rs#L516-L540


morenol (commented):
Those fields are right now mostly for metadata and not being used. I think that we should not expose them in fluvio clients yet

EstebanBorai:
> Those fields are right now mostly for metadata and not being used. I think that we should not expose them in fluvio clients yet

Got it! I'm closing this then! Thanks for the quick prompt!

Issue #2621, opened 2022-09-26
SMDK: Generate new Smart module project 


sehz:
Generate Rust SmartModule project for SmartModule development kit (SMDK).
It has two command, `initialize` and `new`.  

1. New command will create new project in separate directory
```
$ smdk new <project_name>
```
It should prompt for 
* Smart Module type
* Whether to include initialization (if so, generate generate skeleton)

After prompt, it should generate Cargo project plus `SmartModule.toml` metadata

2. Init command will create project in same directory using project name from direct name (similar to Cargo)

It should include cargo profile to set default to `wasm32-unknown-unknown` and LTO flag set

Generate command should be able to generate `smartmodule/sm-regex`

EstebanBorai:
Hi @sehz, I'm working on this feature on #2630 and I wanted to know how do we manage different SmartModule versions?
So far the branch at #2630 generates the project similar to how we do today with `cargo-generate` by running `fluvio smart-module generate` as follows:

```bash
fluvio smart-module generate <NAME>
```

Where would we take versions from in this case? I can see the template on https://github.com/infinyon/fluvio-smartmodule-template has no versions/tags available neither.

sehz:
This depends on getting initial `SMDK` setting up crate

PR #2630, opened 2022-09-29
[Merged by Bors] - feat: generate SmartModules using SMDK

EstebanBorai:
Provides support to generate new SmartModules using `cargo-generate` as a libary.

Resolve: https://github.com/infinyon/fluvio/issues/2621

---

### Generates SmartModule Projects using `smdk generate`

```bash
➜  Desktop smdk generate log-filter
Generating new SmartModule project: log-filter
🔧   Destination: /Users/esteban/Desktop/log-filter ...
🔧   Generating template ...
✔ 🤷   Which type of SmartModule would you like? · filter
✔ 🤷   Want to use SmartModule parameters? · false
🤷   SmartModule Version [default: git = "https://github.com/infinyon/fluvio.git"]: git = "https://github.com/infinyon/fluvio.git"
✔ 🤷   Want to use SmartModule init? · true
Ignoring: /var/folders/0g/l702s31x2mj4zdlw2_43rpyr0000gn/T/.tmpwQ4Yql/cargo-generate.toml
[1/5]   Done: Cargo.toml
[2/5]   Done: README.md
[3/5]   Done: Smart.toml
[4/5]   Done: src/lib.rs
[5/5]   Done: src
🔧   Moving generated files into: `/Users/esteban/Desktop/log-filter`...
💡   Initializing a fresh Git repository
✨   Done! New project created /Users/esteban/Desktop/log-filter
```

### Generated Project

```bash
total 24
drwxr-xr-x  7 esteban  staff   224B Oct  2 13:54 .
drwx------+ 6 esteban  staff   192B Oct  2 13:54 ..
drwxr-xr-x  9 esteban  staff   288B Oct  2 13:54 .git
-rw-r--r--  1 esteban  staff   358B Oct  2 13:54 Cargo.toml
-rw-r--r--  1 esteban  staff   1.8K Oct  2 13:54 README.md
-rw-r--r--  1 esteban  staff   152B Oct  2 13:54 Smart.toml
drwxr-xr-x  3 esteban  staff    96B Oct  2 13:54 src
```

### Generated `src/lib.rs` with the provided options

```rust
use fluvio_smartmodule::dataplane::smartmodule::{SmartModuleExtraParams, SmartModuleInitError};

use once_cell::sync::OnceCell;
use fluvio_smartmodule::eyre;



use fluvio_smartmodule::{smartmodule, Result, Record};

#[smartmodule(filter)]
pub fn filter(record: &Record) -> Result<bool> {
    let string = std::str::from_utf8(record.value.as_ref())?;
    Ok(string.contains('a'))
}




static CRITERIA: OnceCell<String> = OnceCell::new();

#[smartmodule(init)]
fn init(params: SmartModuleExtraParams) -> Result<()> {
    // You can refer to the example SmartModules in Fluvio's GitHub Repository
    // https://github.com/infinyon/fluvio/tree/master/smartmodule
    if let Some(key) = params.get("key") {
        CRITERIA.set(key.clone()).map_err(|err| eyre!("failed setting key: {:#?}", err))
    } else {
        Err(SmartModuleInitError::MissingParam("key".to_string()).into())
    }
}
```

### Building SMDK Project

```bash
➜  Desktop cd ./log-filter
➜  log-filter git:(main) ✗ smdk build
    Updating git repository `https://github.com/infinyon/fluvio.git`
    Updating crates.io index
   Compiling proc-macro2 v1.0.46
```

### Testing SM Project using `smdk test`

```
➜  log-filter git:(main) ✗ smdk test --params "key=a" --text "testing"
project name: "log-filter"
loading module at: target/wasm32-unknown-unknown/release-lto/log_filter.wasm
0 records outputed
➜  log-filter git:(main) ✗ smdk test --params "key=a" --text "tasty"
project name: "log-filter"
loading module at: target/wasm32-unknown-unknown/release-lto/log_filter.wasm
1 records outputed
tasty
```

### DIFF
--- a/crates/smdk/src/generate.rs
+++ b/crates/smdk/src/generate.rs
@@ -1,34 +1,54 @@
 use anyhow::{Error, Result};
 use clap::Parser;
 use cargo_generate::{GenerateArgs, TemplatePath, generate};
+use include_dir::{Dir, include_dir};
 
-const DEFAULT_TEMPLATE: &str =
-    "https://github.com/infinyon/fluvio/tree/master/smartmodule/cargo_template";
+static SMART_MODULE_TEMPLATE: Dir<'_> =
+    include_dir!("$CARGO_MANIFEST_DIR/../../smartmodule/cargo_template");
 
 /// Generate new SmartModule project
 #[derive(Debug, Parser)]
 pub struct GenerateOpt {
+    /// SmartModule Project Name
     name: String,
     /// Template to generate project from.
     ///
     /// Must be a GIT repository
-    #[clap(long, default_value = DEFAULT_TEMPLATE)]
-    temaplate: String,
+    #[clap(long)]
+    template: Option<String>,
 }
 
 impl GenerateOpt {
-    pub(crate) fn process(&self) -> Result<()> {
+    pub(crate) fn process(self) -> Result<()> {
         println!("Generating new SmartModule project: {}", self.name);
-        let template_path = TemplatePath {
-            git: Some(String::from(self.temaplate.as_str())),
-            auto_path: None,
-            subfolder: None,
-            test: false,
-            branch: None,
-            tag: None,
-            path: None,
-            favorite: None,
+
+        let template_path = match self.template {
+            Some(git) => TemplatePath {
+                git: Some(git),
+                auto_path: None,
+                subfolder: None,
+                test: false,
+                branch: None,
+                tag: None,
+                path: None,
+                favorite: None,
+            },
+            None => {
+                let path = SMART_MODULE_TEMPLATE.path().to_str().unwrap().to_string();
+
+                TemplatePath {
+                    git: None,
+                    auto_path: None,
+                    subfolder: None,
+                    test: false,
+                    branch: None,
+                    tag: None,
+                    path: Some(path),
+                    favorite: None,
+                }
+            }
         };
+
         let args = GenerateArgs {
             template_path,
             name: Some(self.name.clone()),
--- a/crates/smdk/src/generate.rs
+++ b/crates/smdk/src/generate.rs
@@ -2,8 +2,9 @@ use anyhow::{Error, Result};
 use clap::Parser;
 use cargo_generate::{GenerateArgs, TemplatePath, generate};
 use include_dir::{Dir, include_dir};
+use tempdir::TempDir;
 
-static SMART_MODULE_TEMPLATE: Dir<'_> =
+static SMART_MODULE_TEMPLATE: Dir<'static> =
     include_dir!("$CARGO_MANIFEST_DIR/../../smartmodule/cargo_template");
 
 /// Generate new SmartModule project
@@ -22,9 +23,9 @@ impl GenerateOpt {
     pub(crate) fn process(self) -> Result<()> {
         println!("Generating new SmartModule project: {}", self.name);
 
-        let template_path = match self.template {
-            Some(git) => TemplatePath {
-                git: Some(git),
+        let template_path = if self.template.is_some() {
+            TemplatePath {
+                git: self.template,
                 auto_path: None,
                 subfolder: None,
                 test: false,
@@ -32,20 +33,22 @@ impl GenerateOpt {
                 tag: None,
                 path: None,
                 favorite: None,
-            },
-            None => {
-                let path = SMART_MODULE_TEMPLATE.path().to_str().unwrap().to_string();
+            }
+        } else {
+            let temp_dir = TempDir::new("smartmodule_template")?;
+            let path = temp_dir.path().to_str().unwrap().to_string();
 
-                TemplatePath {
-                    git: None,
-                    auto_path: None,
-                    subfolder: None,
-                    test: false,
-                    branch: None,
-                    tag: None,
-                    path: Some(path),
-                    favorite: None,
-                }
+            SMART_MODULE_TEMPLATE.extract(temp_dir).map_err(Error::from)?;
+
+            TemplatePath {
+                git: None,
+                auto_path: None,
+                subfolder: None,
+                test: false,
+                branch: None,
+                tag: None,
+                path: Some(path),
+                favorite: None,
             }
         };
 
--- a/crates/smdk/src/generate.rs
+++ b/crates/smdk/src/generate.rs
@@ -19,12 +19,24 @@ pub struct GenerateOpt {
     template: Option<String>,
 }
 
+/// Abstraction on different of template options available for generating a
+/// new SmartModule project.
+///
+/// May hold a reference to a `TempDir` which should not be dropped before
+/// accomplishing the project generation procedure.
 struct Template {
     template_path: TemplatePath,
     _temp_dir: Option<TempDir>,
 }
 
 impl Template {
+    /// Extracts inlined directory contents into a temporary directory and
+    /// builds a `TemplatePath` instance with the `path` pointing to the temp
+    /// directory created.
+    ///
+    /// Is important to hold the reference to the `_temp_dir` until generation
+    /// process is completed, otherwise the temp directory will be deleted
+    /// before reaching the generation process.
     fn inline() -> Result<Self> {
         let temp_dir = TempDir::new("smartmodule_template")?;
         let path = temp_dir.path().to_str().unwrap().to_string();
--- a/crates/smdk/src/generate.rs
+++ b/crates/smdk/src/generate.rs
@@ -1,0 +1,52 @@
+use anyhow::{Error, Result};
+use clap::Parser;
+use cargo_generate::{GenerateArgs, TemplatePath, generate};
+
+const GIT_TEMPLATE: &str =
+    "https://github.com/infinyon/fluvio/tree/master/smartmodule/cargo_template";
+
+/// Generate new SmartModule project
+#[derive(Debug, Parser)]
+pub struct GenerateOpt {
+    name: String,
+}
+
+impl GenerateOpt {
+    pub(crate) fn process(&self) -> Result<()> {
+        println!("Generating new SmartModule project: {}", self.name);
+        let template_path = TemplatePath {
+            git: Some(String::from(GIT_TEMPLATE)),
+            auto_path: None,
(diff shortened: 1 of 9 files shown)


crates/fluvio-cli/src/client/smartmodule/generate.rs
```diff
@@ -0,0 +1,65 @@
...
+}
+
+#[async_trait]
+impl ClientCmd for GenerateSmartModuleOpt {
+    async fn process_client<O: Terminal + Debug + Send + Sync>(
+        self,
+        _out: Arc<O>,
+        _fluvio: &Fluvio,
+    ) -> Result<()> {
+        let template_path = TemplatePath {
```
EstebanBorai, line 30: Neither this `struct` nor `GenerateArgs` implements Default trait. So I had to provide initial values for both.
I'll open a PR to the generate crate to see if they allow us to have a `Default` implementation.

crates/fluvio-cli/src/error.rs
```diff
@@ -28,6 +28,8 @@ pub enum CliError {
     OutputError(#[from] OutputError),
     #[error("Failed to parse format string")]
     TemplateError(#[from] TemplateError),
+    #[error("Failed to generate new SmartModule")]
+    GenerateError(String),
```
EstebanBorai, line 32: The error type used in `cargo-generate` is not exported. Checking on the source code I saw they are using `anyhow` but is not exported, so I'm taking the error message as a `String` and passing it to the bubbled error instance.

EstebanBorai:
Addresses https://github.com/infinyon/fluvio/issues/2631

sehz:
see: https://github.com/infinyon/fluvio/pull/2632

sehz:
Updated #2631 per new Smdk spec

EstebanBorai:
> Updated #2631 per new Smdk spec

Awesome! Here I'm also taking care of the build step by spawning a `cargo build` command.
Heres the file:

https://github.com/infinyon/fluvio/pull/2630/files#diff-953ad69fcfe3250bce68a4d31f2e069b45e53b9448c737fc0d22b6fe77762527

EstebanBorai:
> see: #2632

Great! Will migrate to `crates/smdk/` then. Thanks for sharing!

crates/smdk/src/generate.rs
```diff
@@ -1,14 +1,51 @@
...
 impl GenerateOpt {
     pub(crate) fn process(&self) -> Result<()> {
         println!("Generating new SmartModule project: {}", self.name);
+        let template_path = TemplatePath {
+            git: Some(String::from(GIT_TEMPLATE)),
+            auto_path: None,
+            subfolder: None,
+            test: false,
+            branch: None,
+            tag: None,
```
EstebanBorai, line 22: I think we could keep different versions of the SM templates in https://github.com/infinyon/fluvio-smartmodule-template and use this `tag` field to specify which to use. WDYT?

sehz: No opinion.  There are so many ways to slice templates.  Key factor is reduce duplication so no repeating logic

crates/fluvio-cli/src/error.rs
sehz: Let's use anyhow everywhere.

EstebanBorai: Yeah! I noticed smdk already uses it.
This code doesn't lives here anymore because it was moved to `smdk` crate.

EstebanBorai:
@sehz now we have support for testing the SmartModule build using `smdk build` command.
An example on usage is in the PR comment.

crates/smdk/src/build.rs
```diff
@@ -0,0 +1,32 @@
...
+use std::process::Command;
+use std::fmt::Debug;
+
+/// Builds the SmartModule in the current working directory into a WASM file
+#[derive(Debug, Parser)]
+pub struct BuildOpt;
+
+impl BuildOpt {
+    pub(crate) fn process(&self) -> Result<()> {
+        let mut cmd = Command::new("cargo");
```
morenol, line 12: maybe we should check that `cargo` is installed first

crates/smdk/src/generate.rs
```diff
@@ -1,14 +1,51 @@
+use anyhow::{Error, Result};
 use clap::Parser;
-use anyhow::Result;
+use cargo_generate::{GenerateArgs, TemplatePath, generate};
+
+const GIT_TEMPLATE: &str = "https://github.com/infinyon/fluvio-smartmodule-template";
```
morenol, line 5: I think that we should use: https://github.com/infinyon/fluvio/tree/master/smartmodule/cargo_template instead

crates/smdk/src/build.rs
EstebanBorai: Sure! Now I'm checking `cargo version` and catching any error while running that before going straight to the `cargo build` cmd.

This way if something happens running `cargo version` we acknowledge there's an issue with `cargo` binary instead of the build.

sehz: or other ways it to catch exception if it doesn't exist and display proper error message

crates/smdk/src/generate.rs
sehz: Template should be inlined so there is no dependency to external file

crates/smdk/src/build.rs
```diff
@@ -0,0 +1,45 @@
+use anyhow::{Error, Result};
+use clap::Parser;
+use std::process::Command;
```
sehz, line 3: use our standard import format which is: std, third party,  internal dependencies

crates/smdk/src/build.rs
```diff
@@ -0,0 +1,45 @@
...
+impl BuildOpt {
+    pub(crate) fn process(&self) -> Result<()> {
+        BuildOpt::check_cargo_exists()?;
+
+        let mut cmd = Command::new("cargo");
+        let cwd = std::env::current_dir()?;
+
+        cmd.current_dir(&cwd)
+            .arg("build")
+            .arg("--profile")
```
sehz, line 19: make profile as optional so it can be overridden.  see TestOption.  Default should be made as crate constant

crates/smdk/src/build.rs
```diff
@@ -0,0 +1,45 @@
...
+            .arg("release-lto")
+            .arg("--lib");
+        cmd.arg("--target").arg("wasm32-unknown-unknown");
```
sehz, line 22: `wasm32-unknown-unknown` as constant

crates/smdk/src/generate.rs
```diff
@@ -1,14 +1,52 @@
+use anyhow::{Error, Result};
 use clap::Parser;
-use anyhow::Result;
+use cargo_generate::{GenerateArgs, TemplatePath, generate};
+
+const GIT_TEMPLATE: &str =
+    "https://github.com/infinyon/fluvio/tree/master/smartmodule/cargo_template";
```
sehz, line 6: inline template

sehz: but have option to override

sehz (changes requested):
Can you also attach running output of working?

EstebanBorai:
@sehz code related to building SmartModules lives now on this PR: https://github.com/infinyon/fluvio/pull/2638

crates/smdk/src/generate.rs
```diff
@@ -0,0 +1,119 @@
...
+
+impl Template {
+    /// Extracts inlined directory contents into a temporary directory and
+    /// builds a `TemplatePath` instance with the `path` pointing to the temp
+    /// directory created.
+    ///
+    /// Is important to hold the reference to the `_temp_dir` until generation
+    /// process is completed, otherwise the temp directory will be deleted
+    /// before reaching the generation process.
+    fn inline() -> Result<Self> {
```
sehz, line 40: let's add test to ensure inline unpacking works

sehz (changes requested):
Also, can you add output from generating->build->test including init?
I believe template need handling for `init`

crates/smdk/src/generate.rs
EstebanBorai: Of course, here [d6987fd](https://github.com/infinyon/fluvio/pull/2630/commits/d6987fdd490df93f63c8c770749c05b9f10d5b15) we have tests for each step of the process, making sure the temp directory is created and also one of the files from the template `Smart.toml` is available in such temp directory as well.

EstebanBorai:
> Also, can you add output from generating->build->test including init?
> 
> I believe template need handling for `init` 
> 
> 
> 
> 
> 
> 

Current template has handling for `init`: https://github.com/infinyon/fluvio/blob/f3f381bd3859b73363ac7c3374a72938c46ed257/smartmodule/cargo_template/cargo-generate.toml#L11

But I dont find any effect from such option in the `lib.rs` from the same template.

https://github.com/infinyon/fluvio/blob/master/smartmodule/cargo_template/src/lib.rs

sehz:
That needs to be added. See regex project

smartmodule/cargo_template/Cargo.toml
```diff
@@ -11,7 +11,12 @@ crate-type = ['cdylib']
 fluvio-smartmodule = { {{smartmodule-version}} }
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
+{% if smartmodule-init %}
+{% if smartmodule-type == "filter" %}
+once_cell = "1.13.0"
+{% endif %}
+{% endif %}
```
EstebanBorai, line 18: Used for example filter with init on generation.

EstebanBorai:
> Also, can you add output from generating->build->test including init?
> I believe template need handling for `init`

Handling for `init` is now provided

EstebanBorai:
@sehz im stick to the RegExp project version instead of using the current cargo-generate template for this generation. Also added `once_cell` to provide an example on a filter smart module with `init`.

sehz (approved):
LGTM.  Nice work!

sehz:
Mac VM failures

sehz:
jeez.  Murphy's law strikes...

Issue #2631, opened 2022-09-29
Tracking: SmartModule Development Kit


sehz:
Tracking list for SmartModule Development Kit (SMDK).

SMDK is a new top level CLI designed for SmartModule developer:

```
$ smdk new my-filter
Creating new SmartModule project...


$ smdk build

$ smdk
```

TODO:
- [x] Initial Skeleton
- [x] #2644
- [x] Project generation `smdk new` @EstebanBorai  #2621 
- [x] Build `smdk  build` @EstebanBorai 
- [x] Test `smdk test` @sehz 
- [x] Load `smdk  load` @sehz 
- [x] Publish `smdk publish` @digikata

sehz:
Done!

PR #2632, opened 2022-09-29
[Merged by Bors] - feat(smdk): initial smdk crate project

sehz:
add initial skeleton for: #2631


crates/smdk/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+
+[[bin]]
+name = "smdk"
+path = "src/main.rs"
+doc = false
+
+
+[dependencies]
+tracing = "0.1.19"
+clap = { version = "3.1.8", features = ["std", "derive"], default-features = false }
```
morenol, line 21: it looks like we can upgrade this to 4.X.X

sehz: Let's do in separate PR

morenol: I created https://github.com/infinyon/fluvio/issues/2633 and marked as `good_first_issue` just in case anyone external want to contribute with that

PR #2638, opened 2022-09-30
[Merged by Bors] - feat(smdk): support to build `SmartModules` using `smdk build`

EstebanBorai:
Introduces support to build SmartModules using the `smdk build` command.

---

```bash
➜  sm-regex git:(feat/smdk-build) ../../target/debug/smdk build --help
smdk-build
Builds SmartModule into WASM

USAGE:
    smdk build [OPTIONS]

OPTIONS:
    -h, --help                 Print help information
        --release <RELEASE>    Build release profile [default: release-lto]
```


EstebanBorai:
This PR depends on https://github.com/infinyon/fluvio/pull/2630

sehz:
This should be decouple from #2630.  It should be enough to build: `sm-regex`

EstebanBorai:
> This should be decouple from #2630. It should be enough to build: `sm-regex`

Sure! Let me remove generation bits then so we can merge.

crates/smdk/Cargo.toml
```diff
@@ -23,6 +23,7 @@ dirs = "4.0.0"
 thiserror = "1.0.20"
 anyhow = { version = "1.0.38" }
 cargo_metadata = "0.15.0"
+cargo-generate = "0.16.0"
```
sehz, line 26: not need for build

crates/smdk/src/build.rs
```diff
@@ -0,0 +1,53 @@
...
+    /// Build release profile
+    #[clap(long, default_value = DEFAULT_RELEASE_PROFILE)]
+    release: String,
+}
+
+impl BuildOpt {
+    pub(crate) fn process(&self) -> Result<()> {
+        BuildOpt::check_cargo_exists()?;
+
+        let mut cmd = Command::new("cargo");
```
sehz, line 22: this should be using `check_cargo_exists`.

crates/smdk/src/build.rs
```diff
@@ -0,0 +1,53 @@
...
+            .arg("--lib");
+        cmd.arg("--target").arg(BUILD_TARGET);
+
+        let status = cmd.status()?;
+
+        if status.success() {
+            return Ok(());
+        }
+
+        Err(Error::msg(
```
sehz, line 38: This should just return error from if it is not ok otherwise error will be suppressed

crates/smdk/src/cmd.rs
```diff
@@ -9,13 +10,16 @@ use crate::test::TestOpt;
 pub enum SmdkCommand {
     /// Generate new SmartModule project
     Generate(GenerateOpt),
```
sehz, line 12: remove Generate since ti is in separate PR

crates/smdk/src/generate.rs
```diff
@@ -1,14 +1,52 @@
+use anyhow::{Error, Result};
```
sehz, line 1: remove from this PR

sehz (changes requested):

crates/smdk/src/build.rs
EstebanBorai: Agree. Im creating `anyhow::Error` instances from the actual error from now on so it's not wrapped.

sehz (approved):
LGTM

PR #2659, opened 2022-10-08
[Merged by Bors] - ci: Remove makefile env var check in CI

tjtelan:
Also make sure the correct Fluvio binary is used in CI for packaging

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -71,6 +71,8 @@ jobs:
         run: make download-fluvio-release
 
       - name: Publish artifacts
+        env:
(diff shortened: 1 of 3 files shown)


.github/workflows/publish.yml
```diff
@@ -103,6 +105,8 @@ jobs:
           make curl-install-fluvio
 
       - name: Bump latest version of Fluvio CLI with fluvio-packages
+        env:
+          FLUVIO_BIN: ~/.fluvio/bin/fluvio
```
sehz, line 109: - name: Set up Fluvio binaries
        run: |
          chmod +x ~/bin/fluvio 
          echo "~/bin" >> $GITHUB_PATH

PR #2660, opened 2022-10-08
[Merged by Bors] - ci: Correct the command for package bump

tjtelan:
Remove unnecessary cli install step


tjtelan:
Package bump typo was hard to spot, but if this passes the package bump step, then CD_Dev should also be fixed.

bors r+

PR #2669, opened 2022-10-10
[Merged by Bors] - fix(smart module): fixed failure if sm chain is empty

galibey:
Fixed the issue with the "Unexpected raw input" error message in case the chain is empty.

Also, removed redundant repacking from/to step input and the last step input to output.


sehz (approved):
LGTM.   Love reduction in code and simplification

PR #2687, opened 2022-10-12
feat: support to generate action tokens with other environments

EstebanBorai:
Introduces the `get_action_auth_with_token` method for `HubAccess` to allow tokens in other applications of the ecosystem to generate action tokens similar to how the Hub CLI does.

### DIFF
--- a/crates/fluvio-hub-util/src/hubaccess.rs
+++ b/crates/fluvio-hub-util/src/hubaccess.rs
@@ -92,7 +92,19 @@ impl HubAccess {
         self.get_action_auth(ACTION_PUBLISH).await
     }
 
+    pub async fn get_action_auth_with_token(
+        &self,
+        action: &str,
+        authn_token: &str,
+    ) -> Result<String> {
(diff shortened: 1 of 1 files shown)


crates/fluvio-hub-util/src/hubaccess.rs
```diff
@@ -92,15 +92,26 @@ impl HubAccess {
...
+        &self,
+        action: &str,
+        authn_token: &str,
+    ) -> Result<String> {
+        self.make_action_token(action, authn_token.into()).await
+    }
+
     async fn get_action_auth(&self, action: &str) -> Result<String> {
+        self.make_action_token(action, read_infinyon_token()?).await
+    }
```
EstebanBorai, line 105: To avoid breaking the API we pass the Infinyon token under the hood using the same call to `make_action_token`

sehz:
seems to have instability with other third party dependencies

morenol:
bors retry

PR #2689, opened 2022-10-12
[Merged by Bors] - perf: short circuit wait for metadata

davidbeesley:
1. Added some performance fixes to producer test framework itself.
2. Performance improvements to `sync::lookup_and_wait`
3. Benchmarked results for producer 1 X 100,000 and 3 X 150,000.

Example bench output
```
Testing connection to Fluvio cluster in profile
starting test in child process
Creating the topic: producer-test
topic "producer-test" already exists

Starting Producer test
Producers: 3
# Records: 150000
Starting Producer #0
Starting Producer #1
Starting Producer #2
[producer-2] Produce P99: 6.783µs Peak Throughput: 155189 kB/s
[producer-1] Produce P99: 6.751µs Peak Throughput: 157286 kB/s
[producer-0] Produce P99: 6.783µs Peak Throughput: 157286 kB/s
test passed, signalling success to parent
Test passed
+----------+---------------+
| Pass?    | true          |
|----------+---------------|
| Duration | 23.135994663s |
+----------+---------------+
```

### profiling results
No change

```
# 1 X 100K
[producer-0] Produce P99: 18.175µs Peak Throughput: 65011 kB/s

#3 X 150
[producer-0] Produce P99: 19.199µs Peak Throughput: 62128 kB/s
[producer-1] Produce P99: 19.327µs Peak Throughput: 61341 kB/s
[producer-2] Produce P99: 19.199µs Peak Throughput: 61865 kB/s
```

Test harness improvements only
```
# 1 X 100K
[producer-0] Produce P99: 13.759µs Peak Throughput: 78118 kB/s

#3 X 150
[producer-2] Produce P99: 13.951µs Peak Throughput: 74973 kB/s
[producer-1] Produce P99: 14.015µs Peak Throughput: 76021 kB/s
[producer-0] Produce P99: 14.079µs Peak Throughput: 74448 kB/s
```

Short circuiting
```
# 1 X 100K
[producer-0] Produce P99: 7.103µs Peak Throughput: 147849 kB/s

#3 X 150
[producer-2] Produce P99: 6.783µs Peak Throughput: 155189 kB/s
[producer-1] Produce P99: 6.751µs Peak Throughput: 157286 kB/s
[producer-0] Produce P99: 6.783µs Peak Throughput: 157286 kB/s
```

### DIFF
--- a/crates/fluvio-stream-model/Cargo.toml
+++ b/crates/fluvio-stream-model/Cargo.toml
@@ -27,9 +27,15 @@ once_cell = "1.5"
 k8-types = { version = "0.6.0", optional = true }
 
 [dev-dependencies]
+
 tokio = { version = "1.3.0", features = ["macros"] }
 fluvio-future = { version = "0.4.0", features = ["fixture"] }
 fluvio-stream-model = { path = ".", features = ["fixture"] }
+async-std = { version = "1.8.0", default-features = false, features = [
+    "attributes",
+] }
 
 [package.metadata.cargo-udeps.ignore]
-development = ["fluvio-stream-model"] # installed with `fixture` feature in tests
+development = [
+    "fluvio-stream-model",
+] # installed with `fixture` feature in tests
--- a/crates/fluvio-stream-model/src/store/dual_store.rs
+++ b/crates/fluvio-stream-model/src/store/dual_store.rs
@@ -12,7 +12,6 @@ use async_rwlock::RwLockReadGuard;
 use async_rwlock::RwLockWriteGuard;
 
 use crate::core::{MetadataItem, Spec};
-use self::listener::AtLeastOneChangeListener;
 
 use super::MetadataStoreObject;
 use super::{DualEpochMap, DualEpochCounter, Epoch, EpochChanges};
@@ -174,11 +173,6 @@ where
     /// create new change listener
     pub fn change_listener(self: &Arc<Self>) -> ChangeListener<S, C> {
         ChangeListener::new(self.clone())
-    }
-
-    /// create a new listener to watch for at least one update to the local store
-    pub fn at_least_one_change_listener<'a>(self: &'a Arc<Self>) -> AtLeastOneChangeListener<'a> {
-        AtLeastOneChangeListener::new(self.event_publisher())
     }
 }
 
@@ -399,7 +393,7 @@ mod listener {
     use std::fmt;
     use std::sync::Arc;
 
-    use tracing::{trace, debug, instrument};
+    use tracing::{trace, debug, instrument, error};
 
     use crate::store::event::EventPublisher;
     use crate::store::{
@@ -408,43 +402,43 @@ mod listener {
 
     use super::{LocalStore, Spec, MetadataItem, MetadataChanges};
 
-    /// thin wrapper for an event publisher. blocks until the current change in the event publisher is > 0
-    /// useful for making sure metadata has been delivered before querying store
-    pub struct AtLeastOneChangeListener<'a> {
-        event_publisher: &'a EventPublisher,
-    }
+    // /// thin wrapper for an event publisher. blocks until the current change in the event publisher is > 0
+    // /// useful for making sure metadata has been delivered before querying store
+    // pub struct AtLeastOneChangeListener<'a> {
+    //     event_publisher: &'a EventPublisher,
+    // }
 
-    impl<'a> AtLeastOneChangeListener<'a> {
-        /// only constructed by the associated LocalStore that owns the EventPublisher
-        pub(crate) fn new(event_publisher: &'a EventPublisher) -> Self {
-            Self { event_publisher }
-        }
+    // impl<'a> AtLeastOneChangeListener<'a> {
+    //     /// only constructed by the associated LocalStore that owns the EventPublisher
+    //     pub(crate) fn new(event_publisher: &'a EventPublisher) -> Self {
+    //         Self { event_publisher }
+    //     }
 
-        /// check if there should be any changes
-        /// this should be done before event listener
-        /// to ensure no events are missed
-        #[inline]
-        pub fn has_change(&'a self) -> bool {
-            self.event_publisher.current_change() > 0
-        }
+    //     /// check if there should be any changes
+    //     /// this should be done before event listener
+    //     /// to ensure no events are missed
+    //     #[inline]
+    //     pub fn has_change(&'a self) -> bool {
+    //         self.event_publisher.current_change() > 0
+    //     }
 
-        /// returns when the event_publisher has a current_change greater than 0
-        pub async fn listen(&'a self) {
-            if self.has_change() {
-                return;
-            }
+    //     /// returns when the event_publisher has a current_change greater than 0
+    //     pub async fn listen(&'a self) {
+    //         if self.has_change() {
+    //             return;
+    //         }
 
-            let listener = self.event_publisher.listen();
+    //         let listener = self.event_publisher.listen();
 
-            if self.has_change() {
-                return;
-            }
+    //         if self.has_change() {
+    //             return;
+    //         }
 
-            trace!("waiting for publisher");
+    //         trace!("waiting for publisher");
 
-            listener.await;
-        }
-    }
+    //         listener.await;
+    //     }
+    // }
 
     /// listen for changes local store
     pub struct ChangeListener<S, C>
@@ -493,7 +487,7 @@ mod listener {
         /// this should be done before event listener
         /// to ensure no events are missed
         #[inline]
-        pub fn has_change(&mut self) -> bool {
+        pub fn has_change(&self) -> bool {
             self.event_publisher().current_change() > self.last_change
         }
 
@@ -517,7 +511,17 @@ mod listener {
             self.event_publisher().current_change()
         }
 
-        pub async fn listen(&mut self) {
+        /// returns once there is at least one change recorded by the the event_publisher
+        /// can only be called once so consumes the ChangeListener
+        pub async fn listen_for_at_least_one_change(mut self) {
+            if self.last_change != 0 {
+                error!("listen_for_at_least_one_change should only be called on a newly constructed ChangeListener");
+            }
+            self.last_change = 0;
+            self.listen().await
+        }
+
+        pub async fn listen(&self) {
             if self.has_change() {
                 trace!("before has change: {}", self.last_change());
                 return;
@@ -806,10 +810,10 @@ mod test_notify {
         //  assert_eq!(last_change.load(SeqCst), 4);
     }
     #[fluvio_future::test]
-    async fn test_at_least_one_change_listener_non_blocking() {
+    async fn test_change_listener_non_blocking() {
         let mut timer = sleep(Duration::from_millis(5));
         let store = Arc::new(DefaultTestStore::default());
-        let listener = store.at_least_one_change_listener();
+        let listener = store.change_listener();
 
         // no events, this should timeout
         select! {
@@ -825,7 +829,7 @@ mod test_notify {
     }
 
     #[fluvio_future::test]
-    async fn test_at_least_one_change_listener() {
+    async fn test_change_listener() {
         let topic_store = Arc::new(DefaultTestStore::default());
         let last_change = Arc::new(AtomicI64::new(0));
         let shutdown = SimpleEvent::shared();
@@ -893,7 +897,10 @@ mod test_notify {
         S: Spec,
         C: MetadataItem,
     {
-        store.at_least_one_change_listener().listen().await;
+        store
+            .change_listener()
+            .listen_for_at_least_one_change()
+            .await;
         // Make sure that we never return before we update the store
         assert!(has_been_updated.load(std::sync::atomic::Ordering::Relaxed));
     }
--- a/crates/fluvio-stream-model/src/store/dual_store.rs
+++ b/crates/fluvio-stream-model/src/store/dual_store.rs
@@ -84,7 +84,7 @@ where
 
     /// write guard, this is private, use sync API to make changes
     #[inline(always)]
-    pub async fn write<'a>(
+    async fn write<'a>(
         &'_ self,
     ) -> RwLockWriteGuard<'_, DualEpochMap<S::IndexKey, MetadataStoreObject<S, C>>> {
         self.store.write().await
--- a/crates/fluvio-stream-model/src/store/dual_store.rs
+++ b/crates/fluvio-stream-model/src/store/dual_store.rs
@@ -785,8 +785,8 @@ mod test_notify {
         }
     }
 
-    #[fluvio_future::test]
-    async fn test_wait_for_first_change_assumptions() {
+    #[test]
+    fn test_wait_for_first_change_assumptions() {
         let topic_store = Arc::new(DefaultTestStore::default());
 
         // wait_for_first_change() requires that ChangeListener is initialized with current_change = 0
--- a/crates/fluvio-stream-model/src/store/dual_store.rs
+++ b/crates/fluvio-stream-model/src/store/dual_store.rs
@@ -420,11 +420,29 @@ mod listener {
             Self { event_publisher }
         }
 
+        /// check if there should be any changes
+        /// this should be done before event listener
+        /// to ensure no events are missed
+        #[inline]
+        pub fn has_change(&'a self) -> bool {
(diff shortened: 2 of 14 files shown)


sehz:
@galibey and @tjtelan  Please provide feedback

crates/fluvio/src/sync/mod.rs
```diff
@@ -108,16 +108,26 @@ mod context {
...
             // wait for either changes from store or timeout
             select! {
 
                 _ = listener.listen() => {
+                    // since the listeners change count is always 0 and listener.listen() only returns when change count > 0,
+                    // calling for sync_changes() will always trigger an allocation.
+                    // Since we don't do anything with the changes and sync_changes() doesn't add new changes to the cache, it's wasteful to call
+
                     // this should be full sync
-                    let changes = listener.sync_changes().await;
```
galibey, line 119: `sync_changes` modifies the internal state here: https://github.com/infinyon/fluvio/blob/92cece6daf0c878cabf0210b3f78b00e739006e2/crates/fluvio-stream-model/src/store/dual_store.rs#L539. Are you sure it's completely safe to delete this call here?

crates/fluvio-test/src/tests/mod.rs
```diff
@@ -72,6 +72,11 @@ impl TestRecordBuilder {
         self
     }
 
+    pub fn with_empty_data(mut self, data_size: usize) -> Self {
```
galibey, line 75: It seems like that method is not used in the final version, so there is no sense in keeping it

crates/fluvio-test/src/tests/producer.rs
```diff
@@ -100,30 +104,30 @@ async fn producer_work(
...
-        let record = TestRecordBuilder::new()
-            .with_tag(format!("{}:{}", producer_id, record_n))
-            .with_random_data(test_case.option.record_size)
-            .build();
-        let record_json = serde_json::to_string(&record)
-            .expect("Convert record to json string failed")
-            .as_bytes()
-            .to_vec();
-        let record_size = record_json.len() as u64;
+        // This bottlenecks the system, not good for throughput test
```
galibey, line 114: if the code is not needed so, let's just remove it instead of commenting

crates/fluvio/src/sync/mod.rs
galibey: @sehz Could you pls take a look

davidbeesley: From what I can see, `self.set_last_change(...)` only modifies the `last_change` field on the calling `ChangeListener`. In this case, the `ChangeListener` is only is scope for the `lookup_and_wait` function call, and then gets dropped before the field is ever read.

> `sync_changes` modifies the internal state here:
> 
> https://github.com/infinyon/fluvio/blob/92cece6daf0c878cabf0210b3f78b00e739006e2/crates/fluvio-stream-model/src/store/dual_store.rs#L539
> . Are you sure it's completely safe to delete this call here?

davidbeesley: I think it would be necessary to call `sync_changes()`  if `listen()` was called in a loop wanting to be woken up with new changes, but here is only called once.

crates/fluvio-test/src/tests/producer.rs
davidbeesley: :+1:. It looked like the file was still underdevelopment and I wasn't sure if there was a vision for it in the broader context.

crates/fluvio-test/Cargo.toml
```diff
@@ -20,6 +20,8 @@ clap = { version = "3.1.8", features = [
 ], default-features = false }
 async-trait = "0.1.21"
 rand = "0.8"
+# Cryptographically secure rand is unnecessary and slow enough to affect for benchmark testing
+fastrand = "1.8.0"
```
sehz, line 24: interesting

crates/fluvio/src/sync/mod.rs
sehz: It's been a while since I looked at this part code.
Meanwhile, can you add tests to verify these changes are safe?

crates/fluvio/src/sync/mod.rs
```diff
@@ -108,16 +108,26 @@ mod context {
...
-                    trace!("{} received changes: {:#?}",S::LABEL,changes);
+                    // let changes = listener.sync_changes().await;
+                    // trace!("{} received changes: {:#?}",S::LABEL,changes);
```
sehz, line 130: remove unneeded code

sehz (changes requested):
Thanks for the contribution and nice work.
One point about testing

crates/fluvio/src/sync/mod.rs
davidbeesley: I added a refactor to make it explicit that calling `lookup_by_key` doesn't do anything that could mutate the LocalStore's state. It no longer requires a mutable reference for a `ChangeListener` nor access to LocalStore's internal state. I also tightened up the interfaces a bit for the LocalStore and EventPublisher. 

As far as testing, what I would normally recommend is to have a set of unit tests for all of the public methods on interfaces for the synchronization structs and then protect access to the internals. I didn't remove or touch any preexisting tests for `lookup_by_key` or for the synced state, so if those already exist then the change is safe. If they don't exist, tightening up the interfaces and writing comprehensive unit tests for them would be a bit of a refactor, is that a the direction you want me to take?

sehz: No, just enough unit test to cover changes you proposed here.

crates/fluvio-stream-model/src/store/dual_store.rs
```diff
@@ -403,6 +408,21 @@ mod listener {
 
     use super::{LocalStore, Spec, MetadataItem, MetadataChanges};
 
+    pub struct AtLeastOneChangeListener<'a> {
```
sehz, line 411: add Doc here to explain what this does?

crates/fluvio/src/sync/mod.rs
```diff
@@ -108,17 +108,19 @@ mod context {
             use tokio::select;
             use fluvio_future::timer::sleep;
 
+            // We can short circuit here if already present
+            let found = search(self.store().read().await);
+            if found.is_some() {
+                return Ok(found);
+            }
```
sehz, line 115: Good catch.   Let's tighten code here like:
```
if let Some(found) = search(self.store().read().await) {
     return found
}
```

Surprised clippy didn't catch that

crates/fluvio/src/sync/mod.rs
sehz: But this seems good changes.   Good work

sehz (changes requested):
A minor nit.

crates/fluvio-stream-model/src/store/dual_store.rs
```diff
@@ -282,7 +289,6 @@ where
         drop(write_guard);
 
         self.event_publisher.store_change(epoch);
-        self.event_publisher.notify();
```
sehz, line 285: reasonable for this removal?

crates/fluvio-stream-model/src/store/dual_store.rs
```diff
@@ -373,7 +379,6 @@ where
 
         debug!("notify epoch changed: {}", epoch);
         self.event_publisher.store_change(epoch);
-        self.event_publisher.notify();
```
sehz, line 376: same question

crates/fluvio/src/sync/mod.rs
davidbeesley: Clippy didn't catch because it tightens to
```
if let Some(found) = search(self.store().read().await) {
     return Ok(Some(found))
}
```
I think that's still better. I'll change it.

crates/fluvio-stream-model/src/store/dual_store.rs
davidbeesley: Every time store_change is called, notify() is also called.  I thought it would be simpler and safer to include automatically notify listeners whenever store_change is called instead of requiring users of the interface to remember to notify after storing_changes.

I also removed the increment function since it isn't used anywhere.

sehz: Make sense.  Nice change

crates/fluvio-stream-model/src/store/dual_store.rs
```diff
@@ -403,6 +408,21 @@ mod listener {
...
 
+    pub struct AtLeastOneChangeListener<'a> {
+        event_publisher: &'a EventPublisher,
+    }
+
+    impl<'a> AtLeastOneChangeListener<'a> {
+        pub fn new(event_publisher: &'a EventPublisher) -> Self {
+            Self { event_publisher }
+        }
+        pub async fn listen_for_at_least_one_change(&'a self) {
```
sehz, line 419: is there reason why this is async since there is no await here.

sehz (changes requested):
Most of changes make sense except concern about `AtLeastOneChangeListener`.  I do think it needs unit test

crates/fluvio-stream-model/src/store/dual_store.rs
davidbeesley: > Most of changes make sense except concern about `AtLeastOneChangeListener`. I do think it needs unit test

Agreed. Definitely needs a unit test.

davidbeesley: > is there reason why this is async since there is no await here.

Yes. async is just syntatic sugar for a fn that returns a future. In this case it returns an `EventListener` https://docs.rs/event-listener/2.5.3/event_listener/struct.EventListener.html, which impls `IntoFuture`

We don't want to await here as it would block the thread and disable the timeout, so we return the future.

sehz: Shouldn't  `listen` fn should be like this?
```
pub async fn listen(&'a self) {
      while self.event_publisher.current_change() <= 0 {
              self.event_publisher.listen().await;
      }
 }
```

Otherwise, I don't see how executor can poll since there is no return in while loop.  Maybe unit test can prove that.

davidbeesley: Yes, you are right. It loops, not returns.

davidbeesley: Needs to be fixed.

davidbeesley: Fixed + unit test

crates/fluvio/src/sync/mod.rs
sehz: Great.  Nice testing

sehz:
Can you update benchmark with latest changes?

crates/fluvio-stream-model/src/store/dual_store.rs
davidbeesley: At this point, AtLeastOneChangeListener is just a copy paste of ChangeListener. I think it was better with just one. We can still keep the unit tests.

sehz: Let's do that.

davidbeesley:
> Can you update benchmark with latest changes?

The benchmark didn't change at all because for all iterations except the first it short circuits and that part hasn't changed. I confirmed this by rerunning.

crates/fluvio-stream-model/src/store/dual_store.rs
```diff
@@ -474,7 +473,17 @@ mod listener {
             self.event_publisher().current_change()
         }
 
-        pub async fn listen(&mut self) {
+        /// returns once there is at least one change recorded by the the event_publisher
+        /// can only be called once so consumes the ChangeListener
+        pub async fn listen_for_at_least_one_change(mut self) {
+            if self.last_change != 0 {
+                error!("listen_for_at_least_one_change should only be called on a newly constructed ChangeListener");
```
sehz, line 480: Can you return error instead of logging?

crates/fluvio-stream-model/Cargo.toml
```diff
@@ -27,9 +27,15 @@ once_cell = "1.5"
 k8-types = { version = "0.6.0", optional = true }
 
 [dev-dependencies]
+
 tokio = { version = "1.3.0", features = ["macros"] }
 fluvio-future = { version = "0.4.0", features = ["fixture"] }
 fluvio-stream-model = { path = ".", features = ["fixture"] }
+async-std = { version = "1.8.0", default-features = false, features = [
```
sehz, line 34: Do we still ned this?

sehz (changes requested):
I think we are almost there.  Couple of minor points

crates/fluvio-stream-model/Cargo.toml
davidbeesley: I think so, as one of the tests is using the `async_std::task::JoinHandle` returned by `fluvio_future::spawn`.


 `fluvio_future::spawn` is a reexport of `async_std..::spawn` or `...::spawn_local` depending on arch. `fluvio_future` doesn't also export the `JoinHandle` however, hence the need to import directly.

crates/fluvio-stream-model/src/store/dual_store.rs
davidbeesley: Actually, we can enforce in the interface.

crates/fluvio-stream-model/Cargo.toml
sehz: got it.

sehz (approved):
LGTM.  Thanks for contribution and nice work!

davidbeesley:
> LGTM. Thanks for contribution and nice work!

Thanks. There is still one more loose end I need to tie up before this is ready to ship.
The consumer test used to use the `production-test` topic and I switched it to use `consumer-test`. However, it doesn't populate the topic with any records before running so right now the smoke-test is essentially a no-op. The test should also fail if it didn't get as many records at it expected.

galibey (approved):
LGTM. Good job!

crates/fluvio-test/src/tests/consumer.rs
```diff
@@ -233,8 +233,7 @@ async fn get_multi_stream(
     )
 }
 
-// Default to using the producer test's topic
-#[fluvio_test(name = "consumer", topic = "producer-test")]
+#[fluvio_test(name = "consumer", topic = "consumer-test")]
```
davidbeesley, line 236: This still needs to be addressed for shipping.

crates/fluvio-test/src/tests/producer.rs
```diff
@@ -90,6 +89,11 @@ async fn producer_work(
     let mut throughput_histogram = Histogram::<u64>::new(2).unwrap();
     // Eventually we'll need to store globally to calculate wrt all producers in test
 
+    // We don't do anything with the record struct so we increase the stress on the system by just sending the same data over and over again.
+    // We can only do this because we don't compress the data
+    // Otherwise, test-harness spends much of the time allocating the record
+    let record = "A".repeat(test_case.option.record_size).as_bytes().to_vec();
```
tjtelan, line 95: Workflow-wise, this is fine, but makes some small assumptions that affect the overall utility of the test. This test doesn't get used often, isn't in any automation, and the changes are net-positive. So I wouldn't do anything different at the moment.

This is more neutral feedback, I wouldn't hold up PR for this point because it is a bit out of the scope of your changes:

Enabling the test to be more configurable instead of removing functionality (even if this simple record form was made the default option) would have been my preference for record generation modes, instead of replacing one for the other. 

There are many producer options that have been introduced in the client that haven't been introduced into this test (like compression, and linger). And the ability to vary the shape of test data would further our ability to characterize performance.

crates/fluvio-test/src/tests/consumer.rs
davidbeesley: I pushed a fix for the consumer test, but I think there is a might be a overall bug in the test-driver: https://github.com/infinyon/fluvio/issues/2695

tjtelan (approved):
Changes look good, and the results are even better

crates/fluvio-test/src/tests/producer.rs
davidbeesley: Agree 100%. Thanks for the feedback.

crates/fluvio-test/src/tests/consumer.rs
sehz: k

PR #2697, opened 2022-10-14
feat: fluvio-cli, fluvio hub list, add SMARTMODULES column header

digikata:


ajhunyady (commented):
From what I remember, there is a `table` formatting, that takes into account column widths, etc. It also add a line between the Header and the Rows for better separation.

PR #2710, opened 2022-10-17
[Merged by Bors] - feat: expose all public members from `hubaccess` module

EstebanBorai:
Makes HubAccess module's members accesible if their scope is public.


digikata (approved):
LGTM - I think we'll want to try a less exposed interface later, but that's a much lower priority than integrating these last few items.

PR #2716, opened 2022-10-18
[Merged by Bors] - More smdk metadata paths

superstator:
Makes `load` command more consistent when using the `-p` option


sehz (approved):
LGTM

PR #2717, opened 2022-10-18
[Merged by Bors] - feat: add instrumentation to client producer

nacardin:


crates/fluvio/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio"
-version = "0.14.0"
+version = "0.14.2"
```
sehz, line 3: No need to bump since `0.14.0` is not published yet.

crates/fluvio/src/lib.rs
```diff
@@ -207,6 +207,35 @@ pub async fn consumer<S: Into<String>>(
...
+            opentelemetry::sdk::metrics::selectors::simple::inexpensive(),
+            opentelemetry::sdk::export::metrics::aggregation::cumulative_temporality_selector(),
+            opentelemetry::runtime::AsyncStd,
+        )
+        .with_exporter(opentelemetry_otlp::new_exporter().tonic())
+        .with_period(Duration::from_secs(1))
+        .with_timeout(Duration::from_secs(10))
+        .with_resource(resource)
+        .build()
+        .unwrap();
```
sehz, line 234: need to get rid of unwrap

crates/fluvio/src/producer/partition_producer.rs
```diff
@@ -48,6 +57,13 @@ impl PartitionProducer {
         last_error: Arc<RwLock<Option<ProducerError>>>,
         #[cfg(feature = "stats")] client_stats: Arc<ClientStats>,
     ) -> Self {
+        #[cfg(feature = "otel-metrics")]
+        let meter = global::meter("producer");
+        #[cfg(feature = "otel-metrics")]
+        let records_counter = meter.u64_counter("fluvio.client.records").init();
```
sehz, line 63: this should be more like: `fluvio.producer.records`.

crates/fluvio/src/producer/partition_producer.rs
```diff
@@ -48,6 +57,13 @@ impl PartitionProducer {
...
+        #[cfg(feature = "otel-metrics")]
+        let bytes_counter = meter.u64_counter("fluvio.client.io").init();
```
sehz, line 65: `producer.io`

sehz (changes requested):
Looks reasonable.  Few minor nits

crates/fluvio/src/producer/partition_producer.rs
nacardin: I was looking at convention and examples. https://opentelemetry.netlify.app/docs/reference/specification/metrics/semantic_conventions/database-metrics/

Since the produce and consume will have the same counters. The convention seems to be to use an attribute to distinguish produce and consume. See line 275

Also see `system.network.packets` at https://opentelemetry.netlify.app/docs/reference/specification/metrics/semantic_conventions/system-metrics/. `transmit` and `receive` is the value of the `direction` attribute.

sehz: See http metrics which seems to jive more with our semantics:
https://opentelemetry.netlify.app/docs/reference/specification/metrics/semantic_conventions/http-metrics/
As well as Kafka:
https://opentelemetry.io/docs/reference/specification/metrics/semantic_conventions/instrumentation/kafka/

sehz: so following that it should be something like `messaging.fluvio.producer.outgoing`. etc

crates/fluvio/Cargo.toml
```diff
@@ -50,6 +51,8 @@ quantities = { version = "0.10.0", features = [
 hdrhistogram = { version = "7.5.0", optional = true }
 strum = { version = "0.24", features = ["derive"], optional = true }
 num-traits = { version = "0.2", optional = true }
+opentelemetry = { version = "0.18.0", optional = true, features = ["metrics", "rt-async-std"] }
```
galibey, line 54: Maybe we should introduce a separate crate (e.g. `fluvio-opentelemetry`) where we can hide these 3rd party deps and also some common methods. We will have many crates having such dependencies, with one crate it would be easier to control all deps and features.

crates/fluvio/src/lib.rs
```diff
@@ -207,6 +207,35 @@ pub async fn consumer<S: Into<String>>(
     Ok(consumer)
 }
 
+#[cfg(feature = "otel-metrics")]
```
galibey, line 210: I think there shouldn't any pipeline initialization logic in `fluvio` crate because it is a library. Such methods are end-app level responsibility (user's code or the connector is our case).

crates/fluvio/src/lib.rs
```diff
@@ -207,6 +207,35 @@ pub async fn consumer<S: Into<String>>(
...
+pub fn init_open_telemetry() {
+    use std::time::Duration;
+    use opentelemetry::KeyValue;
+    use opentelemetry::sdk::Resource;
+    use opentelemetry::global;
+
+    let resource_attributes = [
+        KeyValue::new("service.name", "client"),
+        KeyValue::new("service.namespace", "fluvio"),
+    ];
```
tjtelan, line 220: This might need an input parameter to add some way to provide unique identification

crates/fluvio/Cargo.toml
nacardin: fixed

crates/fluvio/src/lib.rs
nacardin: removed

crates/fluvio/src/producer/partition_producer.rs
nacardin: updated

crates/fluvio/Cargo.toml
nacardin: Agreed to look into this improvement in the future

crates/fluvio/src/lib.rs
nacardin: removed

crates/fluvio/src/producer/partition_producer.rs
nacardin: updated

crates/fluvio/Cargo.toml
```diff
@@ -50,6 +51,7 @@ quantities = { version = "0.10.0", features = [
 hdrhistogram = { version = "7.5.0", optional = true }
 strum = { version = "0.24", features = ["derive"], optional = true }
 num-traits = { version = "0.2", optional = true }
+opentelemetry = { version = "0.18.0", optional = true, features = ["rt-async-std"] }
```
sehz, line 54: Do we still need `rt-async-std` ?

sehz (changes requested):
Looks good.  One minor issue

crates/fluvio/Cargo.toml
nacardin: no, removed

sehz (approved):
LGTM

PR #2719, opened 2022-10-19
[Merged by Bors] - feat: `PackageMeta` fully qualified ID support

EstebanBorai:
Provides support to create the fully qualified ID from a Hub's `PackageMeta`
instance.

### DIFF
--- a/crates/fluvio-hub-util/src/packagemeta.rs
+++ b/crates/fluvio-hub-util/src/packagemeta.rs
@@ -3,6 +3,8 @@ use std::io::Read;
 use std::path::Path;
 use std::fs;
 
+use fluvio_controlplane_metadata::smartmodule::FluvioSemVersion;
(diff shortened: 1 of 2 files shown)


crates/fluvio-hub-util/src/packagemeta.rs
```diff
@@ -46,6 +46,13 @@ impl Default for PackageMeta {
 }
 
 impl PackageMeta {
+    /// Retrieves the fully qualified name for the `PackageMeta`.
+    ///
+    /// Eg: `infinyon-example-0.0.1`
+    pub fn id(&self) -> String {
+        format!("{}-{}-{}", self.group, self.name, self.version)
```
morenol, line 53: where is this used?
I see that we currently have `packagename_transform`?

EstebanBorai: Updating approach!

morenol (approved):

PR #2726, opened 2022-10-20
[Merged by Bors] - feat(metrics): added metrics to smartengine

galibey:
So far, due to current implementation details, we can't send the following metrics:
1. Input records count for chain and smartmodules.
2. Output bytes count for chain and smartmodules.
3. Smartmodule name for all metrics.

We will be able to add these metrics later, but it will require changes in the current implementation.

### DIFF
--- a/crates/fluvio-smartengine/src/metrics.rs
+++ b/crates/fluvio-smartengine/src/metrics.rs
@@ -1,0 +1,131 @@
+#[cfg(feature = "otel-metrics")]
+pub(crate) struct ChainMetrics {
+    context: opentelemetry::Context,
+    invocations: opentelemetry::metrics::Counter<u64>,
+    records: opentelemetry::metrics::Counter<u64>,
+    bytes: opentelemetry::metrics::Counter<u64>,
+}
+
+#[cfg(feature = "otel-metrics")]
+impl ChainMetrics {
+    pub(crate) fn new() -> Self {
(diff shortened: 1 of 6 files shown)


sehz (approved):
LGTM.  Nice start

crates/fluvio-smartengine/src/metrics.rs
```diff
@@ -0,0 +1,131 @@
...
+#[cfg(feature = "otel-metrics")]
+impl ChainMetrics {
+    pub(crate) fn new() -> Self {
+        let meter = opentelemetry::global::meter("smartengine");
+        let context = opentelemetry::Context::new();
+        let invocations = meter
+            .u64_counter("fluvio.smartengine.chain.invocations")
+            .init();
+        let records = meter.u64_counter("fluvio.smartengine.chain.records").init();
+        let bytes = meter.u64_counter("fluvio.smartengine.chain.bytes").init();
```
nacardin, line 18: In the conventions, I think they use `io` instead of `bytes` because "bytes" is a unit used in the `unit` field.

nacardin (approved):
If it works, lets ship it!

crates/fluvio-smartengine/src/metrics.rs
galibey: ok. We will change it later. It won't be used in first version on UI, anyway

PR #2731, opened 2022-10-20
[Merged by Bors] - feat: add record counters to spu

nacardin:


crates/fluvio-spu/Cargo.toml
```diff
@@ -40,7 +40,9 @@ thiserror = "1"
 once_cell = "1.5"
 sysinfo = "0.26.0"
 nix = { version = "0.25"}
-
+async-global-executor = { version = "2.3.0", features = ["tokio"] }
```
sehz, line 43: Why do we need to use tokio?

crates/fluvio-spu/src/services/public/metrics.rs
```diff
@@ -0,0 +1,96 @@
...
+    partition: i32,
+}
+
+impl<'c> StorageMetricsTopicPartition<'c> {
+    pub(crate) fn add_records_read(&self, value: u64) {
+        self.topic_metrics.metrics.records.add(
+            &self.topic_metrics.metrics.context,
+            value,
+            &[
+                opentelemetry::KeyValue::new("direction", "read"),
```
sehz, line 54: Can some of these be made into const? This is not show stopper.

crates/fluvio-spu/src/services/public/produce_handler.rs
```diff
@@ -51,12 +53,14 @@ pub async fn handle_produce_request(
...
 ) -> Result<ResponseMessage<ProduceResponse>, Error> {
+    let metrics = super::metrics::StorageMetrics::new();
+
     let (header, produce_request) = request.get_header_request();
     trace!("Handling ProduceRequest: {:#?}", produce_request);
 
     let mut topic_results = Vec::with_capacity(produce_request.topics.len());
     for topic_request in produce_request.topics.into_iter() {
-        let topic_result = handle_produce_topic(&ctx, topic_request).await;
+        let topic_result = handle_produce_topic(&ctx, topic_request, &metrics).await;
```
sehz, line 63: Instead of passing as argument, we can add to context.  not show stopper

sehz (changes requested):
Overall sounds fine.  couple of minor point.
But main concern I have is use `tokio `

galibey (approved):
LGTM

crates/fluvio-spu/src/services/public/produce_handler.rs
nacardin: done

crates/fluvio-spu/src/services/public/metrics.rs
nacardin: Not sure if that would be good in this case, lets defer discussion to later

crates/fluvio-spu/Cargo.toml
nacardin: removed exporting

sehz (approved):
LGTM

Issue #2700, opened 2022-10-14
performance: Improve encoding/decoding of Vec<u8> for WASM


sehz:
SmartModuleWasm encoding/decoding is inefficient because by default it uses Vec<M>.
We should use newtype to improve performance.

PR #2738, opened 2022-10-22
[Merged by Bors] - feat: introduce `ByteBuf` for `SmartModuleSpec`

EstebanBorai:
In order to have a more performant implementation on WASM bytes, a
dedicated type is introduced which provides an specific implementation
for bytes representing a WASM file owned by a `SmartModule` instance.

Resolves: https://github.com/infinyon/fluvio/issues/2700


crates/fluvio-controlplane-metadata/src/smartmodule/spec.rs
```diff
@@ -104,6 +106,40 @@ pub struct SmartModuleWasmSummary {
...
+        T: BufMut,
+    {
+        if dest.remaining_mut() < 1 {
+            return Err(IoError::new(
+                ErrorKind::UnexpectedEof,
+                "Not enough capacity for WasmBytes",
+            ));
+        }
+
+        dest.put_slice(self.0.as_slice());
```
EstebanBorai, line 133: Instead of walking through byte to byte. If the `Encoder` is running on a `WasmBytes` instance, the underlying vector is copied as a slice.

crates/fluvio-controlplane-metadata/src/smartmodule/spec.rs
```diff
@@ -104,6 +106,40 @@ pub struct SmartModuleWasmSummary {
     pub wasm_length: u32,
 }
 
+pub struct WasmBytes(Vec<u8>);
```
EstebanBorai, line 109: Provide a new type to differ from the implementation of `Encoder` for `Vec<T>`.

sehz: Since this is so useful let's move to into `fluvio-protocol-core`.

crates/fluvio-controlplane-metadata/src/smartmodule/spec.rs
```diff
@@ -104,6 +106,40 @@ pub struct SmartModuleWasmSummary {
...
+
+impl From<Vec<u8>> for WasmBytes {
+    fn from(bytes: Vec<u8>) -> Self {
+        Self(bytes)
+    }
+}
+
+impl Encoder for WasmBytes {
+    fn write_size(&self, _version: Version) -> usize {
+        1
```
sehz, line 119: This is wrong.  It should be 4 bytes for length and contents

crates/fluvio-controlplane-metadata/src/smartmodule/spec.rs
```diff
@@ -104,6 +106,40 @@ pub struct SmartModuleWasmSummary {
...
+    }
+
+    fn encode<T>(&self, dest: &mut T, _version: Version) -> Result<(), IoError>
+    where
+        T: BufMut,
+    {
+        if dest.remaining_mut() < 1 {
```
sehz, line 126: use write_size instead of hard coding as 1

crates/fluvio-controlplane-metadata/src/smartmodule/spec.rs
```diff
@@ -104,6 +106,40 @@ pub struct SmartModuleWasmSummary {
...
+            return Err(IoError::new(
+                ErrorKind::UnexpectedEof,
+                "Not enough capacity for WasmBytes",
```
sehz, line 129: include expected size in the error message

crates/fluvio-controlplane-metadata/src/smartmodule/spec.rs
sehz, line 133: need to encode length. See Vec<M> implementation

crates/fluvio-controlplane-metadata/src/smartmodule/spec.rs
```diff
@@ -174,3 +210,26 @@ mod base64 {
...
+    use crate::smartmodule::WasmBytes;
+
+    #[test]
+    fn test_encode_wasmbytes() {
+        let mut dest = Vec::default();
+        let value: WasmBytes = WasmBytes::from(vec![12, 128, 255, 78, 9]);
+        let result = value.encode(&mut dest, 0);
+
+        assert!(result.is_ok());
+        assert_eq!(dest.len(), 5);
```
sehz, line 227: This should be 5 (contents) + 4 (len) = 9

crates/fluvio-controlplane-metadata/src/smartmodule/spec.rs
```diff
@@ -174,3 +210,26 @@ mod base64 {
...
+        assert_eq!(dest[0], 0x0C);
+        assert_eq!(dest[1], 0x80);
+        assert_eq!(dest[2], 0xFF);
+        assert_eq!(dest[3], 0x4E);
+        assert_eq!(dest[4], 0x09);
+        assert_eq!(value.write_size(0), 1);
```
sehz, line 233: should be 9

sehz (changes requested):

crates/fluvio-protocol/src/core/mod.rs
```diff
@@ -1,12 +1,15 @@
...
 mod encoder;
 mod varint;
+mod wasm_bytes;
 mod zerocopy;
 
 pub use self::decoder::Decoder;
 pub use self::decoder::DecoderVarInt;
 pub use self::encoder::Encoder;
 pub use self::encoder::EncoderVarInt;
+pub use self::wasm_bytes::WasmBytes;
```
EstebanBorai, line 11: WasmBytes is now exported from Fluvio Protocol

crates/fluvio-protocol/src/core/wasm_bytes.rs
```diff
@@ -0,0 +1,87 @@
...
+                ErrorKind::UnexpectedEof,
+                format!(
+                    "Not enough capacity for WasmBytes. Expected: {}, Remaining: {}",
+                    expected, remaining
+                ),
+            ));
+        }
+
+        dest.put_u32(self.0.len() as u32);
+        dest.put_slice(self.0.as_slice());
```
EstebanBorai, line 45: Given that we acknowledge the contents of the underlying `Vec`, first puts the length of the `Vec` and then the whole contents as a slice instead of encoding -> pushing to dest.

crates/fluvio-protocol/src/core/wasm_bytes.rs
```diff
@@ -0,0 +1,87 @@
...
+
+use bytes::{Bytes, BufMut};
+
+use crate::{Encoder, Version};
+
+/// Represnts a SmartModule WASM File bytes.
+///
+/// Provides a `Encoder` implementation optimized for WASM files used in
+/// SmartModules.
+pub struct WasmBytes(Vec<u8>);
```
sehz, line 12: Since this can be used anywhere, can we come up with more generic name?  `ByteVec`, `ByteBuf`, etc.

crates/fluvio-protocol/src/core/wasm_bytes.rs
```diff
@@ -0,0 +1,87 @@
...
+    fn from(bytes: Vec<u8>) -> Self {
+        Self(bytes)
+    }
+}
+
+impl Encoder for WasmBytes {
+    fn write_size(&self, version: Version) -> usize {
+        self.0
+            .iter()
+            .fold(4, |sum, val| sum + val.write_size(version))
```
sehz, line 24: this is still inefficient: `val.write_size` since it will recursively calculate.  But since we it know it's byte, we can calculate length

sehz (changes requested):
Much better.  Couple of issues remain

crates/fluvio-protocol/src/core/wasm_bytes.rs
EstebanBorai: `ByteBuf` sounds good to me!

crates/fluvio-protocol/src/core/bytebuf.rs
```diff
@@ -0,0 +1,158 @@
...
+        T: Buf,
+    {
+        let mut len: i64 = 0;
+        len.decode_varint(src)?;
+
+        if len < 1 {
+            return Ok(());
+        }
+
+        self.inner.extend_from_slice(&src.chunk());
```
EstebanBorai, line 48: `chunk` is working for small vectors, but its actually intended for internal usage.
I'm still investigating on a better approach than walking and executing `get_u8` per byte.

EstebanBorai: Replaced with:

```rust
self.inner
            .extend_from_slice(&src.copy_to_bytes(len as usize));
```

Which will consume the amount of bytes from `len`.

EstebanBorai:
Implemented tests to make sure the output is the same as for encoding/decoding from `Vec<u8>`.

crates/fluvio-protocol/src/core/bytebuf.rs
```diff
@@ -0,0 +1,301 @@
...
+            "encoded version doesn't match with expected"
+        );
+        assert_eq!(
+            bytebuf_encoded.inner[3], 0x0A,
+            "the length value doesnt match"
+        );
+    }
+
+    #[test]
+    fn test_bytebuf_decode_transparent_with_vecu8() {
```
EstebanBorai, line 261: This will make sure the output is the same as for `Vec<u8>`. Which will ensure the output is compliant with what we use today. Also applies for the test above on: `test_bytebuf_encodes_transparent_with_vecu8`

crates/fluvio-protocol/src/core/bytebuf.rs
```diff
@@ -0,0 +1,301 @@
...
+
+        Ok(())
+    }
+}
+
+impl Encoder for ByteBuf {
+    fn write_size(&self, version: Version) -> usize {
+        self.inner
+            .iter()
+            .fold(4, |sum, val| sum + val.write_size(version))
```
morenol, line 62: Since this is a vec of `u8` I think that we could do size of `i32.write_size(version)` + self.inner.len()

crates/fluvio-protocol/src/core/bytebuf.rs
```diff
@@ -0,0 +1,301 @@
...
+        ByteBuf { inner: bytes }
+    }
+}
+
+impl Decoder for ByteBuf {
+    fn decode<T>(&mut self, src: &mut T, version: Version) -> Result<(), Error>
+    where
+        T: Buf,
+    {
+        let mut len: i32 = 0;
```
morenol, line 43: using i32 here and u32 in encoder.

crates/fluvio-protocol/src/core/bytebuf.rs
```diff
@@ -0,0 +1,301 @@
...
+///
+/// Provides a `Encoder` and `Decoder` implementation optimized for WASM files
+/// used in SmartModules.
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub struct ByteBuf {
+    inner: Vec<u8>,
+}
+
+impl ByteBuf {
+    pub fn is_empty(&self) -> bool {
```
sehz, line 19: You can use Deref which let you get rid of wrapper methods

crates/fluvio-protocol/src/core/bytebuf.rs
```diff
@@ -0,0 +1,301 @@
...
+        self.inner.is_empty()
+    }
+
+    pub fn len(&self) -> usize {
+        self.inner.len()
+    }
+
+    pub fn inner(&self) -> Vec<u8> {
```
sehz, line 27: it's better to have implement `To` trait

crates/fluvio-protocol/src/core/bytebuf.rs
EstebanBorai: Gotcha! Thanks!

EstebanBorai: Resolved on: 9644e0f by sticking to `u32`

crates/fluvio-protocol/src/core/bytebuf.rs
```diff
@@ -0,0 +1,301 @@
...
+
+        Ok(())
+    }
+
+    fn as_bytes(&self, _version: Version) -> Result<Bytes, Error> {
+        Ok(Bytes::copy_from_slice(self.inner.as_slice()))
+    }
+}
+
+unsafe impl BufMut for ByteBuf {
```
sehz, line 91: Is there way to do without unsafe implementation?

sehz (changes requested):
couple of minor issues.  Critical issue is whether this can be done without unsafe implementation

crates/fluvio-protocol/src/core/bytebuf.rs
EstebanBorai: I agree, it's always better to avoid `unsafe`. I searched for alternatives but theres no derive macro that could apply this for us.

This is required in order to call decode on `ByteBuf`, and given that `BufMut` is already implemented for `Vec<u8>` by the crate authors, I took advantage of such implementation by just calling the respective trait functuons for the `Vec<u8>` in the `inner` prop.

sehz: Can you try Deref?

EstebanBorai: Of course!

crates/fluvio-protocol/src/core/bytebuf.rs
EstebanBorai: Im didn't found any `To` trait, do you have reference to such trait?

crates/fluvio-protocol/src/core/bytebuf.rs
EstebanBorai: Addressed on: https://github.com/infinyon/fluvio/pull/2738/commits/6efaa0b83cbfb1f2901dff646b27a688e2422cbd

crates/fluvio-protocol/src/core/bytebuf.rs
EstebanBorai: Like a charm!

crates/fluvio-protocol/src/core/bytebuf.rs
EstebanBorai: I did try by removing the implementation of `ByteBuf`, but it complains about missing impl even when `Deref` is implemented already.

crates/fluvio-protocol/src/core/bytebuf.rs
EstebanBorai: Addressed on: [e5e660a](https://github.com/infinyon/fluvio/pull/2738/commits/e5e660a2c5875836869d14ee01a91555105ad57e)

crates/fluvio-protocol/src/core/bytebuf.rs
```diff
@@ -0,0 +1,300 @@
...
+
+use crate::{Encoder, Decoder, Version};
+
+/// Represnts a SmartModule WASM File bytes.
+///
+/// Provides a `Encoder` and `Decoder` implementation optimized for WASM files
+/// used in SmartModules.
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub struct ByteBuf {
+    inner: Vec<u8>,
```
sehz, line 16: Since we may need to re-implement much of Bytes, we might just use https://docs.rs/bytes/1.2.1/bytes/struct.Bytes.html instead of `Vec<u8>`.  This also benefits cheap cloning instead of allocating a separate buffer.   This hopefully get rid of the unsafe implementation.

sehz (changes requested):

crates/fluvio-protocol/src/core/bytebuf.rs
EstebanBorai: As I can see in the docs for `BufMut` and `Bytes`, there's no implementation of `BufMut` for `Bytes`, we would have to implement it if we want to use `Bytes` as our inner type. https://docs.rs/bytes/latest/bytes/trait.BufMut.html#foreign-impls.

I have some work on `encoding`, `decoding` implementation using `Bytes` as our inner type, but as I can see we would still have to implement `BufMut` ourselves instead of taking advantage of the implementation provided by crate authors that we have with `Vec<u8>`.

morenol: What about `BytesMut`?

sehz:
rebase.

EstebanBorai:
> rebase.

Rebasing and updating benches

sehz:
that's 50x improvement.  Can you benchmark with a bytes buffer with 100k, which is typical for SmartModule?

morenol:
> @sehz @morenol new implementation is looking very good!
> 
> ```
> test vecu8 encoding ... bench:      414738 ns/iter (+/- 29440)
> 
> test vecu8 decoding ... bench:      355882 ns/iter (+/- 11905)
> 
> test bytebuf encoding ... bench:        7909 ns/iter (+/- 235)
> 
> test bytebuf decoding ... bench:        7684 ns/iter (+/- 293)
> ```
> 
> These are the results benchmarking with Criterion as of [b8f77a5](https://github.com/infinyon/fluvio/pull/2738/commits/b8f77a585b4ac4b72ab3523883ffbe6dedc2da59)

Very promising, I think that we should use ByteBuf in `SmartModuleInput` and maybe other places of the code where we use Vec<u8>

EstebanBorai:
> > @sehz @morenol new implementation is looking very good!
> > ```
> > test vecu8 encoding ... bench:      414738 ns/iter (+/- 29440)
> > 
> > test vecu8 decoding ... bench:      355882 ns/iter (+/- 11905)
> > 
> > test bytebuf encoding ... bench:        7909 ns/iter (+/- 235)
> > 
> > test bytebuf decoding ... bench:        7684 ns/iter (+/- 293)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > These are the results benchmarking with Criterion as of [b8f77a5](https://github.com/infinyon/fluvio/pull/2738/commits/b8f77a585b4ac4b72ab3523883ffbe6dedc2da59)
> 
> Very promising, I think that we should use ByteBuf in `SmartModuleInput` and maybe other places of the code where we use Vec

Totally agree!

morenol:
I think that we could also grab `Encoder` implementation from this PR and add that to `RecordData` https://github.com/infinyon/fluvio/blob/3d5f0ec965eebc07cc656312ec6ddb1b6ccb2559/crates/fluvio-protocol/src/record/data.rs#L169

EstebanBorai:
> that's 50x improvement. Can you benchmark with a bytes buffer with 100k, which is typical for SmartModule?

Im using a SmartModule WASM file for testing. Here is the file in question: https://github.com/EstebanBorai/fluvio/blob/feat/wasmbytes-for-smartmodulespec/crates/fluvio-protocol/fixtures/smartmodule.wasm.

The source for this file is actually a filter SmartModule example. Perhaps we could try with the RegExp example for a more "real-life" example?

sehz:
That's good enough!

crates/fluvio-controlplane-metadata/Cargo.toml
```diff
@@ -36,6 +36,7 @@ fluvio-types = { version = "0.3.1", path = "../fluvio-types" }
 fluvio-stream-model = { path = "../fluvio-stream-model", version = "0.8.0" }
 fluvio-protocol = { path = "../fluvio-protocol", version = "0.8", features = [
     "record",
+    "use_serde",
```
sehz, line 39: we still want to make `serde` optional

crates/fluvio-protocol/src/core/bytebuf.rs
```diff
@@ -0,0 +1,262 @@
...
+
+/// Represnts a SmartModule WASM File bytes.
+///
+/// Provides a `Encoder` and `Decoder` implementation optimized for WASM files
+/// used in SmartModules.
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub struct ByteBuf {
+    inner: Bytes,
+}
+
```
sehz, line 17: should have API to create `ByteBuf` from `Bytes`

crates/fluvio-protocol/src/core/bytebuf.rs
```diff
@@ -0,0 +1,262 @@
...
+        Ok(())
+    }
+
+    fn as_bytes(&self, _version: Version) -> Result<Bytes, Error> {
+        Ok(self.inner.clone())
+    }
+}
+
+#[cfg(feature = "use_serde")]
```
sehz, line 91: I don't think ths belongs to this create.  This is opinonated encoding which can be moved back to controlplane protocol

sehz (changes requested):
Great progress and very nice work.  Just need couple of refinements

crates/fluvio-protocol/src/core/bytebuf.rs
EstebanBorai: Addressed on: [9d73483](https://github.com/infinyon/fluvio/pull/2738/commits/9d734838757f4ce0408a66bfd9b9a145565b6cc8)

crates/fluvio-controlplane-metadata/Cargo.toml
EstebanBorai: Yeah with [630aa06](https://github.com/infinyon/fluvio/pull/2738/commits/630aa06d16fae114d94efae22924cd9060650394) this is back to the current state.

crates/fluvio-protocol/src/core/bytebuf.rs
EstebanBorai: Addressed on: [630aa06](https://github.com/infinyon/fluvio/pull/2738/commits/630aa06d16fae114d94efae22924cd9060650394)

crates/fluvio-protocol/src/core/bytebuf.rs
```diff
@@ -0,0 +1,236 @@
...
+        // Data encoded for Vec<u8> will have the first 4 bytes (32 bits) for
+        // data length, and the remaining bytes represents the actual data
+        let encoded_expect: [u8; 14] = [
+            0x00, 0x00, 0x00, 0x0A, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0A,
+        ];
+
+        let mut encoded_data: Vec<u8> = Vec::default();
+        let encoded_res = Vec::<u8>::from(raw_data).encode(&mut encoded_data, 0);
+
+        assert!(encoded_res.is_ok());
```
morenol, line 171: I think that this assert is not relevant for this test

crates/fluvio-protocol/src/core/bytebuf.rs
```diff
@@ -0,0 +1,236 @@
...
+        assert_eq!(value.write_size(0), dest.len());
+    }
+
+    #[test]
+    fn test_bytebuf_encodes_transparent_with_vecu8() {
+        let raw_data: [u8; 10] = [0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0A];
+
+        // Data encoded for Vec<u8> will have the first 4 bytes (32 bits) for
+        // data length, and the remaining bytes represents the actual data
+        let encoded_expect: [u8; 14] = [
```
morenol, line 164: maybe instead of explicitly defining encoded_expect, we could use: `Vec::<u8>::from(raw_data).encode(&mut encoded_data, 0).expect("vec encoding failed");`

crates/fluvio-protocol/src/core/bytebuf.rs
```diff
@@ -0,0 +1,236 @@
...
+        ];
+
+        let mut encoded_data: Vec<u8> = Vec::default();
+        let is_encode_vecu8_ok = Vec::<u8>::from(raw_data).encode(&mut encoded_data, 0);
+
+        let mut bytebuf_encoded = Vec::default();
+        let is_encode_bytebuf_ok =
+            ByteBuf::from(Vec::from(raw_data)).encode(&mut bytebuf_encoded, 0);
+
+        assert!(is_encode_vecu8_ok.is_ok());
```
morenol, line 210: I think that this assert also is not needed here

crates/fluvio-controlplane-metadata/src/smartmodule/spec.rs
```diff
@@ -109,7 +110,7 @@ pub struct SmartModuleWasmSummary {
 pub struct SmartModuleWasm {
     pub format: SmartModuleWasmFormat,
     #[cfg_attr(feature = "use_serde", serde(with = "base64"))]
-    pub payload: Vec<u8>,
+    pub payload: ByteBuf,
```
morenol, line 113: probably we need to bump `fluvio-controlplane-metadata` version

morenol (commented):
Nice improvement, it looks good to me just a couple of nits.

I see that CI test that checks crate version reports crates that need to be updated but it is not failing, maybe is it an issue with the exit code? CC @tjtelan

crates/fluvio-controlplane-metadata/src/smartmodule/spec.rs
EstebanBorai: I was thinking I could introduce this in a separate PR? Because this PR already has many changes. This way we keep it a bit more progressively.

EstebanBorai: Addressed on: [bd55b27](https://github.com/infinyon/fluvio/pull/2738/commits/bd55b27c6eadc9fd25fe25f90b1b97aca99f552c)

sehz, line 113: let's remove `pub` signature, we don't need to expose it

crates/fluvio-protocol/Cargo.toml
```diff
@@ -36,6 +36,7 @@ tokio-util = { version = "0.7.0", features = [
     "compat",
 ], optional = true }
 semver = { version = "1.0.0", optional = true }
+serde = { version = "1.0.147", features = ['derive'], optional = true }
```
sehz, line 39: Do we still need this? I dont' see ref in the crate anymore

crates/fluvio-controlplane-metadata/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "fluvio-controlplane-metadata"
 edition = "2021"
-version = "0.19.0"
+version = "0.20.0"
```
sehz, line 4: I don't think we need to bump a major bump since we didn't actually change API.   so only patch version increase i s needed

crates/fluvio-protocol/src/core/bytebuf.rs
```diff
@@ -0,0 +1,221 @@
...
+use std::io::ErrorKind;
+use std::ops::Deref;
+
+use bytes::{Buf, Bytes, BufMut};
+
+use crate::{Encoder, Decoder, Version};
+
+/// Represnts a SmartModule WASM File bytes.
+///
+/// Provides a `Encoder` and `Decoder` implementation optimized for WASM files
```
sehz, line 11: remove reference to WASM since this optimization is universal

sehz (changes requested):
few remaining nits

crates/fluvio-controlplane-metadata/src/smartmodule/spec.rs
EstebanBorai: The fluvio-sc-schema uses the field:
https://github.com/infinyon/fluvio/blob/6c57ed0d6d72ddd31d1c9cd59910b6b66f4fa79a/crates/fluvio-sc-schema/src/smartmodule/mod.rs#L21

I think we could expose it via method call (similar to the getter pattern)?

crates/fluvio-protocol/Cargo.toml
EstebanBorai: Thanks! Addressed on: [a63b4b2](https://github.com/infinyon/fluvio/pull/2738/commits/a63b4b25a26acd1d627d7442798ad30fcfe48bb7)

crates/fluvio-controlplane-metadata/Cargo.toml
EstebanBorai: Okay! I thought given that we introduced a different struct this represented a change in API. But as we plan to keep it private and for internal usage makes sense to me as well. Thanks!

crates/fluvio-protocol/src/core/bytebuf.rs
EstebanBorai: Addressed on: [398e679](https://github.com/infinyon/fluvio/pull/2738/commits/398e6792a1ebc8f2e893d574b7bff2d556520a74)

crates/fluvio-controlplane-metadata/src/smartmodule/spec.rs
morenol: we could create a `fn payload_len`

crates/fluvio-sc-schema/Cargo.toml
```diff
@@ -24,9 +24,8 @@ paste = "1.0"
```
sehz, line 24: This crate's version needs to update patch number as well

sehz (changes requested):
Very close.  Last nit

crates/fluvio-controlplane-metadata/src/smartmodule/spec.rs
EstebanBorai: Its good for this use case, but we also consume the `payload` directly for the `SmartModuleInvocationWasm` enum here:

```
error[E0616]: field `payload` of struct `SmartModuleWasm` is private
   --> crates/fluvio-spu/src/smartengine/context.rs:253:78
    |
253 |                 wasm: SmartModuleInvocationWasm::AdHoc(smartmodule.spec.wasm.payload.into()),
```

I think we could think of an API for these use case on a next PR? In order to encapsulate these fields, and
also provide some unit tests to these.

crates/fluvio-sc-schema/Cargo.toml
EstebanBorai: Adressed on: [1ebd052](https://github.com/infinyon/fluvio/pull/2738/commits/1ebd052da8e2352d3f63875a0d6e6e76855421f8)

sehz (approved):
Nice work.  Tremendous improvement.  Ship it!

EstebanBorai:
> Nice work. Tremendous improvement. Ship it!

Thanks! And to everyone involved! :shipit:

Issue #2722, opened 2022-10-19
[Bug]: `fluvio cluster start` fails on Mac


ajhunyady:
I noticed this during issue on a fresh install a few days ago. It was also reported by `xzkxaicxjp` in Discord.
This issue was reported on MacOS Monterey 12.6.

![image](https://user-images.githubusercontent.com/26554085/196742021-0bcf0c01-b2de-4335-908d-b83996d5c4c1.png)

![image](https://user-images.githubusercontent.com/26554085/196742187-4d368b4f-1512-454a-a7ae-50aca6742540.png)

ajhunyady:
Is this a workaround or the solution? Should we update the documentation?

sehz:
This is documented already.

ajhunyady:
It's not... no reference to `proxy` https://www.fluvio.io/docs/get-started/mac/

morenol:
I think that `proxy-addr` is not needed

ajhunyady:
It fails without it.

ajhunyady:
Proof that the `--proxy-addr` os required:

<img width="872" alt="image" src="https://user-images.githubusercontent.com/26554085/196808939-0f39d6f8-f4f1-40e9-9dd3-7aa6f2d61c73.png">

ajhunyady:
@tjtelan @sehz we have 2 options:
* make the installation work without the `--proxy-addr` flag
* update the documentation to include `--proxy-addr` flag

sehz:
For this release, we will update doc.  Will add next to release to automatically configure proxy-addr

EstebanBorai:
Im setting `localhost` as proxy address for k8s environment as default value for macOS here: https://github.com/EstebanBorai/fluvio/commit/c38fa5f19977c1917b6fdc9f8f166ab50274caa8

Im testing locally and will open the PR for review soon. This way users don't have to do it theirselves.

PR #2740, opened 2022-10-22
[Merged by Bors] - fix: use `localhost` as `proxy-addr` for k8s cluster on macOS

EstebanBorai:
When running `fluvio cluster start` on macOS sets `localhost` as proxy
address by default if no other value is set. For other platforms, behavior
remains the same as of today.

Resolve: https://github.com/infinyon/fluvio/issues/2722


sehz:
works for rancher desktop in Mac

PR #2753, opened 2022-10-25
[Merged by Bors] - feat(smartmodule): added chain support to producer

galibey:
This PR is a step towards supporting `transform` section for all types of connectors. Specifically, here we add support of the chain of smartmodules to Fluvio Producer.
Main changes in this PR:
1. Decouple the creation and filling of the chain builder from the engine. Now it is required only in the instantiation phase. 
2. Fluvio Producer can be built with the chain builder passed from external.

### DIFF
--- a/crates/fluvio-smartengine/Cargo.toml
+++ b/crates/fluvio-smartengine/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio-smartengine"
-version = "0.4.0"
+version = "0.4.1"
 edition = "2021"
 license = "Apache-2.0"
 authors = ["Fluvio Contributors <team@fluvio.io>"]
--- a/crates/fluvio-smartengine/Cargo.toml
+++ b/crates/fluvio-smartengine/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio-smartengine"
-version = "0.4.1"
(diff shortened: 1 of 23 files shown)


crates/fluvio-smartengine/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio-smartengine"
-version = "0.4.0"
+version = "0.4.1"
```
sehz, line 3: since API is changed, we need to bump this to 0.5.0

crates/fluvio/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio"
-version = "0.14.0"
```
sehz, line 3: need to bump to `0.15.0` with API change

sehz (changes requested):
Looks very good.  only issue is with that we need to bump up version

crates/fluvio-smartengine/Cargo.toml
galibey: done

crates/fluvio/Cargo.toml
galibey: done

sehz (approved):
LGTM

PR #2754, opened 2022-10-25
[Merged by Bors] - chore: Update changelog for 0.10.0

tjtelan:
Generated with git-cliff, then manually added in.

Still need to clean this up. Was a very large release


sehz (approved):
LGTM

PR #2756, opened 2022-10-26
[Merged by Bors] - fix: Standardize reading records from CLI input

### LINKED ISSUE #2755: Add smdk test input improvements
smdk test should support more of the features that existing the fluvio CLI producer and consumer.

The most important request is the ability to feed files that are processed line by line. This is a requirement to have the ability to test aggregates.

tjtelan:
Resolves #2755 

`smdk test --file <file>` by default will read in file and treat each line as a separate record.

If you want to read file as a single record, you need to pass in additional flag `--raw`

### DIFF
--- a/crates/fluvio-cli-common/Cargo.toml
+++ b/crates/fluvio-cli-common/Cargo.toml
@@ -26,9 +26,14 @@ thiserror = "1.0.20"
 http = { version = "0.2", default-features = false}
 isahc = { version = "1.7", default-features = false, features = ["static-curl"] }
 futures = { version = "0.3" , default-features = false, features = ["std", "io-compat"]}
+tempfile = "3.3.0"
+anyhow = { version = "1.0.38" }
+bytes = { version = "1" }
 
 fluvio-package-index = { path = "../fluvio-package-index", features = ["http_agent"] }
 fluvio-types = { path = "../fluvio-types", optional = true }
+fluvio = { path = "../fluvio", default-features = false }
+fluvio-protocol = { path = "../fluvio-protocol", features=["record","api"] }
 
 [dev-dependencies]
 fluvio-future = { version = "0.4.0", features = ["fs", "io", "subscriber", "native2_tls", "fixture"] }
--- a/crates/fluvio-cli-common/Cargo.toml
+++ b/crates/fluvio-cli-common/Cargo.toml
@@ -14,7 +14,7 @@ path = "src/lib.rs"
 
 [features]
 default = ["fluvio-types"]
-file-records = ["fluvio-protocol"]
+file-records = ["fluvio-protocol/record", "fluvio-protocol/api"]
 
 [dependencies]
 tracing = "0.1.19"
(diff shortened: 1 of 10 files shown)


crates/smartmodule-development-kit/src/test.rs
```diff
@@ -75,20 +81,43 @@ impl TestOpt {
...
-            std::fs::read(json_file)?
+            input.as_bytes().to_vec();
+
+            let input_bytes = input.as_bytes().to_vec();
+
+            input_size += input_bytes.len();
+
+            vec![Record::new_key_value(RecordKey::NULL, input_bytes)]
+        } else if let Some(test_file) = &self.file {
+            if self.file_single {
```
sehz, line 95: can we have unit test for this?  Also, we seems to have similar fn in the consume as well.  It would be good to consolidate into a common util (not show stopper).

sehz (changes requested):

crates/fluvio-cli-common/Cargo.toml
```diff
@@ -26,9 +26,14 @@ thiserror = "1.0.20"
 http = { version = "0.2", default-features = false}
 isahc = { version = "1.7", default-features = false, features = ["static-curl"] }
 futures = { version = "0.3" , default-features = false, features = ["std", "io-compat"]}
+tempfile = "3.3.0"
```
sehz, line 29: this seems to be used only by the test, so should it be moved to `dev-dependencies`?

crates/fluvio-cli-common/Cargo.toml
```diff
@@ -26,9 +26,14 @@ thiserror = "1.0.20"
...
+anyhow = { version = "1.0.38" }
+bytes = { version = "1" }
 
 fluvio-package-index = { path = "../fluvio-package-index", features = ["http_agent"] }
 fluvio-types = { path = "../fluvio-types", optional = true }
+fluvio = { path = "../fluvio", default-features = false }
+fluvio-protocol = { path = "../fluvio-protocol", features=["record","api"] }
```
sehz, line 36: since not all CLI needs to consume (like channel), we should put under separate feature flag

crates/fluvio-cli-common/Cargo.toml
```diff
@@ -26,9 +26,14 @@ thiserror = "1.0.20"
 http = { version = "0.2", default-features = false}
 isahc = { version = "1.7", default-features = false, features = ["static-curl"] }
 futures = { version = "0.3" , default-features = false, features = ["std", "io-compat"]}
+tempfile = "3.3.0"
+anyhow = { version = "1.0.38" }
+bytes = { version = "1" }
```
sehz, line 31: byes can be imported from `fluvio-protocol` so no need to add as dependency

crates/fluvio-cli/Cargo.toml
```diff
@@ -73,6 +73,7 @@ indicatif = "0.17.0"
 async-trait = "0.1.21"
 tokio = { version = "1.3.0", features = ["macros"] }
 async-net = "1.7.0"
+bytes = { version = "1" }
```
sehz, line 76: this can be imported from `fluvio-protocol`

sehz (changes requested):
Generally looks good.  The only issue is that we should put this into the feature flag

crates/smartmodule-development-kit/Cargo.toml
```diff
@@ -13,6 +13,10 @@ name = "smdk"
 path = "src/main.rs"
 doc = false
 
+[features]
+default = ["test-file-io"]
+test-file-io = ["fluvio-cli-common/file-records"]
```
sehz, line 18: is there a reason to make this optional? unless we are building non file-records CLI version.  Just add this feature flag to the dependency.

sehz (changes requested):
Much better.   The remaining question is whether to add to make `producer-file` as an optional feature to CLI's ?  It just seems to add more complexity.

sehz (approved):
LGTM

PR #2759, opened 2022-10-27
[Merged by Bors] - feat(smartmodule): added smartmodule chain support for consumer

galibey:
Added support for multiple smartmodules in `StreamFetchRequest` and removed all legacy types for SmartModule since we have a chance to break compatibility in the current version.


crates/fluvio-protocol-derive/src/ast/prop.rs
```diff
@@ -54,8 +54,7 @@ impl NamedProp {
 
         if let Some(max) = self.attrs.max_version {
             quote! {
```
sehz, line 56: can you bump patch version?

sehz (changes requested):
Generally looks very good.  Can you make sure all released crate version is bump up otherwise will not pass bors.
Also, can you test SmartModule works with stable version?

sehz:
Tested with the stable version of Fluvio.  Works fine.

sehz (approved):

PR #2760, opened 2022-10-27
[Merged by Bors] - feat: capability to validate WASM files

EstebanBorai:
Introduce `validate_binary` to validate WASM files (binary variant).
The validation approach consists on parsing WASM file bytes, when
an error is found the parser will short-circuit by returning an `Err` early.
To ensure the WASM file is valid, the file is parsed to the end.


crates/fluvio-controlplane-metadata/Cargo.toml
```diff
@@ -27,6 +27,7 @@ flate2 = { version = "1.0.22", optional = true }
 semver = "1.0"
 toml = { version = "0.5.9", optional = true }
 lenient_semver = "0.4.2"
+wasmparser = "0.93.0"
```
EstebanBorai, line 30: The current implementation follows the associated function pattern so the validation can be used in other use cases. This opens the possibility of having this dependency as a feature as well.

crates/fluvio-controlplane-metadata/src/smartmodule/spec.rs
```diff
@@ -111,6 +118,35 @@ pub struct SmartModuleWasm {
...
+
+impl SmartModuleWasm {
+    /// Validates a SmartModule's WASM payload to represent a valid WASM file
+    /// in the binary format (*.wasm).
+    pub fn validate_binary(mut data: &[u8]) -> Result<(), ValidateSmartModuleWasmError> {
+        let mut parser = Parser::default();
+
+        loop {
+            match parser
+                .parse(data, true)
```
EstebanBorai, line 130: The `true` value here, belongs to the EOF flag, which represents wether more data will be received.
As we count with all bytes in memory for current use cases we assume bytes are present in memory.

crates/fluvio-controlplane-metadata/src/smartmodule/spec.rs
```diff
@@ -174,3 +210,37 @@ mod base64 {
...
+    }
+
+    #[test]
+    fn validates_smartmodule_wasm_binary_corrupt() {
+        let wasm_bytes = read("tests/sm_need_more_data.wasm")
+            .expect("Failed to find 'sm_need_more_data.wasm' file for tests");
+        let result = SmartModuleWasm::validate_binary(wasm_bytes.as_slice());
+
+        assert!(result.is_err());
+    }
```
EstebanBorai, line 246: For this test use case, I have removed many bytes from the end of the file to make sure the parsing is working as expected. Another test case is above for non WASM files.

digikata (commented):
The code  LGTM.  I like the unit tests cases - those cover some nice basics.  Do we know the full range of users of fluvio-controlplane-metadata.  I don't want to blow up the wasm size for the dashboard for example.  One option is to put this fcn behind a feature.

EstebanBorai:
> The code LGTM. I like the unit tests cases - those cover some nice basics. Have you looked at the full range of users of fluvio-controlplane-metadata. I don't want to blow up the wasm size for the dashboard for example. One option is to put this fcn behind a feature.

Yeah! I thought about the feature approach as well. Let me do that so we can include it only when needed 😄

digikata (changes requested):

digikata (approved):
LGTM

crates/fluvio-controlplane-metadata/Cargo.toml
```diff
@@ -13,6 +13,7 @@ path = "src/lib.rs"
 
 [features]
 smartmodule = ["flate2","toml","use_serde"]
+smartmodule_validation = ["wasmparser"]
```
sehz, line 16: Is there reason to make this feature flag?

sehz (changes requested):

crates/fluvio-controlplane-metadata/Cargo.toml
digikata: I asked for it avoid an increase in the size of the dashboard ui wasm.

EstebanBorai:
I reduced the scope of the validation to Hub Util for our current use case.
Evaluating current API for `SmartModuleWasm` I can see that we have `from_raw_wasm_bytes` which also compresses
the file so I can't reuse it for our Hub Utility which uses another approach.

At some point it would be great to have a single representation we could use for both use cases, so validation is centralized
in a simple but reusable implementation.

crates/fluvio-controlplane-metadata/src/smartmodule/spec.rs
```diff
@@ -141,6 +152,34 @@ impl SmartModuleWasm {
 
         Ok(Self::from_compressed_gzip(buffer))
     }
+
+    #[cfg(feature = "smartmodule_validation")]
+    /// Validates a SmartModule's WASM payload to represent a valid WASM file
+    /// in the binary format (*.wasm).
+    pub fn validate_binary(mut data: &[u8]) -> Result<(), ValidateSmartModuleWasmError> {
+        let mut parser = Parser::default();
```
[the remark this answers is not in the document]
EstebanBorai: Im trying to reuse this in the hub util crate. By making this function private here I couldn't access it from the hub util crate. So I implemented a Hub Util read wasm function which validates internally here: https://github.com/infinyon/fluvio/pull/2760/commits/8ff36fc341463c077b42d6ca1c671b44ec595f2f

crates/fluvio-controlplane-metadata/Cargo.toml
EstebanBorai: I'm resolving this given that implementation has changed.

sehz (approved):
LGTM.  Later, we need to consolidate validation of WASM

EstebanBorai:
> LGTM. Later, we need to consolidate validation of WASM

Totally agree!

PR #2764, opened 2022-10-31
[Merged by Bors] - feat: setup benchmark action

EstebanBorai:
Setups Benchmarks as part of the CI workflow using the [benchmark-action][1].
Benchmarks will be executed after running tests with the `local_cluster_test`
action.

Previous benchmarks are stored using `actions/cache` action with the goal of
comparing them on every change to measure improvements on performance.

[1]: https://github.com/benchmark-action/github-action-benchmark

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -661,8 +661,14 @@ jobs:
       - name: Store benchmark result
         uses: benchmark-action/github-action-benchmark@v1
         with:
+          # What benchmark tool the output.txt came from
           tool: 'cargo'
+          # Where the output from the benchmark tool is stored
           output-file-path: benchmarks.txt
+          # GitHub API token to make a commit comment
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          # Enable alert commit comment
+          comment-on-alert: true
 
   build_image:
     name: Build Fluvio Docker image
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -663,7 +663,7 @@ jobs:
           key: ${{ runner.os }}-benchmark
 
       - name: Run Benchmarks
-        run: cargo +nightly bench | tee benchmarks.txt
+        run: cargo +nightly bench -p fluvio-protocol | tee benchmarks.txt
 
       - name: Store benchmark result
         uses: benchmark-action/github-action-benchmark@v1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -678,7 +678,7 @@ jobs:
           # Enable alert commit comment
           comment-on-alert: true
           # Where the previous data file is stored
-           external-data-json-path: ./benches/benchmark-data.json
+          external-data-json-path: ./benches/benchmark-data.json
 
   build_image:
     name: Build Fluvio Docker image
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -645,7 +645,6 @@ jobs:
   unit-benchmark:
     name: Unit Benchmarks
     runs-on: ubuntu-latest
-    needs: check
     steps:
       - name: Checkout Code
         uses: actions/checkout@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -656,7 +656,7 @@ jobs:
           toolchain: nightly
           profile: minimal
 
-      - name: Download previous benchmark data
+      - name: Setup Cache for Rust and Benchmarks
         uses: Swatinem/rust-cache@v2
         with:
           # Additional non workspace directories, separated by newlines
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -642,6 +642,28 @@ jobs:
           path: diagnostics*.gz
           retention-days: 1
 
+  benchmarks:
+    name: Benchmarks
+    needs: local_cluster_test
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+
+      - name: Download previous benchmark data
+        uses: actions/cache@v1
+        with:
+          path: ./cache
+          key: ${{ runner.os }}-benchmark
+
+      - name: Run Benchmarks
+        run: cargo +nightly bench | tee benchmarks.txt
+
+      - name: Store benchmark result
+        uses: benchmark-action/github-action-benchmark@v1
+        with:
+          tool: 'cargo'
+          output-file-path: benchmarks.txt
+
   build_image:
     name: Build Fluvio Docker image
     needs: build_primary_binaries
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -655,6 +655,13 @@ jobs:
           path: ./cache
           key: ${{ runner.os }}-benchmark
 
+      - name: Install Nightly Toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+            toolchain: nightly
+            override: true
+            components: bench
+
       - name: Run Benchmarks
         run: cargo +nightly bench | tee benchmarks.txt
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -642,8 +642,8 @@ jobs:
           path: diagnostics*.gz
           retention-days: 1
 
-  benchmarks:
-    name: Benchmarks
+  unit-benchmark:
+    name: Unit Benchmarks
     runs-on: ubuntu-latest
     needs: check
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -645,22 +645,21 @@ jobs:
   benchmarks:
     name: Benchmarks
     runs-on: ubuntu-latest
+    needs: check
     steps:
       - name: Checkout Code
         uses: actions/checkout@v2
 
-      - name: Download previous benchmark data
-        uses: actions/cache@v1
-        with:
-          path: ./cache
-          key: ${{ runner.os }}-benchmark
-
-      - name: Install Nightly Toolchain
+      - name: Install Rust
         uses: actions-rs/toolchain@v1
         with:
-            toolchain: nightly
-            override: true
-            components: bench
+          toolchain: nightly
+          profile: minimal
+
+      - name: Download previous benchmark data
+        uses: Swatinem/rust-cache@v2
+        with:
+          key: ${{ runner.os }}-benchmark
 
       - name: Run Benchmarks
         run: cargo +nightly bench | tee benchmarks.txt
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -644,7 +644,7 @@ jobs:
 
   benchmarks:
     name: Benchmarks
-    needs: local_cluster_test
+    runs-on: ubuntu-latest
     steps:
       - name: Checkout Code
         uses: actions/checkout@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -655,11 +655,16 @@ jobs:
           toolchain: nightly
           profile: minimal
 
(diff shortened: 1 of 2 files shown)


crates/fluvio-protocol/benches/bench.rs
```diff
@@ -0,0 +1,71 @@
...
+
+                black_box(|| {
+                    value.encode(&mut dest, 0);
+                });
+            }
+        });
+    }
+
+    #[bench]
+    fn bench_encode_vecu8(b: &mut Bencher) {
```
morenol, line 41: Could you instead or optionally benchmark instead SmartModuleSpec?

.github/workflows/ci.yml
```diff
@@ -642,6 +642,41 @@ jobs:
           path: diagnostics*.gz
           retention-days: 1
 
+  benchmarks:
+    name: Benchmarks
+    needs: local_cluster_test
```
sehz, line 647: since this is only benchmark code, it should not need `local_cluster_test` only primary build

.github/workflows/ci.yml
```diff
@@ -642,6 +642,41 @@ jobs:
...
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+
+      - name: Download previous benchmark data
+        uses: actions/cache@v1
+        with:
+          path: ./cache
+          key: ${{ runner.os }}-benchmark
+
```
sehz, line 657: add rust cache

crates/fluvio-protocol/benches/bench.rs
```diff
@@ -0,0 +1,71 @@
+#![feature(test)]
+
+extern crate test;
```
sehz, line 3: don't think this is still needed.

sehz (changes requested):
missing bunch of github action steps like:
```
runs-on,
strategy
```

crates/fluvio-protocol/benches/bench.rs
```diff
@@ -0,0 +1,71 @@
...
+
+#[cfg(test)]
+mod bench {
+    use std::io::Cursor;
+
+    use fluvio_protocol::{Decoder, Encoder};
+    use test::{Bencher, black_box};
+
+    #[bench]
+    fn bench_decode_bool(b: &mut Bencher) {
```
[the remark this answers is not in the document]
EstebanBorai: I wanted to have a couple benchmarks to see the results and avoid some emptiness. I will remove these anyway given that #2738 will introduce more benchmarks!

.github/workflows/ci.yml
EstebanBorai: I thought if tests doesn't pass, theres no need to benchmark because theres an issue and benchmarking it wouldn't be useful?

sehz: lets make depend on `check` this only excercise code  path

crates/fluvio-protocol/benches/bench.rs
EstebanBorai: I took it from here: https://doc.rust-lang.org/unstable-book/library-features/test.html

And when I try to remove it, Rust is not able to resolve `test` crate.

```
error[E0432]: unresolved import `test`
 --> crates/fluvio-protocol/benches/bench.rs:6:9
  |
6 |     use test::{Bencher, black_box};
  |         ^^^^ use of undeclared crate or module `test`
```

As I understand this is because benchmarks still unstable internally and we have to introduce it to the scope explicitly, theres this note in the link above about it:

> Note the test feature gate, which enables this unstable feature.

EstebanBorai:
> missing bunch of github action steps like:
> 
> ```
> runs-on,
> strategy
> ```

Gotcha! Running on `ubuntu-latest` for now we will use a single platform for benchmarking!

crates/fluvio-protocol/benches/bench.rs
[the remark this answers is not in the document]
EstebanBorai: I have removed `bool` encoding/decoding benches, now we just have for `Vec<u8>`

crates/fluvio-protocol/benches/bench.rs
EstebanBorai: @morenol actually, what do you think about introducing more benchmarks on a separate PR? One where we could focus on adding benchmarks? That way I can speed up progress on #2738?

.github/workflows/ci.yml
```diff
@@ -642,6 +642,40 @@ jobs:
           path: diagnostics*.gz
           retention-days: 1
 
+  benchmarks:
+    name: Benchmarks
+    runs-on: ubuntu-latest
```
EstebanBorai, line 647: Specified `ubuntu-latest` as our platform for running benchmarks.

.github/workflows/ci.yml
```diff
@@ -642,6 +642,40 @@ jobs:
...
+      - name: Download previous benchmark data
+        uses: Swatinem/rust-cache@v2
+        with:
+          key: ${{ runner.os }}-benchmark
+
+      - name: Run Benchmarks
+        run: cargo +nightly bench | tee benchmarks.txt
+
+      - name: Store benchmark result
+        uses: benchmark-action/github-action-benchmark@v1
```
morenol, line 668: I think that we need the parameter `external-data-json-path`

.github/workflows/ci.yml
```diff
@@ -642,6 +642,40 @@ jobs:
           path: diagnostics*.gz
           retention-days: 1
 
+  benchmarks:
```
sehz, line 645: this should be labeled as `unit-benchmark` or something like to indicate this doesn't perform integration or cluster based benchmark

.github/workflows/ci.yml
```diff
@@ -642,6 +642,40 @@ jobs:
...
+      - name: Checkout Code
+        uses: actions/checkout@v2
+
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly
+          profile: minimal
+
+      - name: Download previous benchmark data
```
sehz, line 659: name seems to be different from what it does

crates/fluvio-protocol/benches/bench.rs
sehz: Yes, can always add more benchmarks later.

sehz (changes requested):

.github/workflows/ci.yml
EstebanBorai: Yeah, I just added support to cache a `benches` directory where we will store and gather previous benchmark.

sehz:
CI failure

.github/workflows/ci.yml
```diff
@@ -642,6 +642,44 @@ jobs:
...
+          # What benchmark tool the output.txt came from
+          tool: 'cargo'
+          # Where the output from the benchmark tool is stored
+          output-file-path: benchmarks.txt
+          # GitHub API token to make a commit comment
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          # Enable alert commit comment
+          comment-on-alert: true
+          # Where the previous data file is stored
+           external-data-json-path: ./benches/benchmark-data.json
```
morenol, line 681: looks like there is indentation issue here

sehz: YML is picky.   @EstebanBorai If you have VSCode or other tools, there is YAML plug-in

EstebanBorai: Fixed! Yeah I have to check for linters on this

EstebanBorai: I'm rendering spaces already, but sometimes it is a bit tricky to see

sehz:
dont' see benchmark actions being run

morenol:
> dont' see benchmark actions being run

it runs after `check` which is still in progress. I think that maybe we should remove that dependency

EstebanBorai:
> > dont' see benchmark actions being run
> 
> it runs after `check` which is still in progress. I think that maybe we should remove that dependency

My goal was to make sure benchmarks run only when test passes so these are legit.
But we could skip it anyways because _All CI workflows should pass to enable merging_ :octocat:

morenol:
> 
> My goal was to make sure benchmarks run only when test passes so these are legit. But we could skip it anyways because _All CI workflows should pass to enable merging_ :octocat:

I think that we should not rely entirely on the benchmark results because the performance of the github runner could vary from build to build. But it is at least something that we could use to get some numbers.

Maybe one option to prevent issues with differences between builds is to run the banchmark with master branch and with current branch and compare results with that. That could be handled in other PR

EstebanBorai:
> > My goal was to make sure benchmarks run only when test passes so these are legit. But we could skip it anyways because _All CI workflows should pass to enable merging_ :octocat:
> 
> I think that we should not rely entirely on the benchmark results because the performance of the github runner could vary from build to build. But it is at least something that we could use to get some numbers.
> 
> Maybe one option to prevent issues with differences between builds is to run the banchmark with master branch and with current branch and compare results with that. That could be handled in other PR

This sounds actually great, by default the benchmark action takes previous commit as reference. But It would be a matter of checking if we can checkout master brach or use an special cache key from where we can gather latest `master` benchmark

sehz:
try to supply `-p <package>` to ensure only necessary crates is compiled

.github/workflows/ci.yml
```diff
@@ -642,6 +642,43 @@ jobs:
...
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly
+          profile: minimal
+
+      - name: Setup Cache for Rust and Benchmarks
+        uses: Swatinem/rust-cache@v2
+        with:
+          # Additional non workspace directories, separated by newlines
+          cache-directories: ./benches
```
sehz, line 662: CI is complainig about: 
```
Unexpected input(s) 'cache-directories', valid inputs are ['shared-key', 'key', 'env-vars', 'workspaces', 'cache-on-failure']
```

EstebanBorai:
We are generating benchmarks JSON but still have some issues with caching the file.
I can see Rust's cache does some assumptions on what to cache and that theres some issue with the `cache-directories` option.

```
Data: {
  commit: {
    author: { name: 'EstebanBorai', username: 'EstebanBorai' },
    committer: { name: 'EstebanBorai', username: 'EstebanBorai' },
    id: '4bf6bb0e501de2a24acc37380bb88f08175a968b',
    message: 'feat: setup benchmark action',
    timestamp: '2022-09-14T20:49:36Z',
    url: 'https://github.com/infinyon/fluvio/pull/2764/commits/4bf6bb0e501de2a24acc37380bb88f08175a968b'
  },
  date: 1667338291766,
  tool: 'cargo',
  benches: [
    {
      name: 'bench::bench_decode_vecu8',
      value: 81,
      range: '± 0',
      unit: 'ns/iter'
    },
    {
      name: 'bench::bench_encode_vecu8',
      value: 79,
      range: '± 0',
      unit: 'ns/iter'
    }
  ]
}
``

.github/workflows/ci.yml
morenol, line 662: I think that we shouldn't use rust-cache to cache previous result

sehz: `rust-cache` is for caching rust build artifacts not for other purposes

EstebanBorai: Yeah this warning took me by surprise here: https://github.com/infinyon/fluvio/actions/runs/3373039401/jobs/5597164655#step:4:1

Because I can see in the docs and also in their input implementation that `cache-directories` is actually supported 🤔 .

https://github.com/Swatinem/rust-cache#example-usage

EstebanBorai: Yeah I was trying to accomplish the benchmarks caching as well using the `cache-directories` option from `rust-cache` but given that its not working and that `rust-cache` does a great job caching rust artifacts and deps. I will be using `actions/cache` for the single purpose of caching our benchmark file as shown in the benchmarks repo.

sehz (approved):
LGTM

PR #2770, opened 2022-11-03
feat: benches with criterion and dedicated workflow

EstebanBorai:
Replaces benchmarking with Criterion crate and provides a dedicated
workflow for running benchmarks as part of CI.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -4,6 +4,9 @@ on:
   pull_request:
     branches:
       - master
+  push:
+    branches:
+      - master
 
 jobs:
   unit_fluvio_protocol:
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -50,3 +50,5 @@ jobs:
           comment-always: true
           # Where the previous data file is stored
           external-data-json-path: ./benches_cache/benchmark-data.json
+          # Mention @infinyon/developers in the commit comment
+          alert-comment-cc-users: '@infinyon/developers'
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -1,0 +1,49 @@
+name: Benchmarks
+
+on:
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  unit_fluvio_protocol:
+    name: Fluvio Protocol - Unit Benchmarks
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly
+          profile: minimal
+
+      - name: Cache Rust Cargo files
+        uses: Swatinem/rust-cache@v2
+        with:
+          # Additional non workspace directories, separated by newlines
+          key: ${{ runner.os }}-benchmark
+
+      - name: Cache Benchmark data
+        uses: actions/cache@v1
+        with:
+          path: ./benches_cache
+          key: ${{ runner.os }}-benchmark
+
+      - name: Run Benchmarks
+        run: cargo +nightly bench -p fluvio-protocol | tee benchmarks.txt
+
+      - name: Store benchmark result
+        uses: benchmark-action/github-action-benchmark@v1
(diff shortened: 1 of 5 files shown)


EstebanBorai:
@sehz , @morenol looks like we will have to provide special permissions for Benchmark workflow in order to allow commenting on commit:

```
Run benchmark-action/github-action-benchmark@v1
Error: Resource not accessible by integration
```
[Actual Log](https://github.com/infinyon/fluvio/actions/runs/3382559614/jobs/5617601014#step:7:22)

sehz:
CI failure

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,54 @@
+name: Benchmarks
+
```
sehz, line 2: add concurrency like:
```
concurrency:
  group: ci-${{ github.ref }}
  cancel-in-progress: true
```

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,54 @@
...
+    name: Fluvio Protocol - Unit Benchmarks
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly
```
sehz, line 22: is still need to be nightly with criterion?

sehz (changes requested):

.github/workflows/benchmarks.yml
EstebanBorai: Nope actually!

EstebanBorai:
> CI failure

Yeah, its related to permissions I left this comment here: https://github.com/infinyon/fluvio/pull/2770#issuecomment-1301578625.

I see other people complaining about similar issue here: https://github.com/actions/first-interaction/issues/10

EstebanBorai:
Checking on the GitHub Token Permissions I can see we can only read PR:

```
GITHUB_TOKEN Permissions
  Metadata: read
  PullRequests: read
```

I think we could add an custom token for this workflow which allows write perhaps?

[Actual Log](https://github.com/infinyon/fluvio/actions/runs/3382952790/jobs/5618409078#step:1:17)

I also [opened an issue](https://github.com/benchmark-action/github-action-benchmark/issues/135) on the action repository to ask for permissions scopes we could use for our workflow.

morenol:
I guess that we could make it only run on master branch, I think that for pushes to that branch we won't get permission issues

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,61 @@
+name: Benchmarks
+
+concurrency:
+  group: ci-${{ github.ref }}
```
sehz, line 4: should be `benchmark-${{ github.ref }}`

sehz (changes requested):
maybe one way to solve permission issues is  only cache the benchmark only master

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,61 @@
...
+  cancel-in-progress: true
+
+on:
+  pull_request:
```
morenol, line 8: I think that we should remove from `pull_requests` and try this like cd_dev.yaml, not required for merge but nice to have monitored

morenol (approved):

sehz (approved):

morenol:
ohh, clippy issues!

EstebanBorai:
@morenol reviewing!

EstebanBorai:
> ohh, clippy issues!

Doesn't look related to this PR: https://github.com/infinyon/fluvio/actions/runs/3386673257/jobs/5626387694#step:8:1252
Will fix in a separate one, to also focus tests and checks on such change instead of introducing both things here.

EstebanBorai:
> > ohh, clippy issues!
> 
> Doesn't look related to this PR: https://github.com/infinyon/fluvio/actions/runs/3386673257/jobs/5626387694#step:8:1252 Will fix in a separate one, to also focus tests and checks on such change instead of introducing both things here.

Just saw you took care of this already here: https://github.com/infinyon/fluvio/pull/2771 Great!

morenol:
bors retry

PR #2773, opened 2022-11-03
[Merged by Bors] - feat(bench): benches for SmartModuleInput encoding

EstebanBorai:
Provides benchmarks on current encoding performance for `SmartModuleInput`
in order to have a comparison point against new implementation on [`ByteBuf`](https://github.com/infinyon/fluvio/pull/2738).


.github/workflows/benchmarks.yml
```diff
@@ -42,7 +42,7 @@ jobs:
           key: benches-${{ runner.os }}-unit_fluvio_protocol
 
       - name: Run Benchmarks
-        run: cargo bench -p fluvio-protocol -- --output-format bencher | tee benchmarks.txt
+        run: cargo bench -p fluvio-protocol -- --output-format bencher | tee fluvio_protocol_benches.txt
```
EstebanBorai, line 45: Given that we have different cache keys and jobs for benchmarks I think this may not be needed. WDYT? @sehz , @morenol

sehz: have no idea...

sehz (approved):
LGTM

PR #2794, opened 2022-11-09
cli: don't print in stdout warnings about new version available in `latest` channel

morenol:
in fluvio-connectors we use:

```
fluvio partition list -O json | jq ".[] | select ...
```

But `jq` fails due to those texts are printed to stdout

### DIFF
--- a/crates/fluvio-cli/src/install/update.rs
+++ b/crates/fluvio-cli/src/install/update.rs
@@ -282,9 +282,9 @@ pub async fn prompt_required_update(agent: &HttpAgent) -> Result<()> {
     debug!(%target, %id, "Fetching latest package version:");
     let latest_version = fetch_latest_version(agent, &id, &target, false).await?;
 
-    println!("⚠️ A major update to Fluvio has been detected!");
-    println!("⚠️ You must complete this update before using any 'install' command");
-    println!(
+    eprintln!("⚠️ A major update to Fluvio has been detected!");
+    eprintln!("⚠️ You must complete this update before using any 'install' command");
+    eprintln!(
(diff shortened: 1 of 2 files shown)


sehz:
seems reasonable.  @tjtelan ?

tjtelan (approved):
I don't feel strongly about this change. It looks fine, but it isn't "ideal".

Personally I don't think this is a "correct" behavior for CLI to print non-errors to `stderr`. My worry is that some times scripts/tools sometimes report error when there is content in `stderr`. However, it's a small concern.

The `fluvio` CLI needs to support usage the way Luis described.


This can probably be tech debt for someone to report later, butI see an ideal solution being implemented in 2 different ways:
1.  Before we print announcements, verify if we are going to print to `stdout` or to a `pipe` then change our printing behavior to meet the demands for scripting.
    * In the case of a pipe, we can suppress announcements or anything that could be noise for a script.
2. Add a global flag that suppresses announcements that we can document. Something like `--quiet`, `--suppress-alerts`, `--script-mode`... that sets some expectations for non-interactive usage.

example:
```
$ fluvio --script-mode partition list -O json | jq ".[] | select ...
```

Up to you, Luis, if you want to keep what you've got (Ship it!) or if you feel like approaching how I've suggested.

I can scope and create an issue for beginner contributors depending on your response.

morenol:
@tjtelan My main concern is that it was blocking `fluvio-connectors` CI, I did this workaround that I think that it is ok for now https://github.com/infinyon/fluvio-connectors/pull/406/commits/b5e8c798d0ddb4f26d8165ac9a124e00125f486d 

Now that it was solved, I think that scoping the `--quiet` option is the right way to go

Issue #2593, opened 2022-08-26
Integrate `cargo-vet` for checking third party dependency


sehz:
https://crates.io/crates/cargo-vet.

PR #2795, opened 2022-11-10
[Merged by Bors] - feat: setup cargo-vet for auditing crates

EstebanBorai:
Setups and initializes cargo vet in the project.

Resolves: https://github.com/infinyon/fluvio/issues/2593

---

By default cargo-vet uses the [safe-to-run](https://mozilla.github.io/cargo-vet/built-in-criteria.html#safe-to-run) and [safe-to-deploy](https://mozilla.github.io/cargo-vet/built-in-criteria.html#safe-to-deploy) criteria OOTB, so this [audit criteria][1] is already active. [More details here](https://mozilla.github.io/cargo-vet/specifying-policies.html).

If this gets merged it would be nice to also add this audit to [The Registry][2].

[1]: https://mozilla.github.io/cargo-vet/audit-criteria.html#audit-criteria
[2]: https://mozilla.github.io/cargo-vet/importing-audits.html#the-registry

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -294,6 +294,7 @@ jobs:
   cargo-vet:
     name: Vet Dependencies
     runs-on: ubuntu-latest
+    if: github.event_name == 'push' && github.ref == 'refs/heads/staging'
     env:
       CARGO_VET_VERSION: 0.3.0
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -302,7 +302,10 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Install Rust
-        run: rustup update stable && rustup default stable
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          override: true
 
       - name: Cache Cargo Vet
         uses: actions/cache@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -291,6 +291,34 @@ jobs:
           path: ${{ env.RUST_BIN_DIR }}/${{ matrix.binary }}
           retention-days: 1
 
+  cargo-vet:
+    name: Vet Dependencies
+    runs-on: ubuntu-latest
+    env:
+      CARGO_VET_VERSION: 0.3.0
+
+    steps:
+      - name: Checkout Source
+        uses: actions/checkout@v3
+
+      - name: Install Rust
+        run: rustup update stable && rustup default stable
+
+      - name: Cache Cargo Vet
+        uses: actions/cache@v2
+        with:
+          path: ${{ runner.tool_cache }}/cargo-vet
+          key: cargo-vet-bin-${{ env.CARGO_VET_VERSION }}
+
+      - name: Add "cargo vet" from cache to PATH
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -291,6 +291,34 @@ jobs:
...
+    runs-on: ubuntu-latest
+    env:
+      CARGO_VET_VERSION: 0.3.0
+
+    steps:
+      - name: Checkout Source
+        uses: actions/checkout@v3
+
+      - name: Install Rust
+        run: rustup update stable && rustup default stable
```
sehz, line 305: Can you use `cargo` action?

.github/workflows/ci.yml
```diff
@@ -291,6 +291,34 @@ jobs:
           path: ${{ env.RUST_BIN_DIR }}/${{ matrix.binary }}
           retention-days: 1
 
+  cargo-vet:
+    name: Vet Dependencies
+    runs-on: ubuntu-latest
+    env:
+      CARGO_VET_VERSION: 0.3.0
```
sehz, line 298: Do we need to run in PR or can be just run in staging?

sehz (changes requested):

.github/workflows/ci.yml
EstebanBorai: Of course!

.github/workflows/ci.yml
EstebanBorai: Yeah, the goal is to vet dependencies whenever a PR is opened so we notice before merging if theres any warning.
Im taking in consideration [this scenario](https://mozilla.github.io/cargo-vet/how-it-works.html#adding-new-third-party-code).

.github/workflows/ci.yml
EstebanBorai: Addressed on: https://github.com/infinyon/fluvio/pull/2795/commits/22f4edc921a3cf282dbb659be55b8bb5a5499111

.github/workflows/ci.yml
sehz: I am concerned that this will slow down builds. I proposed that only performs in the staging

sehz (changes requested):

.github/workflows/ci.yml
```diff
@@ -291,6 +291,34 @@ jobs:
           path: ${{ env.RUST_BIN_DIR }}/${{ matrix.binary }}
           retention-days: 1
 
+  cargo-vet:
+    name: Vet Dependencies
+    runs-on: ubuntu-latest
```
EstebanBorai, line 296: @sehz to scope this only to `staging` perhaps we could do:

```suggestion
  cargo-vet:
    name: Vet Dependencies
    runs-on: ubuntu-latest
    if: github.event_name == 'push' && github.ref == 'refs/heads/staging'
```

This way we will only run when pushing to the `staging` branch.

.github/workflows/ci.yml
EstebanBorai: Perhaps the suggestion on [this comment](https://github.com/infinyon/fluvio/pull/2795#discussion_r1019649506) helps reduce the scope to `staging` only.

.github/workflows/ci.yml
sehz: yes that make sense

PR #2813, opened 2022-11-14
Add RetryPolicy while SPU connects to SC

udsamani:
## What does the PR aim to do ?

This PR aims to complete the pending ToDo of adding retry strategy while SPU tries to connect to SC.


udsamani:
@sehz Was going through the code base and saw a pending ToDo of adding retry strategy for SPU connecting to SC. Any advice on what should the behavior be if we run out of maximum retries while connecting SPU to SC ?

sehz:
That's a good question.  Since SPU needs SC, it just needs to continuously retry.

PR #2815, opened 2022-11-15
feat: Reuse StoreContext from stream_dispatcher in fluvio crate

morenol:
This is done in order to reduce code duplication and have the `watch` impl in the fluvio-stream-dispatcher crate

### DIFF
--- a/crates/fluvio-stream-dispatcher/src/store/mod.rs
+++ b/crates/fluvio-stream-dispatcher/src/store/mod.rs
@@ -12,17 +12,13 @@ mod context {
     use std::fmt::{Display, Debug};
     use std::time::Duration;
 
-    use async_rwlock::RwLockReadGuard;
-    use fluvio_stream_model::core::MetadataItem;
-    use fluvio_stream_model::store::DualEpochMap;
-    use tracing::error;
     use async_channel::{Sender, Receiver, bounded, SendError};
+    use async_rwlock::RwLockReadGuard;
+    use fluvio_future::timer::sleep;
+    use fluvio_stream_model::{core::MetadataItem, store::DualEpochMap};
     use once_cell::sync::Lazy;
     use tokio::select;
-    use tracing::instrument;
-    use tracing::{debug, trace};
-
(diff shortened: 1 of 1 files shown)


crates/fluvio-stream-dispatcher/src/store/mod.rs
```diff
@@ -32,16 +35,16 @@ mod context {
...
 
         let var_value = env::var("FLV_DISPATCHER_WAIT").unwrap_or_default();
-        let wait_time: u64 = var_value.parse().unwrap_or(10);
+        let wait_time: Option<u64> = var_value.parse().ok();
         wait_time
     });
 
     #[derive(Debug, Clone)]
-    pub struct StoreContext<S, MetaContext = K8MetaItem>
+    pub struct StoreContext<S, MetaContext = K8MetaItem, const DEFAULT_WAIT_TIME: u64 = 10>
```
sehz, line 47: These changes add more generic arg, which makes it more complex and slows down compile time.

sehz (changes requested):
Good to consolidate the changes.  However, not sure using const is optimal approach

crates/fluvio-stream-dispatcher/src/store/mod.rs
morenol: I added that generic const, just because wait time for `fluvio` `StoreContext` is 60 seconds and for `fluvio-stream-dispatcher` is 10 seconds. Not sure which other approach we could do for that.

crates/fluvio-stream-dispatcher/src/store/mod.rs
```diff
@@ -12,11 +12,14 @@ mod context {
     use std::fmt::{Display, Debug};
     use std::time::Duration;
 
+    use async_rwlock::RwLockReadGuard;
     use fluvio_stream_model::core::MetadataItem;
+    use fluvio_stream_model::store::DualEpochMap;
```
sehz, line 17: let's clean up import order

sehz (changes requested):

Issue #1673, opened 2021-09-22
Cluster:  status to get current state


sehz:
`fluvio cluster status` with something like this:
```
$ fluvio cluster status
📝 Running cluster checks using profile 'rancher-desktop`
    ✅ Kubectl active cluster rancher-desktop at: https://127.0.0.1:6443 found
    ✅ SC is ok
    ✅ 2/2 SPU is ok
    ✅ 10 topics with 25G used
```

PR #2824, opened 2022-11-21
[Merged by Bors] - cli: Add `fluvio cluster status`

crajcan:
### Behavior

Adds `fluvio cluster status` to the CLI

Tries to connect to the SC, reports "none" if unsuccessful, describes where (local/k8s/cloud) cluster is deployed if successful. If the SC is up, it then will report if the SPUs are down, or if they are all empty.

### Comment

I'm not sure if this was still wanted but it looked like a fun first issue. It's also possible I misinterpreted the original request. Specifically, I'm not sure how we could determine if the status should be "stopped (no sc, spu running but data exits)", as I'm not sure how to determine if there is data in the spus if the connection to the sc fails.

I would love some feedback and would be happy to make modifications!

https://github.com/infinyon/fluvio/issues/1673


sehz:
Running changes show following: 
```
$  flvd cluster status
Cluster Running on local k8s
spus have no data
```

sehz:
The response should model something like `fluvio cluster start.`  So something like this would be good:

```
$ fluvio cluster status
📝 Running checks using profile 'rancher-desktop`
    ✅ Kubectl active cluster rancher-desktop at: https://127.0.0.1:6443 found
    ✅ SC is ok
    ✅ 2/2 SPU is ok
    ✅ 10 topics with 25G used
```

I will update original ticket as well

sehz (changes requested):
Nice start.  Updated with better specification

crates/fluvio-cluster/src/cli/status.rs
```diff
@@ -0,0 +1,243 @@
...
+            if let Some(storage) = topic.spec.get_storage() {
+                partition_size = storage.max_partition_size.unwrap();
+            }
+
+            replications * partitions * partition_size
+        }).sum()
+    }
+
+    fn human_readable_size(size: u64) -> String {
+        if size < 1024 {
```
sehz, line 224: can you change to use https://crates.io/crates/bytesize?

crajcan:
> LGTMS. Thanks for contribution!

😄 Thanks for the reviews!

sehz:
clippy failures

PR #2828, opened 2022-11-23
[Merged by Bors] - feat: add hub private packages

digikata:


sehz:
rebase

crates/fluvio-cli/src/client/hub/list.rs
```diff
@@ -5,13 +5,14 @@ use clap::Parser;
...
-use hubutil::PackageList;
+use hubutil::PackageListMeta;
 use hubutil::http;
 
 use crate::{CliError, Result};
+use crate::client::hub::get_hub_access;
 use crate::common::OutputFormat;
 
-const API_LIST: &str = "hub/v0/list";
+const API_LIST_META: &str = "hub/v0/list_with_meta";
```
sehz, line 15: is this compatible with old CLI?

sehz (changes requested):

crates/fluvio-cli/src/client/hub/list.rs
digikata: Yes. The 0.10.1 and older fluvio hub list uses hub/v0/list - so it will continue to operate until a user upgrades the cli 0.10.2, then the cli will access list_with_meta.
list_with_meta was already being used by the ui and the changes are backward compatible with the current ui (which works but doesn't display private / public field). The updated ui will parse and display the visiblity field as private or public.

crates/fluvio-controlplane-metadata/src/smartmodule/package.rs
```diff
@@ -65,6 +65,11 @@ pub struct SmartModulePackage {
     pub api_version: FluvioSemVersion,
     pub description: Option<String>,
     pub license: Option<String>,
+    #[cfg_attr(
+        feature = "use_serde",
+        serde(default = "SmartModulePackage::visibility_if_missing")
+    )]
+    pub visibility: SmartModuleVisibility,
```
sehz, line 72: need to add min_version tag in order to be backward compatible

crates/fluvio-controlplane-metadata/src/smartmodule/package.rs
```diff
@@ -97,6 +106,18 @@ pub enum SmartModuleKeyError {
...
 }
 
+#[derive(Debug, Default, Clone, PartialEq, Eq, Encoder, Decoder)]
+#[cfg_attr(
+    feature = "use_serde",
+    derive(serde::Serialize, serde::Deserialize),
+    serde(rename_all = "lowercase")
+)]
+pub enum SmartModuleVisibility {
+    #[default]
```
sehz, line 116: may need customer encoder/decoder to compatible which assumes public as default

sehz (dismissed):
verify compatibility with previous version of fluvio client and CLI

crates/fluvio-controlplane-metadata/src/smartmodule/package.rs
digikata: Since there aren't any customer public packages yet other than infinyon and our our own demos - it seems better to default to private as requested in the Product RFD. All we need to do is re-publish the public infinyon packages.

sehz: K

crates/fluvio-controlplane-metadata/src/smartmodule/package.rs
```diff
@@ -242,6 +265,29 @@ impl Decoder for FluvioSemVersion {
...
 }
 
+impl std::fmt::Display for SmartModuleVisibility {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::result::Result<(), std::fmt::Error> {
+        let lbl = match self {
+            Self::Private => "private",
+            Self::Public => "public",
+        };
+        write!(f, "{}", lbl)?;
+        Ok(())
```
EstebanBorai, line 275: I think we can do:

```suggestion
        write!(f, "{}", lbl)
```

Here instead, wdyt?

crates/fluvio-hub-util/src/packagemeta.rs
```diff
@@ -310,6 +344,18 @@ fn builds_obj_key_from_package_name() {
...
 
+// this end up mostly for any cli tools printing out the status
+impl std::fmt::Display for PkgVisibility {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::result::Result<(), std::fmt::Error> {
+        let lbl = match self {
+            PkgVisibility::Private => "private",
+            PkgVisibility::Public => "public",
+        };
+        write!(f, "{}", lbl)?;
+        Ok(())
```
EstebanBorai, line 355: I think we can save ourselves the `Ok(())` and the `?` here by taking advantage of `write!` return types.

```suggestion
        write!(f, "{}", lbl)
```

EstebanBorai (commented):
Looking good! Just a suggestion where we can save ourselves a couple lines!

digikata:
> verify compatibility with previous version of fluvio client and CLI

Checked with 0.10.1 fluvio and smdk.  They work with the dev-hub running the private packages fields from smartmodule package struct.

crates/fluvio-controlplane-metadata/src/smartmodule/package.rs
digikata: good catch! I agree

EstebanBorai (approved):
LGTM!

PR #2841, opened 2022-12-01
[Merged by Bors] - feat: introduce fluvio hub util protocol

EstebanBorai:
Splits logic from taxonomy for the Fluvio Hub Util crate by disengaging
type definitions from logic and providing a `Ext`-ish trait to extend
`PackageMeta` and other types behavior.


sehz (approved):
Assumes that this is a just separation

crates/fluvio-hub-util/src/hubaccess.rs
```diff
@@ -6,12 +6,12 @@ use surf::http::mime;
 use surf::StatusCode;
 use tracing::{debug, info};
 
+use fluvio_hub_protocol::{Result, HubUtilError};
```
digikata, line 9: The HubUtilError naming for the error types seems like it should maybe be updated to something else. Maybe just HubError?

EstebanBorai: Addressed on: [2e6191d](https://github.com/infinyon/fluvio/pull/2841/commits/2e6191d8a46f00d69256f3bd9a8ba91869e49c4c)

digikata (approved):
LGTM

PR #2849, opened 2022-12-02
[Merged by Bors] - feat: delivery semantics and isolation for fbm

davidbeesley:
1. Adds `Producer Isolation`, `Consumer Isolation`, and `Delivery Semantic` configuration options to `fbm`
2. Reexports `Isolation` in `fluvio` package so that no need to also depend on `fluvio-spu-schema`
3. Adds defaults for `BenchmarkMatrix` to reduce boilerplate for generating test configurations.


crates/fluvio-benchmark/src/benchmark_config/benchmark_matrix.rs
```diff
@@ -167,3 +240,36 @@ pub fn get_config_from_file(path: &str) -> Vec<BenchmarkMatrix> {
     let file = File::open(path).unwrap();
     vec![serde_yaml::from_reader::<_, BenchmarkMatrix>(file).unwrap()]
 }
+
+#[derive(Debug, Serialize, Deserialize, Copy, Clone, Hash, PartialEq, Eq)]
+pub enum DeliverySemanticStrategy {
+    AtMostOnce,
+    AtLeastOnceFixed,
```
sehz, line 247: Instead of flattening, it would be easier understand by creating At least Once Enum:
```
pub enum DeliverySemanticsStrategy {
    AtMostOnce
    AtLeastOnceStrategy
}

pub enum AtLeastOnceStrategy {
     Fixed,
     Exponential,
     Fibonnaci,
     Custom(RetryPolicy)
}
```

crates/fluvio-benchmark/src/benchmark_config/mod.rs
```diff
@@ -58,13 +58,10 @@ pub struct BenchmarkConfig {
     pub producer_linger: Duration,
```
sehz, line 58: Would prefer separating configuration into it's own domain:
```
     ProducerBenchmarkConfig
```

sehz (changes requested):

crates/fluvio-benchmark/src/benchmark_config/mod.rs
davidbeesley: I agree. The reason it's not already that way is that it was pretty complicated to avoid a massive nested for-loop with generating the cross-product of the dimensions in `BenchmarkMatrix::cross_iterate()`. It's not obvious yet how I would be able to accomplish this change easily without some other big tradeoffs.

crates/fluvio-benchmark/src/benchmark_config/benchmark_matrix.rs
davidbeesley: Agreed.

sehz (approved):

davidbeesley:
Bors r+

davidbeesley:
Bors r+

PR #2850, opened 2022-12-02
[Merged by Bors] - chore: emphasize flush in docs

davidbeesley:


crates/fluvio/src/lib.rs
```diff
@@ -160,6 +163,22 @@ const MINIMUM_PLATFORM_VERSION: &str = "0.9.0";
 /// # }
 /// ```
 ///
+/// # Example: Flushing
+///
+/// Fluvio batches records by default, so it's important to flush the producer before terminating.
+///
+/// ```no_run
+///     # async fn produce_records() -> Result<(), FluvioError> {
```
morenol, line 171: I think that we need to add imports here for `FluvioError` and `RecordKey`. Other than that, it looks good to me

morenol (approved):

PR #2858, opened 2022-12-07
[Merged by Bors] - feat(cdk): added local deployment

galibey:


sehz:
Here is attempt to deploy:
1) Reading config file from: /var/folders/jl/6mn6n7qd1q5cg_b868zgct1w0000gn/T/.tmprZSvoS has wrong path
2) I am expecting to spawn in the background 

```
cdk deploy local --config config-example.yaml 
(base) $> Hello, world!
Reading config file from: /var/folders/jl/6mn6n7qd1q5cg_b868zgct1w0000gn/T/.tmprZSvoS
ConnectorConfig {
    name: "my-json-test-connector",
    type_: "json-test-source",
    topic: "test-topic",
    version: "0.1.0",
    parameters: {
        "interval": String(
            "10s",
        ),
        "template": String(
            "",
        ),
    },
    secrets: {},
    producer: None,
    consumer: None,
    transforms: None,
}
```

connector/json-test-connector/src/main.rs
```diff
@@ -1,3 +1,24 @@
...
+    println!(
+        "Reading config file from: {}",
+        opts.config.to_string_lossy()
+    );
+
+    let config =
+        ConnectorConfig::from_file(opts.config.as_path()).expect("unable to read config file");
+    println!("{:#?}", config);
+}
+
```
sehz, line 19: Can you fill in the actual logic to read from the config so can test and see generalizes it?

sehz (changes requested):
see deployment comments as well

connector/json-test-connector/src/main.rs
galibey: What logic do you mean here? For test purposes, I read the config from a file and print it. Do you mean an assertion that the values are equal to what we expect?

galibey:
>     2. I am expecting to spawn in the background instead of the foreground.

It is spawned in the background, it was just inheriting stdio from the parent process.  I added an output to a log file.

connector/json-test-connector/src/main.rs
sehz: This can be done in separate PR.  Not showstopper

sehz:
Got it

sehz (approved):
LGTM

PR #2861, opened 2022-12-08
[Merged by Bors] - feat: added logic to test-json-connector

galibey:
This is more for discussion of how Rust Connector SDK should look like. Module `common` will be in the common SDK crate.

### DIFF
--- a/connector/json-test-connector/src/common.rs
+++ b/connector/json-test-connector/src/common.rs
@@ -50,8 +50,10 @@ pub(crate) async fn create_producer(
 }
 pub(crate) async fn ensure_topic_exists(config: &ConnectorConfig) -> anyhow::Result<()> {
     let admin = fluvio::FluvioAdmin::connect().await?;
-    let topics = admin.list::<TopicSpec, String>(vec![]).await?;
-    let topic_exists = topics.iter().any(|t| t.name == config.topic.clone());
+    let topics = admin
+        .list::<TopicSpec, String>(vec![config.topic.clone()])
+        .await?;
+    let topic_exists = topics.iter().any(|t| t.name.eq(&config.topic));
     if !topic_exists {
         let _ = admin
             .create(
--- a/connector/json-test-connector/src/common.rs
+++ b/connector/json-test-connector/src/common.rs
@@ -1,0 +1,93 @@
+use fluvio::{Offset, FluvioConfig, metadata::topic::TopicSpec};
+use fluvio_connector_package::config::ConnectorConfig;
+use fluvio_smartengine::transformation::TransformationConfig;
+use futures::stream::LocalBoxStream;
+use anyhow::Result;
+use async_trait::async_trait;
+
+#[async_trait]
+pub(crate) trait Source<'a, I> {
+    async fn connect(self, offset: Option<Offset>) -> Result<LocalBoxStream<'a, I>>;
+}
+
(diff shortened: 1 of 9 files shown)


connector/json-test-connector/src/common.rs
```diff
@@ -0,0 +1,93 @@
...
+
+    if let Some(chain) = create_smart_module_chain(config).await? {
+        Ok(producer.with_chain(chain)?)
+    } else {
+        Ok(producer)
+    }
+}
+pub(crate) async fn ensure_topic_exists(config: &ConnectorConfig) -> anyhow::Result<()> {
+    let admin = fluvio::FluvioAdmin::connect().await?;
+    let topics = admin.list::<TopicSpec, String>(vec![]).await?;
```
sehz, line 53: can use filter argument in the `list` API:  rather than retrieve all and filter

crates/fluvio-connector-package/src/config.rs
```diff
@@ -223,6 +223,21 @@ impl From<BTreeMap<String, String>> for ManagedConnectorParameterValue {
     }
 }
 
+impl ManagedConnectorParameterValue {
```
sehz, line 226: Let's change `ManagedConnectorParameterValue` to `ConnectorParameterValue` since CDK API will be general purpose

sehz (changes requested):
Looks good start.  Couple of minor comments

connector/json-test-connector/src/common.rs
```diff
@@ -0,0 +1,93 @@
+use fluvio::{Offset, FluvioConfig, metadata::topic::TopicSpec};
+use fluvio_connector_package::config::ConnectorConfig;
```
sehz, line 2: we should move this file to a separate crate.  maybe `fluvio-connector-base`  or `fluvio-connector-common` ?
post PR item

sehz (changes requested):

connector/json-test-connector/src/common.rs
galibey: done

crates/fluvio-connector-package/src/config.rs
galibey: done

connector/json-test-connector/src/common.rs
galibey: yes, that was the intention.  I would vote for `fluvio-connector-common`  name.

connector/json-test-connector/src/main.rs
```diff
@@ -1,9 +1,18 @@
...
 
 use clap::Parser;
+use fluvio::{TopicProducer, RecordKey};
 use fluvio_connector_package::config::ConnectorConfig;
+use futures::StreamExt;
+use anyhow::Result;
+
+use crate::{source::TestJsonSource, common::Source};
 
-fn main() {
```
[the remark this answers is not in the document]
galibey: How about `#[connector(source)]` proc macro that wraps:
```
#[connector(source)]
async fn start(config: ConnectorConfig, producer: TopicProducer)) {
    
}
```
into `fn main()`, including all preparations needed.

connector/json-test-connector/src/common.rs
sehz: sounds good

sehz (approved):
LGTM

PR #2874, opened 2022-12-13
[Merged by Bors] - Add pre-release input to release_workflow

morenol:
Only change that it does is to prevent from update `stable` tag for fluvio-cli, not sure which other things are needed. This won't prevent crates publishing workflow from running

### DIFF
--- a/.github/workflows/publish_crates.yml
+++ b/.github/workflows/publish_crates.yml
@@ -5,10 +5,7 @@ permissions:
 
 on:
   workflow_dispatch:
-  workflow_run:
-    workflows: [Release]
-    branches: [master]
-    types: [completed]
+  workflow_call:
 #env:
 #  USE_COMMIT: ${{ github.event.inputs.commit }}
 #  FORCE_RELEASE: ${{ github.event.inputs.force }}
--- a/.github/workflows/publish_crates.yml
+++ b/.github/workflows/publish_crates.yml
@@ -5,7 +5,17 @@ permissions:
 
 on:
   workflow_dispatch:
+    commit:
+      required: false
+      type: string
+      description: 'Fluvio git commit override (latest `master` by default)'
+      default: ''
   workflow_call:
+    commit:
+      required: false
+      type: string
+      description: 'Fluvio git commit override (latest `master` by default)'
+      default: ''
 #env:
 #  USE_COMMIT: ${{ github.event.inputs.commit }}
 #  FORCE_RELEASE: ${{ github.event.inputs.force }}
@@ -48,6 +58,8 @@ jobs:
           override: true
 
       - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.inputs.commit || 'master' }}
 
       - name: Run publish script
         env:
--- a/.github/workflows/publish_crates.yml
+++ b/.github/workflows/publish_crates.yml
@@ -61,7 +61,7 @@ jobs:
 
       - uses: actions/checkout@v3
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -14,6 +14,12 @@ on:
...
         default: ''
+      pre_release:
+        type: boolean
+        required: true
+        description: ''
+        default: true
+
   workflow_call:
     inputs:
       commit:
```
davidbeesley, line 25: If `workflow_call` is triggered, will `pre_release` default to `false`? Or is it set some other way?

nacardin:
`CD_Release` workflow will fail for pre-releases due to stable channel check, but now it will still run the install and upgrade test.

morenol:
> `CD_Release` workflow will fail for pre-releases due to stable channel check, but now it will still run the install and upgrade test.

why the stable channel check would be failing?

morenol:
I saw:
```
Version reported by installer: 0.10.2
Expected version: 0.10.3-dev
```

.github/workflows/release.yml
tjtelan: I would expect `pre_release` will default to not being set at all in this situation. I don't think the inputs between the different triggers behave with a global scope.

.github/workflows/publish_crates.yml
```diff
@@ -51,6 +60,8 @@ jobs:
           override: true
 
       - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.inputs.commit || 'master' }}
```
tjtelan, line 64: I'm not sure we want to default to `master`. We'd want to default to the branch that this workflow is being run against. I think that's the behavior if left empty.

```
    # The branch, tag or SHA to checkout. When checking out the repository that
    # triggered a workflow, this defaults to the reference or SHA for that event.
    # Otherwise, uses the default branch.
    ref: ''
```
https://github.com/actions/checkout/blob/755da8c3cf115ac066823e79a1e1788f8940201b/README.md?plain=1#L28-L31

tjtelan:
> > `CD_Release` workflow will fail for pre-releases due to stable channel check, but now it will still run the install and upgrade test.
> 
> why the stable channel check would be failing?

I'm not sure if this is new, but it looks like the installer check is broken.

It should try to install from the public install script in 3 mutually-exclusive ways, but it seems it is sometimes installing in 2 different ways before doing the check.

1. With version number from `VERSION`: `curl -fsS https://packages.fluvio.io/v1/install.sh | VERSION=$(cat VERSION) bash`
2. With stable tag: `curl -fsS https://packages.fluvio.io/v1/install.sh | VERSION=stable bash`
3. Default: `curl -fsS https://packages.fluvio.io/v1/install.sh | bash`

It seems that option 3 is always run, and since we aren't updating `stable`, options 2 and 3 will always fail.

.github/workflows/publish_crates.yml
nacardin: removed it

.github/workflows/release.yml
nacardin: Where is `workflow_call` for release.yml called from, should this be removed?

tjtelan (approved):
I think there's a little bit of misundestanding that will happen with `CD_Release` in the cases where we're doing pre-release. Overall, based on the most recent release, we need the capability.

The failure is catching the correct issue. The solution is not going to be the most straightforward due to how we trigger the start of `CD_Release`, since we need to skip checks only when the release job has the `pre-release` open set.

PR #2877, opened 2022-12-14
[Merged by Bors] - feat: add connector metadata visibility and binary to deployment

galibey:
Added `visibility` property to connector package metadata.

Also, added an option to specify a binary name in the `deployment` section of connector package metadata.


digikata (approved):
lgtm

PR #2881, opened 2022-12-15
[Merged by Bors] - feat: add connector common and derive crates

galibey:
Added `fluvio-connector-common` crate containing all types needed for connector development. So far, only `source` connectors are supported.

Example of usage is demonstrated in `json-test-connector`.


crates/fluvio-connector-derive/tests/mod.rs
```diff
@@ -0,0 +1,8 @@
+#[test]
+fn ui() {
+    let t = trybuild::TestCases::new();
+    t.compile_fail("tests/ui/sink.rs");
+    t.compile_fail("tests/ui/wrong_direction.rs");
+    t.compile_fail("tests/ui/wrong_input_len.rs");
+    t.compile_fail("tests/ui/not_async.rs");
```
simlay, line 7: We should add some `compile_success`'s to this. I wish there was a version of this that had the expanded rust.

crates/fluvio-connector-common/src/producer.rs
```diff
@@ -0,0 +1,41 @@
+use fluvio::FluvioConfig;
+use crate::{ConnectorConfig, Result};
+
+use crate::{ensure_topic_exists, smartmodule::smartmodule_chain_from_config};
+
+pub async fn producer_from_config(config: &ConnectorConfig) -> Result<fluvio::TopicProducer> {
+    let mut cluster_config = FluvioConfig::load()?;
+    cluster_config.client_id = Some(format!("fluvio_connector_{}", &config.name));
+
+    let fluvio = fluvio::Fluvio::connect_with_config(&cluster_config).await?;
```
morenol, line 10: I wonder if we can somehow keep this Fluvio instance accessible and shared in the connectors so we can use the get the client metrics from that. (We can get them from the TopicProducer struct but that would not be shared if we create a consumer and a producer with different Fluvio instances

morenol:
Could you add something like: https://github.com/infinyon/fluvio-connectors/blob/main/rust-connectors/common/src/monitoring.rs and see how we can start that in the example connector:

Example:

https://github.com/infinyon/fluvio-connectors/blob/a4bcd8139081025b48378a61dfd5ad0684d99f97/rust-connectors/sources/http/src/bin/main.rs#L63

crates/fluvio-connector-derive/src/generator.rs
```diff
@@ -0,0 +1,64 @@
...
+            }
+        }
+    }
+
+    #user_code
+    }
+}
+
+fn generate_sink(_func: &ConnectorFn) -> TokenStream {
+    unimplemented!()
```
simlay, line 63: I think having `todo!()` and a link to a github issue in it would be useful.

crates/fluvio-connector-derive/src/generator.rs
```diff
@@ -0,0 +1,64 @@
...
+                Some(config) => Self {config},
+                None => {
+                    eprintln!("error: The following required arguments were not provided:\n  --config <PATH>");
+                    ::std::process::exit(1)
+                }
+            }
+        }
+    }
+
+    #user_code
```
simlay, line 58: This indentation seems odd to me.

crates/fluvio-connector-common/src/producer.rs
galibey: I added a return Fluvio client along with TopicProducer and added taking metrics from it.

crates/fluvio-connector-derive/src/generator.rs
galibey: Done

crates/fluvio-connector-derive/src/generator.rs
galibey: Fixed

crates/fluvio-connector-derive/tests/mod.rs
galibey: `compile_success` beside the compilation also runs the `main` method that, in our case, won't succeed due to required input parameters. I assume that is why `https://github.com/infinyon/trybuild/commits/check_option` exists. But it's already outdated. With that version, the code doesn't compile. We need to rebase our fork, but it might affect other tests, so it should be done separately. I will try to do this on the next iteration when I work on supporting sink connectors here.

The expanded version is too verbose and will often change to be stored somewhere static.

crates/fluvio-connector-common/src/producer.rs
```diff
@@ -0,0 +1,41 @@
+use fluvio::{FluvioConfig, Fluvio, TopicProducer, TopicProducerConfigBuilder};
+use crate::{ConnectorConfig, Result};
+
+use crate::{ensure_topic_exists, smartmodule::smartmodule_chain_from_config};
+
+pub async fn producer_from_config(config: &ConnectorConfig) -> Result<(Fluvio, TopicProducer)> {
```
morenol, line 6: What do you think about instead of returning a Fluvio instance, accepting a Fluvio instance so we can reuse it if we want to create a consumer or maybe other structs?

galibey: Fluvio instance should be created using the info from ConnectorConfig. Right now, it is only `client_id,` but potentially, it can be more.  Here we control this process. 

I don't anticipate there will be connectors that will need to create both producers and consumers.

galibey: But if there will be such connectors, we will create another methods convenient for them

morenol (approved):

sehz (approved):
I will leave post-PR comments.  good enough

Issue #2509, opened 2022-07-27
[Bug]: SPU fails with "not enough for records"


nacardin:
**What happened**
No response from SPU when producing

**Describe the setup**
- Fluvio Cloud 0.9.31

**How to reproduce it (as minimally and precisely as possible)**
Not sure

**Log output**
SPU:
```
2022-07-27T00:01:34.995869Z ERROR sc_dispatch_loop{socket=14}:handle_update_replica_request{sc_sink=fd(14)}: fluvio_spu::control_plane::dispatcher: error storage Log validation error: not enough for records
2022-07-27T00:01:34.995896Z ERROR sc_dispatch_loop{socket=14}:handle_update_replica_request{sc_sink=fd(14)}: fluvio_spu::control_plane::dispatcher: error storage Log validation error: not enough for records
2022-07-27T00:01:34.995902Z ERROR sc_dispatch_loop{socket=14}:handle_update_replica_request{sc_sink=fd(14)}: fluvio_spu::control_plane::dispatcher: error storage Log validation error: not enough for records
2022-07-27T00:01:34.995907Z ERROR sc_dispatch_loop{socket=14}:handle_update_replica_request{sc_sink=fd(14)}: fluvio_spu::control_plane::dispatcher: error storage Log validation error: not enough for records
```

sehz:
@galibey Can you look into this?

digikata:
fixed w/ release of 0.17.3

PR #2882, opened 2022-12-15
fix: segment validation index bug

nacardin:
Attempting to fix https://github.com/infinyon/fluvio/issues/2509

If batch or index is corrupt, uses the previous batch as latest.


nacardin:
fixed by https://github.com/infinyon/fluvio/pull/2909

PR #2883, opened 2022-12-15
[Merged by Bors] - refactor: fix clippy issues from rust 1.66

morenol:


EstebanBorai:
Is this time of the month again... 😄

release-tools/check-crate-version/toml-diff/src/lib.rs
```diff
@@ -83,6 +83,7 @@ impl<'a> TomlDiff<'a> {
                     let mut a_pairs_it = a_pairs.into_iter().peekable();
                     let mut b_pairs_it = b_pairs.into_iter().peekable();
 
+                    #[allow(clippy::needless_borrowed_reference)]
```
nacardin, line 86: Maybe this is easier to read
```
                    while let (Some(a), Some(b)) =
                        (a_pairs_it.peek(), b_pairs_it.peek())
                    {
                        let (a_key, a_val) = **a;
                        let (b_key, b_val) = **b;
```

nacardin (approved):

release-tools/check-crate-version/toml-diff/src/lib.rs
morenol: applied but with a single `*`

nacardin: I had made `into_iter` into `iter` so that's why I had to double deref. But if it works like this, that's fine.

morenol:
bors retry

PR #2885, opened 2022-12-16
[Merged by Bors] - build: Fix unused dependencies check

morenol:


sehz (approved):

galibey:
Those were not unused dependencies. `cdk` crate doesn't compile without the removed dependency, `fluvio-connector-derive` tests fail because the tests needed those dev dependencies to compile.

@morenol How is the check being performed precisely? Is it a cargo plugin?

morenol:
@galibey They run with `cargo +nigthly udeps` https://github.com/est31/cargo-udeps I guess that we need to use `[package.metadata.cargo-udeps.ignore]` to make that command succeeds, I wonder why CI is not failing after this change

galibey:
Created https://github.com/infinyon/fluvio/pull/2904 as follow-up

PR #2907, opened 2022-12-28
[Merged by Bors] - chore: update version to pre-release 2

EstebanBorai:


nacardin (approved):

VERSION
```diff
@@ -1 +1 @@
-0.10.4-dev-1
\ No newline at end of file
+0.10.4-dev-2
```
morenol, line 1: I think that we need to strip the `\n`

morenol: but not sure if that issue was fixed

nacardin: we should keep the newline

nacardin: we should keep the newline

nacardin (changes requested):

VERSION
morenol: In the past it had compilation errors when we added it, if it works now with the newline I am ok with it

nacardin (approved):

morenol (approved):

PR #2910, opened 2022-12-29
[Merged by Bors] - feat: support custom configs in connector SDK

galibey:
`connector` macro will support any custom configs for connectors.

It is done in the following way:
1. Config file is read and parsed to `serde_yaml::Value`.
2. `ConnectorConfig` is parsed from the yaml value and used to initialize Producer/Consumer.
3. User-defined config is parsed from the yaml value and passed to the user-defined function.

Validation on the deployer side is done only for `ConnectorConfig` type.

Each connector will be responsible for the validation of its own configs.

### DIFF
--- a/connector/json-test-connector/Connector.toml
+++ b/connector/json-test-connector/Connector.toml
@@ -15,8 +15,3 @@ source = true
 image = "infinyon/fluvio-connect-json-test-source:0.1.0"
 
 
-
-[[params]]
-name = "template"
-description = "JSON template"
-type = "string"
--- a/connector/json-test-connector/Connector.toml
+++ b/connector/json-test-connector/Connector.toml
@@ -15,3 +15,16 @@ source = true
 image = "infinyon/fluvio-connect-json-test-source:0.1.0"
 
 
+[custom]
+required = ["template", "interval"]
+
+[custom.properties.template]
+title = "template"
+description = "JSON template"
+type = "string"
+
+[custom.properties.interval]
+title = "Interval"
+description = "Interval of producing value"
+type = "integer"
+
--- a/connector/json-test-connector/src/main.rs
+++ b/connector/json-test-connector/src/main.rs
@@ -1,13 +1,14 @@
 mod source;
 
 use fluvio::{TopicProducer, RecordKey};
-use fluvio_connector_common::{Source, connector, ConnectorConfig, Result};
+use fluvio_connector_common::{Source, connector, Result};
 use futures::StreamExt;
+use serde::Deserialize;
 
 use crate::source::TestJsonSource;
 
 #[connector(source)]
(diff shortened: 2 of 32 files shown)


connector/json-test-connector/Connector.toml
```diff
@@ -15,8 +15,3 @@ source = true
 image = "infinyon/fluvio-connect-json-test-source:0.1.0"
 
 
-
```
sehz, line 18: Why is this removed? This is needed for schema

connector/json-test-connector/src/main.rs
```diff
@@ -1,13 +1,14 @@
...
-use fluvio_connector_common::{Source, connector, ConnectorConfig, Result};
+use fluvio_connector_common::{Source, connector, Result};
 use futures::StreamExt;
+use serde::Deserialize;
 
 use crate::source::TestJsonSource;
 
 #[connector(source)]
-async fn start(config: ConnectorConfig, producer: TopicProducer) -> Result<()> {
+async fn start(config: CustomConfig, producer: TopicProducer) -> Result<()> {
```
sehz, line 11: Why renaming?

galibey: This is a demonstration that the user can have its own type instead of our `ConnectorConfig` (if he wants). It gives the ability to have a custom configuration in free format in the `config.yaml` file. The config file still has to be valid against `ConnectorConfig` type, we parse and use it. However, in the custom config type, the user can use an only subset of  `ConnectorConfig`, or have properties that are not present in the `ConnectorConfig` at all.

Right now, the only way to customize connector configuration is using `parameters` section. This is a very limited and inconvenient approach, which always ends up with manual parsing. Inviting @simlay to the thread to share if it is really painful for connector developers.

This also helps with code decoupling: `ConnectorConfig` from SDK will mostly be used by code from SDK (init Producer/Consumer), and custom user-defined config will be living within the codebase of the connector and will evolve along.

connector/json-test-connector/Connector.toml
galibey: In this approach, `template` is not a parameter but a mandatory config entry, the same as a topic name, for example.

galibey: Added back specification of custom config, but now as OpenApi spec

galibey:
Added an ability to specify the custom configuration in connector package metadata. OpenApi spec is used, so complex objects can be defined as well. The validation honors this specification.

connector/json-test-connector/Connector.toml
sehz: Nice!

sehz (approved):
LGTM.  Great work!

PR #2917, opened 2023-01-06
[Merged by Bors] - feat: separate custom config from common config

galibey:
This PR changes the shape of the connector's config separating fields for custom user-defined config from `ConnectorConfig`. For example:
```yaml
meta:
  version: 0.1.0
  name: my-json-test-connector
  type: json-test-source
  topic: test-topic
json:
  interval: 10  
  template: '{"template":"test"}'  
transforms:
  - uses: infinyon/jolt@0.1.0
    with:
```
sections `meta` and `transforms` go to `ConnectorConfig`, `json` goes to custom config (the name is configurable, by default  is `custom`). If the name of the custom config is overridden, need to specify it in the package metadata also.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2488,6 +2488,7 @@ dependencies = [
  "fluvio-connector-common",
  "proc-macro2",
  "quote",
+ "serde",
  "syn",
  "trybuild 1.0.74",
 ]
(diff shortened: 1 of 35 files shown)


sehz (approved):
LGTM

PR #2920, opened 2023-01-10
[Merged by Bors] - chore: remove unneeded fluvio crate compile triggers

digikata:
In workspace, builds work fine, but any builds out of the workspace, dependent on the fluvio crate, triggered a relink step every time due to a reference to ../../VERSION.

Only emit cargo:rerun-in-changed=../../VERSION if detected in workspace.

### DIFF
--- a/crates/fluvio-cli/build.rs
+++ b/crates/fluvio-cli/build.rs
@@ -2,7 +2,12 @@ use std::process::Command;
 
 fn main() {
     // Copy VERSION file. Do not fail e.g. when built via `cargo publish`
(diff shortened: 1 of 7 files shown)


sehz (approved):

nacardin:
Also `fluvio- socket`

digikata:
Good comment! Anything that is a library crate should be updated. Crates that only build binaries I don't think will ever build out of tree in a way that matters, but I updated them all the same way for consistency.

sehz:
only `fluvial-socket` is meant to be re-used.  Other crates should be dependent of other crates

nacardin (approved):

PR #2923, opened 2023-01-11
[Merged by Bors] - feat: make openssl as option to fluvio

### LINKED ISSUE #2921: Connector builds require native libs
**What happened**
Connector cannot build on systems without `libssl-dev` or `pkg-config`. This is due to the the selected features for e.g. `fluvio-future` pulling in `openssl-sys` even though `rustls` seems possible/available. Might consider switching to it for better dev experience?

**Expected behavior**
I can build a rust based connector without needing something besides rust/cargo.

sehz:
address part of #2921.
Don't hard code `openssl` as hard dependency.   Instead, it will be turn into optional dependency.  Also introduce `rustls` as another optional dependency.

However, `rustls` doesn't work.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -150,10 +150,10 @@ jobs:
             rust: stable
             rust-target: armv7-unknown-linux-gnueabihf
             binary: fluvio
-          - os: ubuntu-latest
-            rust: stable
-            rust-target: x86_64-pc-windows-gnu
-            binary: fluvio.exe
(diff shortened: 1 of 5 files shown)


sehz:
murphy's law strikes

morenol:
shouldn't we merge https://github.com/infinyon/fluvio/pull/2924 first?

sehz:
problem with `gnu-windows` repo.  so it's unlikely #2924 will succeed.   disabled `windows` target in this PR

PR #2934, opened 2023-01-16
additional metrics for smartengine

davidbeesley:
- [x] Add max (in and out) to metric
- [x] Add unit tests

### DIFF
--- a/crates/fluvio-smartengine/src/engine.rs
+++ b/crates/fluvio-smartengine/src/engine.rs
@@ -1,3 +1,4 @@
+use std::cmp::max;
 use std::fmt::{self, Debug};
 
 use anyhow::Result;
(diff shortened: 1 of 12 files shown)


crates/fluvio-smartengine/src/engine.rs
```diff
@@ -109,21 +110,26 @@ impl SmartModuleChainInstance {
...
         if let Some((last, instances)) = self.instances.split_last_mut() {
+            metric.add_records_in(number_of_records);
             let mut next_input = input;
 
             for instance in instances {
                 // pass raw inputs to transform instance
                 // each raw input may result in multiple records
                 let output = instance.process(next_input, &mut self.store)?;
+                smartmodule_usage_out += output.successes.len() as u64;
+                smartmodule_usage_in += output.successes.len() as u64;
```
sehz, line 132: why increase this? this is output

davidbeesley: because the output of previous is the input to next

davidbeesley: notice this is line is not called for the last smartmodule.

sehz: what's definition of usage then?

davidbeesley:
```
 usage = max(total_records_in, total_records_out) where
 total_records_in = Number of records passed to each smartmodule in the chain.
 total_records_out =Number of records returned from each smartmodule in the chain.
```

sehz: Suppose we have the following records count of each stage 10 => 5 => 3 => 2,  total amount of work should be just the sum of all inputs = 10+5+3 = 18

davidbeesley: intent is to capture case where more output than input; 
For example 1 => 10.

sehz: In that case 10 => 5 => 3 => 10,  so total record processed = 10 + 5 + 10.  So the correct definition is
total usage  = sum ( usage in each stage).   usage in each stage = max(input,output)

davidbeesley:
No longer necessary.

PR #2937, opened 2023-01-17
metric-tests

davidbeesley:


crates/fluvio-smartengine/src/engine.rs
```diff
@@ -355,6 +361,9 @@ mod chaining_test {
                 &metrics,
             )
             .expect("process");
+        assert_eq!(metrics.records_in(), 3);
+        assert_eq!(metrics.records_out(), 2);
+        assert_eq!(metrics.smartmodule_usage(), 5);
```
morenol, line 366: i think that this should be 3

crates/fluvio-smartengine/src/engine.rs
```diff
@@ -306,6 +309,9 @@ mod chaining_test {
                 &metrics,
             )
             .expect("process");
+        assert_eq!(metrics.records_in(), 4);
+        assert_eq!(metrics.records_out(), 2);
+        assert_eq!(metrics.smartmodule_usage(), 6);
```
morenol, line 314: I think that this should be 4

crates/fluvio-smartengine/src/engine.rs
morenol: Actually 4 + 3, I think that it should be 7

morenol (approved):

PR #2938, opened 2023-01-17
Dave/traffic in only

davidbeesley:


sehz (approved):

crates/fluvio-protocol/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "fluvio-protocol"
 edition = "2021"
-version = "0.8.3"
```
sehz, line 4: why this need to be bump up?

sehz:
Also, make sure existing metrics is not changed, there might be downstream consequence

crates/fluvio-protocol/Cargo.toml
morenol: It is due to the changes in the other branch

davidbeesley: Yeah this is going into integration branch. We should double check when integration is moving to master that it's correct.

davidbeesley:
No longer necessary.

PR #2940, opened 2023-01-17
[Merged by Bors] - feat: enable fuel

davidbeesley:


morenol (approved):

crates/fluvio-smartengine/src/engine.rs
```diff
@@ -24,7 +24,9 @@ pub struct SmartEngine(Engine);
 #[allow(clippy::new_without_default)]
 impl SmartEngine {
     pub fn new() -> Self {
-        Self(Engine::default())
+        let mut config = wasmtime::Config::default();
+        config.consume_fuel(true);
+        Self(Engine::new(&config).expect("Config is static"))
```
sehz, line 29: should remove expect

sehz (changes requested):

crates/fluvio-smartengine/src/engine.rs
davidbeesley: This expect can be tested before deploy.

davidbeesley: As the config is static.

davidbeesley: So success of Engine::new().expect() is unit tested.

davidbeesley: the default() that was being called previously had an unwrap() inside anyway
```
impl Default for Engine {
    fn default() -> Engine {
        Engine::new(&Config::default()).unwrap()
    }
}
```

davidbeesley: Yes but it can't fail.. it's a static configuration

davidbeesley: Which is unit tested

sehz: Fair enough

sehz (approved):

PR #2942, opened 2023-01-17
[Merged by Bors] - feat: Add Hub support for installing binary packages

### LINKED ISSUE #2894: Install Fluvio CLI extensions from Hub
This is related to the [SmartModule Hub](https://www.fluvio.io/smartmodules/hub/overview/) 

We want to migrate to install CLI extensions from the Hub. This issue should only cover adding in the functionality. The functionality should be introduced with a CLI flag, and not interrupt the current CLI flow.

Migration of the installation workflows require modifying installation scripts and deployment locations. That dimension of the problem is not covered here.

tjtelan:
Resolves #2894 

Adds `--hub` feature flag to `fluvio install`. CLI fetches binaries from the hub instead of packages.fluvio.io

example command installs executable binary `sample-bin-script` to `~/.fluvio/bin/`

```shell
fluvio install sample-bin-script --hub
```

```shell
% sample-bin-script
hello world
```

### DIFF
--- a/crates/fluvio-cli/src/install/plugins.rs
+++ b/crates/fluvio-cli/src/install/plugins.rs
@@ -17,9 +17,8 @@ use crate::install::update::{
 use fluvio_index::{PackageId, HttpAgent, MaybeVersion};
 use fluvio_channel::{LATEST_CHANNEL_NAME, FLUVIO_RELEASE_CHANNEL};
 use fluvio_hub_util as hubutil;
-use hubutil::http;
+use hubutil::{http, HubAccess, INFINYON_HUB_REMOTE, FLUVIO_HUB_PROFILE_ENV};
 use hubutil::http::StatusCode;
-use hubutil::HubAccess;
 pub const HUB_API_BPKG_AUTH: &str = "hub/v0/bpkg-auth"; // copied from hub-tool
 
 #[derive(Parser, Debug)]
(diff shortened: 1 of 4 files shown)


sehz (approved):

crates/fluvio-cli/src/install/plugins.rs
```diff
@@ -28,50 +34,100 @@ pub struct InstallOpt {
...
-                    install_println(format!(
-                        "❕ Package '{}' is not available for target {}, skipping",
-                        self.package.name(),
-                        target
-                    ));
-                    install_println("❕ Consider filing an issue to add support for this platform using the link below! 👇");
-                    install_println(format!(
-                            "❕   https://github.com/infinyon/fluvio/issues/new?title=Support+fluvio-cloud+on+target+{}",
+            let access =
+                HubAccess::default_load(&Some("https://hub-dev.infinyon.cloud".to_string()))
```
digikata, line 72: If this line does a `default_load(None)` then the regular set of hub remote defaults and overrides will apply allowing prod, or dev access with some config.

digikata (changes requested):

crates/fluvio-cli/src/install/plugins.rs
tjtelan: Now there are 2 ways to opt-out of using dev:
1. If env vars `INFINYON_HUB_REMOTE` or `FLUVIO_HUB_PROFILE` are set
2. If `--use-hub-defaults` is used on CLI

Like this
```shell
fluvio install sample-bin-script --hub --use-hub-defaults
```

(When we are ready to make this the default install path, we won't need this CLI flag anymore)

digikata (approved):

PR #2968, opened 2023-02-01
[Merged by Bors] - feat(cdk): enhance deploy options

### LINKED ISSUE #2964: CDK: enhance deploy options
We only can deploy connectors with CDK.   We should have way to shutdown, list as well.
By default, we only support local, this flag can be omitted.

Start:
```
$ cdk deploy start <config_file>
connector cat-fact is deployed
```

Terminate:
```
$ cdk deploy shutdown cat-facts

connector cat-facts is shutdown
```

List:
```
$ cdk deploy list

Connector    status   
cat-facts     Running

```

galibey:
Added `cdk deploy start` for starting connectors, `cdk deploy shutdown` for stopping, `cdk deploy list` for listing and `cdk deploy log` for printing logs of connectors.

Closes #2964

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -827,7 +827,7 @@ dependencies = [
  "serde",
  "tempfile",
  "thiserror",
- "toml",
+ "toml 0.5.11",
  "walkdir",
 ]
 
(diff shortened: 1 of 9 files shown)


sehz (approved):
LGTM.  Will have more feedback after testing it

PR #2970, opened 2023-02-02
[Merged by Bors] - feat: Update fluvio profile export to have the same format that `FluvioConfig`

morenol:
With this change, we can run `fluvio profile export` > profile.json

And then use the generated json to create client in microservices/lambda services, etc

### DIFF
--- a/crates/fluvio-cli/src/profile/export.rs
+++ b/crates/fluvio-cli/src/profile/export.rs
@@ -1,10 +1,9 @@
 use std::sync::Arc;
 
 use clap::Parser;
-use serde::Serialize;
 use anyhow::{anyhow, Result};
 
-use fluvio::config::{ConfigFile, TlsPolicy, TlsConfig};
+use fluvio::config::{ConfigFile};
 use fluvio_extension_common::Terminal;
 use fluvio_extension_common::output::OutputType;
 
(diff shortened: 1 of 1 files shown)


sehz:
how this will work existing profile?

morenol:
I think that this will work with any existing profile, the only difference is in the output of `fluvio profile export`. Not sure why we had that ProfileExport wrapper other than maybe the camelCase change? @nacardin

morenol:
Maybe is for the  TlsConfig::Files case?

```rust
                    TlsConfig::Files(_) => {
                        return Err(anyhow!("Cluster {cluster_name} uses externals TLS certs. Only inline TLS certs are supported."))
                    }
```

nacardin:
> I think that this will work with any existing profile, the only difference is in the output of `fluvio profile export`. Not sure why we had that ProfileExport wrapper other than maybe the camelCase change? @nacardin

This was done to support the swift client. I think we ran into issues with ffi at the time with `swift_bridge`.

nacardin (approved):
Looks good! It will break the workflow for the swift client but we can try to update it to use the new format or if not possible, provide a tool as part of it to do the conversion.

morenol:
bors retry

sehz:
CI failure for Windows:
```
error[E0433]: failed to resolve: use of undeclared crate or module `fluvio_cluster`
 --> crates/fluvio-cli/src/profile/export.rs:7:5
  |
7 | use fluvio_cluster::fluvio_config::TlsPolicy::Verified;
  |     ^^^^^^^^^^^^^^ use of undeclared crate or module `fluvio_cluster`
```

morenol:
> CI failure for Windows:
> 
> ```
> error[E0433]: failed to resolve: use of undeclared crate or module `fluvio_cluster`
>  --> crates/fluvio-cli/src/profile/export.rs:7:5
>   |
> 7 | use fluvio_cluster::fluvio_config::TlsPolicy::Verified;
>   |     ^^^^^^^^^^^^^^ use of undeclared crate or module `fluvio_cluster`
> ```

Fixed in last commit, I was doing an import using a crate that is not installed when building for windows

Issue #2925, opened 2023-01-13
CI: smdk smoke test is very slow


sehz:
This step is very slow.  It takes upwards of 10-20m to execute.  It frequently cause CI failure
```
- name: Run SMDK smoke tests
        timeout-minutes: 20
        if: matrix.cli == 'smdk'
        run: make SMDK_BIN=~/bin/smdk cli-smdk-smoke
```

sehz:
fixed

PR #2972, opened 2023-02-07
[Merged by Bors] - CI: Optimize smdk Smoke Test

crajcan:
https://github.com/infinyon/fluvio/issues/2925

The majority of the run time was spent repeatedly building the same dependencies for each Smart Module. This solution generates each Smart Module within a Cargo workspace, which allows them to share dependencies after building once.

An _alternative solution_ would be to develop a scheme for Github Workflows to cache the dependencies for each Smart Module. The Github Workflows caching solution may be more conventional but has its drawbacks:

- would couple the test suite to Github Workflows
- test would still run slow in local dev environments
- test would still take 10-20 minutes on cache miss


tjtelan:
bors try

tjtelan (approved):
Changes look good. Thanks for the contribution!

bors r+

PR #2975, opened 2023-02-09
[Merged by Bors] - feat: Retry monitoring if there is a failure

morenol:
Also add error logs if there is an issue.

Right now, 
```
    while let Some(stream) = incoming.next().await {
        let mut stream = stream?;
```

could finish if there is an error like broken pipe,

```
2023-02-09T17:56:39.649123Z ERROR fluvio_spu::monitoring: error running monitoring: Broken pipe (os error 32)
```

This adds changes to retry to create the unix socket if there is a failure

### DIFF
--- a/crates/fluvio-connector-common/src/monitoring.rs
+++ b/crates/fluvio-connector-common/src/monitoring.rs
@@ -44,29 +44,37 @@ async fn start_monitoring(metrics: Arc<ConnectorMetrics>) -> Result<(), IoError>
         }
     };
 
-    // check if file exists
-    if let Ok(_metadata) = std::fs::metadata(&metric_out_path) {
-        info!("metric file already exists, deleting: {}", metric_out_path);
-        match std::fs::remove_file(&metric_out_path) {
-            Ok(_) => {}
-            Err(err) => {
-                println!("error deleting metric file: {err}");
-                return Err(err);
+    loop {
+        // check if file exists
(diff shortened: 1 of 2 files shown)


crates/fluvio-connector-common/src/monitoring.rs
```diff
@@ -44,29 +44,37 @@ async fn start_monitoring(metrics: Arc<ConnectorMetrics>) -> Result<(), IoError>
...
 
-    // check if file exists
-    if let Ok(_metadata) = std::fs::metadata(&metric_out_path) {
-        info!("metric file already exists, deleting: {}", metric_out_path);
-        match std::fs::remove_file(&metric_out_path) {
-            Ok(_) => {}
-            Err(err) => {
-                println!("error deleting metric file: {err}");
-                return Err(err);
+    loop {
```
galibey, line 47: This is a very risky construction. Any mistake can lead to an endless loop, which will eat all CPU. Having here counted loop or some delay after an iteration could reduce risks.

morenol: I have a fixed 5 seconds sleep after each loop, but right, I will change to a Exponential backoff approach

galibey: I didn't notice one removed curly bracket :). I think fixed delays are enough here.

galibey (approved):
LGTM

sehz (approved):

PR #3013, opened 2023-02-24
[Merged by Bors] - Remove config from cdk test

### LINKED ISSUE #3011: remove `--config` from CDK test
CDK test should use default `sample-config.yaml` if it is not specified.

```
$ cdk test --config <option_config_path>
```

simlay:
Closes #3011.

### DIFF
--- a/crates/cdk/src/test.rs
+++ b/crates/cdk/src/test.rs
@@ -15,7 +15,7 @@ pub struct TestCmd {
     package: PackageCmd,
 
     /// Path to configuration file in YAML format
-    #[clap(short, long, value_name = "PATH", default_value = "sample-param.yaml")]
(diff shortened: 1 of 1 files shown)


crates/cdk/src/test.rs
```diff
@@ -15,7 +15,7 @@ pub struct TestCmd {
     package: PackageCmd,
 
     /// Path to configuration file in YAML format
-    #[clap(short, long, value_name = "PATH")]
+    #[clap(short, long, value_name = "PATH", default_value = "sample-param.yaml")]
```
ajhunyady, line 18: `param` => `params` to be consistent with the label used in `Connector.toml`

sehz: rename this to `sample-config.yaml`.  This PR needs to be merged after `generate` otherwise there will be no default

galibey (approved):

Issue #2988, opened 2023-02-16
feat: apply SmartModule  for producer


sehz:
Currently, SmartModule transformations can be applied to consumers.   It should apply to producers as well.  This allows more flexible data transformation from CLI.

The transformation runs in the SPU

crajcan:
Is this requesting the client-side transformation described [here](https://github.com/infinyon/fluvio-website/blob/229a54317c240522bff12544da74ca14af94186a/content/docs/concepts/transformations-chain.md?plain=1#L8-L9)?

Or is this requesting that the SPU handle the transformation before committing the data?

crajcan:
A couple questions: 

**Aggregate SmartModules**

Should this feature support aggregate SmartModules? I can think of a few different ways they could be applied:

1. Apply the SmartModule to each batch then commit the accumulator. This would be useful if you wanted to save space by committing the sum of a batch of integers instead of the entire batch, for example.

2. Apply the SmartModule to each record, using the last record in the partition as the initial accumulator. This would be useful if you wanted each record committed to represent a cumulative sum. 

3. Apply the SmartModule to each record, but rely on the producer to provide the initial accumulator.

**Compression**

Should SmartModules be applied when a compression algorithm is specified? Batches would have to be decompressed, transformed, and then compressed again before committing.

sehz:
Nice progress on PR.   Answer to your questions:
```
Aggregation:  I think so, all SmartModule API should be used
Compression:  Yes, it should work in compression cases as well but that is transparent to SmartModule so don't believe there is additional work needed. 
```

crajcan:
[This](https://github.com/infinyon/fluvio/pull/3014) is ready for review now. I'm not sure anyone saw it because it's pretty far down the list.

No rush at all though!

sehz:
Very nice work.  This is tremendous achievements

sehz:
Just to set context.  All transformations happen in the SPU (except connectors) now.   There is option to apply transformation in the fluvio client but it is not enabled by default.

ajhunyady:
There is value to both flavors:
1. Apply to SPU (as done by this PR) when you want to share producer transformations across all clients. Ideally, this type of transformations should be applied to topic to remove ambiguity, however this PR is valuable as is.
2. Apply to Producer client, for edge processing, to filter-out irrelevant data at the source.

sehz:
Done

PR #3014, opened 2023-02-25
[Merged by Bors] - SPU: Apply SmartModule Transformation For Producer

crajcan:
https://github.com/infinyon/fluvio/issues/2988


tests/cli/fluvio_smoke_tests/producer-smartmodule.bats
```diff
@@ -0,0 +1,361 @@
...
+# this should be done before running test
+setup_file() {
+    # Compile the smartmodule examples
+#    pushd "$BATS_TEST_DIRNAME/../../.." && make build_smartmodules && popd
+    SMARTMODULE_BUILD_DIR="$BATS_TEST_DIRNAME/../../../smartmodule/examples/target/wasm32-unknown-unknown/release"
+    export SMARTMODULE_BUILD_DIR
+
+}
+
+@test "invoke map smartmodule in producer by path" {
```
morenol, line 21: In our CI, we run these tests using `stable` and `latest` build for SPU and CLI. I think that we should skip this test for now while we are on `stable` for either CLI and `SPU`.

In PR #3092, this was done to fix that. https://github.com/infinyon/fluvio/pull/3092/commits/35bef9fe30dff63b2d68d03c50995bc4ae25f0a9 

Maybe we can apply something similar for this here. (But we would need to check that either CLI and SPU is on `latest` to run the test.

This will be temporal until we release a new stable version with support to this

crajcan: I see. I added another commit which I think will do this.

morenol: Perfect!

crajcan:
I think the next set of failures are due to the DEV spu trying to decode smartmodule invocations that the STABLE cli doesn't send.

I'll have to tinker with the versioning so that the new spu version can support older cli versions.

crates/fluvio-spu-schema/src/produce/request.rs
```diff
@@ -36,6 +37,10 @@ pub struct ProduceRequest<R> {
 
     /// Each topic to produce to.
     pub topics: Vec<TopicProduceData<R>>,
+
+    #[fluvio(min_version = 7)]
```
morenol, line 41: I think that you have to update this to 8 version, and then  bump `DEFAULT_API_VERSION` and `MAX_API_VERSION` in lines 53-54 from this file

```suggestion
    #[fluvio(min_version = 8)]
```

morenol: Maybe we can use a `const PRODUCER_TRANSFORMATION_API = 8` and use it here and in all the other places where that is used

morenol:
> I think the next set of failures are due to the DEV spu trying to decode smartmodule invocations that the STABLE cli doesn't send.
> 
> I'll have to tinker with the versioning so that the new spu version can support older cli versions.

I think that this comment will help you https://github.com/infinyon/fluvio/pull/3014#discussion_r1149853182

crates/fluvio-spu-schema/src/produce/request.rs
```diff
@@ -46,8 +53,8 @@ where
     const API_KEY: u16 = 0;
 
     const MIN_API_VERSION: i16 = 0;
-    const MAX_API_VERSION: i16 = 7;
-    const DEFAULT_API_VERSION: i16 = 7;
+    const MAX_API_VERSION: i16 = 8;
```
galibey, line 56: It's better to use here the newly introduced const `PRODUCER_TRANSFORMATION_API_VERSION` to prevent mistakes in the future.

galibey (changes requested):
Great job! Thanks for your contribution. Just a couple of minor fixes that are needed to be done before the merge.

.github/workflows/ci.yml
```diff
@@ -994,7 +994,10 @@ jobs:
       - name: Start cluster for stable
         if: matrix.cluster_version == 'stable'
         timeout-minutes: 10
-        run: fluvio cluster start
+        run: |
+          FLUVIO_CLUSTER_RELEASE_CHANNEL=stable
```
sehz, line 998: @tjtelan can you review this?

crates/fluvio-spu/src/smartengine/produce_batch.rs
```diff
@@ -0,0 +1,86 @@
...
+    pub fn new(batches: &'a Vec<Batch<RawRecords>>) -> Self {
+        Self {
+            batches,
+            index: 0,
+            len: batches.len(),
+        }
+    }
+}
+
+impl<'a> Iterator for ProduceBatchIterator<'a> {
```
sehz, line 46: can you add unit test for this?

sehz (changes requested):
Looks good. Just have a minor request

crates/fluvio-spu/src/smartengine/produce_batch.rs
crajcan: Sure!

.github/workflows/ci.yml
sehz: straightforward

crates/fluvio-protocol/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "fluvio-protocol"
 edition = "2021"
-version = "0.9.0"
+version = "0.10.0"
```
sehz, line 4: Don't think this need major bump since we are only adding new API?

crates/fluvio-spu/src/smartengine/produce_batch.rs
sehz: Nice. thanks

sehz (changes requested):
Everything looks ok except need to bump major version.  Can you check if they are really needed or just need patch version upgrade if just adding fields or API

crates/fluvio-protocol/Cargo.toml
crajcan: My understanding is that what I did here is a minor bump.

I think you are right though. The SemVer docs say once the major version is >0, API additions require a minor version bump.

But since the major version is still 0, I think it is still subjective. I can change these all to increment the patch value instead, if you like.

sehz: Yes, let's change to only patch version which should  minimize impact to downstream crates

sehz (approved):
Great work. LGTM

crajcan:
> Great work. LGTM

Thanks!

sehz:
Build failure in windows.   Enable windows job and fix

PR #3030, opened 2023-03-01
[Merged by Bors] - refactor: remove unnecessary bounds for encoder and decoder derive macro

sehz:
Fixed a long standing tech debt around Encoder and Decoder derived macro.
With this fix, no longer need to add bounds to struct as before:
```
pub struct Message<C>
 where
     C: Encoder + Decoder + Debug,
{
     pub header: MsgType,
     pub content: C,
 }
```

Instead, this can be simplified:
```
pub struct Message<C> {
     pub header: MsgType,
     pub content: C,
 }
```

Forcing to add structure bounds cause downstream code to deal with extras bounds which make certain implementation difficult if not outright implementation

In addition, tracing code is only added if enabled.  This should only necessary in development.  This will reduce some code and improve performance slightly.
```
#[derive(Encoder,Decoder,Default]
#[fluvio(trace)]
MyStruct {
...
}
```


morenol:
not sure if we should merge this yet, since not sure if we need to do fluvio release again  `0.10.5`

sehz:
reverting release should be only done in rare circumstances.   next time, we should do another release.

PR #3036, opened 2023-03-01
[Merged by Bors] - fix: backward compatibility for `SmartModuleKind`

### LINKED ISSUE #3035: regression: SmartModule transform
Smartmodule transform doesn't work with `0.10.4` release. 

Steps to reproduce:
1.  Build/download 0.10.5(dev) version of Fluvio
2. Connect to 0.10.4 to Fluvio Cluster
3. Try run Jolt transforms with consume. ```fluvio consume helsinki --transform jolt```

In the 0.10.4 client to 0.10.4 cluster, everything works.   However, connecting 0.10.5 to 0.10.4 doesn't work with transforms.

galibey:
Fixes #3035

### DIFF
--- a/crates/fluvio-spu-schema/src/server/smartmodule.rs
+++ b/crates/fluvio-spu-schema/src/server/smartmodule.rs
@@ -94,6 +94,15 @@ pub enum SmartModuleKind {
     },
     #[fluvio(min_version = ARRAY_MAP_WASM_API)]
     FilterMap,
+    #[fluvio(min_version = SMART_MODULE_API)]
+    #[fluvio(max_version = CHAIN_SMARTMODULE_API)]
+    Join(String),
+    #[fluvio(min_version = SMART_MODULE_API)]
+    #[fluvio(max_version = CHAIN_SMARTMODULE_API)]
+    JoinStream {
+        topic: String,
+        derivedstream: String,
+    },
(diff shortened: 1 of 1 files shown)


crates/fluvio-spu-schema/src/server/smartmodule.rs
```diff
@@ -94,6 +94,15 @@ pub enum SmartModuleKind {
     },
     #[fluvio(min_version = ARRAY_MAP_WASM_API)]
     FilterMap,
+    #[fluvio(min_version = SMART_MODULE_API)]
```
morenol, line 97: i think that we can put min_version and max_version in the same line like in https://github.com/infinyon/fluvio/blob/ed17bb1532cb493cd14d044d16ac18b15b9c4072/crates/fluvio-protocol/tests/version.rs#L8

galibey: You are right. Thanks. Fixed.

morenol:
Not a task for this PR. But maybe we should add max_version here too https://github.com/infinyon/fluvio/blob/ed17bb1532cb493cd14d044d16ac18b15b9c4072/crates/fluvio-spu-schema/src/server/stream_fetch.rs#L74

I see that there are still references to DerivedStreams on CLI

sehz:
confirm fix works.

sehz (approved):
Fix works against 0.10.4

PR #3041, opened 2023-03-02
[Merged by Bors] - feat: set max api version for derived stream

galibey:
This field is going to be removed in the next releases. For now, set max API version for compatibility in the future.


morenol (approved):

morenol:
bors retry

PR #3053, opened 2023-03-06
[Merged by Bors] - feat: use PR Summaries over Commit Comments for Benchmarks

EstebanBorai:
Updates benchmarks config to use [PR Summaries](https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/).

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -55,8 +55,8 @@ jobs:
           output-file-path: fluvio_protocol_benches.txt
           # GitHub API token to make a commit comment
           github-token: ${{ secrets.GITHUB_TOKEN }}
-          # Leave a commit comment comparing the current branch with the previous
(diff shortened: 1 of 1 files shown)


.github/workflows/benchmarks.yml
```diff
@@ -93,7 +93,7 @@ jobs:
 
       - name: Store benchmark result
         uses: benchmark-action/github-action-benchmark@v1
-        if: github.ref == 'refs/heads/master'
+        # if: github.ref == 'refs/heads/master'
```
morenol, line 96: I think that here we should keep the `if`. Only store result from pushes to master

EstebanBorai:
The `[object Object]` instances in the text gets fixed by: https://github.com/benchmark-action/github-action-benchmark/pull/155

morenol (approved):

PR #3054, opened 2023-03-06
[Merged by Bors] - feat: fail fast if socket is stale

morenol:
Right now if socket is stale, we get this error after 300 seconds:

```
    Io {
        source: Custom {
            kind: TimedOut,
            error: "Timed out: 300 secs waiting for response. API_KEY=1003, CorrelationId=27",
        },
        msg: "",
    },
```

This fail fast is a socket stale is detected.

### DIFF
--- a/crates/fluvio-socket/Cargo.toml
+++ b/crates/fluvio-socket/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio-socket"
-version = "0.14.0"
+version = "0.14.1"
 edition = "2021"
 authors = ["Fluvio Contributors <team@fluvio.io>"]
 description = "Provide TCP socket wrapper for fluvio protocol"
--- a/crates/fluvio-socket/Cargo.toml
+++ b/crates/fluvio-socket/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio-socket"
-version = "0.14.1"
+version = "0.14.0"
 edition = "2021"
 authors = ["Fluvio Contributors <team@fluvio.io>"]
 description = "Provide TCP socket wrapper for fluvio protocol"
--- a/crates/fluvio-socket/src/versioned.rs
+++ b/crates/fluvio-socket/src/versioned.rs
@@ -253,15 +253,21 @@ impl VersionedSerialSocket {
         self.socket.is_stale()
     }
 
+    fn check_liveness(&self) -> Result<(), SocketError> {
+        if self.is_stale() {
+            Err(SocketError::SocketStale)
+        } else {
+            Ok(())
+        }
+    }
+
     /// send and wait for reply serially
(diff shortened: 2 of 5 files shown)


crates/fluvio-socket/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio-socket"
-version = "0.14.0"
+version = "0.14.1"
```
sehz, line 3: no need to bump since it's already bumped

crates/fluvio-socket/src/versioned.rs
```diff
@@ -248,12 +248,21 @@ impl VersionedSerialSocket {
...
+        self.socket.is_stale()
+    }
+
     /// send and wait for reply serially
     #[instrument(level = "trace", skip(self, request))]
     pub async fn send_receive<R>(&self, request: R) -> Result<R::Response, SocketError>
     where
         R: Request + Send + Sync,
     {
+        if self.is_stale() {
```
sehz, line 262: let's make it reusable instead of repeating this.  So something like this:

```
/// check to make sure connection is ok
fn check_liveness() -> Result<(),SocketErrro> {
         if self.is_stale() {
        ....
}
```

Then this line just could be: ```
self.check_liveness()?
```

File: crates/fluvio-socket/src/versioned.rs [hunk 3]
```diff
@@ -248,12 +248,21 @@ impl VersionedSerialSocket {
...
+            return Err(SocketError::SocketStale);
```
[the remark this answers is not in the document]
sehz: Definitely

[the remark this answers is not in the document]
sehz: let's create issue

crates/fluvio-socket/src/versioned.rs
```diff
@@ -266,6 +275,10 @@ impl VersionedSerialSocket {
     where
         R: Request + Send + Sync,
     {
+        if self.is_stale() {
```
sehz, line 278: see prev comment

crates/fluvio-socket/src/versioned.rs
```diff
@@ -283,6 +296,10 @@ impl VersionedSerialSocket {
         R: Request + Send + Sync + Clone,
         I: IntoIterator<Item = Duration> + Debug + Send,
     {
+        if self.is_stale() {
```
sehz, line 299: see previous comment

sehz (changes requested):

crates/fluvio-socket/src/versioned.rs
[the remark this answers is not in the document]
morenol: Added in #3059

sehz (approved):
LGTM

PR #3075, opened 2023-03-13
feat(cdk): added target info on `cdk publish`

galibey:


digikata (approved):
LGTM, approved

Not a blocker, but a follow on change may be to add a tag into the published package metadata w/  e.g. ("target", "aarch64-unknown-linux-musl")

sehz (approved):
LGTM

PR #3077, opened 2023-03-13
[Merged by Bors] - fix: use consistent naming for connector config example file

EstebanBorai:
Currently we have `sample-config.yaml` as our connector config template filename and `config-example.yaml` for other connector examples. In order to keep it consistent, rename template to use `config-example.yaml`.


galibey:
`cdk test` uses this name as the default value. You might want to change there too: https://github.com/infinyon/fluvio/blob/f619f5932a8e20e67e33b5c52cac03bc7576e711/crates/cdk/src/test.rs#L18

EstebanBorai:
> `cdk test` uses this name as the default value. You might want to change there too: https://github.com/infinyon/fluvio/blob/f619f5932a8e20e67e33b5c52cac03bc7576e711/crates/cdk/src/test.rs#L18

Thanks so much @galibey!

EstebanBorai:
@galibey PR updated with the rename! Thanks so much for pointing it out!

digikata (approved):
LGTM

PR #3088, opened 2023-03-21
chore: update deps using `cargo update`

EstebanBorai:


EstebanBorai:
Surpassed by https://github.com/infinyon/fluvio/pull/3083

PR #3089, opened 2023-03-21
[Merged by Bors] - chore: Replace tempdir with tempfile

nacardin:
Remove dependency on remove_dir_all 0.5.3, see https://rustsec.org/advisories/RUSTSEC-2023-0018.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -98,3 +98,6 @@ codegen-units = 256
 [profile.release-lto]
 inherits = "release"
 lto = true
+
+[patch.crates-io]
+cargo-generate = { git = 'https://github.com/cargo-generate/cargo-generate.git' }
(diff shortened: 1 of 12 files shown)


EstebanBorai (approved):
LGTM!

Cargo.toml
```diff
@@ -98,3 +98,6 @@ codegen-units = 256
 [profile.release-lto]
 inherits = "release"
 lto = true
+
+[patch.crates-io]
+cargo-generate = { git = 'https://github.com/cargo-generate/cargo-generate.git' }
```
sehz, line 103: is there a reason to dep on git?  if so, lets the lockdown to actual git hash

sehz (changes requested):

Cargo.toml
nacardin: `cargo-generate`'s latest published crate has a dependency on tempfile uses `tempfile = "~3.3` which uses the vulnerable `remove_dir_all`. Will add hash

Issue #3076, opened 2023-03-13
`fluvio produce` options `--key-separator` and `--file` should not be mutually exclusive


tjtelan:
Our docs show an example of using `fluvio produce` with a key separator that saves to file.

<img width="985" alt="image" src="https://user-images.githubusercontent.com/344960/224781099-10f9e0b0-5f87-4fa9-b075-bbb9c03aa2c2.png">

But it no longer works

```
% fluvio produce --key-separator=">" -f records.txt metrics
error: the argument '--key-separator <KEY_SEPARATOR>' cannot be used with '--file <FILE>'
```

This may have been introduced a while ago during the update from Clap v2.

The expectation is the line by line output in `records.txt` should match what gets displayed in stdout

morenol:
I think that `--key` is also not working as expected when using interactive mode:

```
fluvio produce topic-name --key my-key
```

ajhunyady:
Can we add this to `ci` to ensure it won't break in the future?

PR #3092, opened 2023-03-22
[Merged by Bors] - fix: Enabled `--file` and `--key-separator` to be used together, fix `--key` handling when producing lines

ozgrakkurt:
closes https://github.com/infinyon/fluvio/issues/3076

The issue also mentions it should be added to the CI. I'm looking into doing that.

DISCLAIMER: I didn't test this yet


ozgrakkurt:
Added test

ozgrakkurt:
It seems like this is failing because of something with CI, is it not rebuilding the cli?

EstebanBorai:
> ~It seems like this is failing because of something with CI, is it not rebuilding the cli?~
> 
> nevermind I see the point

Hi @ozgrakkurt! Good work so far!

Looks like the issue is here: https://github.com/infinyon/fluvio/actions/runs/4504419808/jobs/7929176514?pr=3092#step:16:72

Due to conflicts with arguments passed on test.

ozgrakkurt:
Hey @EstebanBorai,

Old version fails with those options and part of the PR was making it so that it accepts those options.

So the test I added fails with old version but passes with new version.

Not sure how to integrate this with the test.

sehz:
In that case, run that particular test on new version

ozgrakkurt:
@sehz  I added some hack to do this in the last commit. What would be a proper way of doing this?

sehz:
That's reasonable fix!

EstebanBorai (approved):
LGTM! Nice work @ozgrakkurt!

ozgrakkurt:
@EstebanBorai can you re-check when you have time?

EstebanBorai:
@ozgrakkurt smart way to tune tests for this use case!

ozgrakkurt:
@EstebanBorai thanks!

tests/cli/fluvio_smoke_tests/e2e-file-with-separator.bats
```diff
@@ -0,0 +1,78 @@
...
+teardown_file() {
+    run timeout 15s "$FLUVIO_BIN" topic delete "$TOPIC_NAME"
+    run rm $MULTI_LINE_FILE_NAME
+}
+
+# Create topic
+@test "Create a topic for file message with separator" {
+    if [ "$FLUVIO_CLI_RELEASE_CHANNEL" == "stable" ]; then
+        skip "don't run on stable version"
+    fi
```
morenol, line 50: nice trick!

We should create an issue to update this once we release a new stable version

ozgrakkurt: thanks!

morenol (approved):

ozgrakkurt:
Hey @sehz, can you check when you have time?

sehz (approved):

PR #3097, opened 2023-03-23
[Merged by Bors] - feat(cdk): add `-p package_name` in `cdk publish`

galibey:
Currently, `cdk publish` command does not have the option to specify the package; hence must be executed from the crate's directory. It is inconvenient in CI/CD pipelines. 
This PR adds the support `-p` parameter so the command can be executed from any workspace directory.

 This includes how files listed `package-meta.yaml` are searched: now they are looked up relative to the file `package-meta.yaml` itself instead of the current directory where `cdk` runs.

### DIFF
--- a/crates/cargo-builder/src/package.rs
+++ b/crates/cargo-builder/src/package.rs
@@ -91,9 +91,7 @@ impl PackageInfo {
     }
 
     pub fn package_relative_path<P: AsRef<Path>>(&self, child: P) -> PathBuf {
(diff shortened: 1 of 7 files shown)


sehz (approved):
LGTM for CDK side.  Will defer @digikata for hubutil

digikata:
The search assumption sounds reasonable to me.

sehz:
bors retry

PR #3098, opened 2023-03-23
[Merged by Bors] - chore: disable default `fluvio` features for sdk

galibey:
`fluvio-connector-common` activates `openssl` feature for `fluvio` crate without the actual need for that, and this prevents users from choosing if they want this feature for `fluvio`.


sehz (approved):
LGTM

PR #3101, opened 2023-03-24
[Merged by Bors] - feat: common publish workflow for connectors

galibey:
This a reusable workflow for publishing connectors to the Hub.

### DIFF
--- a/.github/workflows/connector-publish.yml
+++ b/.github/workflows/connector-publish.yml
@@ -9,15 +9,16 @@ on:
         required: true
         type: string
       branch:
-        required: true
         type: string
         default: "main"
       rust-profile:
         type: string
         default: "release"
       cloud-url:
-        required: true  
         type: string
+      fail-fast:
+        type: boolean
+        default: true  
     secrets:
       CLOUD_USER_EMAIL:
         required: true
@@ -29,7 +30,7 @@ jobs:
     name: linux
     runs-on: ubuntu-latest
     strategy:
-      fail-fast: false
+      fail-fast: ${{ inputs.fail-fast }}
       matrix:
         rust-target: [aarch64-unknown-linux-musl, aarch64-unknown-linux-gnu, x86_64-unknown-linux-musl, x86_64-unknown-linux-gnu]  
     env:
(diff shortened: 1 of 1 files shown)


.github/workflows/connector-publish.yml
```diff
@@ -0,0 +1,97 @@
...
+        type: string
+      branch:
+        required: true
+        type: string
+        default: "main"
+      rust-profile:
+        type: string
+        default: "release"
+      cloud-url:
+        required: true
```
sehz, line 19: this should be optional.

.github/workflows/connector-publish.yml
```diff
@@ -0,0 +1,97 @@
...
+permissions: read-all
+
+on:
+  workflow_call:
+    inputs:
+      package-name:
+        required: true
+        type: string
+      branch:
+        required: true
```
sehz, line 12: if there is default, is there reason for making required?

.github/workflows/connector-publish.yml
```diff
@@ -0,0 +1,97 @@
...
+      CLOUD_USER_PASSWORD:
+        required: true
+
+jobs:
+  linux:
+    name: linux
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
```
sehz, line 32: this should be probably made as parameter and made as true for default

.github/workflows/connector-publish.yml
```diff
@@ -0,0 +1,97 @@
...
+        run: |
+          cross build --profile ${{ inputs.rust-profile }}  --target ${{ matrix.rust-target }} -p ${{ inputs.package-name }} 
+          cp target/${{ matrix.rust-target }}/${{ inputs.rust-profile }}/* target/${{ inputs.rust-profile }}/ | true
+      - name: Publish
+        run: |
+          cdk publish --public-yes --target ${{ matrix.rust-target }} -p ${{ inputs.package-name }} 
+
+  macos:
+    name: macos
+    runs-on: macos-latest
```
sehz, line 65: these seems to be same as previous job, why not add as another strategy config?

sehz (changes requested):

.github/workflows/connector-publish.yml
galibey: Changed

.github/workflows/connector-publish.yml
galibey: Removed `required`

.github/workflows/connector-publish.yml
galibey: Done

.github/workflows/connector-publish.yml
galibey: Not exactly the same: 1. runners and targets do not intersect, 2. cargo instead of cross, 3. need to explicitly `rustup target add`.

sehz: fair enough

sehz (approved):

PR #3104, opened 2023-03-27
chore: remove unused code in smartmodule

xxchan:
It seems these code are never used. Remove them to avoid confusion.


xxchan:
Oops it's used 🤯

PR #3110, opened 2023-03-28
[Merged by Bors] - fix: fix toolchain specification

xxchan:
Previous filename and channel name are wrong

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,2 @@
+[toolchain]
+channel = "1.68"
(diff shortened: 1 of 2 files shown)


rust-toolchain.toml
```diff
@@ -0,0 +1,2 @@
+[toolchain]
+channel = "1.68"
```
sehz, line 2: thanks for changes.  this should be `1.68.2`

sehz (changes requested):

rust-toolchain.toml
xxchan, line 2:
```suggestion
channel = "1.68.2"
```

sehz (approved):
LGTM

PR #3122, opened 2023-04-05
[Merged by Bors] - refactor(smartengine): reuse code for transforms except agg

xxchan:
All transforms except agg shares the same code, so this PR unifies them into one shared implementation. By the way, in `smartmodule-derive`, all these transforms except agg shares the same implementation via `generate_transform`, so I think it also makes sense to reuse code in smartengine.

### DIFF
--- a/crates/fluvio-smartengine/src/transforms/mod.rs
+++ b/crates/fluvio-smartengine/src/transforms/mod.rs
@@ -18,8 +18,10 @@ mod instance {
     };
 
     use super::{
-        simple_transform::SimpleTansform, aggregate::SmartModuleAggregate,
-        simple_transform::SimpleTansformKind,
+        simple_transform::{
+            SimpleTansform, FILTER_FN_NAME, MAP_FN_NAME, FILTER_MAP_FN_NAME, ARRAY_MAP_FN_NAME,
+        },
+        aggregate::SmartModuleAggregate,
     };
 
     pub(crate) fn create_transform(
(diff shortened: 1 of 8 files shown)


xxchan:
image

Hmmm, why do we need to bump version for such change? 👀

xxchan:
@sehz PTAL

crates/fluvio-smartengine/src/transforms/mod.rs
```diff
@@ -18,29 +18,32 @@ mod instance {
...
+        simple_transform::SimpleTansformKind,
     };
 
     pub(crate) fn create_transform(
         ctx: &SmartModuleInstanceContext,
         initial_data: SmartModuleInitialData,
         store: &mut impl AsContextMut,
     ) -> Result<Box<dyn DowncastableTransform>> {
-        if let Some(tr) = SmartModuleFilter::try_instantiate(ctx, store)?
+        if let Some(tr) = SimpleTansform::try_instantiate(SimpleTansformKind::Filter, ctx, store)?
```
sehz, line 30: Since these are not dynamic, maybe create static instead of enum?

xxchan: Sorry I didn't get it. :eyes: I think enum is also just a set of static values. BTW, since the `kind` is just for getting `name`, I think it doesn't differ a lot for different choices.

sehz: Enum is probably overkill and not much add value since it is only used once.  In that case it could be just a const:
```

const MAP_FN: &str = "map";
const FILTER_FN: &str = "filter";

...
SimpleTransform: try_instantiate(FILTER_FN,ctx).....

```

Otherwise good simplification

xxchan: Ok. Updated

crates/fluvio-smartengine/src/transforms/simple_transform.rs
```diff
@@ -0,0 +1,83 @@
...
+use anyhow::Result;
+
+use crate::{
+    instance::{SmartModuleInstanceContext, SmartModuleTransform},
+    state::WasmState,
+};
+
+type WasmFn = TypedFunc<(i32, i32, u32), i32>;
+
+pub const FILTER_FN_NAME: &str = "filter";
```
sehz, line 14: change to pub(crate).  They don't need to be exported

crates/fluvio-smartengine/src/transforms/simple_transform.rs
```diff
@@ -0,0 +1,83 @@
...
+pub const MAP_FN_NAME: &str = "map";
+pub const FILTER_MAP_FN_NAME: &str = "filter_map";
+pub const ARRAY_MAP_FN_NAME: &str = "array_map";
+
+pub struct SimpleTansform {
```
sehz, line 19: pub(crate)

sehz (changes requested):

crates/fluvio-smartengine/src/transforms/simple_transform.rs
xxchan: `pub(crate)` is unnecessary since `simple_transform` is a private mod.

xxchan: But I can change to that if you still prefer that style :)

sehz: make it more proof against future changes

crates/fluvio-smartengine/src/transforms/simple_transform.rs
```diff
@@ -0,0 +1,83 @@
...
+
+pub(crate) struct SimpleTansform {
+    f: WasmFn,
+    name: String,
+}
+
+impl std::fmt::Debug for SimpleTansform {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self.name.as_str() {
+            FILTER_FN_NAME => write!(f, "FilterFn"),
```
sehz, line 27: no need to distinguish fn name. just write it out

sehz (changes requested):

crates/fluvio-smartengine/src/transforms/simple_transform.rs
xxchan: Just wanted to keep it the same as before :) Updated.

sehz (approved):
LGTM.  Nice simplification

PR #3131, opened 2023-04-13
[Merged by Bors] - feat: add support to detect secret on sequences

morenol:
Also added `Default` and `Clone` impl to SecretString


galibey (approved):
LGTM

Issue #3124, opened 2023-04-07
[Bug]: CLI consumer & producers fail silently when they are disconnect from the cluster


ajhunyady:
Star 2 CLI sessions, one consumer and one producer to the cloud cluster.
Some some records then leave them idle.

In a while... (I don't know how long)... they get disconnected, but there is no indication that the connection is down.

Consumer example:
```
[6:52](https://infinyon.slack.com/archives/CN1D204C9/p1680789158903939)
 2023-04-05T22:14:47.009Z {"forks":134,"stars":1725}
⠦                                                             
⠖
```

Producer Example:
```
> test
 Ok!
> test
hello world
test
test
```

I would expect to get notified if the client fails to reconnect. 
Furthermore, the CLI should exit interactive mode on error.

galibey:
> My behavior is different, I don't get an error at the CLI

Are you sure the connectivity was lost for more than 5 minutes?

ajhunyady:
I left the computer on overnight, but since there was no message when the connection went down I don't know. All I know is that when I tried to use the producer, it didn't work. After restarting the consumer, I noticed that it didn't receive messages.

I cannot answer this question if I don't get a message when the connection is dropped.

ajhunyady:
I can disconnect the wifi for 5 minutes to see what happens.

galibey:
When you left the computer overnight, was the producer sending any data? If not, it is not relevant. The producer just didn't do anything, that's why no messages or errors. When you send a message, it initiates a new "send operation" and 5 minutes start counting from that time.

ajhunyady:
no, the producer was idle

ajhunyady:
Here is what I got from the Wifi off test:

<img width="643" alt="image" src="https://user-images.githubusercontent.com/26554085/232121383-1453636f-b5c6-40ea-ba8c-7ce1dc1030e9.png">

producer eventually errored out, consumer didn't.

galibey:
> producer eventually errored out, consumer didn't.

That is expected, when connection is up and something will be produced, consumer will recover and consume

ajhunyady:
The consumer didn't recover. I still didn't get `wifi back on`, and it's been almost 10 minutes since I resumed wifi

galibey:
Ok, it is an issue, I will try to reproduce it again.

ajhunyady:
[...]

PR #3137, opened 2023-04-14
[Merged by Bors] - feat: explicit set of errors to retry

galibey:
The current default time to wait for Fluvio Socket is 5 minutes, and until that time, if there is a network connectivity issue, it does not report an error. It seems too coarse, it doesn't align well with retries. For the producer, for example, we set a whole timeout for all retries is 5 minutes.

This PR is a proposal to decrease this value and explicitly define the set of socket errors that we want to retry.

Related #3124

### DIFF
--- a/crates/fluvio-socket/Cargo.toml
+++ b/crates/fluvio-socket/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio-socket"
-version = "0.14.0"
+version = "0.15.0"
 edition = "2021"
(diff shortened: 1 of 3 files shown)


sehz (approved):
LGTM.  check if need to bump up crate version.

sehz (commented):
Think just patch version changes is required

morenol (approved):

galibey:
I will merge it after the 0.10.7 release is finished.

galibey:
bors retry

Issue #3127, opened 2023-04-10
feature:  string interpolation for connector config


sehz:
In the connector config, we may need to  inject parameters as constant like:
```
http: 
     intervals: 10s
```

However,  these constants may change depending on context.  We need to give the ability for these parameters to be substituted until like this:

```
http:
     intervals: ${{ env.interval }}
```

It could also be used in secrets like:
```
http:
    headers:
          - "Authorization: ${{ secrets.token}}'
```

This can be anywhere the string values are defined.

ref: https://github.com/infinyon/http-connector/issues/25

morenol:
So, from the examle I see that we should resolve `secrets.token`  using the current `secrets` feature. But not so sure about `env.interval` I think that we don't have a way to inject `envs` in connectors either in cloud or locally

sehz:
done

PR #3138, opened 2023-04-14
[Merged by Bors] - feat: add support to string interpolation

morenol:
Solves #3127 

With this change we could use something like:

```
http:
  headers:
   - "Authorization: JWT ${{ secrets.JWT_TOKEN }}"
```

And that should be resolved as

   - "Authorization: JWT MYSECRET_TOKEN"

### DIFF
--- a/crates/fluvio-connector-common/src/config.rs
+++ b/crates/fluvio-connector-common/src/config.rs
@@ -1,15 +1,14 @@
-pub use fluvio_connector_package::config::ConnectorConfig;
+use std::io::Read;
+
+use anyhow::{Result, Context};
+use serde::de::DeserializeOwned;
+use serde_yaml::Value;
 use tracing::trace;
 
-use std::{path::PathBuf, fs::File};
+pub use fluvio_connector_package::config::ConnectorConfig;
 
-use serde::de::DeserializeOwned;
-use anyhow::{Result, Context};
-use serde_yaml::Value;
-
-pub fn value_from_file<P: Into<PathBuf>>(path: P) -> Result<Value> {
-    let file = File::open(path.into())?;
-    serde_yaml::from_reader(file).context("unable to parse config file into YAML")
+pub fn value_from_reader<R: Read>(reader: R) -> Result<Value> {
+    serde_yaml::from_reader(reader).context("unable to parse config file into YAML")
 }
 
 pub fn from_value<T: DeserializeOwned>(value: Value, root: Option<&str>) -> Result<T> {
--- a/crates/fluvio-connector-derive/src/generator.rs
+++ b/crates/fluvio-connector-derive/src/generator.rs
@@ -113,11 +113,15 @@ fn init_and_parse_config(config_type_path: &Path) -> TokenStream {
 
         ::fluvio_connector_common::tracing::info!("Reading config file from: {}", opts.config.to_string_lossy());
 
-        let config_value = ::fluvio_connector_common::config::value_from_file(opts.config.as_path())?;
-        ::fluvio_connector_common::tracing::trace!("{:#?}", config_value);
+        let config_str = ::std::fs::read_to_string(opts.config.as_path())?;
+        ::fluvio_connector_common::tracing::debug!(%config_str, "input config");
+
+        /// Resolve any secrets/env in the config
+        let config_str_resolved =::fluvio_connector_common::render_config_str(&config_str)?;
+
+        let config_value = ::fluvio_connector_common::config::value_from_reader(config_str_resolved.as_bytes())?;
 
         let common_config = ::fluvio_connector_common::config::ConnectorConfig::from_value(config_value.clone())?;
-        ::fluvio_connector_common::tracing::debug!("{:#?}", common_config);
 
         let user_config: #config_type_path = ::fluvio_connector_common::config::from_value(config_value, Some(#config_type_path::__config_name()))?;
 
@@ -129,6 +133,7 @@ pub(crate) fn generate_connector_config(item: &ConnectorConfigStruct) -> TokenSt
     let config_struct = item.item_struct;
     let ident = &item.item_struct.ident;
     let config_name = &item.config_name;
+
     quote! {
         #[derive(serde::Deserialize)]
         #config_struct
@@ -137,6 +142,7 @@ pub(crate) fn generate_connector_config(item: &ConnectorConfigStruct) -> TokenSt
             pub fn __config_name() -> &'static str {
                 #config_name
             }
+
         }
     }
 }
--- a/crates/fluvio-connector-package/Cargo.toml
+++ b/crates/fluvio-connector-package/Cargo.toml
@@ -14,6 +14,7 @@ default = ["toml"]
 anyhow = { workspace = true }
 bytesize = { workspace = true}
 humantime-serde = "1.1.1"
+minijinja = { git = "https://github.com/mitsuhiko/minijinja" , rev = "c4b48dc5be9de0143e06bef97618c7ed029fb155", default-features = false, features = ["custom_syntax", "fuel"] }
 openapiv3 = { git = "https://github.com/galibey/openapiv3", rev = "bdd22f046d2bc19ede257504645d31f835545222", default-features = false }
 once_cell = { workspace = true }
 serde = { workspace = true, default-features = false, features = ["derive"] }
--- a/crates/fluvio-connector-package/src/lib.rs
+++ b/crates/fluvio-connector-package/src/lib.rs
@@ -1,3 +1,6 @@
 pub mod metadata;
 pub mod config;
 pub mod secret;
+mod render;
+
+pub use render::render_config_str;
--- a/crates/fluvio-connector-package/src/render/mod.rs
+++ b/crates/fluvio-connector-package/src/render/mod.rs
@@ -1,0 +1,209 @@
+use minijinja::{Environment};
+
+use crate::{
+    secret::{self},
+};
+
+use self::{
+    context::{Context, ContextStore},
+    syntax::ConnectorTemplateSyntax,
+};
+
+mod context;
+mod syntax;
+
+/// Config renderer. This is the main entry point for rendering a config.
+/// It is responsible for resolving secrets and rendering the config.
+pub(crate) struct ConfigRenderer {
+    inner_renderer: Environment<'static>,
+    context_stores: Vec<Box<dyn ContextStore>>,
+}
+
+impl ConfigRenderer {
+    fn new(
+        inner_renderer: Environment<'static>,
+        context_stores: Vec<Box<dyn ContextStore>>,
+    ) -> Self {
+        Self {
+            inner_renderer,
+            context_stores,
+        }
+    }
+
+    fn new_with_context_stores(context_stores: Vec<Box<dyn ContextStore>>) -> anyhow::Result<Self> {
+        let mut inner_renderer = Environment::new();
+        inner_renderer.set_undefined_behavior(minijinja::UndefinedBehavior::Strict);
+        inner_renderer.set_syntax(ConnectorTemplateSyntax::default().into())?;
+        inner_renderer.set_fuel(Some(500));
+
+        Ok(Self::new(inner_renderer, context_stores))
+    }
+
+    fn new_with_default_stores() -> anyhow::Result<Self> {
+        let context_stores = Self::default_stores()?;
+
+        Self::new_with_context_stores(context_stores)
+    }
+
+    fn render_str(&self, input: &str) -> anyhow::Result<String> {
+        let context = self.build_context(input)?;
+        match self.inner_renderer.render_str(input, context) {
+            Ok(rendered) => Ok(rendered),
+            Err(_) => Err(anyhow::anyhow!("failed to render: `{}`.", input)),
+        }
+    }
+
+    fn default_stores() -> anyhow::Result<Vec<Box<dyn ContextStore>>> {
+        Ok(vec![Box::new(secret::default_secret_store()?)])
+    }
+
+    fn build_context(&self, input: &str) -> anyhow::Result<Context> {
+        let mut context = Context::default();
+        for store in self.context_stores.iter() {
+            store.add_to_context(&mut context, input)?;
+        }
+        Ok(context)
+    }
+}
+
+/// Render a config from a string.
+/// This is the main entry point for rendering a config.
+/// It is responsible for resolving secrets and rendering the config.
+pub fn render_config_str(input: &str) -> anyhow::Result<String> {
+    let renderer = ConfigRenderer::new_with_default_stores()?;
+
+    let value = renderer.render_str(input)?;
+
+    Ok(value)
+}
+
+#[cfg(test)]
+mod test {
+    use std::{io::Write};
+
+    use minijinja::value::Value;
+
+    use crate::{
+        secret::{FileSecretStore},
+    };
+
+    use super::{ConfigRenderer, context::ContextStore};
+
+    // mock secret store
+    pub struct SecretTestStore;
+
+    impl ContextStore for SecretTestStore {
+        fn extract_context_values(&self, _input: &str) -> anyhow::Result<Value> {
+            let values = [
+                ("foo", "bar"),
+                ("api_key", "my_api_key"),
+                ("interval", "10s"),
+            ]
+            .into_iter()
+            .collect();
+            Ok(values)
+        }
+
+        fn context_name(&self) -> &'static str {
+            "test_secrets"
+        }
+    }
+
+    #[test]
+    fn test_build_context() {
+        let store = Box::new(SecretTestStore);
+
+        let renderer = ConfigRenderer::new_with_context_stores(vec![store]).unwrap();
+        let context: super::context::Context = renderer.build_context("").unwrap();
+
+        assert_eq!(
+            context
+                .0
+                .get("test_secrets")
+                .unwrap()
+                .get_attr("foo")
+                .unwrap(),
+            "bar".into()
+        );
+        assert_eq!(context.0.get("test_secrets").unwrap().len().unwrap(), 3);
+    }
+
+    #[test]
+    fn test_render_str() {
+        let store = Box::new(SecretTestStore);
+        let renderer = ConfigRenderer::new_with_context_stores(vec![store]).unwrap();
+        let output = renderer
+            .render_str("hello ${{ test_secrets.foo }}")
+            .unwrap();
+        assert_eq!(output, "hello bar");
+    }
+
+    #[test]
+    fn test_render_str_undefined_secret() {
+        let store = Box::new(SecretTestStore);
+
+        let renderer = ConfigRenderer::new_with_context_stores(vec![store]).unwrap();
+        let output = renderer.render_str("hello ${{ test_secrets.undefined_var }}");
+        assert!(output.is_err());
+    }
+
+    #[test]
+    fn test_invalid_syntax() {
+        let store = Box::new(SecretTestStore);
+
+        let renderer = ConfigRenderer::new_with_context_stores(vec![store]).unwrap();
+        let output = renderer.render_str("hello ${{");
+        assert!(output.is_err());
+    }
+
+    #[test]
+    fn test_undefined_context_variable() {
+        let store = Box::new(SecretTestStore);
+
+        let renderer = ConfigRenderer::new_with_context_stores(vec![store]).unwrap();
+        let output = renderer.render_str("hello ${{ some_undefined }}");
+        assert!(output.is_err());
+    }
+
+    #[test]
+    fn test_resolve_config() {
+        use super::*;
+        let mut file = tempfile::NamedTempFile::new().expect("failed to create tmp file");
+        file.write_all(b"foo=bar\napi_key=my_api_key\ninterval=10s")
+            .expect("file to write");
+        let _ = secret::set_default_secret_store(::std::boxed::Box::new(FileSecretStore::from(
+            file.path(),
+        )));
+
+        let value_str = r#"
+            meta:
+                name: test
+                version: 0.1.0
+                topic: test
+                type: http-source
+                consumer:
+                    partition: 0
+                secrets:
+                    - name: api_key
+                    - name: interval
+            my_service:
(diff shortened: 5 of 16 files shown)


crates/fluvio-connector-package/Cargo.toml
```diff
@@ -11,11 +11,19 @@ description = "Fluvio Connector Package"
...
-serde_yaml = "0.8.18"
+anyhow = { workspace = true }
+bytesize = { workspace = true}
 humantime-serde = "1.1.1"
+lazy_static = "1.4.0"
+openapiv3 = { git = "https://github.com/galibey/openapiv3", rev = "bdd22f046d2bc19ede257504645d31f835545222", default-features = false }
+once_cell = { workspace = true }
+regex = { workspace = true }
+serde = { workspace = true, default-features = false, features = ["derive"] }
+serde_yaml = "0.8.18"
```
morenol, line 22: this should be updated to use workspace dependency (0.9.0) not doing in this PR because I think that it would require some code changes

galibey:
This implementation only covers string interpolation for `SecretString` type, right? But we need it for other types as well. At minimum other String as well.

We need to come up with a such abstraction that it will be easier later to add other sources of data, not only `secrets` but `envs`, for example.

morenol:
> This implementation only covers string interpolation for `SecretString` type, right? But we need it for other types as well. At minimum other String as well.
> 
> We need to come up with a such abstraction that it will be easier later to add other sources of data, not only `secrets` but `envs`, for example.

Yes, only for SecretString

morenol:
@galibey regarding the String case, I added a new method to the `SecretStore` trait. The method is `resolve_str` where you can pass any String and it will resolve it.

morenol:
> We need to come up with a such abstraction that it will be easier later to add other sources of data, not only `secrets` but `envs`, for example.

Maybe creating a new Wrapper type for String, were we could call SecretStore::resolve, EnvStore::resolve, and any other resolvers as needed?

galibey:
> @galibey regarding the String case, I added a new method to the `SecretStore` trait. The method is `resolve_str` where you can pass any String and it will resolve it.

I meant it works with yaml string only if its Rust counterpart is `SecretString`, not `String`. 
And other yaml data types is not covered, but according to the linked issue, they should.

morenol:
> I meant it works with yaml string only if its Rust counterpart is `SecretString`, not `String`. And other yaml data types is not covered, but according to the linked issue, they should.

The `find_secrets` function will work as expected (We will be able to detect which are the secret names), but yes, we don't automatically resolve them on connector config. Do you think that we should do an automatic `resolve()` for every string in the input?

galibey:
> Maybe creating a new Wrapper type for String, were we could call SecretStore::resolve, EnvStore::resolve, and any other resolvers as needed?
Yes, this is a possible approach. The other one could be is to parse the config data into string, replace variables and then parse the resulting string into config type.

morenol:
> 



> > Maybe creating a new Wrapper type for String, were we could call SecretStore::resolve, EnvStore::resolve, and any other resolvers as needed?
> 
> Yes, this is a possible approach. The other one could be is to parse the config data into string, replace variables and then parse the resulting string into config type.

Implemented that behavior in [f6331950a1a834b32bb9bb9db0e3425c55b99a29](https://github.com/infinyon/fluvio/pull/3138/commits/f6331950a1a834b32bb9bb9db0e3425c55b99a29)

crates/fluvio-connector-package/src/lib.rs
```diff
@@ -1,3 +1,15 @@
...
 pub mod secret;
+
+pub fn resolve_config(input: Value) -> anyhow::Result<Value> {
+    let str_input = serde_yaml::to_string(&input)?;
+    let str_result = resolve_str_secrets(&str_input)?;
+    // Here we can resolve envs, files, etc.
+
+    let result = serde_yaml::from_str(&str_result)?;
+    Ok(result)
+}
```
morenol, line 15: @galibey I think that we can add to this function all the resolvers that we need. For now I only added secret resolver

morenol:
> This implementation only covers string interpolation for `SecretString` type, right? But we need it for other types as well. At minimum other String as well.
> 
> We need to come up with a such abstraction that it will be easier later to add other sources of data, not only `secrets` but `envs`, for example.

With the last update it should work with any string in the connector config. What I don't like about it is that we don't sanitize the replace value. (It could insert new lines in config). I am thinking now in a way to solve it

morenol:
> With the last update it should work with any string in the connector config. What I don't like about it is that we don't sanitize the replace value. (It could insert new lines in config). I am thinking now in a way to solve it

Fixed by removing the serde_yaml::Value -> String step:

Previously, it was serde_yaml::Value -> String -> String replacement -> serde_yaml::Value

Now it is Serde_yaml::Value -> Secret Resolution -> serde_yaml::Value

crates/fluvio-connector-package/src/secret.rs
```diff
@@ -456,16 +595,28 @@ config:
...
+    key: value
 "#;
 
         //when
         let secrets = detect_secrets_from_str(raw_config).expect("invalid raw config string");
 
         //then
-        assert_eq!(secrets.len(), 4);
+        assert_eq!(secrets.len(), 7);
+        println!("{:?}", secrets);
```
morenol, line 613:
```suggestion
```

crates/fluvio-connector-package/src/secret.rs
```diff
@@ -75,6 +93,19 @@ impl From<Secret> for SecretString {
 
 pub trait SecretStore: Send + Sync {
     fn read(&self, name: &str) -> Result<String>;
+
+    fn resolve_str(&self, input: &str) -> Result<String> {
+        let result = SECRETS_RE
+            .replace_all(input, |caps: &regex::Captures| {
+                let secret_name = caps
+                    .get(1)
+                    .expect("should not fail since SECRETS_RE, contains a capture group")
```
galibey, line 102: Even if it is expected to never happen, I would change it for proper error propagation instead of panicking. This is a third-partly library, they can change the behavior.

galibey (approved):
LGTM!. Just a minor comment.

crates/fluvio-connector-package/src/secret.rs
```diff
@@ -5,13 +5,25 @@ use std::{
...
 use once_cell::sync::OnceCell;
+use regex::Regex;
 use serde::{Serialize, Deserialize, de::Visitor, Deserializer};
-use anyhow::{Result, anyhow};
-use serde_yaml::Value;
+use serde_yaml::{Value, Mapping, Sequence};
 
 static SECRET_STORE: OnceCell<Box<dyn SecretStore>> = OnceCell::new();
 
+lazy_static! {
```
sehz, line 17: avoid lazy_static.  should use once_cell

sehz (changes requested):

crates/fluvio-connector-package/src/secret.rs
morenol: Resolved with https://github.com/infinyon/fluvio/pull/3138/commits/c65e6ddfcabbed2bfe1a2c0adc8f47de3f933ff7

crates/fluvio-connector-package/src/secret.rs
morenol: updated

crates/fluvio-connector-common/src/config.rs
```diff
@@ -13,6 +13,11 @@ pub fn value_from_file<P: Into<PathBuf>>(path: P) -> Result<Value> {
 }
 
 pub fn from_value<T: DeserializeOwned>(value: Value, root: Option<&str>) -> Result<T> {
```
sehz, line 15: should have unit tests for these

crates/fluvio-connector-derive/src/generator.rs
```diff
@@ -119,7 +119,11 @@ fn init_and_parse_config(config_type_path: &Path) -> TokenStream {
         let common_config = ::fluvio_connector_common::config::ConnectorConfig::from_value(config_value.clone())?;
         ::fluvio_connector_common::tracing::debug!("{:#?}", common_config);
```
sehz, line 121: can we update unit test for derive macro?

crates/fluvio-connector-package/src/secret.rs
```diff
@@ -108,6 +151,12 @@ impl<T: AsRef<Path>> From<T> for FileSecretStore {
     }
 }
 
+fn get_or_init_secrets_re() -> Result<&'static Regex> {
```
sehz, line 154: need unit tests for this

sehz (changes requested):

crates/fluvio-connector-derive/src/generator.rs
morenol: not sure how we could test this from the UI tests. I will add a unit test to fluvio_connector_common::resolve_config tho

crates/fluvio-connector-package/src/secret.rs
morenol: addded a test for this in https://github.com/infinyon/fluvio/pull/3138/commits/cda51fa1e06684af9fcc0ddf2d7c0b622d0225d5

crates/fluvio-connector-common/src/config.rs
morenol: Added a test in https://github.com/infinyon/fluvio/pull/3138/commits/cda51fa1e06684af9fcc0ddf2d7c0b622d0225d5

morenol:
@sehz could you take another look?

crates/fluvio-connector-derive/src/generator.rs
sehz: Yes, let's add to fluvio_connector_common::resolve_config

crates/fluvio-connector-package/src/secret.rs
```diff
@@ -453,16 +613,35 @@ config:
...
+        assert!(secrets.get("secret7").is_some());
+    }
+
+    #[test]
+    fn test_secret_re() {
+        let regex = get_or_init_secrets_re().expect("invalid regex");
+        assert_eq!(2, regex.captures_len());
+        assert!(regex.is_match("${{ secrets.my_secret_name }}"));
+        assert!(!regex.is_match("${{ env.my_env }}"));
     }
```
sehz, line 646: we need negative test cases as well.  For example what if input like:
```
{{ xxxx
```

What if there are two values? and so forth

sehz (changes requested):

crates/fluvio-connector-derive/src/generator.rs
morenol: I added a test to it on `crates/fluvio-connector-package/src/lib.rs` fluvio_connector common` is re-exporting the function

crates/fluvio-connector-package/src/secret.rs
morenol: `{{ xxxx` will just be keep as is, since it will not match the regex. That will just be an string. If there are multiple matches they will all be replaced because we are iterating over all the matches. Will add a test for it.

crates/fluvio-connector-package/src/secret.rs
```diff
@@ -75,6 +90,34 @@ impl From<Secret> for SecretString {
...
 pub trait SecretStore: Send + Sync {
     fn read(&self, name: &str) -> Result<String>;
+
+    fn resolve_str(&self, input: &str) -> Result<String> {
+        let mut content = String::new();
+        let mut last = 0;
+
+        for cap in get_or_init_secrets_re()?.captures_iter(input) {
+            let range = cap.get(0).ok_or(anyhow!(
+                "Unreachable error, capture should have at least a match"
```
sehz, line 100: more meaningful message to API consumer

crates/fluvio-connector-package/src/secret.rs
```diff
@@ -75,6 +90,34 @@ impl From<Secret> for SecretString {
...
+            ))?;
+            content.push_str(&input[last..range.start()]);
+            let secret_name = cap
+                .get(1)
+                .ok_or(anyhow!(
+                    "Unreachable error, capture has defined a non-empty capture group"
```
sehz, line 106: same.  need to hide regex related which is only relevant to regex implementation

sehz (changes requested):

crates/fluvio-connector-package/src/secret.rs
sehz: any non conforming syntax should trigger error

morenol: That would require a more sophisticated lexer. Here my plan was just to support string interpolation for `${{ secrets.something }}`.  I can work on a lexer if that is needed

morenol: I see that there are several of them, but not sure about one generic that could let me use `${{` instead of `{{`. Checking it now

morenol:
Can you take a new look? @galibey @sehz 

I updated it to use `minijinja`. I choose `minijinja` because it supports [custom delimiters](https://github.com/mitsuhiko/minijinja/commit/78b9b87eaa68ac4c1b1c814d420a8e48e054fefc) (`${{` instead of `{{` normal jinja delimiter). In `yaml` we need to use something like `${{` because `{{` is parsed as a `Mapping` by `serde_yaml`.

I think that this implementation is more extensible now. We can add new elements to the `Context` struct and use them on `render` time. 

I thought that maybe in the future we could support something like this to decide the version of the connector?

```yaml
meta:
   type: http-source
   version: ${{ globals.HTTP_SOURCE_LATEST_VERSION }}
   ...
http:
   endpoint: https://example.com/api/v1/something?token=${{ secrets.api_token }}
```

crates/fluvio-connector-package/src/renderer.rs
```diff
@@ -0,0 +1,174 @@
...
+        let mut file = tempfile::NamedTempFile::new().expect("failed to create tmp file");
+        file.write_all(b"foo=bar\napi_key=my_api_key\ninterval=10s")
+            .expect("file to write");
+        secret::set_default_secret_store(::std::boxed::Box::new(FileSecretStore::from(
+            file.path(),
+        )))
+        .expect("failed to set default secret store");
+
+        let renderer = ConfigRenderer::new_with_default_stores().unwrap();
+        let context: crate::context::Context =
```
sehz: Why do you need ConfigRenderer here?

crates/fluvio-connector-package/src/renderer.rs
```diff
@@ -0,0 +1,174 @@
...
+
+use crate::{
+    secret::{SecretStore, detect_secrets_from_str},
+    context::Context,
+};
+
+/// Config renderer. This is the main entry point for rendering a config.
+/// It is responsible for resolving secrets and rendering the config.
+pub(crate) struct ConfigRenderer {
+    secret_store: &'static dyn SecretStore,
```
sehz, line 23: This should handle arbitrary context not just these two.  Maybe trait?

crates/fluvio-connector-package/src/context.rs
```diff
@@ -0,0 +1,11 @@
+use std::collections::HashMap;
+
+use serde::Serialize;
+
+/// Context for the template engine
+/// This is the data that is available to the template engine
+/// when it is rendering the template.
+#[derive(Serialize)]
+pub(crate) struct Context {
```
galibey, line 9: Maybe we should put this type into `renderer` module, wdyt? It is used only there and seems like impl detail specific only to renderer while occupies such a common module name as `context`.

crates/fluvio-connector-common/src/config.rs
```diff
@@ -21,5 +26,39 @@ pub fn from_value<T: DeserializeOwned>(value: Value, root: Option<&str>) -> Resu
...
+        )
+        .unwrap();
+        let config: MetaConfig = from_value(value, Some("meta")).unwrap();
+        assert_eq!(config.name, "test");
+        assert_eq!(config.topic, "test");
+        assert_eq!(
+            config
+                .consumer
+                .expect("expected some consume config")
+                .partition,
```
sehz, line 60: we also need to make sure config works in non YAML source as well

crates/fluvio-connector-package/src/context.rs
```diff
@@ -0,0 +1,11 @@
...
+    pub(crate) secrets: HashMap<String, String>,
```
sehz: Shouldn't secret be using SecretString?

sehz (changes requested):
first pass

crates/fluvio-connector-package/src/renderer.rs
morenol: because `build_context` is a method of `ConfigRenderer` that uses internal secret_store to build the context

crates/fluvio-connector-package/src/context.rs
morenol: This is the actual context used to render the template. So, for template engine we need that this serializes to:
```json
{
  "secrets": {
      "foo": "bar"
   }
}

So, templeta engine knows that `${{ secrets.foo }}` should be replaced by `bar`

   ```

File: crates/fluvio-connector-package/Cargo.toml [hunk 17]
```diff
@@ -14,10 +14,14 @@ default = ["toml"]
 anyhow = { workspace = true }
 bytesize = { workspace = true}
 humantime-serde = "1.1.1"
+# Move to crates.io release, once https://github.com/mitsuhiko/minijinja/commit/78b9b87eaa68ac4c1b1c814d420a8e48e054fefc is released
```
morenol, line 17:
```suggestion
```

crates/fluvio-connector-package/src/secret.rs
```diff
@@ -453,16 +493,60 @@ config:
...
 
         //then
-        assert_eq!(secrets.len(), 4);
+
+        assert_eq!(secrets.len(), 9);
         assert!(secrets.get("secret1").is_some());
         assert!(secrets.get("secret2").is_some());
         assert!(secrets.get("secret3").is_some());
         assert!(secrets.get("secret4").is_some());
+        assert!(secrets.get("secret5").is_some());
```
sehz, line 520: shouldn't this check against actual secret value?

crates/fluvio-connector-package/src/secret.rs
```diff
@@ -453,16 +493,60 @@ config:
...
+        assert!(secrets.get("secret6").is_some());
+        assert!(secrets.get("secret7").is_some());
```
sehz, line 522: shouldn't this be none? since there is no secret7?

crates/fluvio-connector-package/src/secret.rs
```diff
@@ -5,12 +5,18 @@ use std::{
...
-use anyhow::{Result, anyhow};
-use serde_yaml::Value;
+use serde_yaml::{Value};
 
 static SECRET_STORE: OnceCell<Box<dyn SecretStore>> = OnceCell::new();
+/// Regex for detecting secret references in the config string.
+/// This is used to capture the secret names in the config string.
+/// {{ secrets.SECRET_NAME }}
+/// it returns a capture group with the value SECRET_NAME.
+static SECRETS_RE: OnceCell<Regex> = OnceCell::new();
```
sehz, line 19: not quire sure why regex is being used?

crates/fluvio-connector-package/src/secret.rs
morenol: this is a test for `detect_secrets_from_str`. In this function we only detect which are the `Secret names` that we need to add to the context, but we don't resolve them here.

crates/fluvio-connector-package/src/secret.rs
morenol: There is one reference to secret7: 
```
string_interpolation_without_spaces: ${{secrets.secret7}}
```

crates/fluvio-connector-package/src/secret.rs
morenol: We use it to detect which are the secrets that we need to resolve/add to the context

sehz: but that should be done at context level?

sehz: secrets resolution should be at SecretStore which is nothing to do with syntax.  SecretStore can be treated like key/value which is orthogonal to Interpolation.

morenol: yes, I am using this `detect_secrets_from_str` in ContextStore implementation for SecretStore

crates/fluvio-connector-package/src/secret.rs
```diff
@@ -143,28 +156,55 @@ pub fn detect_secrets_from_str(raw_config: &str) -> Result<HashSet<String>> {
 }
 
 fn find_secrets<'a>(set: &mut HashSet<&'a str>, src: &'a Value) {
```
sehz, line 158: This API should be changed.  This should be moved to SecretStore

crates/fluvio-connector-package/src/secret.rs
morenol: We also use this on cloud to detect which are secrets that we should expose to each connector

sehz: That is up to SecretStore implementation.  Still not sure why regex is needed.  Interpolation engine should have resolved all syntax.  So it doesn't matter `${{ env.my_value}}` vs `${{ secret.SECRET1}}`.  The context look up should resolve whether we are injecting simple string or Secret.

crates/fluvio-connector-package/src/render/mod.rs
```diff
@@ -0,0 +1,209 @@
...
+                secrets:
+                    - name: api_key
+                    - name: interval
+            my_service:
+                api_key: ${{secrets.api_key}}
+                interval: ${{ secrets.interval }}
+            "#;
+
+        let value_str = render_config_str(value_str).unwrap();
+        let value: serde_yaml::Value = serde_yaml::from_str(&value_str).expect("should be yaml");
```
sehz, line 195: need test with actual MetaConfig as well

sehz (changes requested):
Overall looks good to me.  One minor ask

sehz (approved):
LGTM

PR #3140, opened 2023-04-14
[Merged by Bors] - build: use workspace deps for serde_yaml

### LINKED ISSUE #3139: Use serde_yaml workspace dependency
Update crates like `fluvio-connector-package` to use workspace dependency for serde_yaml. Also check if other crates are not using the workspace dependency

morenol:
Fixes #3139


sehz (approved):
LGTM

Issue #3145, opened 2023-04-14
[Bug]: CD_Dev using outdated install.sh input parameter


digikata:
#3117 changed VERSION to FLV_VERSION and CD_Dev checks dev version against stable version as a result - always failing.

digikata:
Analyze .github actions to find uses of VERSION as an input to install.sh

morenol:
Probably we also need to update https://github.com/infinyon/fluvio/blob/66e43d7e8f1c9a3242f156b02973b046bc27952d/actions/action-install-fluvio-cluster.sh#L25

digikata:
This also means we need to really carefully inspect all uses of VERSION too, the core change to FLV_VERSION is a good idea, but VERSION is overloaded.

digikata:
https://github.com/infinyon/fluvio/blob/66e43d7e8f1c9a3242f156b02973b046bc27952d/Makefile#L1 too.

digikata:
VERSION->FLV_VERSION is good to sepearate out, reintroduce the commit after release, while also working through all the parameter name changes.

PR #3117, opened 2023-04-04
[Merged by Bors] - refactor!: replaced `$VERSION` with `$FLV_VERSION` to make it unambigous on install.sh

viniarck:
Closes #3020

After refactoring, I also explored and executed `./install.sh` using these `FLV_VERSION` values `inexistent`,`latest`, and not set. I've also confirmed that if a `VERSION=inexistent` is set it doesn't interfere as expected.

```
❯ FLV_VERSION=inexistent ./install.sh
fluvio: ⏳ Downloading Fluvio inexistent for x86_64-unknown-linux-musl...
curl: (22) The requested URL returned error: 404
fluvio: ❌ Failed to download Fluvio, could not find custom FLV_VERSION=inexistent
fluvio:     Make sure this is a valid version, or unset FLV_VERSION to download stable
fluvio: 
fluvio: If you believe this is a bug (or just need help),
fluvio: please feel free to file an issue on Github ❤️
fluvio:     https://github.com/infinyon/fluvio/issues/new
```

```
❯ FLV_VERSION=latest ./install.sh
fluvio: ⏳ Downloading Fluvio 0.10.7-dev-1+18c80fbd0ec5a9162aae5024d327097fb593fd46 for x86_64-unknown-linux-musl...
fluvio: ⬇️  Downloaded Fluvio, installing...
fluvio: ✅ Successfully installed /home/viniarck/.fluvio/bin/fluvio-latest
fluvio: ⏳ Downloading Fluvio channel frontend 0.10.7-dev-1+18c80fbd0ec5a9162aae5024d327097fb593fd46 for x86_64-unknown-linux-musl...
fluvio: ✅ Successfully installed /home/viniarck/.fluvio/bin/fluvio
fluvio: ☁️ Installing Fluvio Cloud...
Current channel: stable
fluvio: 🎣 Fetching latest version for package: fluvio/fluvio-cloud...
fluvio: ⏳ Downloading package with latest version: fluvio/fluvio-cloud:0.2.8...
fluvio: 🔑 Downloaded and verified package file

💡 An update to Fluvio is available!
💡     Run 'fluvio update' to install v0.10.6 of Fluvio
fluvio: ☁️ Installing Fluvio Runner...
Current channel: stable
fluvio: ⏳ Downloading package with provided version: fluvio/fluvio-run:latest...
fluvio: 🔑 Downloaded and verified package file

💡 An update to Fluvio is available!
💡     Run 'fluvio update' to install v0.10.6 of Fluvio
[...]

PR #3146, opened 2023-04-14
fix: install script input name change

digikata:
restore commit "refactor!: replaced `$VERSION` with `$FLV_VERSION` to make it unambigous on install.sh (#3117)" with safety deprecation warning + VERSION -> FLV_VERSION copy if detected.

partial fixes for #3145 

will need to watch for warnings and fix more on other downstream CI before


ajhunyady:
why not call this FLUVIO_VERSION, FLV was a keyword we abandoned long time ago.

sehz:
Let's pause on this.  We need to rethink the installation process.

PR #3150, opened 2023-04-15
[Merged by Bors] - fix: unblock cd_dev.yaml

digikata:
narrow workaound, use FLV_VERSION because last dev version install.sh is being used in CD_Dev in order to release next install.sh


morenol:
I think with a pre-release, the script will be updated so this won't be needed

morenol (approved):

digikata:
I think even w/ pre-release set, the release ci checks that the last CD_Dev suceeded. This should get the reverted install.sh installed onto the latest channel, so everything else downstream will work.

morenol:
bors retry

morenol:
bors retry

PR #3151, opened 2023-04-15
refactor(smartengine): add a layer of indirection to support alternative wasm runtimes

xxchan:
Context of this PR: I'm working on integrating WasmEdge as an alternative runtime for Fluvio smartengine. (https://github.com/WasmEdge/WasmEdge/issues/2231) 

Currently, the code of smartengine is highly coupled with Wasmtime, so it would be a little bit hard (or dirty) to support alternative wasm engines. This PR improves the situation in 2 ways:
1. There are some reusable code not coupled with the engine. So I extracted them into a new module `config`.
2. The public APIs of the `smartengine` crate, which are depended on by other crates, aren't very clear. So I reorganized it: Public structs are defined in `engine/mod.rs`. The implementation details are represented as an `inner` field, and can be imported according to which feature is enabled. This style is learned from backtrace-rs and getrandom (e.g., https://github.com/rust-lang/backtrace-rs/blob/b3e5bb857773fcc6fc8247374264bd4c8acd5387/src/backtrace/mod.rs)

P.S. here's the WIP work of WasmEdge support https://github.com/xxchan/fluvio/pull/1. It's almost finished. But I think splitting it into smaller ones would be easier to review, and get quicker feedback. Besides, I think the refactoring is generally trivial, and makes some sense whether or not WasmEdge is accepted as a new engine, so I submitted this PR first :)


crates/fluvio-smartengine/src/engine/config.rs
xxchan, line 1: This file is extracted from engine.rs without any change.

crates/fluvio-smartengine/src/lib.rs
```diff
@@ -1,21 +1,9 @@
...
-        pub(crate) mod transforms;
-        pub(crate) mod init;
-        pub(crate) mod error;
-        pub mod metrics;
         mod engine;
-        mod state;
         pub use engine::*;
-        pub mod instance;
     }
 }
```
xxchan, line 6: The new `engine` mod will correspond to the feature `engine`. The implementations are moved to `engine::wasmtime_engine`.

xxchan:
@sehz PTAL. Thanks 🙏

crates/fluvio-smartengine/Cargo.toml
```diff
@@ -12,11 +12,11 @@ description = "The official Fluvio SmartEngine"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
-engine = ["wasmtime"]
+engine = ["wasmtime-engine"]
 wasi = ["wasmtime-wasi", "engine"]
 transformation = ["serde_json", "serde_yaml"]
 default = ["engine"]
-
+wasmtime-engine = ["wasmtime", "wasmtime-wasi"]
```
morenol, line 19: with this update, `wasmtime-wasi` dependency will be pulled by default, could you update it to keep current behavior?

xxchan: Ok, this should be a mistake.

sehz:
Thanks for the work.   It's still bit large.    What I suggest is to split this into separate PR:
(1) re-org without changing semantics.  
(2) WASI related changes

xxchan:
@sehz I think the semantic is not changed. But I've created a new PR https://github.com/infinyon/fluvio/pull/3152 without the changes of Cargo features and the "imp" style, if this is what you mean.

PR #3152, opened 2023-04-15
[Merged by Bors] - refactor: improve modularity of smartengine

xxchan:
This PR is part of https://github.com/infinyon/fluvio/pull/3151 (see more context there), containing only trivial code reorganizations, without any actual changes.

To summarize:
- a new `engine` mod, corresponding to the feature `engine`. The implementations are moved to `engine::wasmtime_engine`.
- There are some reusable code not coupled with the engine. So I extracted them into a new module `config`.


xxchan:
@sehz PTAL

crates/fluvio-smartengine/src/engine/wasmtime_engine/engine.rs
```diff
@@ -1,22 +1,17 @@
-use std::fmt::{self, Debug};
-
 use anyhow::Result;
```
sehz, line 3: directory `wasmtime_engine` can be shortened to `wasmtime` since this is already engine

crates/fluvio-smartengine/src/engine/wasmtime_engine/engine.rs
```diff
@@ -77,27 +72,12 @@ impl SmartModuleChainBuilder {
...
-impl<T: Into<Vec<u8>>> From<(SmartModuleConfig, T)> for SmartModuleChainBuilder {
-    fn from(pair: (SmartModuleConfig, T)) -> Self {
-        let mut result = Self::default();
-        result.add_smart_module(pair.0, pair.1.into());
-        result
-    }
-}
-
 /// SmartModule Chain Instance that can be executed
 pub struct SmartModuleChainInstance {
```
sehz: Since this Wasmtime specific, this should be named `SmartModuleWasmTimeChainInstance` or something like that.

crates/fluvio-smartengine/src/engine/wasmtime_engine/instance.rs
```diff
@@ -12,10 +12,10 @@ use fluvio_smartmodule::dataplane::smartmodule::{
...
-use crate::error::EngineError;
-use crate::init::SmartModuleInit;
-use crate::{WasmSlice, memory};
-use crate::state::WasmState;
+use super::error::EngineError;
+use super::init::SmartModuleInit;
+use super::{WasmSlice, memory};
+use super::state::WasmState;
 
 pub(crate) struct SmartModuleInstance {
```
sehz, line 20: Same comment, This should be named like `SmartModuleWasmInstance` ?

sehz (changes requested):
Initial comments

crates/fluvio-smartengine/src/engine/wasmtime_engine/engine.rs
xxchan: I use `wasmtime_engine` because I doubt it will conflict with the `wasmtime` crate. But it seems it won't be a problem. I can use `self::wasmtime`

```
  --> crates/fluvio-smartengine/src/engine/mod.rs:17:9
   |
17 | pub use wasmtime::{SmartEngine, SmartModuleChainBuilder, SmartModuleChainInstance};
   |         ^^^^^^^^ ambiguous name
   |
   = note: ambiguous because of multiple potential import sources
   = note: `wasmtime` could refer to a crate passed with `--extern`
   = help: use `::wasmtime` to refer to this crate unambiguously
note: `wasmtime` could also refer to the module defined here
  --> crates/fluvio-smartengine/src/engine/mod.rs:16:1
   |
16 | mod wasmtime;
   | ^^^^^^^^^^^^^
   = help: use `self::wasmtime` to refer to this module unambiguously

```

crates/fluvio-smartengine/src/engine/wasmtime_engine/engine.rs
xxchan: 1. This PR doesn't introduce wasm abstractions (The "imp" style), renaming it would break public APIs. Or would you like `pub use self::wasmtime:: SmartModuleWasmTimeChainInstance as SmartModuleChainInstance`?
2. It's already in `wasmtime` directory, so I think adding `Wasmtime` to the name is optional.

How do you think?

crates/fluvio-smartengine/src/engine/mod.rs
```diff
@@ -0,0 +1,17 @@
...
+mod error;
+
+#[cfg(test)]
+mod fixture;
+
+pub type WasmSlice = (i32, i32, u32);
+pub type Version = i16;
+
+mod wasmtime;
+pub use self::wasmtime::{SmartEngine, SmartModuleChainBuilder, SmartModuleChainInstance};
```
sehz, line 17: this should be gated using feature flag wasmtime

sehz (changes requested):

crates/fluvio-smartengine/src/engine/mod.rs
xxchan: Hmmm, I thought you don't want to change that. If you want to add new flags for different engines, why not review #3151 directly? This PR is intended to be minimal.

xxchan: I would suggest merge this PR AS-IS, and discuss more about such things in https://github.com/infinyon/fluvio/pull/3151

xxchan: To clarify, this PR hasn't considered "how to add alternative engines", and just contains *trivial* refactoring to make that one step easier. So I don't want to add new feature flags here. And I think how to add feature flags can be nontrivial and break things and might need further discussion, so I'd rather discuss it in a PR where we introduce alternative engines.

sehz: K. That's fair enough.  Let's do min changes here

sehz (approved):
Good enough.   We can merge after 0.10.7 is out.

sehz (changes requested):
need to bump up patch version with release

sehz (approved):
LGTM

Issue #3113, opened 2023-03-31
make rustls as default TLS


sehz:
With latest release, rustls support IP address certs which in the past, prevent being used in the Kubernetes env.

ozgrakkurt:
related to https://github.com/infinyon/fluvio/issues/3091

ozgrakkurt:
@sehz Do we want to keep optional `openssl` or should it be `rustls` only?

Also it is not possible to get rid of `openssl` from repo because we depend on `cargo-generate` which doesn't seem to be able to work without openssl.

sehz:
We should keep openssl.  Not sure relevance to cargo generate

smndtrl:
Just wanted to let you know I stumbled upon issues with `openssl` in our builds as there's no `libssl` in our new builder and base images. For now that's not a huge problem as we only intend to run our `connectors` as static binaries in containers starting first release in CY23 Q3

Looking forward to this

digikata:
Probably good to tackle this a crate at a time

digikata:
A new release https://github.com/rustls/rustls/releases/tag/v/0.22.0 (mentioned by Seyho here https://infinyon.slack.com/archives/CQ8UX6SBH/p1701539652032509)

Making rustls the default build config would reduce maintenance for cross platform builds, as well as surface area for security updates in dependencies.

digikata:
This should be revisited - rustls support needs a reintegration with a fixed up https://github.com/infinyon/future-aio

fraidev:
Solved by https://github.com/infinyon/fluvio/pull/4569!

Issue #3091, opened 2023-03-22
[Feature Request]: Use `reqwest` instead of `isahc`


ozgrakkurt:
Hey!

Currently windows build is a little cumbersome because the project depends on OpenSSL which seems to be because it uses isahc.
It could depend on `surf` with the `rustls` option to remove OpenSSL dependency. This would improve build experience in case the user doesn't have OpenSSL on their system already.

I can do this if it makes sense

digikata:
Improving the build experience on windows is always nice and in general openssl as a dep can be annoying to get correctly configured in various build environments. So we'd definitely take a look at any contributions in that area.

I think you're talking about isahc in fluvio-cli-common?  It would be nice to switch it over to surf with the hyper_client feature on, which IIRC uses rustls by default.   From fluvio-hub-util, we're getting a warning with surf in default config with nom, so there could be a similar change there too.

ozgrakkurt:
@digikata would using `surf` with `async-h1-rustls` be good enough? `hyper` is more `tokio` oriented afaik.

[async-h1](https://crates.io/crates/async-h1)

digikata:
As an experiment, if you create a standalone crate with a surf dep configured to hyper-client, no openssl is pulled in. 
So openssl may get pulled in from some other part of the dep tree? It may be  something we need to chip away at a little at a time.

ozgrakkurt:
It pulls it in for me, it looks like `hyper_client` -> `http_client/hyper_client` -> ` hyper-tls` -> `native-tls` -> `openssl`

digikata:
Thanks for checking that, I re-checked on macos it is not used in that config, but w/ linux it is. Macos must use a different native tls lib.  It's possible to use reqwest w/ rustls (which I can confirm doesn't pull in openssh), but it can get tricky in mixing it into dependencies in various crates and workspace in the total fluvio ecosystem.  

You are welcome to give it a try and we can try to review and work through the issues, but it will take a little more time than a simple dep reconfig.

ozgrakkurt:
I checked `reqwest`, it seems like it depends on `tokio` so we would have to enable `tokio03` feature on `async-std` and the more I look into it, it seems like the main issue is `async-std` isn't being developed much. I would propose to use `hyper` which might be too low level or switching to tokio which might be too much work so not sure. Also `fluvio-hub-util` already depends on `surf/async-h1` so might be ok to just use it maybe?

digikata:
[...]

Issue #3070, opened 2023-03-10
migrate from http-client to other http client


sehz:
This is caused by having old dep to `deadpool`.  Seems like this project is not being maintained.  Let's move to more maintained project

ozgrakkurt:
related to https://github.com/infinyon/fluvio/issues/3091

digikata:
Closed - completed for first phase

PR #3156, opened 2023-04-17
feature: make `rustls` default in `fluvio`, move to `reqwest` from `surf`

ozgrakkurt:
- Made `rustls` feature on `fluvio` the default
- Made most components use `rustls` if it is enabled. It used to use `openssl` anyway because of a compatibility problem.
-  Removed usage of `surf` because it is not maintained and added `reqwest` instead.

closes https://github.com/infinyon/fluvio/issues/3091
closes https://github.com/infinyon/fluvio/issues/3113
closes https://github.com/infinyon/fluvio/issues/3070

Seems like a bunch of things need to be manually checked using the cli before we are sure this works?. One problem might be that we need `async-std/tokio1` enabled and I enabled it in `fluvio` crate but this will be problem if the sub-crates are used somewhere else where `async-std/tokio1` isn't enabled.


ozgrakkurt:
@digikata it isn't possible to get rid of `openssl` entirely for now because we depend on `cargo-generate` which doesn't seem to give an option to use `rustls` or run without tls.

Cargo.toml
```diff
@@ -97,13 +97,31 @@ colored = "2.0.0"
...
 which = "4.1.0"
+reqwest = { version = "0.11", default-features = false, features = ["json", "gzip"] }
 
 # fluvio dep
 k8-client = { version = "10.0.0" }
 k8-types = { version = "0.8.0" }
 k8-config = { version = "2.0.0"}
 k8-metadata-client = { version = "5.1.0" }
-fluvio-future = { version = "0.4.0" }
+fluvio-future = { version = "0.4", default-features = false, features = [
```
sehz, line 107: don't add features at workspace dependency

sehz (changes requested):

Cargo.toml
ozgrakkurt: done

Cargo.toml
```diff
@@ -97,13 +97,31 @@ colored = "2.0.0"
 semver = "1.0.3"
 url = "2.1.1"
 which = "4.1.0"
+reqwest = { version = "0.11", default-features = false, features = ["json", "gzip"] }
```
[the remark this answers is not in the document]
ozgrakkurt: done

ozgrakkurt:
@sehz moved features from workspace to individual crates, made everything use `rustls` as much as I can and the rest uses `openssl`/`native-tls`.

removed tls features on `fluvio` crate.

ozgrakkurt:
seems like there are errors about doing `fluvio-future = { workspace = true }`, how should I handle these? Do we need the version to be specified in each of these crates?

sehz:
you don't need to specify version but need specific feature enabled in each crate that need optional features

ozgrakkurt:
@sehz bumped versions also fixed some mistakes. Also this is a breaking change since I removed the `openssl` and `rustls` features from `fluvio` crate

digikata (commented):
Looks really good so far, after it settles down more I wonder if it might be worth splitting apart the PR to some boundaries to phase it in. It's a large number of changes, so separating possible side effects to smaller sets of changes makes it easier on downstream projects like fluvio-cloud-ops.  e.g. the fluvio-future syncup to a workspace dep might be worth peeling off and merging to master before this.

ozgrakkurt:
> Looks really good so far, after it settles down more I wonder if it might be worth splitting apart the PR to some boundaries to phase it in. It's a large number of changes, so separating possible side effects to smaller sets of changes makes it easier on downstream projects like fluvio-cloud-ops. e.g. the fluvio-future syncup to a workspace dep might be worth peeling off and merging to master before this.

Makes sense, probably gonna have to do this like `reqwest` change and then another PR for the TLS. My bad on this PR

digikata:
> Makes sense, probably gonna have to do this like `reqwest` change and then another PR for the TLS. My bad on this PR

No worries, the exploratory work had to be done to find out what the extent of the changes might be.

Issue #3045, opened 2023-03-03
Re-enable crates publish after Release workflow


tjtelan:
Related to #2998 

It appears this job no longer runs after release. Should run after release completes successfully

sehz:
Other option is top publishing to crates.io and use explicit GitHub dependency

drc-infinyon:
Decision needed on publishing crates in the future.

morenol:
It seems that also bump-stable job is being skipped

PR #3157, opened 2023-04-17
[Merged by Bors] - ci: parse input as boolean

morenol:
>> Note: The workflow will also receive the inputs in the github.event.inputs context. The information in the inputs context and github.event.inputs context is identical except that the inputs context preserves Boolean values as Booleans instead of converting them to strings.

https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_dispatchinputs

This should fix https://github.com/infinyon/fluvio/issues/3045

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -243,7 +243,7 @@ jobs:
 
   publish_crates:
     name: Publish Crates
-    if: ${{ !github.event.inputs.pre_release }}
+    if: ${{ !inputs.pre_release }}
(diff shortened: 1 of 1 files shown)


morenol:
The input was parsed as `string`, so it seems that `!` was being applied to a non-empty string instead of to a boolean value

digikata:
This seems to get a lot of people, see https://github.com/actions/runner/issues/1483

digikata (approved):
LGTM

PR #3159, opened 2023-04-17
[Merged by Bors] - chore: post release 0.10.7 ver bump

digikata:


VERSION
```diff
@@ -1 +1 @@
-0.10.7
\ No newline at end of file
+0.10.8-alpha.1
```
morenol, line 1: I think that this should be dev-1 instead of alpha.1

digikata: I was debating that.  Updated

PR #3163, opened 2023-04-18
[Merged by Bors] - feat: introduce utilily to check k8s resource names

EstebanBorai:
Provides `shared::is_valid_k8s_resource_name` which checks if the provided
string is a k8s resource name compliant value. This could be used for topics,
connectors and other resources relying on k8s. In the future is likely we want to
use a single function all around to apply the same rules on a centralized approach.

### DIFF
--- a/crates/fluvio-cli/src/client/smartmodule/create.rs
+++ b/crates/fluvio-cli/src/client/smartmodule/create.rs
@@ -5,11 +5,12 @@ use std::sync::Arc;
 use tracing::debug;
 use async_trait::async_trait;
 use clap::Parser;
-use anyhow::Result;
+use anyhow::{anyhow, Result};
 
 use fluvio::Fluvio;
 use fluvio_controlplane_metadata::smartmodule::{SmartModuleWasm, SmartModuleSpec};
 use fluvio_extension_common::Terminal;
+use fluvio_sc_schema::shared::{MAX_RESOURCE_NAME_LEN, is_valid_resource_name};
 
 use crate::client::cmd::ClientCmd;
 
@@ -68,6 +69,11 @@ impl ClientCmd for CreateSmartModuleOpt {
             (None, BTreeMap::new())
         };
         */
+
+        if !is_valid_resource_name(&self.name) {
+            debug!(name = self.name, "Invalid name provided for SmartModule");
+            return Err(anyhow!("Invalid name for SmartModule {}. Names must have a maximum of {} characters, and don't include any special characters nor spaces.", self.name, MAX_RESOURCE_NAME_LEN));
+        }
 
         let raw = std::fs::read(self.wasm_file)?;
 
--- a/crates/fluvio-cli/src/client/smartmodule/create.rs
+++ b/crates/fluvio-cli/src/client/smartmodule/create.rs
@@ -70,7 +70,7 @@ impl ClientCmd for CreateSmartModuleOpt {
         };
         */
 
-        if !is_valid_resource_name(&self.name) {
+        if is_valid_resource_name(&self.name).is_err() {
             debug!(name = self.name, "Invalid name provided for SmartModule");
             return Err(anyhow!("Invalid name for SmartModule {}. Names must have a maximum of {} characters, and don't include any special characters nor spaces.", self.name, MAX_RESOURCE_NAME_LEN));
         }
--- a/crates/fluvio-cli/src/client/smartmodule/create.rs
+++ b/crates/fluvio-cli/src/client/smartmodule/create.rs
@@ -72,11 +72,7 @@ impl ClientCmd for CreateSmartModuleOpt {
 
         if let Err(err) = validate_resource_name(&self.name) {
             debug!(name = self.name, "Invalid name provided for SmartModule");
-            return Err(anyhow!(
-                "Invalid name for SmartModule {}. {}",
-                self.name,
-                err.to_string()
-            ));
+            return Err(anyhow!("Invalid name for SmartModule {}, {err}", self.name));
         }
 
         let raw = std::fs::read(self.wasm_file)?;
--- a/crates/fluvio-cli/src/client/smartmodule/create.rs
+++ b/crates/fluvio-cli/src/client/smartmodule/create.rs
@@ -10,7 +10,7 @@ use anyhow::{anyhow, Result};
 use fluvio::Fluvio;
 use fluvio_controlplane_metadata::smartmodule::{SmartModuleWasm, SmartModuleSpec};
 use fluvio_extension_common::Terminal;
-use fluvio_sc_schema::shared::{MAX_RESOURCE_NAME_LEN, is_valid_resource_name};
+use fluvio_sc_schema::shared::validate_resource_name;
 
 use crate::client::cmd::ClientCmd;
 
@@ -70,9 +70,13 @@ impl ClientCmd for CreateSmartModuleOpt {
         };
         */
 
-        if is_valid_resource_name(&self.name).is_err() {
+        if let Err(err) = validate_resource_name(&self.name) {
             debug!(name = self.name, "Invalid name provided for SmartModule");
-            return Err(anyhow!("Invalid name for SmartModule {}. Names must have a maximum of {} characters, and don't include any special characters nor spaces.", self.name, MAX_RESOURCE_NAME_LEN));
+            return Err(anyhow!(
+                "Invalid name for SmartModule {}. {}",
+                self.name,
+                err.to_string()
+            ));
         }
 
         let raw = std::fs::read(self.wasm_file)?;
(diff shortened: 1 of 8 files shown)


morenol (approved):
LGTM, could you use it in https://github.com/infinyon/fluvio/blob/3a0ad8dc95af9034eaefdb4ffbe14946ef7a384b/crates/fluvio-cli/src/client/topic/create.rs#L135

And also in smartmodule create?

crates/fluvio-sc-schema/src/shared.rs
```diff
@@ -0,0 +1,78 @@
+const K8S_MAX_RESOURCE_NAME_LEN: usize = 63;
+
+pub fn is_valid_k8s_resource_name(name: &str) -> bool {
```
sehz, line 3: rename to make generic since we don't to expose k8 directly.

sehz (changes requested):

sehz:
this should be used by other object such as topic, etc

EstebanBorai:
> this should be used by other object such as topic, etc

Agree!

EstebanBorai:
> LGTM, could you use it in
> 
> https://github.com/infinyon/fluvio/blob/3a0ad8dc95af9034eaefdb4ffbe14946ef7a384b/crates/fluvio-cli/src/client/topic/create.rs#L135
> 
> And also in smartmodule create?

Just propagated the new function all around! I still conserve the `valid_topic_name` function to avoid breaking other projects, but perhaps we could also annotate it with a `deprecated` trait for the next version? WDTY?

crates/fluvio-cli/src/client/smartmodule/create.rs
```diff
@@ -69,6 +70,11 @@ impl ClientCmd for CreateSmartModuleOpt {
         };
         */
 
+        if !is_valid_resource_name(&self.name) {
+            debug!(name = self.name, "Invalid name provided for SmartModule");
+            return Err(anyhow!("Invalid name for SmartModule {}. Names must have a maximum of {} characters, and don't include any special characters nor spaces.", self.name, MAX_RESOURCE_NAME_LEN));
```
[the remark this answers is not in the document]
EstebanBorai: I think one good approach would be to have an `enum` with error codes.

Two variants I think of for this enum are:

- Max Length Exceeded
- Invalid Character Encountered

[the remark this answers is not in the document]
EstebanBorai: Just introduced an `enum` for error descriptions here: [5dcfe11](https://github.com/infinyon/fluvio/pull/3163/commits/5dcfe11fe6fe6e1905611126f6062641d8a4c6f6). To keep it compatible with current API, in certain locations we are checking for the `Result` to be `ok` or `err`. We would still have to do a granular error handling in the next iteration.

crates/fluvio-cli/src/client/smartmodule/create.rs
```diff
@@ -69,6 +70,11 @@ impl ClientCmd for CreateSmartModuleOpt {
         };
         */
 
+        if is_valid_resource_name(&self.name).is_err() {
+            debug!(name = self.name, "Invalid name provided for SmartModule");
```
sehz, line 74: just return error

crates/fluvio-cli/src/client/topic/create.rs
```diff
@@ -132,8 +132,8 @@ impl CreateTopicOpt {
             })
         };
 
-        let is_valid = valid_topic_name(&self.topic);
-        if !is_valid {
+        let is_valid = is_valid_resource_name(&self.topic);
+        if is_valid.is_err() {
```
sehz, line 136: just return error

crates/fluvio-sc-schema/src/shared.rs
```diff
@@ -0,0 +1,98 @@
...
+    InvalidCharacterEncountered,
+}
+
+/// Checks if the Resource Name is valid for internal resources.
+///
+/// ```test
+/// let name = "prices-list-scrapper";
+/// assert!(is_valid_resource_name(name));
+/// ```
+pub fn is_valid_resource_name(name: &str) -> Result {
```
sehz, line 17: since this return error, it should be something like: `check_resource_name`

crates/fluvio-sc-schema/src/shared.rs
```diff
@@ -0,0 +1,98 @@
+pub type Result = std::result::Result<(), ValidateResourceNameError>;
+
+pub const MAX_RESOURCE_NAME_LEN: usize = 63;
+
+#[derive(Copy, Clone, Debug)]
```
sehz, line 5: Use `thiserror` to define error

sehz (changes requested):

crates/fluvio-sc-schema/src/shared.rs
EstebanBorai: Fixed!

crates/fluvio-cli/src/client/topic/create.rs
EstebanBorai: Sure!

crates/fluvio-cli/src/client/smartmodule/create.rs
EstebanBorai: Applied!

crates/fluvio-cli/src/client/smartmodule/create.rs
```diff
@@ -69,6 +70,15 @@ impl ClientCmd for CreateSmartModuleOpt {
         };
         */
 
+        if let Err(err) = validate_resource_name(&self.name) {
+            debug!(name = self.name, "Invalid name provided for SmartModule");
+            return Err(anyhow!(
+                "Invalid name for SmartModule {}. {}",
+                self.name,
+                err.to_string()
```
sehz, line 78: don't need `err.to_string()`.   you can also interpolation like:
```
anyhow!("Invalid name for SmartModule {}, {err}", self.name}):
```

crates/fluvio-cli/src/client/topic/create.rs
```diff
@@ -132,12 +132,12 @@ impl CreateTopicOpt {
...
 
-        let is_valid = valid_topic_name(&self.topic);
-        if !is_valid {
-            return Err(CliError::InvalidArg(
-                "Topic name must only contain lowercase alphanumeric characters or '-'."
-                    .to_string(),
-            )
+        if let Err(err) = validate_resource_name(&self.topic) {
+            return Err(CliError::InvalidArg(format!(
+                "Invalid Topic name {}. {}",
```
sehz, line 137: same comment as above

crates/fluvio-sc-schema/Cargo.toml
```diff
@@ -18,7 +18,7 @@ use_serde = ["fluvio-controlplane-metadata/use_serde", "serde"]
 log = "0.4.8"
 tracing = "0.1.19"
 serde = { version = "1.0.0", features = ['derive'], optional = true }
-thiserror = "1.0.20"
+thiserror = "1.0.40"
```
sehz, line 21: use workspace dep

crates/fluvio-sc/src/services/public_api/topic/create.rs
```diff
@@ -80,12 +80,11 @@ pub async fn handle_create_topics_request<AC: AuthContext>(
...
     debug!("validating topic: {}", name);
 
-    let valid_name = valid_topic_name(name);
-    if !valid_name {
+    if let Err(err) = validate_resource_name(name) {
         return Status::new(
             name.to_string(),
             ErrorCode::TopicInvalidName,
-            Some(format!("Invalid topic name: '{name}'. Topic name can contain only lowercase alphanumeric characters or '-'.")),
+            Some(format!("Invalid topic name: '{name}'. {}", err.to_string())),
```
sehz, line 87: see previously comment about interpolation

crates/fluvio-sc-schema/src/topic/mod.rs
```diff
@@ -1,46 +1,6 @@
 pub use fluvio_controlplane_metadata::topic::*;
-pub mod validate {
```
sehz, line 2: if we remove this API, we should bump up major version

sehz (changes requested):

sehz (approved):
LGTM.

EstebanBorai:
> LGTM.  

Thanks for all the feedback! 🙌

EstebanBorai:
Okay, deprecation `macro` wont let me pass `clippy`. Looks like I will have to iterate by migrating from the function internally and then deprecate.

```
warning: use of deprecated function `topic::validate::valid_topic_name`: Will be removed in the next version. Use `fluvio_sc_schema::shared::validate_resource_name` instead.
  --> crates/fluvio-sc-schema/src/topic/mod.rs:19:37
   |
19 |         use crate::topic::validate::valid_topic_name;
   |                                     ^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(deprecated)]` on by default
```

PR #3170, opened 2023-04-19
[Merged by Bors] - fix: properly close multiplexer

### LINKED ISSUE #3124: [Bug]: CLI consumer & producers fail silently when they are disconnect from the cluster

galibey:
The PR fixes two issues in `socket multiplexer`:
1. In case of an error, the thread is abandoned, but awaiting senders are not notified. Only timeout releases them if it happens.
2. Serial senders are never notified if the thread is terminated.

Fixes #3124

### DIFF
--- a/crates/fluvio-socket/src/multiplexing.rs
+++ b/crates/fluvio-socket/src/multiplexing.rs
@@ -40,7 +40,8 @@ use crate::FluvioStream;
 
 pub type SharedMultiplexerSocket = Arc<MultiplexerSocket>;
 
-type SharedMsg = (Arc<Mutex<Option<Bytes>>>, Arc<Event>);
+#[derive(Clone)]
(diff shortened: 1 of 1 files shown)


sehz (approved):
Looks good.  Nice cleanup!
We need way to test these path as separate effort.

PR #3176, opened 2023-04-20
[Merged by Bors] - post release 0.10.7 notes

digikata:
- chore: post-release release_checklist.md update
- chore: Release.md, add pointer to integrated platform release docs

### DIFF
--- a/.github/ISSUE_TEMPLATE/release_checklist.md
+++ b/.github/ISSUE_TEMPLATE/release_checklist.md
@@ -18,14 +18,19 @@ For more detail, refer to [`RELEASE.md`](https://github.com/infinyon/fluvio/blob
 - [ ] Inform team that release is about to begin
 - [ ] Ensure no merges are in flight
 - [ ] Create a PR for release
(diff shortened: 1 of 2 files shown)


RELEASE.md
```diff
@@ -1,5 +1,7 @@
 # Release process
 
+See [Platform Release doc](https://github.com/infinyon/fluvio-cloud-ops/blob/master/docs/release/RELEASE.md) for integrated platform release process steps.
+
```
morenol, line 4: maybe we should not mention this here?

digikata: I'm was undecided about this, but it put it in for discussion.  We should only have one release procedure. Maybe we should remove most of this RELEASE file and depend on the fc-ops one? It was nicer to only have to look at one prodcedure (+ issues).

sehz: remove cloud ops release

digikata: removed!

sehz (approved):

PR #3180, opened 2023-04-20
[Merged by Bors] - build: fix toolchain reference

morenol:
Without this change it fails at this way:

```
error: the 'cargo' binary, normally provided by the 'cargo' component, is not applicable to the '1.69-aarch64-apple-darwin' toolchain
```


morenol:
bors retry

PR #3193, opened 2023-04-25
[Merged by Bors] - fix: use `.hub` over `hub` and delete if exists

EstebanBorai:
Ensures the Package Metadata Directory is recreated before publishing
a package to the hub.

### DIFF
--- a/crates/smartmodule-development-kit/src/publish.rs
+++ b/crates/smartmodule-development-kit/src/publish.rs
@@ -1,4 +1,5 @@
 use std::path::{Path, PathBuf};
+use std::fs::remove_dir_all;
 
 use anyhow::{Result, anyhow};
 use cargo_builder::package::PackageInfo;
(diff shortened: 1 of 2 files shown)


crates/smartmodule-development-kit/src/publish.rs
```diff
@@ -44,17 +45,19 @@ pub struct PublishCmd {
...
         let hubdir = package_info.package_relative_path(DEF_HUB_INIT_DIR);
-        if !hubdir.exists() {
-            init_package_template(&package_info)?;
-        } else if !self.public_yes {
-            check_package_meta_visiblity(&package_info)?;
+
+        if hubdir.exists() {
+            // Delete the `.hub` directory if already exists
+            tracing::warn!("Removing directory at {:?}", hubdir);
+            remove_dir_all(&hubdir)?;
```
EstebanBorai, line 55: Given that we are removing directories under the hood, perhaps we could prompt the user before proceeding?
And allowing a flag like `-y` if we want to allow for CI use cases? WDYT @sehz , @digikata?

sehz: dont think thats necessary.  user can back before publishing

sehz (approved):
LGTM

PR #3194, opened 2023-04-25
[Merged by Bors] - feat: Secrets in connector meta

morenol:
Previously, we defined possible secrets used by a connector on connector package spec. This approach makes less customizable the experience for the user.

In this PR we are moving this definition to the connector meta configuration, users can define explicitly the name of the secrets that they want and use them explicitly.

This change also implies that secret names are not longer detected from parsing the connector specific configuration and instead only parsing the secrets section on the connector configuration.

### DIFF
--- a/crates/fluvio-connector-package/src/metadata.rs
+++ b/crates/fluvio-connector-package/src/metadata.rs
@@ -1,4 +1,4 @@
-use std::{collections::BTreeMap, ops::Deref, fmt::Display};
+use std::{ops::Deref, fmt::Display};
 
 use anyhow::{anyhow, Context};
 use openapiv3::{Schema, AnySchema, ReferenceOr, Type, SchemaKind};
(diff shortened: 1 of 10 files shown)


crates/fluvio-connector-package/src/metadata.rs
```diff
@@ -270,7 +270,7 @@ impl ConnectorMetadata {
             serde_yaml::from_reader(reader).context("unable to parse config into YAML format")?;
         validate_custom_config(&self.custom_config, &value)
             .context("custom config validation failed")?;
-        validate_secrets(&self.secrets, &value)?;
+        //  validate_secrets(&self.secrets, &value)?;
```
morenol, line 273:
```suggestion
```

sehz:
LGTM

crates/fluvio-connector-package/src/metadata.rs
```diff
@@ -313,19 +313,6 @@ fn validate_deployment(deployment: &Deployment, config: &ConnectorConfig) -> any
     Ok(())
 }
 
-fn validate_secrets(secrets: &Secrets, config: &serde_yaml::Value) -> anyhow::Result<()> {
```
galibey, line 316: We should also probably delete `secrets` field and type from `ConnectorMetadata` as it is no longer being used.

morenol: Removed in https://github.com/infinyon/fluvio/pull/3194/commits/9ccec8ba81ace408a4dae24599c054289ce212e9

PR #3195, opened 2023-04-25
ci: replace unmaintained actions-rs/toolchain by dtolnay/rust-toolchain

morenol:


morenol:
That github action does not support `rust-toolchain.toml`

PR #3198, opened 2023-04-27
[Merged by Bors] - fix: fix fluvio hub utils

morenol:
In the update of `pem` https://github.com/infinyon/fluvio/commit/fdcfce51067a44c06a91bc8e4aab518f0a193145 to `2.0.0`. It was changed the tag from `PUBLIC KEY` and `PRIVATE KEY` to `PUBLIC_KEY` and `PRIVATE_KEY`. This changes uses a global const so we use the same tag when we write to file and when we read from file

### DIFF
--- a/crates/fluvio-hub-util/Cargo.toml
+++ b/crates/fluvio-hub-util/Cargo.toml
@@ -47,4 +47,4 @@ features = ["h1_client", "rustls"]
 default-features = false
 
 [dev-dependencies]
-tracing-subscriber = { workspace = true,  features = ["env-filter"] }
+tracing-subscriber = { workspace = true,  features = ["env-filter", "fmt"] }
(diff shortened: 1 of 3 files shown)


digikata:
The global const is fine, but the reason for "PUBLIC KEY" and "PRIVATE KEY" with spaces is that those tags matched openssh generated key formats.

morenol:
> The global const is fine, but the reason for "PUBLIC KEY" and "PRIVATE KEY" with spaces is that those tags matched openssh generated key formats.

Right! I updated the consts to have the space in the tag. The issue right now is the difference between

```rust 
let pem = Pem::new("PRIVATE_KEY", self.kp.secret.to_bytes().to_vec());
``` 
with:
```rust
        if pem.tag() != "PRIVATE KEY" {
```

digikata:
I took a longer look at it this morning and last night I was looking at the wrong keys, I think its good!

digikata (approved):

PR #3200, opened 2023-04-27
[Merged by Bors] - tests: add tests to `SecretName` struct

morenol:
Use `FromStr` instead of `From<String>` since it is more generic, also add Deref impl to `&str`


digikata (approved):
LGTM

PR #3206, opened 2023-04-28
[Merged by Bors] - feat: Introduce `apiVersion` to `ConnectorConfig` in order to allow backward c…

### LINKED ISSUE #3204: Add apiVersion to connector config
So we can handle schema changes in a better way

morenol:
…ompatible schema changes

Fixes #3204

### DIFF
--- a/crates/fluvio-connector-package/test-data/connectors/error-batchsize.yaml
+++ b/crates/fluvio-connector-package/test-data/connectors/error-batchsize.yaml
@@ -1,4 +1,4 @@
-schemaVersion: v1
+apiVersion: v1
 meta:
   version: 0.1.0
   name: my-test-mqtt
--- a/crates/fluvio-connector-package/test-data/connectors/error-batchsize.yaml
+++ b/crates/fluvio-connector-package/test-data/connectors/error-batchsize.yaml
@@ -1,4 +1,4 @@
-apiVersion: v1
+apiVersion: 0.1.0
 meta:
   version: 0.1.0
   name: my-test-mqtt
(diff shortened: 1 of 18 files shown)


crates/fluvio-connector-package/test-data/connectors/error-batchsize.yaml
```diff
@@ -1,3 +1,4 @@
+schemaVersion: v1
```
sehz, line 1: We are using `apiVersion` in other configs like connector and SmartModule.   I suggest we keep the same.  Also they are using server which seems to be more easy to manage

sehz (changes requested):
Overall looks good.  The only issue is the name of the new field which I suggest keeping same as others.

crates/fluvio-connector-package/test-data/connectors/error-batchsize.yaml
morenol: Got it, In the issue I put apiVersion, but then changed because that name seemed more accurate to me but I don't have any strong opinion about it. Changing back to `apiVersion`

sehz: `schemaVersion` will be confused with the new schema mgt feature

sehz: and use semversion

sehz (approved):

morenol:
bors retry

digikata (approved):
lgtm

PR #3209, opened 2023-05-01
[Merged by Bors] - feat: add max_bytes to connector consumer option

### LINKED ISSUE #3210: add max_bytes to connector consumer option

morenol:
Closes #3210

### DIFF
--- a/crates/fluvio-connector-package/test-data/connectors/full-config.yaml
+++ b/crates/fluvio-connector-package/test-data/connectors/full-config.yaml
@@ -11,7 +11,7 @@ meta:
     compression: gzip
   consumer:
     partition: 10
-    max-bytes: "1 MB"
+    max_bytes: "1 MB"
   secrets:
     - name: secret1
 transforms:
--- a/crates/fluvio-connector-package/test-data/connectors/full-config.yaml
+++ b/crates/fluvio-connector-package/test-data/connectors/full-config.yaml
@@ -11,6 +11,7 @@ meta:
     compression: gzip
   consumer:
     partition: 10
+    max-bytes: "1 MB"
   secrets:
     - name: secret1
 transforms:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -718,6 +718,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "bytesize-serde"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "86d1eb2fd2668859e9785b99700c66b7ea9fdeda35d99f95827a4e5493440178"
+dependencies = [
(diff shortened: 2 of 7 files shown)


nacardin (commented):
Wouldn't it be cleaner to implement a custom deserializer/serialzier for `ByteSize` instead of a separate string field?

You could mark the field with `#[serde(with = "module")] ` https://serde.rs/field-attrs.html#with

morenol:
> Wouldn't it be cleaner to implement a custom deserializer/serialzier for `ByteSize` instead of a separate string field?
> 
> You could mark the field with `#[serde(with = "module")] ` https://serde.rs/field-attrs.html#with

addressed in https://github.com/infinyon/fluvio/pull/3209/commits/692b9ff26e6b25fdafe430a41829389608e2f874

Now field is `max_bytes`. I don't like the inconsistency between `max_bytes` and `batch-size`. Maybe I should update it too for `batch_size`?

morenol:
> > Wouldn't it be cleaner to implement a custom deserializer/serialzier for `ByteSize` instead of a separate string field?
> > You could mark the field with `#[serde(with = "module")] ` https://serde.rs/field-attrs.html#with
> 
> addressed in [692b9ff](https://github.com/infinyon/fluvio/commit/692b9ff26e6b25fdafe430a41829389608e2f874)
> 
> Now field is `max_bytes`. I don't like the inconsistency between `max_bytes` and `batch-size`. Maybe I should update it too for `batch_size`?

Addressed my concern on https://github.com/infinyon/fluvio/pull/3209/commits/b7bc0e233daf07c061f6a19744a55c98063b323e

nacardin (approved):
Nice!

crates/fluvio-connector-package/test-data/connectors/full-config.yaml
```diff
@@ -11,6 +11,7 @@ meta:
     compression: gzip
   consumer:
     partition: 10
+    max_bytes: "1 MB"
   secrets:
```
sehz, line 15: Can we add test cases for just regular numbers? For example, this should work
```
consumer:
      max_bytes: 13933
```
```

morenol: Ok, I will update then to use:  https://docs.rs/bytesize-serde/latest/bytesize_serde/ that seems to handle that case

morenol: Addressed in https://github.com/infinyon/fluvio/pull/3209/commits/869d4ef8b0b8dac2543311fdde36a80b359c2aec (also added test there)

sehz (approved):
LGTM

PR #3211, opened 2023-05-01
[Merged by Bors] - Update generate to use new version

morenol:
http-sink initial commit has this very old version. Updating this and adding to checklist so we keep using new versions


sehz (approved):
LGTM

nacardin (approved):

PR #3230, opened 2023-05-05
[Merged by Bors] - feat(cdk): ensure readme is included in manifest

EstebanBorai:
Ensures the README is included into the `ipkg`

### DIFF
--- a/.github/workflows/connector-publish.yml
+++ b/.github/workflows/connector-publish.yml
@@ -19,9 +19,12 @@ on:
       fail-fast:
         type: boolean
         default: true
-      readme_path:
+      readme:
         type: string
         default: './README.md'
+      public:
+        type: boolean
+        default: true
     secrets:
       CLOUD_USER_EMAIL:
         required: true
@@ -68,10 +71,10 @@ jobs:
       - name: Publish
         run: |
           cdk publish \
-            --public-yes \
+            ${{ inputs.public && '--public-yes' || '' }} \
             --target ${{ matrix.rust-target }} \
             -p ${{ inputs.package-name }} \
-            --readme ${{ inputs.readme_path }}
+            ${{ (inputs.readme && '--readme ' + inputs.readme) || '' }}
 
   macos:
     name: macos
@@ -104,4 +107,7 @@ jobs:
           cargo build --profile ${{ inputs.rust-profile }}  --target ${{ matrix.rust-target }} -p ${{ inputs.package-name }}
       - name: Publish
         run: |
-          cdk publish --public-yes --target ${{ matrix.rust-target }} -p ${{ inputs.package-name }}
+          cdk publish ${{ inputs.public && '--public-yes' || '' }} \
+            --target ${{ matrix.rust-target }} \
+            -p ${{ inputs.package-name }} \
+            ${{ (inputs.readme && '--readme ' + inputs.readme) || '' }}
--- a/.github/workflows/connector-publish.yml
+++ b/.github/workflows/connector-publish.yml
@@ -19,6 +19,9 @@ on:
       fail-fast:
         type: boolean
         default: true
+      readme_path:
+        type: string
+        default: './README.md'
     secrets:
       CLOUD_USER_EMAIL:
         required: true
@@ -64,7 +67,11 @@ jobs:
           cross build --profile ${{ inputs.rust-profile }}  --target ${{ matrix.rust-target }} -p ${{ inputs.package-name }}
       - name: Publish
         run: |
-          cdk publish --public-yes --target ${{ matrix.rust-target }} -p ${{ inputs.package-name }}
+          cdk publish \
+            --public-yes \
+            --target ${{ matrix.rust-target }} \
+            -p ${{ inputs.package-name }} \
+            --readme ${{ inputs.readme_path }}
 
   macos:
     name: macos
--- a/crates/cdk/src/publish.rs
+++ b/crates/cdk/src/publish.rs
@@ -23,7 +23,6 @@ use tracing::{debug, info};
 use crate::cmd::PackageCmd;
 
 pub const CONNECTOR_TOML: &str = "Connector.toml";
-pub const README_MD: &str = "README.md";
 
 /// Publish Connector package to the Hub
 #[derive(Debug, Parser)]
@@ -197,7 +196,7 @@ pub fn init_package_template(
     let readme_md_relative_path = package_meta_relative_path(&package_meta_path, &readme_path);
     pm.manifest.push(
         readme_md_relative_path
-            .clone()
+            
             .unwrap_or_else(|| readme_path.to_string_lossy().to_string()), // if failed to get relative path, use absolute)
     );
 
@@ -247,16 +246,6 @@ pub(crate) fn find_connector_toml(package_info: &PackageInfo) -> Result<PathBuf>
     }
 
     Err(anyhow::anyhow!("No \"{}\" file found", CONNECTOR_TOML))
-}
-
-pub(crate) fn find_readme_md(package_info: &PackageInfo) -> Result<PathBuf> {
-    let readme_md = package_info.package_relative_path(README_MD);
-
-    if readme_md.exists() {
-        return Ok(readme_md);
-    }
-
-    Err(anyhow::anyhow!("No \"{}\" file found", README_MD))
 }
 
 fn verify_public_or_exit() -> Result<()> {
--- a/crates/cdk/src/publish.rs
+++ b/crates/cdk/src/publish.rs
@@ -1,6 +1,7 @@
 //!
 //! Command for hub publishing
 
+use std::env::current_dir;
 use std::fs::remove_dir_all;
 use std::path::Path;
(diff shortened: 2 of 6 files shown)


crates/cdk/src/publish.rs
```diff
@@ -215,15 +222,25 @@ fn from_connectorvis(cv: &ConnectorVisibility) -> PkgVisibility {
...
-        return Ok(smartmodule_toml);
+    if connector_toml.exists() {
+        return Ok(connector_toml);
     }
 
     Err(anyhow::anyhow!("No \"{}\" file found", CONNECTOR_TOML))
 }
 
+pub(crate) fn find_readme_md(package_info: &PackageInfo) -> Result<PathBuf> {
+    let readme_md = package_info.package_relative_path(README_MD);
```
morenol, line 235: This assumes the Readme is in the same file than Connector.toml? Can we add capability to override where the readme is?

EstebanBorai: I think we could have a param for the publish command which takes the path and uses that if provided? Otherwise fallback to this?

EstebanBorai: I think this connector: https://github.com/infinyon/http-sink-connector is a good use case for this, wdyt?

morenol: Sounds good to me. Also for instance, in the kafka connector we right now have a big readme for both source and sink. Maybe we should split them and put them on each connector directory

EstebanBorai: Great insights! Will provide the option!

tests/cli/cdk_smoke_tests/cdk-basic.bats
```diff
@@ -67,5 +67,18 @@ setup_file() {
...
+    mkdir ../testing
+    echo "# My Cool Connector" > ../testing/README.md
+
+    cd $CONNECTOR_DIR
+    run $CDK_BIN publish --pack --readme ../testing/README.md
+    assert_success
 
+    cat ./.hub/package-meta.yaml | grep '../../testing/README.md'
+    assert_output --partial "../../testing/README.md"
+    assert_success
```
EstebanBorai, line 83: **Side Note**: The output is canonicalized! Thats why we went from `../testing/README.md` to `../../testing/README.md`

crates/cdk/src/publish.rs
```diff
@@ -52,6 +53,10 @@ pub struct PublishCmd {
 
     #[clap(long, hide_short_help = true)]
     remote: Option<String>,
+
+    /// Relative path to this connector package README
+    #[clap(long)]
+    readme: Option<PathBuf>,
```
morenol, line 59: maybe we should add in the future a way to other files too

morenol:
Could you update https://github.com/infinyon/fluvio/blob/3d212b696884ba60997c8189cebdbb9373db71c8/.github/workflows/connector-publish.yml#L8

We need to add a new input parameter and properly handle it. That way we can update our workflow in our connector repos to specify where the readme is

crates/cdk/src/publish.rs
sehz, line 59: why this is optional? by default, we should add README

sehz (changes requested):

crates/cdk/src/publish.rs
morenol: if it is not specified, it is taking the README from the same directory where the Connector.toml file is

EstebanBorai: `readme` is optional for the CLI, but by default CDK will find the README in the same directory as the `Connector.toml`.

If we happen to have a Readme in other directory, then we can overwrite the default behavior using the `readme` option.

If the readme is not present/found CDK will complain.

.github/workflows/connector-publish.yml
```diff
@@ -19,6 +19,9 @@ on:
       fail-fast:
         type: boolean
         default: true
+      readme_path:
+        type: string
+        default: './README.md'
```
EstebanBorai, line 24: By default take the readme in the same directory

.github/workflows/connector-publish.yml
```diff
@@ -64,7 +67,11 @@ jobs:
           cross build --profile ${{ inputs.rust-profile }}  --target ${{ matrix.rust-target }} -p ${{ inputs.package-name }}
       - name: Publish
         run: |
-          cdk publish --public-yes --target ${{ matrix.rust-target }} -p ${{ inputs.package-name }}
+          cdk publish \
+            --public-yes \
+            --target ${{ matrix.rust-target }} \
+            -p ${{ inputs.package-name }} \
+            --readme ${{ inputs.readme_path }}
```
EstebanBorai, line 74: The workflow will always specify a README, this may break current CI 🤔 
I would like to just skip if theres no `readme_path` specified by I didn't found a clean way to do it,
thinking if this needs to grow now into its own action package?

sehz: That's ok. If README is not there, should expect publish to fail

morenol: I think that it is ok, we just need to create an issue in each repo to update if needed and we should do it before next release

crates/cdk/src/publish.rs
sehz: if it is mandatory then we should make it so and provide default here instead in the GitHub Action

crates/cdk/src/publish.rs
```diff
@@ -142,9 +147,21 @@ pub fn package_push(opts: &PublishCmd, pkgpath: &str, access: &HubAccess) -> Res
...
 
-pub fn init_package_template(package_info: &PackageInfo, binary_arch: &str) -> Result<()> {
+pub fn init_package_template(
+    package_info: &PackageInfo,
+    binary_arch: &str,
+    readme_path: &Option<PathBuf>,
+) -> Result<()> {
     let connector_toml_path = find_connector_toml(package_info)?;
     let connector_metadata = ConnectorMetadata::from_toml_file(&connector_toml_path)?;
+    let readme_md_path = if let Some(readme_path) = readme_path {
```
sehz, line 157: This would be easier if option is mandatory

crates/cdk/src/publish.rs
```diff
@@ -176,6 +193,11 @@ pub fn init_package_template(package_info: &PackageInfo, binary_arch: &str) -> R
         binary_relative_path.unwrap_or_else(|| binary_path.to_string_lossy().to_string()), // if failed to get relative path, use absolute
     );
 
+    let readme_md_relative_path = package_meta_relative_path(&package_meta_path, &readme_md_path);
+    pm.manifest.push(
+        readme_md_relative_path.unwrap_or_else(|| readme_md_path.to_string_lossy().to_string()), // if failed to get relative path, use absolute)
```
sehz, line 198: path should take Path

sehz (changes requested):

.github/workflows/connector-publish.yml
EstebanBorai: Here is the set of issues to track progress:

- https://github.com/infinyon/kafka-connector/issues/37
- https://github.com/infinyon/sql-connector/issues/35
- https://github.com/infinyon/http-source-connector/issues/58
- https://github.com/infinyon/mqtt-connector/issues/35
- https://github.com/infinyon/http-sink-connector/issues/28

crates/cdk/src/publish.rs
EstebanBorai: For the CLI is not mandatory, what we do is:
1. If user provides a readme path then we use that
2. If the user doesn't provides a readme path, then we perform the current approach, which is checking for the README.md in the same path as the `Connector.toml`

Here is the code in question:

https://github.com/infinyon/fluvio/blob/0af53e3044be35acecd908e64c15bd63eed40203/crates/cdk/src/publish.rs#L157-L164

morenol:
There is a CDK test failing with `# output : Error: No "README.md" file found` Maybe we need to add a README to the connector test folder so it passes

crates/cdk/src/publish.rs
EstebanBorai: This is the commit in question: https://github.com/infinyon/fluvio/pull/3230/commits/b735e845d42d2700bdfdbab105c4c922a3ba1881

tests/cli/cdk_smoke_tests/cdk-basic.bats
```diff
@@ -60,11 +60,13 @@ setup_file() {
 @test "Pack connector" {
     # Pack when package meta doesn't exist
     cd $CONNECTOR_DIR
+    echo "# My Cool Connector" > README.md
```
morenol, line 63: I think that instead of this we should: either put a readme in connector template and put a readme in `connector/json-test-connector`

EstebanBorai: Well thought!

sehz (approved):
LGTM.  Nice work!

EstebanBorai:
bors retry

EstebanBorai:
bors retry

Issue #3169, opened 2023-04-19
[Bug]: Move `x records outputed` from `smdk test` to --verbose


ajhunyady:
Since most of the use cases we are currently facing are `json` it is be useful to pipe the result of `smdk test` to `jq`.

But the `x records outputed` like breaks this functionality.

Add a new flag `--verbose`, to `smdk test` and move `x records outputed` under this flag:

```
smdk test ... --verbose
x records outputed
...
```

TanNgocDo:
so, without "verbose" the output should be :
`
smdk test ...
record_1
record_2
.....

`
with verbose : 
`
smdk test ... --verbose
x records outputed
record_1
record_2
.....
`
right? @ajhunyady  ?

ajhunyady:
Hi @TanNgocDo, yes, the example provided looks right.
Thanks for looking into this.

TanNgocDo:
Thanks @ajhunyady , I have just made a simple PR

ajhunyady:
Thank you for the PR @TanNgocDo! The engineering team will review soon.

sehz:
done

PR #3234, opened 2023-05-07
[Merged by Bors] - Move x records outputed from smdk test to --verbose #3169

TanNgocDo:
This is to add verbose flag into smdk test : #3169

### DIFF
--- a/crates/smartmodule-development-kit/src/main.rs
+++ b/crates/smartmodule-development-kit/src/main.rs
@@ -10,6 +10,7 @@ mod set_public;
 use std::path::{PathBuf};
 use clap::Parser;
 use anyhow::Result;
+use tracing::debug;
 
 use cmd::SmdkCommand;
 
@@ -23,6 +24,6 @@ fn main() -> Result<()> {
 }
 
 pub(crate) fn read_bytes_from_path(path: &PathBuf) -> Result<Vec<u8>> {
-    println!("loading module at: {}", path.display());
+    debug!("loading module at: {}", path.display());
     std::fs::read(path).map_err(|err| anyhow::anyhow!("error reading wasm file: {}", err))
 }
--- a/crates/smartmodule-development-kit/src/main.rs
+++ b/crates/smartmodule-development-kit/src/main.rs
@@ -8,6 +8,7 @@ mod hub;
 mod set_public;
 
 use std::path::{PathBuf};
+
 use clap::Parser;
 use anyhow::Result;
 use tracing::debug;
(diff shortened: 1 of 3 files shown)


ajhunyady (changes requested):
Can you please also add a test for the `non-verbose` case?

```
run $SMDK_BIN test --text '["foo", "bar"]'
```

TanNgocDo:
> Can you please also add a test for the `non-verbose` case?
> 
> ```
> run $SMDK_BIN test --text '["foo", "bar"]'
> ```

Hi @ajhunyady , I have just added 2 non-verbose test. Thanks

TanNgocDo:
**Update** : I think I should change the **Generate and test filter-map**  



```  
# Test
    run $SMDK_BIN test --verbose --text '2'
    assert_output --partial "1 records outputed"
    assert_output --partial "1" #changed from 2 -> 1 (Have tested in my local the generated wasm smart module)
    assert_success
```
Thanks

ajhunyady:
@morenol please review.

morenol (commented):
LGTM

crates/smartmodule-development-kit/src/main.rs
```diff
@@ -10,6 +10,7 @@ mod set_public;
 use std::path::{PathBuf};
 use clap::Parser;
```
sehz, line 11: Minor nit. can you separate std lib from third party?

crates/smartmodule-development-kit/src/main.rs
```diff
@@ -23,6 +24,6 @@ fn main() -> Result<()> {
 }
 
 pub(crate) fn read_bytes_from_path(path: &PathBuf) -> Result<Vec<u8>> {
-    println!("loading module at: {}", path.display());
+    debug!("loading module at: {}", path.display());
```
sehz, line 27: use standard tracing format.  so it should be `debug(path = path.display(),"loading module")`

sehz (changes requested):
Looks good.  couple of minor nits

ajhunyady (approved):

crates/smartmodule-development-kit/src/main.rs
TanNgocDo: I have just updated it. Thanks

crates/smartmodule-development-kit/src/main.rs
TanNgocDo: updated it. Thanks

sehz (approved):
LGTM

sehz:
bors retry

PR #3236, opened 2023-05-08
[Merged by Bors] - improvement: add helper method to get utf8 from data

sehz:
Make it easier to get the utf8 string from record data, especially for consumers.   This helps unnecessary conversion to owned unless it is necessary.   Replaced existing examples with improved patterns to reduce unnecessary allocations and establish best practices for string conversion.  

Also, add  `get_key()` and `get_value()` to ConsumerRecord, which should be used rather than the existing `value()`, which returns `&[u8]`.   Minor clean-up to remove unnecessary abstractions 

Before:
```
        let string = String::from_utf8_lossy(record.value());
        println!("{string}");
 }
```

After:
```
      println!("{}", record.get_value().as_utf8_lossy_string());
}
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2840,7 +2840,7 @@ dependencies = [
 
 [[package]]
 name = "fluvio-protocol"
-version = "0.9.2"
+version = "0.9.3"
 dependencies = [
  "bytes 1.4.0",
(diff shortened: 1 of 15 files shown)


sehz:
Create separate issue for that

PR #3242, opened 2023-05-09
chore: Add message about record too large on produce cli

### LINKED ISSUE #3208: add more explanation to `ProducerError::RecordTooLarge`
Taken from discord:

<pre>
it's also easy to run into this with the CLI. I suspect that fluvio produce blah --file=some-large-file.json might be one of the first things the average user will try and that gave me
```
the given record is larger than the buffer max_size (16384 bytes)
```
which then sent me to rolling a custom connector instead where I ran into the issue mentioned above
might be worth special-casing that error in the CLI and suggest raising the batch size since (at least for me) it wasn't really obvious 
</pre>

morenol:
Fixes #3208


crates/fluvio-cli/src/client/produce/mod.rs
```diff
@@ -16,6 +16,7 @@ mod cmd {
     use std::path::PathBuf;
 
     use async_trait::async_trait;
+    use fluvio_protocol::link::ErrorCode;
```
sehz, line 19: move ErrorCode after third party

crates/fluvio-cli/src/client/produce/mod.rs
```diff
@@ -211,6 +213,35 @@ mod cmd {
...
+        ) -> Result<()> {
+            self.process_client_impl(_out, fluvio).await.map_err(|err| {
+                error!(%err, "Error producing messages",);
+
+                let err = if let Some(producer_error) =
+                    err.downcast_ref::<ProducerError>()
+                {
+                    match producer_error {
+                        ProducerError::RecordTooLarge ( size ) => {
+                            CliError::CollectedError(format!("Record too large. Try increasing the batch size ({} bytes) or reducing the record size enabling a compression algorithm.", size)).into()
```
sehz, line 225: instead of doing here.  Why not just add custom msg for `Producer::RecordTooLarge`?

sehz (changes requested):

crates/fluvio/src/producer/error.rs
```diff
@@ -28,3 +28,18 @@ pub enum ProducerError {
     #[error("the batch enqueue timeout limit reached")]
     BatchQueueWaitTimeout,
 }
+
+impl ProducerError {
+    /// Suggest possible fixes to the errors
+    pub fn user_message(&self) -> Option<String> {
+        match self {
+            Self::RecordTooLarge(size) => {
```
sehz, line 36: You can add direct into Error itself using `thiserror`

morenol: I though that that is for the display representation, therefore not sure if it was a good place to put the suggestion.Updating this to use it

sehz: A single unified message would be good enough

tests/cli/fluvio_smoke_tests/produce-error.bats
```diff
@@ -40,6 +40,19 @@ teardown_file() {
...
+@test "Produce message with internal error Record too large" {
+    run bash -c "yes a | tr -d "\n" |head -c 40000000 > $TOPIC_NAME.txt"
+    run bash -c 'timeout 15s "$FLUVIO_BIN" produce "$TOPIC_NAME" --raw --file $TOPIC_NAME.txt'
+    assert_failure
+    if [[ "$FLUVIO_CLI_RELEASE_CHANNEL" == "stable" ]]; then
+        skip "don't check output on stable version"
+    fi
+
+    # it can be Record too large or Message too large message.
+    assert_output --partial "Record too large"
```
morenol, line 53:
```suggestion
    assert_output --partial "Try increasing the producer batch size"
```

PR #3252, opened 2023-05-12
[Merged by Bors] - feat(cdk): prebuild artifacts for publish and deploy

EstebanBorai:
Abstracts the build command used for `build`, `test`, `deploy` and `publish` and then
makes sure commands relying on binaries from build process output (Artifacts) are present
by spawning the `build` process before proceeding.


galibey (approved):
LGTM

PR #3266, opened 2023-05-19
[Merged by Bors] - chore: improve error logs while rendering interpolated strings

morenol:
With this change, in case of undefined value, the log will be something like:

```
Error: undefined value at line 1:`hello ${{ some_undefined }}`
```


EstebanBorai (approved):

morenol (commented):
bors r+

PR #3276, opened 2023-05-24
[Merged by Bors] - feat: added `look_back` to smartmodule proc macro

### LINKED ISSUE #3273: Add `lock_back` function support to smartmodule proc macro
The new `lock_back` function will provide API for reading existing records from a SmartModule.

galibey:
Added `#[smartmodule(look_back)]` proc macro to support generating `look_back` function in the SmartModule. 

Fixes #3273

### DIFF
--- a/smartmodule/examples/filter_look_back/transforms.yaml
+++ b/smartmodule/examples/filter_look_back/transforms.yaml
@@ -1,0 +1,6 @@
+transforms:
+  - uses: infinyon/fluvio-smartmodule-filter-lookback@0.1.0
+    with:
+      lookback:
+        last: 10
+
--- a/smartmodule/examples/filter_look_back/transforms.yaml
+++ b/smartmodule/examples/filter_look_back/transforms.yaml
@@ -1,6 +1,5 @@
 transforms:
   - uses: infinyon/fluvio-smartmodule-filter-lookback@0.1.0
-    with:
-      lookback:
-        last: 10
+    lookback:
+      last: 10
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2812,7 +2812,7 @@ dependencies = [
 
 [[package]]
 name = "fluvio-protocol"
-version = "0.10.0"
+version = "0.10.1"
 dependencies = [
  "bytes 1.4.0",
  "content_inspector",
@@ -2955,7 +2955,7 @@ dependencies = [
 
 [[package]]
 name = "fluvio-smartmodule"
-version = "0.6.0"
+version = "0.7.0"
 dependencies = [
  "criterion",
  "eyre",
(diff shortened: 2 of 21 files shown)


digikata:
LGTM!

One nitpick, is there a recommended error for user look_back logic to use and attach "application" level context to? A ui-test where user code explicitly returns an error in the lookback fn might be good.

digikata (approved):

galibey:
> LGTM!
> 
> One nitpick, is there a recommended error for user look_back logic to use and attach "application" level context to? A ui-test where user code explicitly returns an error in the lookback fn might be good.

Added an example in ui-tests here https://github.com/infinyon/fluvio/pull/3276/commits/310f6a791a5041dce29a74ad1940c449b493cb2c

smartmodule/examples/filter_look_back/transforms.yaml
```diff
@@ -0,0 +1,6 @@
+transforms:
+  - uses: infinyon/fluvio-smartmodule-filter-lookback@0.1.0
+    with:
+      lookback:
```
sehz, line 4: There is no loopback param in the code

sehz: Not part of this issue but we should have ability to test loopback using SMDK

galibey: > There is no `loopback` in the code example. It should be removed. The loopback will be controlled by SPU

This parameter is consumed on SmartEngine and SPU levels. It will come in the next PRs.
The loopback is controlled by SPU, that's correct. But it still needs to get information of how many records need to be read and passed to the SM via `look_back` call, and that is where this parameter comes to play.

sehz: Can't rely on SmartModule on having parameters since it is dependent on author.   This need to be assume be defined by SPU

galibey: It is supposed to be defined not by the author but by the user of the SmartModule. Some SmartModules need only the last record (github stars, deduplication by monotonic ids, etc), some SmartModules need to read 1000 last records to build some data structure (bloom filters). Later, we may want to introduce asking for records based on time, etc. **This is a concern of the configuration of the SmartModule**. SPU has no this information.

galibey: This is how it was discussed in RFD https://github.com/infinyon/infinyon-rfd/pull/23 and already implemented

galibey: I updated example

sehz (approved):
LGTM.  Great job

sehz (commented):
Not showstopper but what happens in if there are other filters and transforms in chain.  They may have their own lookback param.  So maybe lookback should be associated with chain rather than individual transform steps

In that case transform may more look like this

transforms:
   lookback:
      last: 10
   steps:
      - mytranform
      - transform2
      - bloom3

PR #3283, opened 2023-05-25
[Merged by Bors] - add: fluvio-schema, batch id

digikata:
Add ability to set, encode and decode a schema id to a record batch

part of fluvio-schema tasks
this portion can be merged as a standalone feature

### DIFF
--- a/crates/fluvio-protocol/Cargo.toml
+++ b/crates/fluvio-protocol/Cargo.toml
@@ -72,3 +72,4 @@ fluvio-future = { workspace = true, features = [
     "net",
 ] }
 futures = { workspace = true }
+tracing-subscriber = { workspace = true,  features = ["env-filter", "fmt"] }
--- a/crates/fluvio-protocol/Cargo.toml
+++ b/crates/fluvio-protocol/Cargo.toml
@@ -72,4 +72,3 @@ fluvio-future = { workspace = true, features = [
     "net",
 ] }
 futures = { workspace = true }
-tracing-subscriber = { workspace = true,  features = ["env-filter", "fmt"] }
--- a/crates/fluvio-protocol/src/record/batch.rs
+++ b/crates/fluvio-protocol/src/record/batch.rs
@@ -23,6 +23,8 @@ const ATTR_SCHEMA_PRESENT: i16 = 0x10;
 const ATTR_COMPRESSION_CODEC_MASK: i16 = 0x07;
 pub const NO_TIMESTAMP: i64 = -1;
 
+const SCHEMA_ID_NULL: SchemaId = SchemaId(0u32);
+
 pub trait BatchRecords: Default + Debug + Encoder + Decoder + Send + Sync {
     /// how many bytes does record wants to process
     #[deprecated]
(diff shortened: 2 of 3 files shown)


crates/fluvio-protocol/Cargo.toml
```diff
@@ -72,3 +72,4 @@ fluvio-future = { workspace = true, features = [
     "net",
 ] }
 futures = { workspace = true }
+tracing-subscriber = { workspace = true,  features = ["env-filter", "fmt"] }
```
sehz, line 75: why this is needed here?

digikata: It was a needed dev-dependency to debug w/ some of the tracing messages running unit test during development. But not regularly needed. remove?

sehz: yes if not needed

crates/fluvio-protocol/src/record/batch.rs
```diff
@@ -67,11 +68,24 @@ pub const BATCH_PREAMBLE_SIZE: usize = size_of::<Offset>()     // Offset
 
 pub const BATCH_FILE_HEADER_SIZE: usize = BATCH_PREAMBLE_SIZE + BATCH_HEADER_SIZE;
 
+#[derive(Clone, Default, Debug, Encoder, PartialEq)]
+pub struct SchemaId(u32);
+
+impl Decoder for SchemaId {
```
sehz, line 74: is this needed?

crates/fluvio-protocol/src/record/batch.rs
```diff
@@ -67,11 +68,24 @@ pub const BATCH_PREAMBLE_SIZE: usize = size_of::<Offset>()     // Offset
...
+        Ok(())
+    }
+}
+
 #[derive(Default, Debug)]
 pub struct Batch<R = MemoryRecords> {
     pub base_offset: Offset,
     pub batch_len: i32, // only for decoding
     pub header: BatchHeader,
+    pub schema_id: Option<SchemaId>,
```
sehz, line 88: need version

sehz: it might be better to just use SchemaId instead of using option since SchemaId can tell if schema is present

crates/fluvio-protocol/src/record/batch.rs
```diff
@@ -318,15 +353,22 @@ where
...
         self.decode_from_file_buf(src, version)?;
-
-        let batch_len = self.batch_len as usize - BATCH_HEADER_SIZE;
-        let mut buf = src.take(batch_len);
-        if buf.remaining() < batch_len {
+        let rec_len = if self.header.has_schema() {
+            let mut sid = SchemaId(0);
+            sid.decode(src, version)?;
+            self.schema_id = Some(sid);
+            trace!("schema_id: {:?}", self.schema_id());
```
sehz, line 360: use trace!(schema_id,...)

sehz (changes requested):

crates/fluvio-protocol/src/record/batch.rs
digikata: Sounds good, the struct in memory is a little more compact w/o the option too.

sehz (approved):

PR #3311, opened 2023-06-07
[Merged by Bors] - feat: measure metrics on `look_back` calls

galibey:
We will measure `bytes_in` and `invocations` for `look_back` in the same way as for `transform`.


sehz (approved):
LGTM

PR #3317, opened 2023-06-08
[Merged by Bors] - feat(smdk): support `lookback` in test

### LINKED ISSUE #3315: Add support of `lock_back` to SMDK tool
We should provide an ability to test `look_back` functions in SMDK tool.

galibey:
Added an ability to specify the lookback parameter and provide existing records to simulate data in the topic.

For example, here we run `test` and pass input record "1" and provide other records "2" and "3" as they would exist in the topic before:
```bash
smdk test --text "1" --lookback-last 1 --record "2" --record "3" 
```
the `look_back` function will receive record "3".

Closes #3315

### DIFF
--- a/crates/fluvio-smartengine/src/engine/config.rs
+++ b/crates/fluvio-smartengine/src/engine/config.rs
@@ -61,6 +61,10 @@ impl SmartModuleConfig {
     pub(crate) fn version(&self) -> i16 {
         self.version.unwrap_or(DEFAULT_SMARTENGINE_VERSION)
     }
+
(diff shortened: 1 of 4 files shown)


sehz (changes requested):
Can we add CLI test?

galibey:
> Can we add CLI test?

added in https://github.com/infinyon/fluvio/pull/3317/commits/d1d962aaf5502334b4c02441f626a9c73e16b02a

sehz (approved):
Nice

PR #3319, opened 2023-06-08
[Merged by Bors] - chore: remove old connector crd

morenol:


sehz (changes requested):
Need to bump sys chart version

sehz (approved):

PR #3324, opened 2023-06-12
fix: fluvio-socket, improve connect error message

digikata:
output a nicer messages instead of just printing "socket"


crates/fluvio-socket/src/error.rs
```diff
@@ -12,9 +12,7 @@ pub enum SocketError {
 
 impl From<IoError> for SocketError {
     fn from(err: IoError) -> Self {
-        SocketError::Io {
-            source: err,
-            msg: "".to_string(),
-        }
+        let msg = err.to_string();
+        SocketError::Io { source: err, msg }
```
sehz, line 16: this could be shortend without intermediate var
```
SocketError::Io { source: err, msg: err_to_string() }
```

crates/fluvio-socket/src/socket.rs
```diff
@@ -95,7 +95,14 @@ impl FluvioSocket {
     ) -> Result<Self, SocketError> {
         debug!("connecting to addr at: {}", addr);
 
-        let (write, read, fd) = connector.connect(addr).await?;
+        let (write, read, fd) = connector.connect(addr).await.map_err(|e| {
+            let emsg = e.to_string();
+            SocketError::Io {
+                source: e,
```
sehz, line 101:
```
source: e.to_string()
```

sehz (changes requested):
couple of nits

crates/fluvio-socket/src/error.rs
digikata: Hmm, that gives me a borrow error because err is moved to source.

sehz: Oh right.  silly rust borrow checker limitation

sehz (approved):

PR #3333, opened 2023-06-19
[Merged by Bors] - Use separate argument to specify dev/prod in smartmodule-publish workflow

ozgrakkurt:


sehz (changes requested):
Since this is dependent by other repo, we should create another version rather than replace it

sehz:
sounds good

sehz (approved):

.github/workflows/smartmodule-publish.yml
```diff
@@ -44,7 +40,10 @@ jobs:
       - name: Install Fluvio SMDK
         run: fluvio install smdk
       - name: Fluvio Login
-        run: fluvio cloud login --email ${{ secrets.CLOUD_USER_EMAIL }} --password ${{ secrets.CLOUD_USER_PASSWORD }} ${{ inputs.cloud-url != '' && '--remote' || '' }} ${{ inputs.cloud-url }}
+        run: |
+          fluvio cloud login --email ${{ inputs.target_prod && secrets.HUB_USER_EMAIL || secrets.DEV_HUB_USER_EMAIL }} \
+          --password ${{ inputs.target_prod && secrets.HUB_USER_PASSWORD || secrets.DEV_HUB_USER_PASSWORD }} \
+          ${{ !inputs.target_prod && ('--remote ' + vars.DEV_CLOUD_URL) || '' }}
```
EstebanBorai, line 46: I think `secrets` must be introduced as input given that these are parameters.
Perhaps its not necessary to put them behind the `secrets` object,

Take a look at how GitHub does here: https://github.com/actions/add-to-project/blob/main/action.yml#L10-L12

The `github-token` input is just another string. I think we could use two regular inputs for User Email and Password.

sehz: see: https://docs.github.com/en/actions/security-guides/encrypted-secrets#creating-encrypted-secrets-for-a-repository.   GITHUB_TOKEN is special...

sehz: and https://docs.github.com/en/actions/using-workflows/reusing-workflows#using-inputs-and-secrets-in-a-reusable-workflow.

sehz (approved):

EstebanBorai (approved):

PR #3346, opened 2023-06-23
[Merged by Bors] - fix: use new variables in smartmodule-publish workflow

ozgrakkurt:


.github/workflows/smartmodule-publish.yml
```diff
@@ -50,8 +48,8 @@ jobs:
         run: fluvio install smdk
       - name: Fluvio Login
         run: |
-          fluvio cloud login --email ${{ inputs.target_prod && secrets.HUB_USER_EMAIL || secrets.DEV_HUB_USER_EMAIL }} \
-          --password ${{ inputs.target_prod && secrets.HUB_USER_PASSWORD || secrets.DEV_HUB_USER_PASSWORD }} \
+          fluvio cloud login --email ${{ secrets.ORG_HUB_SA_NAME }} \
+          --password ${{ !inputs.target_prod && secrets.ORG_HUB_SA_PASSWD_DEV || secrets.ORG_HUB_SA_PASSWD_PROD }} \
```
digikata, line 52: The passwords and PROD and DEV are swapped.  Looks good otherwise

digikata (changes requested):

.github/workflows/smartmodule-publish.yml
ozgrakkurt: should be ok because of the `!`

sehz (approved):

.github/workflows/smartmodule-publish.yml
digikata: Oh wait, why invert inputs.targets_prod?

sehz:
Generally it would be easier to test workflow in specific repo and then abstract into a common reusable

ozgrakkurt:
> Generally it would be easier to test workflow in specific repo and then abstract into a common reusable

makes sense, I'll do it like that next time

Issue #3353, opened 2023-06-28
Error rework to make errors easier to understand


ozgrakkurt:
Currently most error types are like this: https://github.com/infinyon/fluvio/blob/693a2923a36fe4cd0222d70ef34fcc9a40243009/crates/fluvio-auth/src/error.rs#L8

which hides the underlying error and would just print `io error` in the console.

There is a similar error where `fluvio` prints `socket` into the console in place of an error.

Maybe we can do an error rework to fix these issues?

digikata:
Yes, fixing these a little at a time should improve our errors. Some of the socket errors were already fixed in this release https://github.com/infinyon/fluvio/pull/3324/files

PR #3354, opened 2023-06-28
feat: print causes for errors

ozgrakkurt:
closes https://github.com/infinyon/fluvio/issues/3353

This is WIP, but requesting reviews to get feedback on the approach.

I'm just doing this pattern:
```
#[error("Failed to parse format string: {0}")]
TemplateError(#[from] TemplateError),
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1149,7 +1149,6 @@ version = "7.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9ab77dbd8adecaf3f0db40581631b995f312a8a5ae3aa9993188bb8f23d83a5b"
 dependencies = [
- "crossterm 0.26.1",
  "strum",
  "strum_macros",
  "unicode-width",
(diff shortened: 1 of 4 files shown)


digikata:
I think this is fine for this error, but I'd mention that each error needs to be looked at individually. Sometimes you don't want to print the error info directly, and the code injecting the error or passing it along should inject context.

sehz:
This is an improvement.  However, some of these errors are unnecessary if we convert to anyhow

ozgrakkurt:
To keep PR simple, is it ok to just do this on all errors for this PR and then do a selective deletion of extra info?

sehz:
Converting to anyhow will be more productive

sehz:
Yes, unless there is an explicit need for modeling error.  The vast majority of these are just passed thru which are unnecessary

ozgrakkurt:
> Yes, unless there is an explicit need for modeling error. The vast majority of these are just passed thru which are unnecessary

got it, makes sense, we don't even match on any of them probably

nacardin (approved):
LGTM

digikata (approved):

ozgrakkurt:
closing this in favor of smaller PRs per crate

PR #3369, opened 2023-07-04
[Merged by Bors] - feat: support time bound in lookback engine

### LINKED ISSUE #3357: Support time bound in lookback engine
We need to support the lookback engine querying existing records by "age" - read records that are not older than the given age parameter:
```yaml
transforms:
  - uses: infinyon/bloom-filter@0.1.0
    with:
      lookback:
        age: 2d
```

galibey:
Closes #3357

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2821,7 +2821,7 @@ dependencies = [
 
 [[package]]
 name = "fluvio-protocol"
-version = "0.10.2"
+version = "0.10.3"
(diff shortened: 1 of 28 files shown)


digikata (approved):
LGTM, there might be some interaction with incoming timestamp changes from Luis, but I like the FileRecordIterator approach. Had to read the age, age+look logic carefully to make sure I understood what was going on there.

Issue #3365, opened 2023-07-03
[Bug]: `smdk publish` with `--push` flag is broken


ozgrakkurt:
When I do:
- smdk build
- smdk publish --pack

And then do:
- smdk publish --push

I consistently get this error:
```
Error: Package missing file package-meta.yaml
```
Even though the file is there in the ipkg file.

Probably due to some logic error in `smdk`.

PR #3370, opened 2023-07-04
[Merged by Bors] - fix: fix --push flag on smdk and cdk publish

ozgrakkurt:
closes https://github.com/infinyon/fluvio/issues/3365

### DIFF
--- a/crates/cdk/src/publish.rs
+++ b/crates/cdk/src/publish.rs
@@ -9,6 +9,7 @@ use std::path::PathBuf;
 use anyhow::Context;
 use anyhow::Result;
 use cargo_builder::package::PackageInfo;
+use cargo_builder::package::PackageOption;
 use clap::Parser;
 
 use fluvio_connector_package::metadata::ConnectorMetadata;
@@ -64,17 +65,50 @@ impl PublishCmd {
     pub(crate) fn process(&self) -> Result<()> {
         let access = HubAccess::default_load(&self.remote)?;
 
-        if !self.pack && self.push {
-            let pkgfile = self
-                .ipkg_file
-                .as_ref()
-                .ok_or_else(|| anyhow::anyhow!("need to specify --ipkg-file when using --push"))?;
-            package_push(self, pkgfile, &access)?;
-            return Ok(());
+        match (self.pack, self.push) {
+            (false, false) | (true, true) => {
+                let opt = self.package.as_opt();
+                let hubdir = self.run_in_cargo_project(&opt)?;
+                let pkgmetapath = self.package_meta_path(&hubdir);
+                let pkgdata = package_assemble(pkgmetapath, &opt.target, &access)?;
+                package_push(self, &pkgdata, &access)?;
+                self.cleanup(&hubdir)?;
+            }
+
+            // --pack only
+            (true, false) => {
+                let opt = self.package.as_opt();
+                let hubdir = self.run_in_cargo_project(&opt)?;
+                let pkgmetapath = self.package_meta_path(&hubdir);
+                package_assemble(pkgmetapath, &opt.target, &access)?;
+                self.cleanup(&hubdir)?;
+            }
+
+            // --push only, needs ipkg file
+            (false, true) => {
+                let pkgfile = self.ipkg_file.as_ref().ok_or_else(|| {
+                    anyhow::anyhow!("need to specify --ipkg-file when using --push")
+                })?;
+                package_push(self, pkgfile, &access)?;
+            }
         }
 
-        let opt = self.package.as_opt();
-        let package_info = PackageInfo::from_options(&opt)?;
+        Ok(())
+    }
+
+    fn package_meta_path(&self, hubdir: &Path) -> PathBuf {
+        self.package_meta
+            .as_ref()
+            .map(PathBuf::from)
+            .unwrap_or_else(|| hubdir.join(hubutil::HUB_PACKAGE_META))
+    }
+
+    // This gets run only if the command should be run in the cargo project folder
+    // of the connector
+    //
+    // returns hubdir
+    fn run_in_cargo_project(&self, opt: &PackageOption) -> Result<PathBuf> {
+        let package_info = PackageInfo::from_options(opt)?;
         let hubdir = package_info.package_relative_path(DEF_HUB_INIT_DIR);
 
         info!(
@@ -83,7 +117,7 @@ impl PublishCmd {
             package_info.package_path().to_string_lossy()
         );
 
-        self.cleanup(&package_info)?;
+        self.cleanup(&hubdir)?;
 
         if !self.no_build {
             build_connector(&package_info, BuildOpts::with_release(opt.release.as_str()))?;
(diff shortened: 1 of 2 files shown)


sehz:
Not quit understanding issue.  Whats need to modify CLI?

crates/cdk/src/publish.rs
```diff
@@ -101,11 +115,7 @@ impl PublishCmd {
 
             // --push only, needs ipkg file
             (false, true) => {
-                let pkgfile = &self
-                    .package_meta
```
galibey, line 105: We should put the logic from lines 68-72 here and remove `unreachable`:
```
            // --push only, needs ipkg file
            (false, true) => {
                let pkgfile = &self
                    .ipkg_file
                    .as_ref()
                    .ok_or_else(|| anyhow::anyhow!("need to specify --ipkg-file when using --push"))?;
                package_push(self, pkgfile, &access)?;
            }
```

We should also make `ipkg_file` required if  `push` is specified using clap annotations.

galibey (changes requested):

ozgrakkurt:
> Not quit understanding issue. Whats need to modify CLI?

Currently, it tries to run `PackageInfo::from_options(&opt)` even when only `--push` is specified which should only require the `ipkg` file. Also there is no way to specify path to the ipkg file

crates/cdk/src/publish.rs
ozgrakkurt: we should decide on how to organize this and apply it to both smdk and cdk. Currently cdk has `no_build` and smdk doesn't have it as well

galibey: It is probably the time to re-organize the whole function. Split it into separate functions for better readability.

Because the current change with one more if and `unreachable` macro doesn't look good.

ozgrakkurt: I refactored using functions but don't thing there is a way to specify to require an arg if one flag is true and other is false in clap

ozgrakkurt:
> Why is ipkg required? Shouldn't publish create ipkg?

Effectively the push stage only takes the ipkg that is produced by pack stage and publishes it. So when run with only `--push` we should only require the ipkg file

crates/cdk/src/publish.rs
ozgrakkurt: created issue about `no-build` flag https://github.com/infinyon/fluvio/issues/3371

crates/cdk/src/publish.rs
```diff
@@ -59,70 +64,76 @@ pub struct PublishCmd {
...
+            .as_ref()
+            .map(PathBuf::from)
+            .unwrap_or_else(|| hubdir.join(hubutil::HUB_PACKAGE_META))
+    }
+
+    // This gets run only if the command should be run in the cargo project folder
+    // of the connector
+    //
+    // returns hubdir
+    fn run_in_cargo_project(&self, opt: &PackageOption) -> Result<PathBuf> {
```
galibey, line 110: The function is called `run_in_cargo_project` but it does: cleanup, building connector, init package, check visibility.

crates/cdk/src/publish.rs
```diff
@@ -59,70 +64,76 @@ pub struct PublishCmd {
...
+            build_connector(&package_info, BuildOpts::with_release(opt.release.as_str()))?;
+        }
+
+        init_package_template(&package_info, &self.readme)?;
+        check_package_meta_visiblity(&package_info)?;
+
+        Ok(hubdir)
+    }
+
+    fn cleanup(&self, hubdir: &Path) -> Result<()> {
```
galibey, line 132: it doesn't need to have a reference to `self`

crates/cdk/src/publish.rs
galibey: I was thinking more of having high level methods `push`, `pack` and `build` and hiding specific things inside.

galibey (changes requested):

crates/cdk/src/publish.rs
ozgrakkurt: it is more of a `if` so this gets run if we are running in a cargo project

ozgrakkurt:
> We should smdk and cdk consistent.
> 
> Build - produce binary for testing Publish - create hub pkg and upload

Then we can't separate building the artifact from publishing it from what I am seeing on the current code this is what is intended but the statements before the match break this because it expects to run in the cargo project but only needs/uses ipkg file when only --push is specified

crates/cdk/src/publish.rs
ozgrakkurt: high level methods don't really work without this function because we want to run these in two cases and not run them on one (only --push) case. High level methods would be the statements in the match expression

galibey: Ok, let's return to your first version but instead of `unreachable` just print and return an error.

galibey: We can come up with a new structure later, for now, let's just fix a bug with minimal refactoring.

ozgrakkurt: tbh, I don't want to revert because imo this is cleaner and also reverting will take more time

galibey: Again, a function that does a lot of things, and returns a path to hub directory and with name `run_in_cargo_project` is very far from clean and readable code.

galibey:
`publish --push` should only look for a ipkg file, that is correct, but why there is an option to specify ipkg path? It should use only one location, where is pack operation writes to.

ozgrakkurt:
> `publish --push` should only look for a ipkg file, that is correct, but why there is an option to specify ipkg path? It should use only one location, where is pack operation writes to.

User should be able to publish with only a single ipkg file, this is the case in reusable workflows for example, it just needs to pull the ipkg artifact and not care about folder structure or anything else. One thing to do is we can make `--ipkg` optional and look under the `.hub` folder by default.

crates/cdk/src/publish.rs
```diff
@@ -33,6 +34,10 @@ pub struct PublishCmd {
 
     pub package_meta: Option<String>,
 
+    /// path to the ipkg file, used when --push is specified
+    #[arg(long)]
+    ipkg: Option<String>,
```
sehz, line 39: This is only should be used in conjunction with `--push`

crates/cdk/src/publish.rs
```diff
@@ -59,70 +64,76 @@ pub struct PublishCmd {
 impl PublishCmd {
     pub(crate) fn process(&self) -> Result<()> {
         let access = HubAccess::default_load(&self.remote)?;
```
sehz, line 66: consolidate publish with `smdk` so have consistent publishing exp

crates/cdk/src/publish.rs
```diff
@@ -59,70 +64,76 @@ pub struct PublishCmd {
...
+                let opt = self.package.as_opt();
+                let hubdir = self.run_in_cargo_project(&opt)?;
+                let pkgmetapath = self.package_meta_path(&hubdir);
                 let pkgdata = package_assemble(pkgmetapath, &opt.target, &access)?;
                 package_push(self, &pkgdata, &access)?;
+                Self::cleanup(&hubdir)?;
             }
 
             // --pack only
             (true, false) => {
```
sehz, line 79: use common pack as in the `smdk`

crates/cdk/src/publish.rs
```diff
@@ -59,70 +64,76 @@ pub struct PublishCmd {
...
+            build_connector(&package_info, BuildOpts::with_release(opt.release.as_str()))?;
+        }
+
+        init_package_template(&package_info, &self.readme)?;
+        check_package_meta_visiblity(&package_info)?;
+
+        Ok(hubdir)
+    }
+
+    fn cleanup(hubdir: &Path) -> Result<()> {
```
sehz, line 132: same as `smdk`.  can move this to hubutil

sehz (changes requested):
The publishing logic is almost almost same as smdk.  let's consolidate them

ozgrakkurt:
> The publishing logic is almost almost same as smdk. let's consolidate them

I don't think we should do it in this PR, we can just merge this as it is functional an needed and create issue for this. I also need this fix for publish workflows

crates/cdk/src/publish.rs
ozgrakkurt: It is debatable if it is cleaner than putting the if at the start of the function and having a needless arm in match that references the if at the end of the function. It is code to delete/create folders and make http requests in the end don't need to work days to make it perfect

crates/cdk/src/publish.rs
ozgrakkurt: Yes but we also allow both `--pack` and `--push` at the same time, in that case this would mean nothing since push will use the artifact from pack step

digikata:
Let's move this forward with this plan:
- fix the default path/search with `--ipkg` so it works w/o a path
- keep accepting a path as override for the ipkg for workflow automation &  testing
- otherwise merge as is
- open an issue so we can identify longer term refactoring, and if needed rfd ui enhancements to describe common ui between smdk and cdk.

PR #3378, opened 2023-07-06
[Merged by Bors] - chore: update syn on fluvio-smartmodule-derive

morenol:
`param` is not being passed to smartmodule main function for a while, so I removed these tests and the references to that


sehz (approved):
LGTM

crates/fluvio-smartmodule-derive/src/ast.rs
```diff
@@ -30,30 +34,27 @@ pub enum SmartModuleKind {
...
+            "init" => Some(Self::Init),
+            "look_back" => Some(Self::LookBack),
+            _ => None,
+        };
+
+        let ss_type = maybee_ss_type
+            .ok_or_else(|| SynError::new(meta.path.span(), "Invalid SmartModule type"))?;
 
-        Ok(ss_type)
+        Ok(Some(ss_type))
```
galibey, line 57: If this function is capable to return always value, why do we make it optionable and then deal with option instead of value?

sehz (commented):
Not sure best practice with new sync API.  However at least for this fn signature, it is result of option which force to return.  Maybe it is possible to combine last two expressions to make it more elegant

Issue #3235, opened 2023-05-08
Tech debt: SPU storage separation


sehz:
As part of Materialize view, SPU needs to take more workloads.   To prepare that,  Storage Replication should be moved to separate crates

PR #3385, opened 2023-07-11
[Merged by Bors] - feat: added topic-level deduplication mechanism

galibey:
Upd.: Split into two parts: [first](https://github.com/infinyon/fluvio/pull/3392)

Added topic-level deduplication mechanism. 
Key impacted components:
 - Topic Config
 - Fluvio CLI (topic creation, topic list, topic describe)
 - SPU (replica and producer handler)
 - SC
 - K8s CRDs (topic and partition)

The feature requires for Dedup SmartModule filter to be provided. It is coming, but the one from the examples can be used to try it out:
```bash
cd smartmodule/examples/filter_hashset
smdk build
smdk load
```
then create a file `topic.yaml`:
```yaml
version: 0.1.0
meta:
  name: dedup-topic
deduplication:
  bounds:
    count: 10
  filter:
    transform:
      uses: infinyon/fluvio-smartmodule-filter-hashset@0.1.0
```
and 
```bash
fluvio topic create -c topic.yaml
```
with such configuration the window will be last 10 records.

Closes infinyon/roadmap#114

### DIFF
--- a/crates/fluvio-spu/src/core/global_context.rs
+++ b/crates/fluvio-spu/src/core/global_context.rs
@@ -11,7 +11,7 @@ use tracing::{debug, error, instrument};
 
 use fluvio_controlplane_metadata::partition::Replica;
 use fluvio_types::SpuId;
-use fluvio_storage::{ReplicaStorage};
+use fluvio_storage::ReplicaStorage;
 
 use crate::config::SpuConfig;
 use crate::replication::follower::FollowersState;
@@ -159,7 +159,7 @@ where
 mod file_replica {
 
     use fluvio_controlplane::{ReplicaRemovedRequest, UpdateReplicaRequest};
-    use fluvio_storage::{FileReplica};
+    use fluvio_storage::FileReplica;
     use flv_util::actions::Actions;
     use tracing::{trace, warn};
 
@@ -186,7 +186,12 @@ mod file_replica {
                 old_leader = old_replica.leader
             )
         )]
-        pub async fn promote(&self, new_replica: &Replica, old_replica: &Replica) {
+        pub async fn promote(
+            &self,
+            new_replica: &Replica,
+            old_replica: &Replica,
+        ) -> Vec<ReplicaChange> {
+            let mut outputs = Vec::new();
             if let Some(follower_replica) = self
                 .followers_state()
                 .remove_replica(old_replica.leader, &old_replica.id)
@@ -197,17 +202,24 @@ mod file_replica {
                     "old follower replica exists, promoting to leader"
                 );
 
-                self.leaders_state()
+                if let Err(err) = self
+                    .leaders_state()
                     .promote_follower(
                         self.config().into(),
                         follower_replica,
                         new_replica.clone(),
                         self.status_update_owned(),
+                        self,
                     )
-                    .await;
+                    .await
+                {
+                    error!("follower promotion failed: {err}");
+                    outputs.push(ReplicaChange::StorageError(err));
+                };
             } else {
                 error!("follower replica {} didn't exists!", old_replica.id);
             }
+            outputs
         }
 
         pub async fn apply_replica_update(
@@ -294,7 +306,8 @@ mod file_replica {
                             // check for leader change
                             if new_replica.leader != old_replica.leader {
                                 if new_replica.leader == local_id {
-                                    self.promote(&new_replica, &old_replica).await
+                                    outputs
+                                        .append(&mut self.promote(&new_replica, &old_replica).await)
                                 } else {
                                     // we are follower
                                     // if we were leader before, we demote out self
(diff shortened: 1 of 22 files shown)


crates/fluvio-spu/src/core/global_context.rs
```diff
@@ -197,17 +202,24 @@ mod file_replica {
...
+                    error!("follower promotion failed: {err}");
+                    outputs.push(ReplicaChange::StorageError(err));
+                };
             } else {
                 error!("follower replica {} didn't exists!", old_replica.id);
             }
+            outputs
         }
 
         pub async fn apply_replica_update(
```
sehz, line 225: Is these formatting changes only?  If so, would prefer to move out of this PR since it's not related to this feature

crates/fluvio-spu/src/replication/leader/leaders_state.rs
```diff
@@ -1,19 +1,14 @@
-use std::{
-    ops::{Deref},
-};
+use std::ops::Deref;
```
sehz, line 1: Same comment as previously. let's move formatting changes out of this PR

sehz (changes requested):
first pass

sehz:
This is a rather large PR.  Can this break it into smaller chunks? Suggest
1) CRD and API-related changes
2) SPU related changes

crates/fluvio-spu/src/core/global_context.rs
galibey: This is not formatting. This is propagating errors up from replica initialization.

crates/fluvio-spu/src/replication/leader/leaders_state.rs
galibey: If I modify a file where there are warnings from the rust-analyzer, I fix them. In that way, such warnings will be gradually fixed. Nobody ever will fix those warnings separately unless all tasks on the project are gone.

galibey:
> This is a rather large PR. Can this break it into smaller chunks? Suggest
> 
>     1. CRD, API, CLI related changes (all configuration related)
> 
>     2. SC/SPU changes

Ok, this is big, indeed. I am going to split it into parts.

galibey:
This PR only contains SPU/SC changes

crates/fluvio-spu/src/replication/leader/replica_state.rs
```diff
@@ -311,11 +320,16 @@ where
...
     /// then update our follower's leo
     #[instrument(skip(self, records, notifiers))]
-    pub async fn write_record_set<R: BatchRecords>(
+    pub async fn write_record_set(
         &self,
-        records: &mut RecordSet<R>,
+        records: &mut RecordSet<RawRecords>,
         notifiers: &FollowerNotifier,
     ) -> Result<(Offset, Offset, usize)> {
+        self.transform(records).await?;
```
sehz, line 328: Could you tell me why transformation is performed here instead of `produce_handler`? Replication logic should be decoupled event processing.  Meaning all write goto leader

sehz (changes requested):

crates/fluvio-spu/src/replication/leader/replica_state.rs
galibey: - `produce_handler` handles only input from producers, but potentially, it could be some topic to topic data movements (derived topic or something similar) which also should uphold deduplication guarantees.
 - `produce_handler` doesn't have access to a particular replica's lifecycle (whether it's added or promoted or removed), so keeping track of all SM contexts in there will be hard and error-prone.
 - in general, we plan to use this replica sm context for schema management as well, there could be more. From the architectural point of view, it will be too much overloading for  `produce_handler` which essentially is a service, which main responsibility is to handle API requests.

sehz: This is per design.  SPU's `produce` lifecycle shouldn't deal with the replica's lifecycle.  Our replication model is a single master writer.  The replication should not deal with SM, which is only relevant for a writer.  Otherwise, this will result in inconsistent replication.  The promotion/demotion replication cycle should not impact write processing.

In the future, replication needs to move to its own crate and needs to be firewalled from the rest of SPU similar to storage.

galibey: > The replication should not deal with SM

This is a question to the entire feature design then, then to the code. In the design https://github.com/infinyon/roadmap/blob/main/rfd/deduplication-rfd.md it was clearly stated that `LeaderReplicaState` owns the SM context.

> Otherwise, this will result in inconsistent replication

The replication model is not affected at all. The SM runs only on LeaderReplicaState. The transformation happens on write before data is sent to the storage. All replication events read from the storage.

> The promotion/demotion replication cycle should not impact write processing.

It doesn't. Why do you think it does?

sehz: After offline discussion, this approach seems to be good enough for now since we don't have good leader replica abstraction which is necessary for https://github.com/infinyon/fluvio/issues/3235.

sehz (approved):
LGTM.   Great work interleaving Smartmodule in existing replica lifecycle

PR #3391, opened 2023-07-12
[Merged by Bors] - Copy keys-value output behavior from `fluvio consume` to `smdk test`

### LINKED ISSUE #3375: Add `  -k, --key-value` to smdk test
**What happened**
Deduplication uses `record-key` to remove duplicates.
In some cases, users need to generate a `record-key` from the record value with a smartmodule `map`.

While this is possible today,`smdk test` does not return the key in the output, which makes the result difficult to test.

**Expected behavior**
`smdk test` command should follow the same pattern as `fluvio consume`, where we can use the following parameter to print the key & value:

```  
   -k, --key-value
        Print records in "[key] value" format, with "[null]" for no key
```

The ask it to implement the same parameter in `smdk test`:

```
smdk test -k

[null] {"record": "one"} => if there is no key
 key1  {"record": "one"} => if there is a key
```

tjtelan:
Fixes #3375

### DIFF
--- a/crates/smartmodule-development-kit/src/test.rs
+++ b/crates/smartmodule-development-kit/src/test.rs
@@ -41,6 +41,10 @@ pub struct TestCmd {
     /// Key to use with the test record(s)
     key: Option<String>,
 
+    /// Print records in "[key] value" format, with "[null]" for no key
+    #[arg(short, long)]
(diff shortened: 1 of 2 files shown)


digikata (approved):
LGTM, clean update

PR #3392, opened 2023-07-12
[Merged by Bors] - feat: added topic deduplication mechanism 1/2

galibey:
This is the first part of [Dedup Impl PR](https://github.com/infinyon/fluvio/pull/3385) split for a more convenient review process.

### DIFF
--- a/crates/fluvio-controlplane-metadata/src/topic/deduplication.rs
+++ b/crates/fluvio-controlplane-metadata/src/topic/deduplication.rs
@@ -1,0 +1,65 @@
+use std::{time::Duration, collections::BTreeMap};
+
+use derive_builder::Builder;
+use fluvio_protocol::{Encoder, Decoder};
+
+#[derive(Debug, Default, Builder, Clone, PartialEq, Eq, Encoder, Decoder)]
+#[cfg_attr(
+    feature = "use_serde",
+    derive(serde::Serialize, serde::Deserialize),
+    serde(rename_all = "kebab-case")
+)]
+pub struct Deduplication {
+    pub bounds: Bounds,
+    pub filter: Filter,
+}
+
+#[derive(Debug, Default, Builder, Clone, PartialEq, Eq, Encoder, Decoder)]
+#[cfg_attr(
+    feature = "use_serde",
+    derive(serde::Serialize, serde::Deserialize),
+    serde(rename_all = "kebab-case")
+)]
+pub struct Bounds {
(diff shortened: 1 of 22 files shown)


crates/fluvio-protocol/src/link/error_code.rs
```diff
@@ -182,6 +182,14 @@ pub enum ErrorCode {
     #[fluvio(tag = 9000)]
     #[error("a compression error occurred in the SPU")]
     CompressionError,
+
+    // Deduplication
+    #[fluvio(tag = 10000)]
```
sehz, line 187: We should avoid adding new error codes until a new strategy is decided.  Not show stopper

k8-util/helm/fluvio-sys/templates/crd_topic.yaml
```diff
@@ -90,6 +90,33 @@ spec:
                     maxPartitionSize:
                       type: integer
                       minimum: 2048
+                deduplication:
```
sehz, line 93: Bit concerned about duplicating schema.   let's create issue to see if we can have better reusability

sehz (approved):
a couple of minor comments.  otherwise LGTM.  Nice work

crates/fluvio-controlplane-metadata/src/topic/deduplication.rs
```diff
@@ -0,0 +1,65 @@
...
+    #[cfg_attr(
+        feature = "use_serde",
+        serde(
+            default,
+            skip_serializing_if = "Option::is_none",
+            with = "humantime_serde"
+        )
+    )]
+    pub age: Option<Duration>,
+}
```
digikata, line 35: Not something to change for this PR, but I wonder if there is some commonality with Lookback logic we can reuse between the two (slightly different) uses as a generic "record (pre-?)selection"

digikata (approved):
LGTM, nice work!

crates/fluvio-protocol/src/link/error_code.rs
galibey: Ok. Especially, on SPU/SC there is no much sense in enum of errors.

crates/fluvio-controlplane-metadata/src/topic/deduplication.rs
galibey: Although Bounds and Lookback are common now, they might differ later. Bound covers both preselection(Lookback) and limits for runtime stream processing (via SmartModule init params).

PR #3399, opened 2023-07-14
feat: use `async-std`@`1.12.0` and `async-channel`@`1.5`

EstebanBorai:
Updates `async-std` to use [latest version available](https://crates.io/crates/async-std), in order to
tackle compile-time issues due to dependency conflicts.

`async-std`@`1.12.0` only supports `async-channel`@`1.5`, so we could use latest `async-std`
by downgrading `async-channel` to `1.5`.

(_Let's hope this doesn't ends in Yak Shaving 😅_)

## Related

- [ ] https://github.com/infinyon/sql-connector/pull/76

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,7 +65,7 @@ async-io = "1.3.1"
 async-lock = "2.4.0"
 async-net = { version = "1.7.0", default-features = false  }
 async-rwlock = "1.3.0"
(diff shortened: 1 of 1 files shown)


EstebanBorai:
Closing, using: https://github.com/infinyon/sql-connector/pull/77 instead!

PR #3405, opened 2023-07-20
[Merged by Bors] - fix: properly handle produce operation failure

galibey:
The failure of produce operation was mistakenly treated as a failure of the transportation layer, which is not. That led to non-helpful error reporting to the user:
```bash
Producer error: failed to get record metadata
```
Fixed by not handling return error code too early.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2288,7 +2288,7 @@ dependencies = [
 
 [[package]]
 name = "fluvio"
-version = "0.19.3"
+version = "0.19.4"
 dependencies = [
  "anyhow",
  "async-channel",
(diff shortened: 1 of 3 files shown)


digikata (approved):
The change and description LGTM.  

Just for my information, I tracked the error code thru the Poll::Ready. Either::Right logic, but I'm not sure I have a full handle on all the places on the path the error code is handled now? Is BatchMetadata::base_offset the only place?  Just doing some extra reading because the (offset,error) pair has to be a little bit more explictly handled vs a Result type.

galibey:
> Is BatchMetadata::base_offset the only place? 

To my understanding, yes. This method is called by `wait` function which you call when you want to know the result of your batch operation (Fluvio CLI does). Inside this  `BatchMetadata::base_offset` if error_code is not None the batch metadata turns itself to failed state and `Err` is propagated up.

PR #3427, opened 2023-07-28
[Merged by Bors] - feat: use `SmartModuleRecord` instead of `Record` for `smdk generate`

EstebanBorai:


EstebanBorai:
I think this will be introduced after the next release!

sehz (approved):

PR #3428, opened 2023-07-28
[`smdk`] Report SmartModule errors

### LINKED ISSUE #3367: [Bug]: SMDK test doesn't report error from SmartModule
**What happened**
when running `smdk test`, it doesn't report error from SmartModule

**Expected behavior**
any error from SmartModule should be output to SMDK

**Describe the setup**
Rust deve env

- Set up `smdk`
- Write SmartModule that returns error
- run `smdk test`

**Log output**
Nothing which si bug

 - OS: [e.g. Mac with Apple Silicon]
 - Fluvio Version [N/A]

**Additional context**

c410-f3r:
Fix #3367

Reports SmartModule errors and also creates an integration test to assert the expected output.

### DIFF
--- a/crates/smartmodule-development-kit/src/test.rs
+++ b/crates/smartmodule-development-kit/src/test.rs
@@ -158,12 +158,18 @@ impl TestCmd {
         let metrics = SmartModuleChainMetrics::default();
 
         let test_records: Vec<Record> = test_data.into();
+        let test_records_len = test_records.len();
         let mut sm_input =
             SmartModuleInput::try_from_records(test_records, DEFAULT_SMARTENGINE_VERSION)?;
 
         sm_input.set_base_timestamp(Utc::now().timestamp_millis());
 
         let output = chain.process(sm_input, &metrics)?;
+
+        if let Some(error) = output.error {
+            print!("Failed to process {test_records_len} record(s) with errors:\n\n1. {error}");
+            return Ok(());
+        }
 
         if self.verbose {
             println!("{:?} records outputed", output.successes.len());
--- a/smartmodule/examples/filter_error/src/lib.rs
+++ b/smartmodule/examples/filter_error/src/lib.rs
@@ -1,0 +1,7 @@
+use fluvio_smartmodule::{Record, Result, smartmodule};
+
+#[smartmodule(filter)]
+pub fn filter(_: &Record) -> Result<bool> {
+    let _a: u32 = "dasdas".parse()?;
+    Ok(true)
+}
--- a/tests/cli/smdk_smoke_tests/smdk-basic.bats
+++ b/tests/cli/smdk_smoke_tests/smdk-basic.bats
@@ -981,3 +981,23 @@ setup_file() {
     assert_output --partial "\"Cranberry\"_$DATE_NOW_YEAR-$DATE_NOW_MONTH-$DATE_NOW_DAY"
     assert_success
 }
+
+@test "`smdk test` redirects SmartModule error to stdout" {
+    cd "$(pwd)/smartmodule/examples/filter_error/"
+
+    # Build
+    run $SMDK_BIN build
+
+    # Test
+    run $SMDK_BIN test --text foo
+    assert_output "Failed to process 1 record(s) with errors:
+
+1. invalid digit found in string
+
+SmartModule Info: 
+    Type: Filter
+    Offset: 0
+    Key: NULL
+    Value: foo"
+    assert_success
+}
--- a/smartmodule/examples/Cargo.toml
+++ b/smartmodule/examples/Cargo.toml
@@ -7,6 +7,7 @@ members = [
     "aggregate-init",
     "aggregate_with_timestamp",
     "filter",
+    "filter_error",
     "filter_init",
     "filter_look_back",
     "filter_look_back_with_timestamps",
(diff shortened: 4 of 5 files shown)


crates/smartmodule-development-kit/tests/cmd.rs
```diff
@@ -0,0 +1,32 @@
+use std::process::Command;
```
c410-f3r, line 1: `smartmodule-development-kit` is a CLI application and therefore it seems appropriate to verify what the user will actually see.

crates/smartmodule-development-kit/src/test.rs
```diff
@@ -165,6 +165,11 @@ impl TestCmd {
 
         let output = chain.process(sm_input, &metrics)?;
 
+        if let Some(error) = output.error {
+            print!("SmartModule error: {error}");
```
c410-f3r, line 169: The reasoning here is that `stderr` should be populated only if `process` returns a `Err`, leaving `stdout` as a way to tell that the SmartModule execution wasn't a success but the involving command acted as expected.

Regardless, it is possible to print to `stderr` using `eprint` if desired.

crates/smartmodule-development-kit/src/test.rs
```diff
@@ -165,6 +165,11 @@ impl TestCmd {
...
+            return Ok(());
```
c410-f3r, line 170: It seems reasonable to return early but I am not aware if it possible to have successful records along side errors in the same output.

crates/smartmodule-development-kit/tests/cmd.rs
```diff
@@ -0,0 +1,32 @@
...
+
+fn handle_cmd_outputs(cmd: &mut Command) -> (String, String) {
+    let output = cmd.output().unwrap();
+    if !output.status.success() {
+        panic!("Command `{:?}` returned an error", cmd.get_program());
+    }
+    (
+        String::from_utf8(output.stderr).unwrap(),
+        String::from_utf8(output.stdout).unwrap(),
+    )
```
c410-f3r, line 31: Leaving both `stderr` and `stdout` so that future tests can use them

crates/smartmodule-development-kit/tests/cmd.rs
```diff
@@ -0,0 +1,32 @@
...
+        "run",
+        "--manifest-path",
+        &format!("{}/Cargo.toml", cargo_manifest_dir),
+        "--",
+        "test",
+        "--text",
+        "foo",
+        "--wasm-file",
+        &format!("{}/fixtures/filter_error.wasm", cargo_manifest_dir),
+    ]));
```
c410-f3r, line 16: It is possible to mimic the `fluvio-smartengine` tests using `#[ignore]` along side `make` dependencies but I think it is easier to just call `cargo` directly.

https://github.com/infinyon/fluvio/blob/ac4453b51320713700fcbedee5d1d469575fac36/makefiles/check.mk#L45

c410-f3r (commented):
Let me know if there's anything missing

crates/smartmodule-development-kit/fixtures/filter_error.wasm
c410-f3r, line 1:
```
use fluvio_smartmodule::{Record, Result, smartmodule};

#[smartmodule(filter)]
pub fn filter(_: &Record) -> Result<bool> {
    let _a: u32 = "dasdas".parse()?;
    Ok(true)
}
```

Blob is optimized but unfortunately has 242K. If such weight is not desired, should I put the project in `smartmodule/examples`?

EstebanBorai, line 1: It would be nice to have the source of this WASM module so we could perform build tests on it on the current and upcoming versions.

crates/smartmodule-development-kit/src/test.rs
EstebanBorai: Sounds good to me! I think we could have outputs like:

```
Failed to process 1 record with errors:

1. <Error Message>
```

c410-f3r:
Thank you for the review @EstebanBora, all comments were addressed accordingly.

smartmodule/examples/filter_error/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+use fluvio_smartmodule::{Record, Result, smartmodule};
+
+#[smartmodule(filter)]
+pub fn filter(_: &Record) -> Result<bool> {
+    let _a: u32 = "dasdas".parse()?;
+    Ok(true)
+}
```
c410-f3r, line 7: `smartmodule/examples` doesn't seem like the best place so let me know if there is a better location.

tests/cli/smdk_smoke_tests/smdk-basic.bats
```diff
@@ -981,3 +981,23 @@ setup_file() {
...
+
+1. invalid digit found in string
+
+SmartModule Info: 
+    Type: Filter
+    Offset: 0
+    Key: NULL
+    Value: foo"
+    assert_success
+}
```
c410-f3r, line 1003: `bats` interprets `\n` and `\\n` as literals leaving explicit new lines as the only option AFAICT.

c410-f3r:
@EstebanBorai Let me know if there's anything missing.

Issue #3441, opened 2023-08-03
WatchAPI error: return a better user error when versions are not compatible


digikata:
Currently WatchAPI error is returned the the fluvio client and cluster do not support a common version, user error should probably ask to upgrade either the client or the cluster as well as giving relevant version info for both

The error should be something like:

```
Error: Fluvio Client {client_version} and Cluster {cluster_version} versions are not compatible
Please upgrade client to {cluster_version}
```

ajhunyady:
Marking high priority as it led to customer issue.

fraidev:
https://github.com/infinyon/fluvio/pull/4149 fix it

PR #3455, opened 2023-08-07
Improve error message for client cluster version mismatch

Thomasdezeeuw:
I'm not sure if maybe return the `MaximumPlatformVersion` variant for cases where the client is too new, see https://github.com/infinyon/fluvio/blob/b7a2765b8aa2263838a46fd2ce55921bfdf64736/crates/fluvio/src/error.rs#L39-L54 for the error message. That would mean we have to define a `MAXIMUM_PLATFORM_VERSION`, similar to [`MINIMUM_PLATFORM_VERSION`](https://github.com/infinyon/fluvio/blob/b7a2765b8aa2263838a46fd2ce55921bfdf64736/crates/fluvio/src/lib.rs#L48).

Fixes https://github.com/infinyon/fluvio/issues/3441

### DIFF
--- a/crates/fluvio/src/admin.rs
+++ b/crates/fluvio/src/admin.rs
@@ -135,8 +135,8 @@ impl FluvioAdmin {
             })
         } else {
             let cluster_version = versions.platform_version().clone();
-            let client_version = Version::parse(crate::PLATFORM_VERSION)
-                .expect("PLATFORM_VERSION must be semver");
+            let client_version =
+                Version::parse(crate::PLATFORM_VERSION).expect("PLATFORM_VERSION must be semver");
             if client_version > cluster_version {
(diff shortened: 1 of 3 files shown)


digikata:
> I'm not sure if maybe return the `MaximumPlatformVersion` variant for cases where the client is too new...

I think in this case, returning a MaximumPlatformVersion error with the current client version would be acceptable.

Thomasdezeeuw:
@digikata I pushed another commit using the `VERSION` file, I hope this is correct. If I need to use `CARGO_PKG_VERSION` instead please let me know.

Thomasdezeeuw:
For some reason my automatic formatting in my code editor didn't realise the project was using Rust 2021 and thus couldn't properly for the code. So I had to manually fix that in a follow up commit.

PR #3464, opened 2023-08-09
[Merged by Bors] - Add stdin input support for `smdk test`

### LINKED ISSUE #3318: `smdk` support for input from stdin
I think it's inconvenient to use `smdk test` if I want to test with dynamic json output. The only options are from  that input is either provided with `--input` or `--file`.

I would like to be able to pipe in input from curl

`curl https://myservice/api | smdk test ...` 

I've seen patterns that use stdin if the file is `-` - e.g. `--file -`

Or  a more verbose `--stdin` flag. I don't have a strong preference for either, but it should be explicit

matheus-consoli:
Introduces a new `--stdin` flag to `smdk test` to support passing the test data via stdin.

```text
>> smdk test --help
Test SmartModule

Usage: smdk test [OPTIONS] [KEY]

Arguments:
  [KEY]  Key to use with the test record(s)

Options:
      --text <TEXT>
          Provide test input with this flag
      --stdin
          Read the test input from the StdIn (e.g. Unix piping)
      ...
```

It's now possible to pipe data as the test input:

```sh
>> echo fluvio | smdk test --stdin
 fluvio
```

Closes #3318

### DIFF
--- a/crates/fluvio-cli-common/src/user_input.rs
+++ b/crates/fluvio-cli-common/src/user_input.rs
@@ -9,6 +9,7 @@ use fluvio_protocol::bytes::Bytes;
 
 pub enum UserInputType {
     Text { key: Option<Bytes>, data: Bytes },
+    StdIn { key: Option<Bytes>, data: Bytes },
     File { key: Option<Bytes>, path: PathBuf },
     FileByLine { key: Option<Bytes>, path: PathBuf },
 }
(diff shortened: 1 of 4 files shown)


matheus-consoli:
The basic implementation is done and it took me less than 10m to write and do some manual tests.

I'm marking this PR as a draft because I still need to write the automatic tests to cover everything, but I'm having some difficulties running the existing tests. Running `make smoke-test-local` fails to connect to the cluster (`Connecting to cluster failed: Socket io Connection refused (os error 111), can't connect to localhost:9003`) - I will investigate it further tomorrow, but any help will be much appreciated (cc @digikata).

digikata:
We can run through CI for this. To run locally that check typically expects a kubernetes setup that is locally accessible via nodeport, though in general it would be nicer if some of our tests would support a k8 that is separated in a another host so we could run vs a local vm (like a lot of the macos k8s) or a cloud install.

matheus-consoli:
Ok, I managed to get everything working out, Thank you!

I modified two tests to use `--stdin` instead of `--text`

tjtelan (approved):
Simple addition to input options and the new test updates pass! Looks good to me

EstebanBorai (approved):
LGTM

ajhunyady:
@matheus-consoli, please add the new feature to the documentation. https://github.com/infinyon/fluvio-website/tree/master/content/smartmodules/smdk

PR #3466, opened 2023-08-11
[Merged by Bors] - Remove lazy_static in favor of once_cell

matheus-consoli:
`fluvio-package-index` was the only package using `lazy_static` - this small change favors `once_cell` instead.

This way we have one less dependency and it will be easier to update once [`std::sync::LazyLock`](https://doc.rust-lang.org/std/sync/struct.LazyLock.html) is stabilized.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2791,7 +2791,7 @@ name = "fluvio-package-index"
 version = "0.7.4"
 dependencies = [
  "http",
- "lazy_static",
+ "once_cell",
  "semver 1.0.18",
  "serde",
  "serde_json",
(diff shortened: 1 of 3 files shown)


sehz (approved):
LGTM

matheus-consoli:
Build failed on the crate version check - it detected a code change without a version change. Is it ok to ignore it and merge (since nothing meaningful changed), or is it preferable to bump a minor?

sehz:
All merges need to go thru bors (there is no exception).  So in order to make bors happy, need to bump up minor version.  Should create issue to skipped version check if changes are only related to tests

PR #3475, opened 2023-08-18
[Merged by Bors] - feat: print more error causes

ozgrakkurt:
Print more error causes. This is short term minimal change solution that just adds printing causes to `thiserror` derivations.

I made changes for a couple of crates, will do other ones in next PRs.

I realise this might be printing too much but I think it is still better than not printing enough in some cases. Later on we can refine it more using `anyhow`.

### DIFF
--- a/crates/fluvio-auth/src/error.rs
+++ b/crates/fluvio-auth/src/error.rs
@@ -4,14 +4,14 @@ use thiserror::Error;
 /// Possible errors from Auth
 #[derive(Error, Debug)]
 pub enum AuthError {
-    #[error("IoError")]
+    #[error("IoError: {0}")]
     IoError(#[from] IoError),
 }
 
 impl From<AuthError> for IoError {
     fn from(e: AuthError) -> Self {
-        match e {
-            AuthError::IoError(source) => source,
+        match &e {
+            AuthError::IoError(source) => IoError::new(source.kind(), e),
         }
     }
 }
--- a/crates/fluvio-cli/src/error.rs
+++ b/crates/fluvio-cli/src/error.rs
@@ -16,33 +16,33 @@ use crate::common::target::TargetError;
 pub enum CliError {
     #[error(transparent)]
     OutputError(#[from] OutputError),
-    #[error("Failed to parse format string")]
+    #[error("Failed to parse format string: {0}")]
     TemplateError(#[from] TemplateError),
 
     #[cfg(feature = "k8s")]
-    #[error("Fluvio cluster error")]
+    #[error("Fluvio cluster error: {0}")]
     ClusterCliError(#[from] ClusterCliError),
 
-    #[error("Target Error")]
+    #[error("Target Error: {0}")]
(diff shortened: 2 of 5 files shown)


crates/fluvio-cluster/src/error.rs
```diff
@@ -18,47 +18,47 @@ use crate::runtime::local::LocalRuntimeError;
...
     /// Failed to execute a command
     #[error(transparent)]
     CommandError(#[from] CommandError),
     /// One or more pre-checks (successfully) failed when trying to start the cluster
-    #[error("Pre-checks failed during cluster startup")]
+    #[error("Pre-checks failed during cluster startup: {0:#?}")]
     FailedPrecheck(CheckStatuses),
     /// Encountered an error while performing one or more pre-checks
-    #[error("Failed to perform one or more pre-checks")]
+    #[error("Failed to perform one or more pre-checks: {0:#?}")]
```
EstebanBorai, line 58: These kind of formats `{#?}` could span many lines causing upper logs to be missed by the Terminal buffer.

For these cases I think we could have a one liner, perhaps implementing a `Display` trait on them and have a summary of the details in a single line.

EstebanBorai (commented):
Overall I love the idea I also think the more details the better!
Im just concerned about cases where formatting with `{:#?}` may take too many lines.

crates/fluvio-cluster/src/error.rs
ozgrakkurt: makes sense! let me try to fix this

ozgrakkurt: Removed debug prints for now. In next iteration we can implement custom `Display`, I didn't want to do it now because it is typedef as `Vec<_>` so can't implement trait on it directly, need to make it a struct which will be a breaking change.

crates/fluvio-auth/src/error.rs
```diff
@@ -4,14 +4,14 @@ use thiserror::Error;
 /// Possible errors from Auth
 #[derive(Error, Debug)]
 pub enum AuthError {
-    #[error("IoError")]
+    #[error("IoError: {0}")]
```
sehz, line 7: let's get rid of this and move to anyhow

crates/fluvio-cli/src/error.rs
```diff
@@ -16,33 +16,33 @@ use crate::common::target::TargetError;
 pub enum CliError {
```
sehz, line 16: Let's remove this and move to anyhow

crates/fluvio-auth/src/error.rs
ozgrakkurt: I want to do anyhow, but then it requires changes in other places in the code. Want to make this change like this and do other PRs for anyhow in future

morenol (approved):
LGTM

ozgrakkurt:
merging this, and working on `anyhow` PR now

PR #3485, opened 2023-08-23
Issue-3479: Refactor: Fetch Platform Version

### LINKED ISSUE #3479: Validate `VERSION` will be parsable by `semver` at runtime
Related: https://github.com/infinyon/fluvio/issues/2949

Due to how we include the platform version number in the CLI, if there is any leading/trailing whitespace, parsing for semver may fail. We handle this case inconsistently.

https://github.com/infinyon/fluvio/blob/d2b7ccf1cf2237c4999e2f97520ac37f7dea395a/crates/fluvio-cli/src/lib.rs#L16

---

`fluvio version` can handle whitespace
https://github.com/infinyon/fluvio/blob/d2b7ccf1cf2237c4999e2f97520ac37f7dea395a/crates/fluvio-cli/src/version.rs#L23

But `fluvio install/update` (and many other places) do not include `.trim()`
https://github.com/infinyon/fluvio/blob/d2b7ccf1cf2237c4999e2f97520ac37f7dea395a/crates/fluvio-cli/src/install/update.rs#L282

There is no known workaround after compilation, but `VERSION` requirements should also be less strict. We should support basic whitespace trimming and address the root issue.

And rather than handle case-by-case, we should validate that `VERSION` is valid during buildtime. But the CLI build should fail in cases when `VERSION` can't be parsed due to actual semver validity.

jamrok:
Hey @tjtelan,

I added a common library (`fluvio-common`) to parse the `VERSION` file and made all relevant crates refer to it. I saw that there was `fluvio-cli-common`, but It seemed more specific to `fluvio-cli` versus being a light dependency that can be used by all. 

Please let me know if there's another place that you recommend that the change be made.

Thanks

---
Resolves #3479

### DIFF
--- a/crates/fluvio-types/build.rs
+++ b/crates/fluvio-types/build.rs
@@ -1,0 +1,17 @@
+use semver::Version;
+
+fn main() {
+    if let Ok(verpath) = std::fs::canonicalize("../../VERSION") {
+        if verpath.exists() {
+            println!("cargo:rerun-if-changed=../../VERSION");
+        }
+    }
+
+    println!("cargo:rerun-if-changed=build.rs");
+
+    // Parse the Fluvio Platform Version
+    let version = include_str!("../../VERSION").trim();
+    let version = Version::parse(version)
+        .unwrap_or_else(|_| panic!("Fluvio Platform VERSION '{version}' is not a valid semver"));
+    println!("cargo:rustc-env=FLUVIO_PLATFORM_VERSION={version}");
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2998,6 +2998,7 @@ dependencies = [
  "fluvio-future",
  "fluvio-sc",
  "fluvio-spu",
+ "fluvio-types",
  "semver 1.0.19",
  "serde_json",
  "thiserror",
@@ -3415,6 +3416,8 @@ version = "0.4.4"
 dependencies = [
  "event-listener 3.0.0",
  "fluvio-future",
+ "once_cell",
+ "semver 1.0.19",
  "thiserror",
  "tokio",
  "tracing",
--- a/crates/fluvio-cli/Cargo.toml
+++ b/crates/fluvio-cli/Cargo.toml
@@ -101,6 +101,5 @@ fluvio-future = { workspace = true, features = ["fs", "io", "subscriber", "nativ
 fluvio-sc-schema = { workspace = true,  features = ["use_serde"], optional = true }
 fluvio-spu-schema = { workspace = true, optional = true }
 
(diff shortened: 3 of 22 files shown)


tjtelan:
@jamrok Thanks for the quick turnaround!

I think there are 2 points of feedback:

1. Your implementation looks good, however not quite the direction I had in mind. The desired outcome is that compilation should be prevented upon discovering errors with parsing `VERSION`, instead of allowing the condition to be found in runtime. I think validation of `VERSION` should be implemented within a `build.rs` build script, in a similar fashion to how we include the current git sha of builds. 

2. `FLUVIO_PLATFORM_VERSION` should live in the `fluvio-types` crate for the purposes of sharing with other crates

jamrok:
Thanks for the feedback @tjtelan!

1. Updated as requested. Please also confirm that the `build.rs` for `fluvio-test-util` is done as expected.
    - I moved the compile time check for the optional `FLUVIO_VERSION_SUFFIX` env var to `build.rs` and set a different ENV variable (`FLUVIO_PLATFORM_TEST_VERSION`) so it doesn't conflict with the `FLUVIO_PLATFORM_VERSION` variable.
2. Updated as requested. I originally had `FLUVIO_PLATFORM_VERSION` in `fluvio-types`, but then moved it because `fluvio-types` was published and I was thinking that "types" didn't seem like the right place to put it.
   - Are there any implications with it being published?
   - Should it be placed in `fluvio_types::defaults` and exported at the crate level (`fluvio_types::FLUVIO_PLATFORM_VERSION`), or placed somewhere else?

 Please let me know if you have any other suggestions.

Thanks

jamrok:
Hey @tjtelan, just checking on this. I had a couple questions in my last comment.

I also rebased and resolved the merge conflicts.

crates/fluvio-types/build.rs
```diff
@@ -0,0 +1,17 @@
...
+    if let Ok(verpath) = std::fs::canonicalize("../../VERSION") {
+        if verpath.exists() {
+            println!("cargo:rerun-if-changed=../../VERSION");
+        }
+    }
+
+    println!("cargo:rerun-if-changed=build.rs");
+
+    // Parse the Fluvio Platform Version
+    let version = include_str!("../../VERSION").trim();
```
sehz, line 13: Why parsing this again? This won't work since it is public crate

tjtelan: I think this is my bad. At the time of [my feedback](https://github.com/infinyon/fluvio/pull/3485#issuecomment-1690070729) I didn't consider that this was a public crate

PR #3500, opened 2023-08-29
[Merged by Bors] - chore: cleanup format in some crates and fix fluvio cluster status wh…

morenol:
…en there is not access to the underlying k8 context


sehz (approved):
LGTM.
Check crate versions.

Also, we need to migrate cluster crate to anyhow and get rid of unnecessary error

PR #3525, opened 2023-09-08
[Merged by Bors] - feat: Add capability to start SC on read-only mode. Read-only mode me…

morenol:
…ans, all write/update service endpoints are disabled.

Also added smoke test that validates that a topic cannot be created in this mode


sehz:
still in draft mode

crates/fluvio-sc/src/services/public_api/create.rs
```diff
@@ -24,13 +24,23 @@ pub async fn handle_create_request<AC: AuthContext, C: MetadataItem>(
...
-    let status = if let Some(create) = req.downcast()? as Option<CreateRequest<TopicSpec>> {
-        super::topic::handle_create_topics_request(create, auth_context).await?
-    } else if let Some(create) = req.downcast()? as Option<CreateRequest<SpuGroupSpec>> {
+
+    let status = if let Some(create) = req.downcast()? as Option<CreateRequest<SpuGroupSpec>> {
         super::spg::handle_create_spu_group_request(create, auth_context).await?
     } else if let Some(create) = req.downcast()? as Option<CreateRequest<CustomSpuSpec>> {
         super::spu::RegisterCustomSpu::handle_register_custom_spu_request(create, auth_context)
             .await
+    } else if auth_context.global_ctx.config().read_only_metadata {
```
sehz, line 33: There is already `auth_ctx`, which checks permission.  Instead of adding another case, could create read-only auth_ctx so there is no need to create different path

crates/fluvio-sc/src/services/public_api/delete.rs
```diff
@@ -32,6 +32,17 @@ pub async fn handle_delete_request<AC: AuthContext, C: MetadataItem>(
 
     debug!(?del_req, "del request");
 
+    if auth_ctx.global_ctx.config().read_only_metadata {
```
sehz, line 35: same comments regarding auth

sehz (changes requested):
overall looks good.  comments about using auth context rather than custom path

crates/fluvio-sc/src/services/public_api/create.rs
morenol: updated

crates/fluvio-sc/src/services/auth/mod.rs
```diff
@@ -96,4 +96,50 @@ mod common {
...
+        pub fn new() -> Self {
+            Self {}
+        }
+    }
+
+    #[derive(Debug)]
+    pub struct ReadOnlyAuthContext {}
+
+    #[async_trait]
+    impl AuthContext for ReadOnlyAuthContext {
```
sehz, line 126: Can you add a unit test for this?

sehz (changes requested):
Looks good.  Just need to add unit test

crates/fluvio-sc/src/services/auth/mod.rs
morenol: added unit test and also added no-k8 behavior

sehz (approved):
LGTM.  Nice work

PR #3529, opened 2023-09-13
[Merged by Bors] - chore: remove unnecessary generics

sehz:
convert many generic based argument conversion like `<S: Into<String>>` into `impl Into<String>`.    This remove unnecessary generic signature and hopefully reduce build time and make code cleaner

### DIFF
--- a/crates/fluvio/Cargo.toml
+++ b/crates/fluvio/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio"
-version = "0.21.0"
+version = "0.22.0"
 edition = "2021"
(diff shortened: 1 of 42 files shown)


crates/fluvio/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio"
-version = "0.21.0"
+version = "0.22.0"
```
morenol, line 3: I think that 0.21.0 is not released

sehz: fluvio version is bump up due to change in generic signature

PR #3535, opened 2023-09-14
feat: added embedded k8s for local cluster start

galibey:
This should not be merged until k8s and etcd extensions are shipped during the installation process!

CLI and integration tests will come in the next PRs.

### DIFF
--- a/crates/fluvio-cli-common/src/http.rs
+++ b/crates/fluvio-cli-common/src/http.rs
@@ -1,7 +1,16 @@
-use std::fmt::Debug;
+use std::{
+    fmt::Debug,
+    time::{Duration, Instant},
+    path::Path,
+};
 
+use fluvio_future::timer::sleep;
 use http::uri::Scheme;
-use isahc::{AsyncBody, Request, Response};
+use isahc::{
+    AsyncBody, Request, Response,
+    config::{ClientCertificate, CaCertificate, PrivateKey},
+    prelude::Configurable,
+};
 use tracing::{debug, error, instrument};
 use anyhow::Result;
 
@@ -40,6 +49,58 @@ pub async fn read_to_end(response: Response<AsyncBody>) -> std::io::Result<Vec<u
     let mut body = Vec::with_capacity(async_body.len().unwrap_or_default() as usize);
     async_body.read_to_end(&mut body).await?;
     Ok(body)
+}
+
+pub async fn wait_http_ready(url: &str, timeout: Duration) -> Result<bool> {
+    let started = Instant::now();
+    while started.elapsed() < timeout {
+        let request = http::Request::get(url).body(())?;
+        if let Ok(response) = isahc::send_async(request).await {
+            let status = response.status();
+            if !status.is_client_error() && !status.is_server_error() {
+                return Ok(true);
+            }
+        }
+        sleep(Duration::from_secs(1)).await;
+    }
+    Ok(false)
+}
+
+pub async fn wait_https_ready(
+    url: &str,
+    timeout: Duration,
+    ca_cert: Option<&Path>,
+    client_cert: Option<&Path>,
+    client_private_key: Option<&Path>,
+) -> anyhow::Result<bool> {
+    let started = Instant::now();
+    let ca_cert = ca_cert.map(CaCertificate::file);
+    let client_cert = client_cert.map(|p| {
+        ClientCertificate::pem_file(
+            p,
+            client_private_key.map(|p_path| PrivateKey::pem_file(p_path, None)),
+        )
+    });
+    while started.elapsed() < timeout {
+        let builder = http::Request::get(url);
+        let builder = match ca_cert.clone() {
+            Some(ca_cert) => builder.ssl_ca_certificate(ca_cert),
+            None => builder,
+        };
+        let builder = match client_cert.clone() {
+            Some(client_cert) => builder.ssl_client_certificate(client_cert),
+            None => builder,
+        };
+        let request = builder.body(())?;
+        if let Ok(response) = isahc::send_async(request).await {
+            let status = response.status();
+            if !status.is_client_error() && !status.is_server_error() {
+                return Ok(true);
+            }
+        }
+        sleep(Duration::from_secs(1)).await;
+    }
+    Ok(false)
 }
 
 #[cfg(test)]
--- a/crates/fluvio-cli-common/src/http.rs
+++ b/crates/fluvio-cli-common/src/http.rs
@@ -51,56 +51,57 @@ pub async fn read_to_end(response: Response<AsyncBody>) -> std::io::Result<Vec<u
     Ok(body)
 }
 
-pub async fn wait_http_ready(url: &str, timeout: Duration) -> Result<bool> {
-    let started = Instant::now();
-    while started.elapsed() < timeout {
-        let request = http::Request::get(url).body(())?;
-        if let Ok(response) = isahc::send_async(request).await {
-            let status = response.status();
-            if !status.is_client_error() && !status.is_server_error() {
-                return Ok(true);
-            }
-        }
-        sleep(Duration::from_secs(1)).await;
-    }
-    Ok(false)
+#[derive(Debug, derive_builder::Builder)]
+pub struct HttpsReadinessCheck<'a, 'b> {
+    #[builder(default = "Duration::from_secs(60)")]
+    timeout: Duration,
+    #[builder(default = "Duration::from_secs(1)")]
+    interval: Duration,
+    url: &'a str,
+    #[builder(setter(strip_option), default)]
+    ca_cert: Option<&'b Path>,
+    #[builder(setter(strip_option), default)]
+    client_cert: Option<&'b Path>,
+    #[builder(setter(strip_option), default)]
+    client_private_key: Option<&'b Path>,
 }
 
-pub async fn wait_https_ready(
-    url: &str,
-    timeout: Duration,
-    ca_cert: Option<&Path>,
-    client_cert: Option<&Path>,
-    client_private_key: Option<&Path>,
-) -> anyhow::Result<bool> {
-    let started = Instant::now();
-    let ca_cert = ca_cert.map(CaCertificate::file);
-    let client_cert = client_cert.map(|p| {
-        ClientCertificate::pem_file(
-            p,
-            client_private_key.map(|p_path| PrivateKey::pem_file(p_path, None)),
-        )
-    });
-    while started.elapsed() < timeout {
-        let builder = http::Request::get(url);
-        let builder = match ca_cert.clone() {
-            Some(ca_cert) => builder.ssl_ca_certificate(ca_cert),
-            None => builder,
-        };
-        let builder = match client_cert.clone() {
-            Some(client_cert) => builder.ssl_client_certificate(client_cert),
-            None => builder,
-        };
-        let request = builder.body(())?;
-        if let Ok(response) = isahc::send_async(request).await {
-            let status = response.status();
-            if !status.is_client_error() && !status.is_server_error() {
-                return Ok(true);
-            }
-        }
-        sleep(Duration::from_secs(1)).await;
+impl<'a, 'b> HttpsReadinessCheck<'a, 'b> {
+    pub fn builder() -> HttpsReadinessCheckBuilder<'a, 'b> {
+        HttpsReadinessCheckBuilder::default()
     }
-    Ok(false)
+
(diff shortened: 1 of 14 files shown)


crates/fluvio-cli-common/src/http.rs
```diff
@@ -42,6 +51,58 @@ pub async fn read_to_end(response: Response<AsyncBody>) -> std::io::Result<Vec<u
     Ok(body)
 }
 
+pub async fn wait_http_ready(url: &str, timeout: Duration) -> Result<bool> {
```
sehz, line 54: Not show stopper, but this could be abstracted out into utility struct since this is related to URL. So something like this

```
let url_util = HttpUrlUtil::new(ur);
url_util.wait_ready()?;
```

crates/fluvio-cluster/src/error.rs
```diff
@@ -162,6 +162,12 @@ pub enum LocalInstallError {
     ClusterCheckError(#[from] ClusterCheckError),
     #[error("Progress Error: {0}")]
     ProgressError(#[from] TemplateError),
+    /// Timed out when waiting for etcd.
```
sehz, line 165: Need to migrate to anyhow.  doesn't need for this

crates/fluvio-cluster/src/progress.rs
```diff
@@ -6,6 +6,8 @@ use crate::render::{ProgressRenderedText, ProgressRenderer};
 
 #[derive(Debug)]
 pub(crate) enum InstallProgressMessage {
+    LaunchingK8s,
```
sehz, line 9: Should hide internal embedded.  Maybe `Metadata`?

crates/fluvio-cluster/src/progress.rs
```diff
@@ -20,6 +22,10 @@ impl ProgressRenderedText for InstallProgressMessage {
         use colored::*;
 
         match self {
+            InstallProgressMessage::LaunchingK8s => {
+                format!("🚢 {}", "Starting k8s server".bold())
```
sehz, line 26: change to `Installing Metadata server`

crates/fluvio-cluster/src/runtime/local/etcd.rs
```diff
@@ -0,0 +1,47 @@
+use std::{
+    fs::File,
+    path::PathBuf,
+    process::{Command, Stdio},
+};
+
+use fluvio_command::CommandExt;
```
sehz, line 7: move this down bit

crates/fluvio-cluster/src/runtime/local/etcd.rs
```diff
@@ -0,0 +1,47 @@
...
+use tracing::info;
+
+use super::LocalRuntimeError;
+
+pub struct EtcdProcess {
```
sehz, line 12: `pub(crate)`.  I know we were not consistent in this

crates/fluvio-cluster/src/runtime/local/etcd.rs
```diff
@@ -0,0 +1,47 @@
...
+    pub log_dir: PathBuf,
+    pub data_dir: PathBuf,
+    pub base: PathBuf,
+}
+
+impl EtcdProcess {
+    pub fn start(&self) -> Result<String, LocalRuntimeError> {
+        let outputs = File::create(format!("{}/flv_etcd.log", self.log_dir.display()))?;
```
sehz, line 20: separate out config file directory

crates/fluvio-cluster/src/runtime/local/k8s.rs
```diff
@@ -0,0 +1,211 @@
...
+use k8_config::{K8Config, User, UserDetail, Cluster, ClusterDetail, Context, ContextDetail};
+use rcgen::{Certificate, CertificateParams, IsCa, DistinguishedName, DnType, DnValue, KeyUsagePurpose};
+use tracing::{info, debug};
+
+use super::LocalRuntimeError;
+
+const DEFAULT_K8S_CONTEXT_NAME: &str = "fluvio-local";
+const DEFAULT_K8S_CLUSTER_NAME: &str = "fluvio-local-cluster";
+const DEFAULT_K8S_USER_NAME: &str = "fluvio-local-client";
+const DEFAULT_K8S_NAMESPACE: &str = "default";
```
sehz, line 17: This should be `fluivo-system` or something like that?

crates/fluvio-cluster/src/runtime/local/k8s.rs
```diff
@@ -0,0 +1,211 @@
...
+pub struct K8sProcess {
+    pub log_dir: PathBuf,
+    pub data_dir: PathBuf,
+    pub base: PathBuf,
+    pub etcd_endpoint_url: String,
+}
+
+impl K8sProcess {
+    pub fn start(&self) -> Result<(String, Certs), LocalRuntimeError> {
+        let logs_path = self.log_dir.join("flv_k8s.log");
```
sehz, line 28: separate out log directory path

crates/fluvio-cluster/src/runtime/local/k8s.rs
```diff
@@ -0,0 +1,211 @@
...
+        let outputs = File::create(&logs_path)?;
+        let errors = outputs.try_clone()?;
+
+        let port = portpicker::pick_unused_port()
```
sehz, line 32: This seems to be repeated.  can we just make it common util?

crates/fluvio-cluster/src/runtime/local/k8s.rs
```diff
@@ -0,0 +1,211 @@
...
+    dn.push(
+        DnType::CommonName,
+        DnValue::PrintableString("fluvio.io local ca cert".to_string()),
+    );
+    let mut root_params = CertificateParams::new(["127.0.0.1".into(), "localhost".into()]);
+    root_params.is_ca = IsCa::Ca(rcgen::BasicConstraints::Unconstrained);
+    root_params.distinguished_name = dn;
+    root_params.key_usages = vec![KeyUsagePurpose::CrlSign, KeyUsagePurpose::KeyCertSign];
+    let cert = Certificate::from_params(root_params)?;
+    debug!("writing root private key to {private_key:?}");
```
sehz, line 183: `debug(private_key, "writing to root private key")`

crates/fluvio-cluster/src/runtime/local/k8s.rs
```diff
@@ -0,0 +1,211 @@
...
+    std::fs::write(private_key, cert.serialize_private_key_pem())?;
+    debug!("writing root public key to {public_key:?}");
```
sehz, line 185: same

crates/fluvio-cluster/src/runtime/local/k8s.rs
```diff
@@ -0,0 +1,211 @@
...
+        kube_cert,
+        service_account_private_key,
+        service_account_cert,
+        root_cert,
+        client_private_key,
+        client_cert,
+    })
+}
+
+fn generate_root_cert(private_key: &Path, public_key: &Path) -> anyhow::Result<Certificate> {
```
sehz, line 171: Suggest change `private_key` to `private_key_path` and same for public_key.  Got confused reading code so it was key value itself

crates/fluvio-cluster/src/runtime/local/k8s.rs
```diff
@@ -0,0 +1,211 @@
...
+    std::fs::write(public_key, cert.serialize_pem()?)?;
+    Ok(cert)
+}
+
+fn generate_cert(
+    root: &Certificate,
+    private_key: &Path,
+    public_key: &Path,
+) -> anyhow::Result<Certificate> {
+    let cert = rcgen::generate_simple_self_signed(["127.0.0.1".into(), "localhost".into()])?;
```
sehz, line 195: let's make loopback and localhost as constant (or it is defined somewhere)

crates/fluvio-cluster/src/runtime/local/k8s.rs
```diff
@@ -0,0 +1,211 @@
...
+
+            config.current_context = context.name.clone();
+            config.put_user(user);
+            config.put_cluster(cluster);
+            config.put_context(context);
+
+            config.save()?;
+
+            info!(
+                "k8s context succesfully added  to {}",
```
sehz, line 131: `info(path = config.path_to_string_lossy(),,`

sehz (changes requested):
Overall, it looks good.  Should update progress message.
There are a few minor nits & clean up log messages and separate out major config.
Also, I didn't see a clean-up of embedded processes and files when cluster is deleted

crates/fluvio-cli-common/src/http.rs
galibey: Done. However, I called it "HttpsReadinessCheck" instead of "HttpUrlUtil"

crates/fluvio-cluster/src/error.rs
galibey: It should be migrated all together. For now, it is better to keep the same style.

crates/fluvio-cluster/src/progress.rs
galibey: Done

crates/fluvio-cluster/src/progress.rs
galibey: Done

crates/fluvio-cluster/src/runtime/local/etcd.rs
galibey: Done

crates/fluvio-cluster/src/runtime/local/etcd.rs
galibey: Done

crates/fluvio-cluster/src/runtime/local/etcd.rs
galibey: Not sure I understood what you mean here. This is a log file that will reside in the same directory as other logs. The directory itself is managed upper in the stack.

crates/fluvio-cluster/src/runtime/local/k8s.rs
galibey: This k8s api will be dedicated to fluvio only and only one namespace will be used, so there will be no need for any separation, and the "default" is convenient, because if you want to look at it by `kubectl`, it is just less typing there, no need to type namespace name.

crates/fluvio-cluster/src/runtime/local/k8s.rs
galibey: Why? They are all in the same directory. Fewer jumps are needed if you want to read them.

crates/fluvio-cluster/src/runtime/local/k8s.rs
galibey: Done

crates/fluvio-cluster/src/runtime/local/k8s.rs
galibey: Done

crates/fluvio-cluster/src/runtime/local/k8s.rs
galibey: Done

crates/fluvio-cluster/src/runtime/local/k8s.rs
galibey: The '_path' suffix would be needed if the type was 'String', but here type is 'Path' which excludes any confusion.

crates/fluvio-cluster/src/runtime/local/k8s.rs
galibey: done

crates/fluvio-cluster/src/runtime/local/k8s.rs
galibey: done

galibey:
>  Also, I didn't see a clean-up of embedded processes and files when cluster is deleted

All files are created within the `data_dir` folder (by default it is $HOME/.fluvio/data) that is created on `cluster start` and deleted on `cluster delete` . 
Processes are shut by name here: https://github.com/infinyon/fluvio/blob/a4bf600d3f77a52807cd6500942981ab19baa6b0/crates/fluvio-cluster/src/delete.rs#L183

crates/fluvio-cluster/src/runtime/local/k8s.rs
sehz: fair enough

sehz (approved):
LGTM.  Nice job

galibey:
Marked as a draft as we want to explore other alternatives.

PR #3542, opened 2023-09-18
feat: add in memory client for store context

morenol:
Added MemoryClient in `fluvio-stream-dispatcher` crate. Also, added unit test for trait impl used

### DIFF
--- a/crates/fluvio-stream-dispatcher/src/store/memory.rs
+++ b/crates/fluvio-stream-dispatcher/src/store/memory.rs
@@ -1,0 +1,534 @@
+use std::collections::HashMap;
+use std::sync::Arc;
+use core::fmt::Display;
+
+use async_channel::{Sender, Receiver, bounded};
+use async_lock::Mutex;
+use async_lock::RwLock;
+use tracing::debug;
+use futures_util::FutureExt;
+use futures_util::StreamExt;
+use futures_util::stream::BoxStream;
+use serde::Serialize;
+use serde::de::DeserializeOwned;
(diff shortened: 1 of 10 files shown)


crates/fluvio-stream-dispatcher/src/store/memory.rs
```diff
@@ -0,0 +1,534 @@
...
+            drop(lock);
+            st2.watch_stream()
+        };
+
+        ft_stream.flatten_stream().boxed()
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum MemoryClientError {
```
sehz, line 379: I think it is better to migrate to anyhow

sehz (changes requested):
overall looks good.  couple of comments.
one critical thing need is validate lock design.  Can you try this https://amanieu.github.io/parking_lot/parking_lot/deadlock/index.html?.

crates/fluvio-stream-dispatcher/src/store/memory.rs
morenol: The issue is that the trait `MetadataClientError` has several [constraints](https://github.com/infinyon/k8-api/blob/577ad7e3894ea2cf3317461ec1a43dafa841a916/src/k8-metadata-client/src/client.rs#L97C1-L103C1): , and using anyhow would require 

    type MetadataClientError: MetadataClientError
        + Send
        + Display
        + From<IoError>
        + From<DiffError>
        + From<SerdeJsonError>;

PR #3546, opened 2023-09-19
[Merged by Bors] - build: update k8 crates

morenol:


crates/fluvio-cli/src/error.rs
```diff
@@ -32,10 +32,6 @@ pub enum CliError {
     #[error("Kubernetes config error: {0}")]
     K8ConfigError(#[from] k8_config::ConfigError),
```
sehz, line 33: do we still need this one?

crates/fluvio-cli/src/profile/sync/k8.rs
```diff
@@ -112,22 +112,22 @@ pub async fn set_k8_context(opt: K8Opt, external_addr: String) -> Result<Profile
...
         .retrieve_item::<ServiceSpec, _>(&InputObjectMeta::named("fluvio-sc-public", ns))
         .await
     {
         Ok(svc) => svc,
-        Err(err) => match err {
-            k8_client::ClientError::ApiResponse(status)
-                if status.code == Some(StatusCode::NOT_FOUND.as_u16()) =>
+        Err(err) => {
+            if let Some(MetaStatus {
+                code: Some(404), ..
```
sehz, line 123: use http crate?

sehz (changes requested):
Nice cleanup.  couple of nits

crates/fluvio-cli/src/error.rs
morenol: I didn't update it, but it seems to compile successfully without this

crates/fluvio-cli/src/profile/sync/k8.rs
morenol: Not sure what you mean. 

This is just replacing the old

```
k8_client::ClientError::ApiResponse(status)
                if status.code == Some(StatusCode::NOT_FOUND.as_u16()) =>
```
                
using the anyhow downcasting to extract the status code from the inner error.

sehz (approved):
LGTM

PR #3556, opened 2023-09-25
feat: added k8s extensions publishing

galibey:
Include k8s extensions to the fluvio artifacts publishing process.


sehz:
also, we seem to repeating `gh release ...`,  should centralize this later

galibey:
Marked as a draft as we want to explore other alternatives.

PR #3566, opened 2023-09-29
[Merged by Bors] - feat: fvm API Client and updated type definitions

EstebanBorai:
Provides the FVM API Client which serves as the Channel Manager.
Also provides unit tests for `Channel` and updates type definitions.


crates/fluvio-hub-util/src/fvm/api.rs
```diff
@@ -0,0 +1,116 @@
+//! Hub FVM API Client
+
+use anyhow::{Error, Result};
+use surf::get;
+use url::Url;
+
+use super::{Channel, PackageSet, RustTarget};
+
+/// Base URL for `fmv` related API requests
+pub const API_BASE_URL: &str = "https://hub.infinyon.cloud/hub/v1/fvm";
```
sehz, line 10: this should be using same url for entire crate

crates/fluvio-hub-util/src/fvm/api.rs
```diff
@@ -0,0 +1,116 @@
...
+
+/// HTTP Client for interacting with the Hub FVM API
+pub struct Client {
```
sehz, line 13: can we reuse same client instead of duplicating?

crates/fluvio-hub-util/src/fvm/api.rs
```diff
@@ -0,0 +1,116 @@
...
+impl Client {
+    pub fn new(api_url: Url) -> Self {
+        Self { api_url }
+    }
+
+    /// Fetches a [`PackageSet`] from the Hub with the specific [`Channel`]
+    #[allow(unused)]
+    pub async fn fetch_package_set<S: AsRef<str>>(
+        &self,
+        name: S,
```
sehz, line 33: use `impl AsRef`

crates/fluvio-hub-util/src/fvm/api.rs
```diff
@@ -0,0 +1,116 @@
...
+        channel: &Channel,
+        arch: &RustTarget,
+    ) -> Result<PackageSet> {
+        let url = self.make_fetch_package_set_url(name, channel, arch)?;
+        let mut res = get(url).await.map_err(|err| {
+            tracing::error!("Failed to fetch package set: {}", err);
```
sehz, line 39: remove redudant error

crates/fluvio-hub-util/src/fvm/api.rs
```diff
@@ -0,0 +1,116 @@
...
+            tracing::error!("Failed to parse package set: {}", err);
+            Error::msg(err.to_string())
+        })?;
+
+        Ok(pkg)
+    }
+
+    /// Builds the URL to the Hub API for fetching a [`PackageSet`] using the
+    /// [`Client`]'s `api_url`.
+    fn make_fetch_package_set_url<S: AsRef<str>>(
```
sehz, line 52: this seems like separate util?

sehz: concern that this URL has too many nested so make it hard to maintain.
name, channel, arch seems to be qualifier.

sehz (changes requested):

crates/fluvio-hub-util/src/fvm/api.rs
EstebanBorai: Here I map it to `anyhow::Error` because `surf::Error` doesn't implement `StdError` so we cant use the `?` operator straight.

crates/fluvio-hub-util/src/fvm/api.rs
EstebanBorai: I did it like this to be flexible for development as we work on Hub's dev instance as well.

By merging FVM API Client with `HubAccess` client we would be inheriting authentication protocols from the Hub Access which are not relevant for FVM given that FVM is a public API and Hub Access requires authentication and token signing.

```
    pub fn new() -> Self {
        HubAccess {
            remote: HUB_REMOTE.to_string(),
            hubid: String::new(),
            pkgkey: String::new(),
            pubkey: String::new(),
        }
    }
```

I could move the URI to the `make_url...` methods as well and reuse `HUB_REMOTE` URL as the de-facto base URL.

crates/fluvio-hub-util/src/fvm/api.rs
sehz: I meant, remove `tracing:error` since it will be converted to upstream error

crates/fluvio-hub-util/src/fvm/api.rs
EstebanBorai: This builds the URL for the request separately so it can be isolated when testing, is strongly tied to the client given that is only used for this request.

Each of the params should be provided by the consumer:

- Name for the package set to fetch
- Version/Channel
- Architecture of the host machine to download compatible binaries

crates/fluvio-hub-util/src/fvm/api.rs
EstebanBorai: Oh! Of course!

crates/fluvio-hub-util/src/fvm/api.rs
```diff
@@ -0,0 +1,115 @@
...
+
+/// HTTP Client for interacting with the Hub FVM API
+pub struct Client {
+    api_url: Url,
+}
+
+impl Default for Client {
+    fn default() -> Self {
+        // Safe to `unwrap`, `API_BASE_URL` is a valid URL and it's a constant
+        Self::new(Url::parse(HUB_REMOTE).unwrap())
```
sehz, line 19: remove unwrap

sehz (changes requested):

sehz (approved):
LGTM.  there are couple of `AsRef` that could be converted to `impl` but not show stopper

EstebanBorai:
> LGTM. there are couple of `AsRef` that could be converted to `impl` but not show stopper

Actually lets fix them to have a clean codebase! :D. Thanks @sehz!
https://github.com/infinyon/fluvio/pull/3566/commits/f7e1dded7fdaafeaec1a0e632c077aee0e17469a

PR #3570, opened 2023-09-30
[Merged by Bors] - feat: FVM `self` subcommand

EstebanBorai:
Introduces a subcommand for installing/uninstalling and in the short term also updating
FVM. This command is in charge of preparing the environment for FVM before installing
and switching Fluvio Versions.

---

## Demo

CleanShot 2023-09-30 at 19 21 45


crates/fluvio-version-manager/Cargo.toml
```diff
@@ -0,0 +1,38 @@
+[package]
+name = "fluvio-version-manager"
+version = "0.1.0"
```
sehz, line 3: set version to `0.0.0` of unpublished crate

crates/fluvio-version-manager/Cargo.toml
```diff
@@ -0,0 +1,38 @@
...
+repository = "https://github.com/infinyon/fluvio"
+license = "Apache-2.0"
+publish = false
+
+[[bin]]
+name = "fvm"
+path = "src/main.rs"
+
+[dependencies]
+dialoguer = "0.11.0"
```
sehz, line 16: move this to workspace

crates/fluvio-version-manager/Cargo.toml
```diff
@@ -0,0 +1,38 @@
...
+colored = { workspace = true }
+color-eyre = { workspace = true, default-features = false }
+comfy-table = { workspace = true }
+dirs = { workspace = true }
+hex = { workspace = true }
+semver = { workspace = true }
+serde = { workspace = true }
+serde_json = { workspace = true }
+sha2 = { workspace = true }
+surf = { workspace = true }
```
sehz, line 30: this should not have access to `surf` since all downloading thru should delegated to `hub-util`

crates/fluvio-version-manager/Cargo.toml
```diff
@@ -0,0 +1,38 @@
...
+tempfile = { workspace = true }
+tracing = { workspace = true }
+tracing-subscriber = { workspace = true, features = ["env-filter", "fmt"] }
```
sehz, line 33: use `fluvio-future`

crates/fluvio-version-manager/src/command/itself/install.rs
```diff
@@ -0,0 +1,102 @@
+use std::env::current_exe;
+use std::fs::{create_dir, copy, write};
+use std::path::PathBuf;
+
+use clap::Parser;
+use color_eyre::eyre::Result;
```
sehz, line 6: use anyhow

crates/fluvio-version-manager/src/command/itself/install.rs
```diff
@@ -0,0 +1,102 @@
...
+    *:"$HOME/.fvm/bin":*)
+        ;;
+    *)
+        export PATH="$PATH:$HOME/.fvm/bin"
+        ;;
+esac
+"#;
+
+#[derive(Clone, Debug, Parser)]
+pub struct InstallOpt {
```
sehz, line 27: probably rename Ito `SelfInstall` to make it obvious

crates/fluvio-version-manager/src/command/itself/install.rs
```diff
@@ -0,0 +1,102 @@
...
+            ?fvm_dir,
+            "Copied the FVM binary to the FVM home directory with success"
+        );
+
+        // Creates the package set directory
+        let fvm_pkgset_dir = fvm_pkgset_path()?;
+        create_dir(fvm_pkgset_dir)?;
+
+        // Creates the `env` file
+        let fvm_env_file_path = fvm_dir.join("env");
```
sehz, line 91: what is purpose of this?

crates/fluvio-version-manager/src/command/itself/install.rs
```diff
@@ -0,0 +1,102 @@
...
+        write(fvm_env_file_path, FVM_ENV_FILE_CONTENTS)?;
+
+        Ok(fvm_dir)
+    }
+}
+
```
sehz, line 97: please add unit test and CLI test

crates/fluvio-version-manager/src/command/itself/mod.rs
```diff
@@ -0,0 +1,48 @@
+//! FVM Management Commands
+
+pub mod install;
+pub mod uninstall;
+
+use color_eyre::eyre::Result;
```
sehz, line 6: use anyhow

crates/fluvio-version-manager/src/command/itself/mod.rs
```diff
@@ -0,0 +1,48 @@
...
+        match &self.command {
+            ItselfCommand::Install(cmd) => cmd.process().await?,
+            ItselfCommand::Uninstall(cmd) => cmd.process().await?,
+        }
+
+        Ok(())
+    }
+}
+
+impl Notify for ItselfOpt {
```
sehz, line 44: why this is repeated here?

crates/fluvio-version-manager/src/command/itself/uninstall.rs
```diff
@@ -0,0 +1,56 @@
+use std::fs::remove_dir_all;
+
+use clap::Parser;
+use color_eyre::eyre::Result;
```
sehz, line 4: use anyhow

crates/fluvio-version-manager/src/command/itself/mod.rs
```diff
@@ -0,0 +1,48 @@
...
+pub enum ItselfCommand {
+    /// Install `fvm` and setup the workspace
+    Install(InstallOpt),
+    /// Uninstall `fvm` and removes the workspace
+    Uninstall(UninstallOpt),
+}
+
+/// The `install` command is responsible of installing the desired Package Set
+#[derive(Debug, Parser)]
+pub struct ItselfOpt {
```
sehz, line 25: Change to `SelfInstallOpt`

crates/fluvio-version-manager/src/command/itself/uninstall.rs
```diff
@@ -0,0 +1,56 @@
...
+
+        if workdir_path.exists() {
+            if Confirm::with_theme(&ColorfulTheme::default())
+                .with_prompt(&format!(
+                    "Are you sure you want to uninstall FVM from {}?",
+                    workdir_path.display().italic()
+                ))
+                .interact()?
+            {
+                remove_dir_all(&workdir_path)?;
```
sehz, line 33: probably have error handling here to give warning message if remove dir didn't work

crates/fluvio-version-manager/src/command/itself/uninstall.rs
```diff
@@ -0,0 +1,56 @@
...
+                self.notify_done(format!(
+                    "Fluvio Version Manager was removed from {}",
+                    workdir_path.display().italic()
+                ));
+            }
+
+            return Ok(());
```
sehz, line 40: better use if else to make it obvious

crates/fluvio-version-manager/src/common/mod.rs
```diff
@@ -0,0 +1,16 @@
+pub mod notify;
+pub mod settings;
+pub mod workdir;
+
+use std::path::PathBuf;
+
+use color_eyre::eyre::{Error, Result};
```
sehz, line 7: use anyhow

crates/fluvio-version-manager/src/common/mod.rs
```diff
@@ -0,0 +1,16 @@
...
+
+/// Wrapper on `dirs::home_dir` which returns `anyhow::Error` instead of `Option`.
+pub(super) fn home_dir() -> Result<PathBuf> {
+    if let Some(home_dir) = dirs::home_dir() {
```
sehz, line 11: use if else

crates/fluvio-version-manager/src/common/notify.rs
```diff
@@ -0,0 +1,29 @@
+use colored::Colorize;
+
+pub trait Notify {
+    fn is_quiet(&self) -> bool;
+
+    fn notify_info(&self, message: impl AsRef<str>) {
```
sehz, line 6: instead of creating another trait use terminal trait in other CLI

crates/fluvio-version-manager/src/common/settings.rs
```diff
@@ -0,0 +1,66 @@
...
+
+pub const SETTINGS_TOML_FILENAME: &str = "settings.toml";
+
+/// The `settings.toml` is in charge of keeping track of the active version
+/// through the default key, which holds the name of the directory under
+/// `~/.fvm/pkgset/default/versions` for the desired default version.
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Settings {
+    /// The active `channel` for the Fluvio Installation
+    pub channel: Option<Channel>,
```
sehz, line 20: this should be enum? Since it'e either channel or version

crates/fluvio-version-manager/src/common/settings.rs
```diff
@@ -0,0 +1,66 @@
...
+    /// The specific version in use
+    pub version: Option<Version>,
+}
+
+impl Settings {
+    /// Used to create an empty `settings.toml` file. This is used when the user
+    /// installs FVM but no version is set yet.
+    pub fn init() -> Result<Self> {
```
sehz, line 28: pub(crate) since this will not be exposed.
rename to `load` which would be more obvious?

crates/fluvio-version-manager/src/common/settings.rs
```diff
@@ -0,0 +1,66 @@
...
+        Ok(())
+    }
+
+    /// Retrieves the path to the `settings.toml` file for this host
+    fn settings_file_path() -> Result<PathBuf> {
+        let fvm_path = fvm_workdir_path()?;
+        let settings_path = fvm_path.join(SETTINGS_TOML_FILENAME);
+
+        Ok(settings_path)
+    }
```
sehz, line 65: please add unit test

crates/fluvio-version-manager/src/common/workdir.rs
```diff
@@ -0,0 +1,41 @@
+//! The `Workdir` is the directory used by Fluvio Version Manager (FVM) to
+//! store its files and binaries.
+
+use std::path::PathBuf;
+
+use color_eyre::eyre::Result;
```
sehz, line 6: use anyhow

crates/fluvio-version-manager/src/common/workdir.rs
```diff
@@ -0,0 +1,41 @@
...
+/// FVM Binary Name
+pub const FVM_BINARY_NAME: &str = "fvm";
+
+/// FVM Packages Set Directory Name
+pub const FVM_PACKAGES_SET_DIR: &str = "pkgset";
+
+/// Retrieves the path to the `~/.fvm` directory in the host system
+pub fn fvm_workdir_path() -> Result<PathBuf> {
+    let home = home_dir()?;
+    let fvm_path = home.join(FVM_HOME_DIR);
```
sehz, line 22: this can be condensed into `Ok(home.join(FVM_HOME_DIR)`

crates/fluvio-version-manager/src/common/workdir.rs
```diff
@@ -0,0 +1,41 @@
...
+    Ok(fvm_binary_path)
+}
+
+/// Retrieves the path to the `~/.fvm/pkgset` directory in the host system
+pub fn fvm_pkgset_path() -> Result<PathBuf> {
+    let fvm_workdir_path = fvm_workdir_path()?;
+    let fvm_pkget_path = fvm_workdir_path.join(FVM_PACKAGES_SET_DIR);
+
+    Ok(fvm_pkget_path)
+}
```
sehz, line 41: please add unit test

crates/fluvio-version-manager/src/main.rs
```diff
@@ -0,0 +1,79 @@
...
+
+use clap::{Args, Parser};
+use color_eyre::eyre::Result;
+
+use self::command::itself::ItselfOpt;
+
+#[async_std::main]
+async fn main() -> Result<()> {
+    color_eyre::install()?;
+    let _ = tracing_subscriber::fmt()
```
sehz, line 12: use fluvio future

sehz (changes requested):

sehz:
please add CLI tests in addition to unit test

crates/fluvio-version-manager/Cargo.toml
EstebanBorai: Gotcha! Yes Im not using it here, its from migration C/P. Removing immediately!

crates/fluvio-version-manager/src/command/itself/install.rs
EstebanBorai: Oh! I moved it to `color_eyre` given that it provides good UserEx for error reporting.
Now that `fvm` CLI only holds presentational logic, errors bubble up to `main`, the `color_eyre` `Error` includes Error reporting features.

Here is an example from docs:

```
color-eyre on  hooked [$!] is 📦 v0.5.0 via 🦀 v1.44.0
❯ cargo run --example custom_section
    Finished dev [unoptimized + debuginfo] target(s) in 0.04s
     Running `target/debug/examples/custom_section`
Error:
   0: Unable to read config
   1: cmd exited with non-zero status code

Stderr:
   cat: fake_file: No such file or directory

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ SPANTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

   0: custom_section::output2 with self="cat" "fake_file"
      at examples/custom_section.rs:14
   1: custom_section::read_file with path="fake_file"
      at examples/custom_section.rs:58
   2: custom_section::read_config
      at examples/custom_section.rs:63

Suggestion: try using a file that exists next time
```

Source: https://docs.rs/color-eyre/latest/color_eyre/#tldr

crates/fluvio-version-manager/src/command/itself/install.rs
EstebanBorai: Adds the `env` file to the `~/.fvm` directory. The `env` file is a Bash script used to add `fvm` to the path.

The user will be able to decide if it wants to add `fvm` to the path by running:

`source $HOME/.fvm/env`

This is the same approach used for `rustup` after installation. The rustup counterpart can be found at: `$HOME/.cargo/env`.

crates/fluvio-version-manager/src/command/itself/install.rs
EstebanBorai: Of course! I would like to add them before merging so we address all improvements on impl and just focus on test files then

crates/fluvio-version-manager/src/command/itself/mod.rs
EstebanBorai: Because this provides `ItselfOpt` context to determine if output should be printed.
I plan to do a derive macro later so we reduce boilerplate.

This makes the `GlobalConfig` `is_quiet` prop acknowledged when printing to stdout.

crates/fluvio-version-manager/src/command/itself/uninstall.rs
EstebanBorai: Sure! `color_eyre` will make it easier to debug as well by providing spantrace and error messages by layer.

crates/fluvio-version-manager/src/command/itself/install.rs
sehz: This kind of stack trace is only relevant for developers.   This has been removed from fluvio CLI.    Formatting of error should be decoupled from error handling itself.

crates/fluvio-version-manager/src/command/itself/install.rs
sehz: k

crates/fluvio-version-manager/src/command/itself/mod.rs
sehz: Do we need to this now?  Just add print when needed.  Otherwise you can just add to logging.

EstebanBorai: Its for user feedback. `Notify` provides  methods to keep it standardized on all commands

crates/fluvio-version-manager/src/command/itself/uninstall.rs
EstebanBorai: Achually the `?` will short circuit and return early in this case, with the error message, adding another message on screen would be redundant no?

crates/fluvio-version-manager/src/common/settings.rs
EstebanBorai: Yeah, it is `enum`. Channel is an enum with either `Stable`, `Latest` or `Version`.

Here is the source for when we introduced it in the previous PR: https://github.com/infinyon/fluvio/commit/f72af5d9c35fa91359c013445651f76a98fe9206#diff-c1995b0a457b64145b5f1912f0b961dfc7116178a89908cb2580ba3ad903f142R35-R42

crates/fluvio-version-manager/src/common/settings.rs
EstebanBorai: `load` would mean reading from file no? `init` similar to for example `git init` is more related to creation of a workspace or files.

I would expect `load` to fail if `settings.toml` doesn't exists

crates/fluvio-version-manager/src/command/itself/mod.rs
EstebanBorai: Another advantage is the fact that is config aware, every command has access to the top-level `GlobalConfig`, `Notify` takes advantage of this to use as global flag for no output which is helpful in CI Tests if we want to focus on certain output.

Looking for other impls in `fluvio-cli`, I found uses on `println!`

https://github.com/infinyon/fluvio/blob/master/crates/fluvio-cli/src/install/plugins.rs#L115-L126

The disadvantage is that the user doesn't have an option to suppress this, other than piping output somewhere else.

sehz: Let's get something working and then improve rather than trying to do everything once.  If this is absolutely needed then let's use(improve) the `terminal` interface which is in common.   No need to have two different ways of doing things

crates/fluvio-version-manager/src/command/itself/uninstall.rs
sehz: just catch and provide user friendly message

crates/fluvio-version-manager/src/common/settings.rs
sehz: Then why it is repeated here?

crates/fluvio-version-manager/src/common/settings.rs
```diff
@@ -0,0 +1,84 @@
...
+    #[test]
+    fn test_settings_file_path() {
+        let settings_path =
+            Settings::settings_file_path().expect("Failed to get settings.toml path");
+        let home = home_dir().expect("Failed to get home directory");
+
+        assert!(settings_path.is_absolute());
+        assert!(settings_path.starts_with(home));
+        assert!(settings_path.ends_with(SETTINGS_TOML_FILENAME));
+    }
```
sehz, line 83: have test for loading and parsing

sehz (changes requested):

ajhunyady:
@EstebanBorai do we also need an `update` self command.

crates/fluvio-version-manager/src/common/settings.rs
```diff
@@ -0,0 +1,142 @@
...
+
+        let settings_str =
+            read_to_string(&settings_path).expect("Failed to read settings.toml file");
+
+        assert!(settings_str.contains("channel = \"stable\""));
+        assert!(settings_str.contains("version = \"0.11.0\""));
+
+        remove_file(&settings_path).expect("Failed to remove settings.toml file");
+        delete_fvm_dir();
+    }
```
EstebanBorai, line 141: Is important to note that these tests will create/delete the directory `~/.fvm`, it would be nice to run them only on CI.

EstebanBorai: Perhaps we could guard them behind a environment variable? cc: @sehz

crates/fluvio-version-manager/src/common/settings.rs
EstebanBorai: Is imported from Hub's FVM

sehz: Let's have a single definition

crates/fluvio-version-manager/src/common/settings.rs
sehz: Yes, that's fine FMV directory path can be overridden in env VAR

.github/workflows/ci.yml
```diff
@@ -341,6 +367,16 @@ jobs:
...
         run: make build-fbm
 
+      - name: Build fvm
+        timeout-minutes: 40
+        if: matrix.binary == 'fvm'
+        run: make build-fvm
+
+      - name: Build fvm.exe
+        timeout-minutes: 40
+        if: matrix.binary == 'fvm.exe'
```
sehz, line 377: this can be combine as ```matrix.binary == `fvm` || matrix.binary == `fvm.exe`

crates/fluvio-version-manager/src/common/settings.rs
sehz: this needs to be fixed.  it is probably most critical part of this PR

crates/fluvio-version-manager/src/common/settings.rs
```diff
@@ -0,0 +1,142 @@
...
+            "the settings file should exist at this point"
+        );
+
+        remove_file(settings_path).expect("Failed to remove settings.toml file");
+        delete_fvm_dir();
+    }
+
+    #[test]
+    fn update_settings_file() {
+        create_fvm_dir();
```
sehz, line 122: why this is calling this? shouldn't installation step should take care of it?

sehz (changes requested):
Still missing CLI / Unit tests.

I would expect following tests:
```
$ ../fvm self-install
```
```
$ .../fvm uninstall
```

And test to ensure directory is created correctly (install) with stable channel. and test to ensure everything is removed correctly

crates/fluvio-version-manager/src/common/settings.rs
EstebanBorai: But the definition its being reused no? The one in Hub Util crate.

Or are you referring to some other?

crates/fluvio-version-manager/src/common/settings.rs
EstebanBorai: This is testing tho Settings file creation only, these ale unit tests on the Settings struct

EstebanBorai:
> Still missing CLI / Unit tests.
> 
> 
> 
> I would expect following tests:
> 
> ```
> 
> $ ../fvm self-install
> 
> ```
> 
> ```
> 
> $ .../fvm uninstall
> 
> ```
> 
> 
> 
> And test to ensure directory is created correctly (install) with stable channel. and test to ensure everything is removed correctly
> 
> 

I think the tests you are mentioning are here: https://github.com/infinyon/fluvio/blob/8053a155b08ffc92b5e08ee8cae26dd9863c02ac/tests/cli/fvm_smoke_tests/fvm_basic.bats perhaps?

sehz:
These tests doesnt do anything since they dont have any validation

crates/fluvio-version-manager/src/common/settings.rs
EstebanBorai: As a mater of fact, the `Channel` struct present here is the one used by FVM to determine 3 possible states:

- Latest
- Stable
- A specific SemVer Compliant version number

This is the same enum used for FVM API Client.

The implementation lives in the Hub Util crate and is being imported to be used here, given that is also used in the Settings toml

sehz: I am still trying to understand why there are two different semantics difference

sehz: This is a struct while the other is an enum.   And it seems to have duplicate field `version` associated with it.

sehz: it would be good to have tests to show what it does

EstebanBorai:
> These tests doesnt do anything since they dont have any validation

Oh! Sorry about that! I just updated tests, now I also check the presence of the `~/.fvm/bin/fvm`. 

Im testing the following assertions:

### `fvm self install`

1. Create the `~/.fvm` directory
2. Create a `settings.toml`, but as no version of Fluvio has been installed already, the `settings.toml` remains empty as of this phase
3. Copies the `fvm` binary to `~/.fvm/bin/fvm`
4. Prints the command to add `fvm` to the path for the user

### `fvm self uninstall`

1. Delete the `~/.fvm` directory.

Previously I was checking the settings file to be there, is important to keep in mind that `fvm self install` doesnh have any side effects, no fluvio versions will be installed as of this phase. The `settings.toml` content is empty as no versions has been installed so far.

In the following iteration, with `fvm install`, `settings.toml` will be updated to have the
versions and channels that are active.

crates/fluvio-version-manager/src/common/settings.rs
EstebanBorai: > shouldn't installation step should take care of it?
Yeah! The command takes care of this but the `Settings` struct is not responsible of creating the directory.

In this context we just create the `settings.toml` file, se I have to prepare the environment for unit tests in this context

EstebanBorai:
@sehz just checked for existence using `test` command.
Checking for each file to be present, lets see how this goes through CI!

crates/fluvio-version-manager/src/common/settings.rs
```diff
@@ -0,0 +1,142 @@
...
+
+/// The `settings.toml` is in charge of keeping track of the active version
+/// through the default key, which holds the name of the directory under
+/// `~/.fvm/pkgset/default/versions` for the desired default version.
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Settings {
+    /// The active `channel` for the Fluvio Installation
+    pub channel: Option<Channel>,
+    /// The specific version in use
+    pub version: Option<Version>,
```
digikata, line 22: While a channel label (stable or latest) is optionally there, isn't there always a version number?

EstebanBorai: Yeah! But as of this point, as no Fluvio version has been installed, both `channel` and `version` will be `None`

tests/cli/fvm_smoke_tests/fvm_basic.bats
```diff
@@ -0,0 +1,53 @@
...
+    test -d ~/.fvm
+    assert_success
+
+    # Ensure the `fvm` binary is available
+    test -f ~/.fvm/bin/fvm
+    assert_success
+
+    # Ensure the `settings.toml` is available
+    test -f ~/.fvm/settings.toml
+    assert_success
```
sehz, line 32: Much better.  Still need to check if setting is correct

crates/fluvio-version-manager/src/common/settings.rs
```diff
@@ -0,0 +1,142 @@
...
+                settings_path.display()
+            )));
+        }
+
+        let initial = Self {
+            channel: None,
+            version: None,
+        };
+
+        initial.save()?;
```
sehz, line 43: Is there a reason this needs to be set to empty? why not save when channel is set

sehz (changes requested):

crates/fluvio-version-manager/src/common/settings.rs
sehz: This also could be lazily loaded like fluvio config

tests/cli/fvm_smoke_tests/fvm_basic.bats
EstebanBorai: At this point is an empty file, given that no version is installed already.
As of today `toml` doesn't support `null`/`nil` it just leaves the file empty.

sehz: After FVM is installed, should be assumed that the channel is set to stable

crates/fluvio-version-manager/src/common/settings.rs
EstebanBorai: Mmm yeah is part of setup files, so `fvm install` is not responsible of checking the existence of `settings.toml`. The other approach that comes to my mind would involve checking if the `settings.toml` exists and then write otherwise create.

tests/cli/fvm_smoke_tests/fvm_basic.bats
sehz: Or just not set at all. let channel manager set

digikata: I think we should leave that to install.sh and a test around that script to default to stable, this is self install and shouldn't have side effect logic.

digikata (approved):

EstebanBorai:
Thanks for your time @sehz and @digikata! Moving into `fvm install` now!
Refer: #3573

sehz:
There is an issue with the setting config.  Please address them on next PR

PR #3571, opened 2023-10-02
[Merged by Bors] - Add `version` command to `smdk` and `cdk`

### LINKED ISSUE #3552: `smdk version` and `cdk version`
Currently, the only way to know which version we're running is by reinstalling the binary.

We should:

- add a `smdk version` to display the current `smdk` version
- add a `cdk version` command to show the current `cdk` version

matheus-consoli:
Closes #3552

Add a `version` command to `smdk` and `cdk` displaying general information:

## Example output

```text
>> smdk version
SMDK CLI: 0.11.0-dev-1
SMDK CLI Arch: x86_64-unknown-linux-gnu
SMDK CLI SHA256: 77deb47876af0a57abbf01811be48e8a4a002b5c3e82aa044cb9bf1449c49b95
OS Details: Arch Linux rolling (kernel 6.5.3-arch1-1)
```


Cargo.toml
```diff
@@ -82,7 +82,7 @@ content_inspector = { version = "0.2" }
 crc32c = { version = "0.6"}
 criterion = "0.5.0"
 crossbeam-channel = "0.5"
-current_platform = { version = "0.2" }
+current_platform = { workspace = true }
```
morenol, line 85: maybe this line changed is not needed? not sure about the behavior of using this in the workspace Cargo.toml

crates/fluvio-cli-common/src/version_cmd.rs
```diff
@@ -0,0 +1,86 @@
...
+        println!("Fluvio CLI: {}", VERSION.trim());
+        println!("Fluvio CLI Arch: {CURRENT_PLATFORM}");
+
+        if let Some(sha) = self.format_cli_sha() {
+            println!("Fluvio CLI SHA256: {}", sha);
+        }
+
+        if let Some(sha) = self.format_frontend_sha() {
+            println!("Fluvio channel frontend SHA256: {}", sha);
+        }
```
tjtelan, line 38: I don't know if we want to do it now, but when `fvm` is introduced, but this will need to be removed soon. Your call for how to deal with it

The main problem right now is we're likely to always print this line even though its context is only valid for `fluvio` and that can be a little confusing.

Cargo.toml
matheus-consoli: oh, true! I didn't realize I did it, thanks.
til that cargo accepts that

tjtelan (commented):
Looks good so far. Some feedback on the printed output

crates/fluvio-cli-common/src/version_cmd.rs
```diff
@@ -0,0 +1,86 @@
...
+    ///
+    /// The following information is displayed:
+    /// - Release channel, if available;
+    /// - CLI version;
+    /// - Platform arch;
+    /// - CLI SHA256, if available;
+    /// - channel frontend SHA256, if available;
+    /// - Git hash, if available;
+    /// - OS details, if available;
+    pub fn process(self) -> Result<()> {
```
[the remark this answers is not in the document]
matheus-consoli: done, I chose to receive a name to display as an arg to "process" instead of reading the bin name to have a consistent display even if the user renames the binary - let me know if you think is better otherwise

crates/fluvio-cli-common/src/version_cmd.rs
matheus-consoli: removed

crates/fluvio-cli-common/src/version_cmd.rs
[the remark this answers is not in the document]
tjtelan: Feels a little weird, but it's reasonable for the consistency. I don't have an alternative to suggest

crates/smartmodule-development-kit/build.rs
tjtelan, line 1: Should this file be with `fluvio-cli-common`?

matheus-consoli: oh, yes! good catch, thanks

tjtelan (approved):

PR #3574, opened 2023-10-04
Use utc timestamp in cli-smdk-basic-test

matheus-consoli:
Some tests were failing in some parts of the day because they used `date`, which gives the local time, to compare with a UTC timestamp.


sehz (approved):
LGTM.  Good catch

PR #3576, opened 2023-10-04
[Merged by Bors] - feat(fvm): install fluvio versions

EstebanBorai:
Introduces `fvm install`, similar to `nvm install`, downloads artifacts for the specific
Fluvio Version and stores them in `~/.fvm/versions` for future use.

Along with downloading the artifacts, other routines are involved in this process:

1. Check binary Sha256 Sums
2. Provide executable permissions on Unix systems
3. Creating a `manifest.json` which helps determine the binaries versions

At this point no Fluvio Version is set as active/default, only the download of binaries is addressed as of this PR.

---

## Demo

https://github.com/infinyon/fluvio/assets/34756077/93b2115f-09ae-476f-b2ba-fbb1f9db2461


crates/fluvio-hub-util/src/fvm/api.rs
```diff
@@ -15,9 +13,8 @@ pub struct Client {
...
 impl Client {
     /// Creates a new [`Client`] with the default Hub API URL
-    #[allow(unused)]
-    pub fn new() -> Result<Self> {
-        let api_url = Url::parse(HUB_REMOTE)?;
+    pub fn new(url: &str) -> Result<Self> {
+        let api_url = url.parse::<Url>()?;
 
         Ok(Self { api_url })
     }
```
EstebanBorai, line 20: This allows us to set different URLs

sehz: This can be passed as env var rather than hard coding.  So this means
1. Pass base URL.
2. Set base URL in the CLI (this will automatically read env var)

EstebanBorai: Agree! Im now removing the hardcoded `HUB_REMOTE` constant and allowing the user to
pass the `url` to use.

As can be seen in the demo I just posted above, the `--registry` option can be passed
to determine the URL to use when fetching the Package Set from the Hub.

Additionally I can add an environment variable: `HUB_REGISTRY_URL` as a second approach
to pass the Hub URL, perhaps this is helpful for CI as well!

ajhunyady:
Does this command work with out parameters? I assume `fvm install` works (not in the video).

EstebanBorai:
> Does this command work with out parameters? I assume `fvm install` works (not in the video).

Yeah! By default `fvm install` will download `stable`

tests/cli/fvm_smoke_tests/fvm_basic.bats
```diff
@@ -87,5 +87,22 @@ setup_file() {
...
+
+    run $FVM_BIN install
+    assert_success
+
+    # Check for `stable` dir to be present under "versions"
+    test -d ~/.fvm/versions/stable
+    assert_success
+
+    # Check on manifest contents
+    cat ~/.fvm/versions/stable/manifest.json
```
sehz, line 106: why do you need manifest?

EstebanBorai: The `manifest.json` file holds both:

- Tag
- Channel

When we want to know the tag behind a channel we use `manifest.json` to get such info.
This is also helpful for the `fvm show` command, which renders the list of installed Fluvio Versions.

We can use each `Manifest.json` to tell the tag for channel versions in such case as well.

An alternative could be keeping this data in `settings.toml`, like a list of installed versions with its tags, we would have to write against that on each `fvm install`.

sehz: But you can scan directories for the installed version.  I am concerned that there are two sources of truth.

EstebanBorai: Yeah I get you, I thought about doing the same at first, but then I noticed, how can I tell the version for a channel dir?

If I find `latest` or `stable` how can I tell the exact version number to tell if theres an update?

I thought that by having a manifest I could get that information.

sehz: As just metadata, it is fine.

crates/fluvio-version-manager/src/command/install.rs
```diff
@@ -0,0 +1,285 @@
...
+/// The `install` command is responsible of installing the desired Package Set
+#[derive(Debug, Parser)]
+pub struct InstallOpt {
+    #[command(flatten)]
+    global_opts: GlobalOptions,
+    /// Version to install
+    #[arg(index = 1, default_value_t = Channel::Stable)]
+    version: Channel,
+    /// Registry used to fetch Fluvio Versions
+    #[arg(long, env = "HUB_REGISTRY_URL", default_value = HUB_REMOTE)]
```
sehz, line 50: Since another command will use this, move into common arg.   Now show stopper for this PR though

crates/fluvio-version-manager/src/command/install.rs
```diff
@@ -0,0 +1,285 @@
...
+                tmp_dir.path().join(&artif.name),
+                version_path.join(&artif.name),
+            )?;
+        }
+
+        Ok(version_path)
+    }
+
+    /// Writes out the manifest for the specified `PackageSet`
+    async fn write_out_manifest(
```
sehz, line 221: Separate manifest logic from here. This should be moved to Channel obj so can consolidate it's behavior

crates/fluvio-version-manager/src/command/install.rs
```diff
@@ -0,0 +1,285 @@
...
+        };
+        let contents = serde_json::to_string(&manifest)?;
+
+        write(manifest_path, contents)?;
+
+        Ok(())
+    }
+
+    /// Generates the Sha256 checksum for the specified file
+    fn shasum256(file: &File) -> Result<String> {
```
sehz, line 240: Ditto. Move this into the pkg download and check sum.

crates/fluvio-version-manager/src/command/install.rs
EstebanBorai: Good catch! Will move it to a global scope!

crates/fluvio-version-manager/src/command/install.rs
```diff
@@ -0,0 +1,285 @@
...
+
+        Ok(())
+    }
+}
+
+impl Notify for InstallOpt {
+    fn is_quiet(&self) -> bool {
+        self.global_opts.quiet
+    }
+}
```
sehz, line 280: Need to have unit tests for checksum, executable, and manifest logic.

crates/fluvio-version-manager/src/common/fluvio.rs
```diff
@@ -0,0 +1,57 @@
...
+use anyhow::Result;
+
+use super::home_dir;
+
+/// Home Directory for the Fluvio installation
+pub const FLUVIO_HOME_DIR: &str = ".fluvio";
+
+/// Retrieves the path to the `~/.fluvio` directory in the host system
+pub fn fluvio_path() -> Result<PathBuf> {
+    Ok(home_dir()?.join(FLUVIO_HOME_DIR))
```
sehz, line 13: The Fluvio path can be overridden by env var in the current installer.  We should preserve that behavior

crates/fluvio-version-manager/src/common/fluvio.rs
```diff
@@ -0,0 +1,57 @@
...
+    Ok(flv_dir)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_fvm_workdir_path() {
+        let fluvio_path = fluvio_path().expect("Failed to get fluvio path");
```
sehz, line 44: Need concrete assertion.  this should test directory doesn't exist

crates/fluvio-version-manager/src/common/fluvio.rs
```diff
@@ -0,0 +1,57 @@
...
+        let home = home_dir().expect("Failed to get home directory");
+
+        assert_eq!(fluvio_path, home.join(FLUVIO_HOME_DIR));
+    }
+
+    #[test]
+    fn test_fvm_bin_path() {
+        let fluvio_bin_path = fluvio_bin_path().expect("Failed to get fluvio bin path");
```
sehz, line 52: ditto

crates/fluvio-version-manager/src/common/mod.rs
```diff
@@ -6,6 +7,12 @@ use std::path::PathBuf;
 
 use anyhow::{Error, Result};
 
+/// The Target Architecture of the current build (e.g. "aarch64-apple-darwin")
+///
+/// This is injected by Cargo during the build process. Refer to `build.rs` for
+/// more information.
+pub const TARGET: &str = env!("TARGET");
```
sehz, line 14: Same issue.  These config items should be passed as parameters from CLI env var not from const

crates/fluvio-version-manager/src/common/workdir.rs
```diff
@@ -70,10 +66,10 @@ mod tests {
...
     #[test]
-    fn test_fvm_pkgset_path() {
-        let fvm_pkgset_path = fvm_pkgset_path().expect("Failed to get fvm pkgset path");
+    fn test_fvm_versions_path() {
+        let fvm_version_path = fvm_versions_path().expect("Failed to get fvm pkgset path");
         let fvm_path = fvm_workdir_path().expect("Failed to get fvm path");
 
-        assert_eq!(fvm_pkgset_path, fvm_path.join(FVM_PACKAGES_SET_DIR));
+        assert_eq!(fvm_version_path, fvm_path.join(FVM_VERSIONS_DIR));
     }
```
sehz, line 74: See the same comments from the fluvio bin directory.  Add actual assertion.  Don't just assume code does the right thing.

tests/cli/fvm_smoke_tests/fvm_basic.bats
```diff
@@ -51,3 +53,56 @@ setup_file() {
     ! test -d ~/.fvm
     assert_success
 }
+
+@test "Install Fluvio at 0.10.15" {
+    run $FVM_BIN self install
+    assert_success
```
sehz, line 59: How do you know this succeeds?  should check for output of bin

tests/cli/fvm_smoke_tests/fvm_basic.bats
```diff
@@ -51,3 +53,56 @@ setup_file() {
...
+
+    run $FVM_BIN install 0.10.15
+    assert_success
```
sehz, line 62: Check version number output

tests/cli/fvm_smoke_tests/fvm_basic.bats
```diff
@@ -51,3 +53,56 @@ setup_file() {
...
+
+    # Ensure the stable version dir is available
+    test -d ~/.fvm/versions/0.10.15
```
sehz, line 65: since version number would be change in future, make version number configurable instead of hard coding

tests/cli/fvm_smoke_tests/fvm_basic.bats
```diff
@@ -51,3 +53,56 @@ setup_file() {
...
+    assert_success
+
+    # Verify fluvio binary is present
+    test -f ~/.fvm/versions/0.10.15/fluvio
```
sehz, line 69: check if this is actual fluvio.  assert output

tests/cli/fvm_smoke_tests/fvm_basic.bats
```diff
@@ -51,3 +53,56 @@ setup_file() {
...
+    assert_success
+
+    # Verify fluvio-run binary is present
+    test -f ~/.fvm/versions/0.10.15/fluvio-run
```
sehz, line 73: ditto

tests/cli/fvm_smoke_tests/fvm_basic.bats
```diff
@@ -51,3 +53,56 @@ setup_file() {
...
+    assert_success
+
+    # Verify fluvio-cloud binary is present
+    test -f ~/.fvm/versions/0.10.15/fluvio-cloud
```
sehz, line 77: ditto

tests/cli/fvm_smoke_tests/fvm_basic.bats
```diff
@@ -51,3 +53,56 @@ setup_file() {
...
+    assert_success
+
+    # Verify cdk binary is present
+    test -f ~/.fvm/versions/0.10.15/cdk
```
sehz, line 81: ditto

tests/cli/fvm_smoke_tests/fvm_basic.bats
```diff
@@ -51,3 +53,56 @@ setup_file() {
...
+    assert_success
+
+    # Verify smdk binary is present
+    test -f ~/.fvm/versions/0.10.15/smdk
```
sehz, line 85: ditto

tests/cli/fvm_smoke_tests/fvm_basic.bats
```diff
@@ -51,3 +53,56 @@ setup_file() {
...
+    assert_success
+
+    # Check on manifest contents
+    cat ~/.fvm/versions/0.10.15/manifest.json
+    assert_output --partial "{\"channel\":\"0.10.15\",\"version\":\"0.10.15\"}"
+}
```
sehz, line 91: also check make sure number of binaries matches

tests/cli/fvm_smoke_tests/fvm_basic.bats
```diff
@@ -51,3 +53,56 @@ setup_file() {
...
+
+@test "fvm install straight installs `fluvio` stable" {
+    # Make sure `stable` is not installed already
+    ! test -d ~/.fvm/versions/stable
+    assert_success
+
+    run $FVM_BIN install
+    assert_success
```
sehz, line 99: check output

tests/cli/fvm_smoke_tests/fvm_basic.bats
```diff
@@ -51,3 +53,56 @@ setup_file() {
...
+
+    # Check for `stable` dir to be present under "versions"
+    test -d ~/.fvm/versions/stable
+    assert_success
+
+    # Check on manifest contents
+    cat ~/.fvm/versions/stable/manifest.json
+    assert_output --partial "{\"channel\":\"stable\""
```
sehz, line 107: also repeat all binaries are present

sehz (changes requested):

sehz:
Also, the download could fail at any time.  So in that case, the installation would need to be repeated

crates/fluvio-version-manager/src/command/install.rs
EstebanBorai: Here I added to the global scope: 0e39b0c0bf835696ad99b8e1f4975e68dd7dda95

crates/fluvio-version-manager/src/command/install.rs
```diff
@@ -0,0 +1,285 @@
...
+        self.install_package().await?;
+        Ok(())
+    }
+
+    async fn install_package(&self) -> Result<()> {
+        let pkgset = self.fetch_pkgset().await?;
+        let version_path = fvm_versions_path()?.join(self.version.to_string());
+
+        if version_path.exists() {
+            self.notify_warn(format!(
```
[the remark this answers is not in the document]
EstebanBorai: Actually yes! I did like this before to check on SemVer in case a channel was provided, so I can tell if an update is available. But now that we will have `fvm update` its not required anymore!

EstebanBorai:
> Also, the download could fail at any time. So in that case, the installation would need to be repeated

To be sure, like adding a retry?

crates/fluvio-version-manager/src/command/install.rs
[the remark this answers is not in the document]
EstebanBorai: Now we fetch after checking 71bfaac

sehz:
Assume that the user should be able to retry the installation.  that should be covered under CLI test

crates/fluvio-version-manager/src/command/install.rs
```diff
@@ -0,0 +1,285 @@
...
+                "Version {} is already installed. Aborting installation.",
```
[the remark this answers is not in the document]
EstebanBorai: This is for simplicity, the goal is to avoid side effects and have atomic commands.
Given that we also have a `fvm update` is easier to maintain the update as a separate command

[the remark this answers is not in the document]
sehz: That's confusing.   See comments about failed install; it's better to have the user resume if something goes wrong.  Ideally, all operation should be idempotent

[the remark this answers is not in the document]
sehz: There are other failure modes we don't know about.  So it's better to have the ability to re-install anytime

crates/fluvio-version-manager/src/command/install.rs
```diff
@@ -0,0 +1,285 @@
...
+                self.version.to_string().bold()
+            ));
+
+            if matches!(self.version, Channel::Stable) || matches!(self.version, Channel::Latest) {
```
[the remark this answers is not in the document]
EstebanBorai: This attempts to check for the instance of version to be a channel instead of a explicit version tag, to reduce verbosity I've introduced `is_version_tag` method on `Channel`.

This is because for cases where the user wants to install `latest` or `stable`, the underlying version tag must be printed.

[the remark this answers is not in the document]
EstebanBorai: Here I added the method in question: https://github.com/infinyon/fluvio/pull/3576/commits/6eb7d2f1a4a0384216fa87bb52c0210936090e11

crates/fluvio-version-manager/src/command/install.rs
```diff
@@ -0,0 +1,285 @@
...
+                self.notify_help(format!(
+                    "Use {} to update the local installation for this channel",
```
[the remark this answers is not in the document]
EstebanBorai: The goal here is to tell the user to employ `fvm update` instead of `fvm install` given that the version it is requiring to download is already downloaded. This is because the update procedure is different to download when it comes to manage the `.fvm/versions` directory.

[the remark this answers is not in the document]
sehz: this just adds more complexity and confusion.  let's remove

crates/fluvio-version-manager/src/command/install.rs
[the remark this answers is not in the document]
sehz: don't think that's relevant.  just add more complexity.  consolidate into a simple info

crates/fluvio-version-manager/src/command/install.rs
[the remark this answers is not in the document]
EstebanBorai: So if we have: `stable` version, and `fvm install stable` is executed, the command will:

- Delete the existing `versions/stable` directory
- Continue as `fvm install` usually does

Theres two observations I have about this approach:

- Users will be asked to run `fvm install stable` to update fluvio, perhaps not as intuitive as `fvm update`/`fvm update stable`

- Users will not keep a local copy of the previous version in case they want to stick to the previous, instead, users will have to explicitly install it and then use `fvm switch` to use such version.

Is it ok?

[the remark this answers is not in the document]
sehz: Don't delete. Make it idempotent.  Create the directory if it doesn't exist—only download the missing pkg.  So `fluvio install <channel>` is same as `fluvio update`

crates/fluvio-version-manager/src/command/install.rs
```diff
@@ -0,0 +1,285 @@
...
+
+            return Ok(());
+        }
+
+        // The `tmp_dir` must be dropped after copying the binaries to the
+        // destination directory. By dropping `tmp_dir` the directory will be
+        // deleted from the filesystem.
+        let tmp_dir = TempDir::new()?;
+
+        self.download_artifacts(&tmp_dir, &pkgset).await?;
```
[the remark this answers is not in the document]
EstebanBorai: Theres some special behavior going on here, the `TempDir` represents the temporal directory similar to a Database ORM's active model.

If I [drop `TempDir`, then the directory is deleted](https://docs.rs/tempdir/latest/tempdir/struct.TempDir.html#method.new), so by having the `TempDir` owned by the `download_artifacts` it would have to copy the downloaded binaries to a destination directory before returning.

To keep it simple/explicit, I rather specifying the out dir and handle it here.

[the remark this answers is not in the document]
sehz: That's fine, but downloading/caching logic should be isolated from the installer.  All the installer needs to do is check and ask the downloader to see if pkg exists.  The downloader will return if it is cached; otherwise, it will download from the hub.   It should be probably moved to hub util.

EstebanBorai:
@sehz I think it would be nice to add some logic to delete existing version directory before copying if the directory already exists. What do you think?

CleanShot 2023-10-05 at 22 21 17

EstebanBorai:
@sehz I did different changes here: [d07daa5](https://github.com/infinyon/fluvio/pull/3576/commits/d07daa5eebd7d7eddec9d59bc3baecba4291affa), they are listed in the commit message. Will continue with remaining! Let we know your thoughts!

EstebanBorai:
Im addressing tests at the end, when all behavior changes are applied!

sehz:
As a general principle, the least changes to the user's file system are better.   deleting directory is destructive operation which should be avoided much as possible

EstebanBorai:
> As a general principle, the least changes to the user's file system are better.   deleting directory is destructive operation which should be avoided much as possible

Agree! In that example Im copying the binaries but as its seen FS complains. I will try with atomic move and see if theres a different behavior.

EstebanBorai:
@sehz here 81255f94f6c25e2cf311184ae995aa07ad1136b0 we are handling the already existent version.

crates/fluvio-version-manager/src/command/install.rs
```diff
@@ -0,0 +1,285 @@
...
+        self.checksum_artifacts(&tmp_dir, &pkgset).await?;
+
+        let allocation_path = self.allocate_artifacts(&tmp_dir, &pkgset).await?;
```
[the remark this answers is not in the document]
EstebanBorai: This stores the downloaded binaries. I changed it to be `store_artifacts`.

crates/fluvio-version-manager/src/command/install.rs
```diff
@@ -0,0 +1,285 @@
...
+
+        self.write_out_manifest(allocation_path, &pkgset, &self.version)
```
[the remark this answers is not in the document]
EstebanBorai: Here I provided a dedicated module for this logic: [d07daa5](https://github.com/infinyon/fluvio/pull/3576/commits/d07daa5eebd7d7eddec9d59bc3baecba4291affa).

crates/fluvio-version-manager/src/command/install.rs
```diff
@@ -0,0 +1,285 @@
...
+            .await?;
+
+        self.notify_help(format!(
+            "Use {} to switch to this version",
```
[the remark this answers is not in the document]
EstebanBorai: Feedback to the user on how to use downloded version as the active one. `fvm switch` is not implemented here yet but you can find it in the original PR here #3562

crates/fluvio-version-manager/src/command/install.rs
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    ///  Performs the installation of the specified `PackageSet`
+    async fn fetch_pkgset(&self) -> Result<PackageSet> {
+        let client = Client::new(self.registry.as_str())?;
+        let pkgset = client
+            .fetch_package_set(DEFAULT_PKGSET, &self.version, TARGET)
+            .await?;
+
+        if matches!(self.version, Channel::Stable) || matches!(self.version, Channel::Latest) {
```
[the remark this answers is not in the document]
EstebanBorai: I've introduced a method on `Channel` for this purpose

crates/fluvio-version-manager/src/command/install.rs
```diff
@@ -0,0 +1,285 @@
...
+            self.notify_info(format!(
+                "Found {total} packages for {DEFAULT_PACKAGE_SET_DISPLAY_NAME}@{channel}. Current {channel} is {version}.",
+                channel = self.version.to_string().bold(),
+                total = pkgset.artifacts.len(),
+                version = pkgset.version.to_string().bold(),
+            ));
+        } else {
+            self.notify_info(format!(
```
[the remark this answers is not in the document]
EstebanBorai: Just consolidated into one!

crates/fluvio-version-manager/src/command/install.rs
```diff
@@ -0,0 +1,285 @@
...
+                total = pkgset.artifacts.len(),
+            ));
+
+            let mut res = surf::get(&artf.download_url)
+                .await
+                .map_err(|err| Error::msg(err.to_string()))?;
+
+            if res.status() == StatusCode::Ok {
+                let out_path = tmp_dir.path().join(&artf.name);
+                let mut file = File::create(&out_path).unwrap();
```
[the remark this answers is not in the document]
EstebanBorai: Gotcha! Thanks!

crates/fluvio-version-manager/src/command/install.rs
```diff
@@ -0,0 +1,285 @@
...
+                artf.name.bold(),
+                artf.version.to_string().bold()
+            ));
+        }
+
+        Ok(())
+    }
+
+    /// Verifies downloaded artifacts checksums against the upstream checksums
+    async fn checksum_artifacts(&self, tmp_dir: &TempDir, pkgset: &PackageSet) -> Result<()> {
```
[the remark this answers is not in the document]
EstebanBorai: This is for readability, to keep functions atomic. But I can merge them.

crates/fluvio-version-manager/src/command/install.rs
```diff
@@ -0,0 +1,285 @@
...
+                }
+            }
+        }
+
+        Ok(())
+    }
+
+    /// Allocates artifacts in the FVM `pkgset` directory for future use.
+    /// Returns the path to the allocated version directory.
+    async fn allocate_artifacts(&self, tmp_dir: &TempDir, pkgset: &PackageSet) -> Result<PathBuf> {
```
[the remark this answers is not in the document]
EstebanBorai: Stores artifacts into the `~/.fvm/versions` directory

crates/fluvio-version-manager/src/command/install.rs
EstebanBorai: This is only used by FVM and is not part of the `Channel` logic. Manifest uses `Channel` as value but `Channel` does not need `Manifest`

crates/fluvio-version-manager/src/common/mod.rs
EstebanBorai: The purpose of this is to gather the target triple to use for artifacts download.
This is not an env var that can be set at runtime, but instead is build time.

This is later used when fetching the registry to retrieve the compatible artifacts.
We could have an environment variable to overwrite this.

crates/fluvio-hub-util/src/fvm/api/download.rs
```diff
@@ -37,13 +37,14 @@ pub trait Download {
...
-    async fn download(&self, target_dir: PathBuf) -> Result<()>;
+    /// and returns the path to the downloaded artifact
+    async fn download(&self, target_dir: PathBuf) -> Result<PathBuf>;
 }
 
 #[async_trait]
 impl Download for Artifact {
     #[instrument(skip(self, target_dir))]
-    async fn download(&self, target_dir: PathBuf) -> Result<()> {
+    async fn download(&self, target_dir: PathBuf) -> Result<PathBuf> {
```
EstebanBorai, line 47: Added a minor improvement to have access to downloaded binary path

crates/fluvio-version-manager/src/command/install.rs
EstebanBorai: Checksum is now in the previous PR #3589. Im just missing Executable and Manifest!

EstebanBorai:
Any suggestions on this output would be nice! Im trying to keep it clean but insightful.

CleanShot 2023-10-07 at 18 00 04

sehz (changes requested):
Looks like almost everything is addressed except version config for assertion

EstebanBorai:
> Looks like almost everything is addressed except version config for assertion

Improved tests to check for version tags using `jq` also reduced scope by tests a bit to make them simpler

tests/cli/fvm_smoke_tests/fvm_basic.bats
sehz: This is still not resolved.  Make version configurable so that CI doesn't break next time version is upgraded

sehz (changes requested):

tests/cli/fvm_smoke_tests/fvm_basic.bats
```diff
@@ -35,19 +39,123 @@ setup_file() {
...
-    run $FVM_BIN self uninstall --yes
+    run bash -c '$FVM_BIN self uninstall --yes'
     assert_success
 
     # Ensure the `~/.fvm/` directory is not available anymore
     ! test -d ~/.fvm
     assert_success
 }
+
+@test "Install Fluvio at 0.10.15" {
```
EstebanBorai, line 59: This tests a specific version by passing the tag as argument

tests/cli/fvm_smoke_tests/fvm_basic.bats
```diff
@@ -35,19 +39,123 @@ setup_file() {
...
+    run bash -c '~/.fvm/versions/0.10.15/fluvio version > flv_version_0.10.15.out && cat flv_version_0.10.15.out | head -n 1 | grep "0.10.15"'
+    assert_output --partial "0.10.15"
+    assert_success
+
+    # Removes FVM
+    run bash -c '$FVM_BIN self uninstall --yes'
+    assert_success
+}
+
+@test "Install Stable Fluvio" {
```
EstebanBorai, line 111: This installs the stable Fluvio version as of the moment of executing the test. We use GitHub to make assertions on versions.

sehz (approved):
LGTM.  Nice CLI testing

EstebanBorai:
> LGTM.  Nice CLI testing

Thanks for the support!

.github/workflows/ci.yml
```diff
@@ -1130,7 +1130,8 @@ jobs:
       - name: Run FVM smoke tests
         if: matrix.test == 'fvm'
         timeout-minutes: 20
-        run: make FVM_BIN=~/bin/fvm cli-fvm-smoke
+        run: |
+          make FVM_BIN=~/bin/fvm HUB_REGISTRY_URL="https://hub-dev.infinyon.cloud" cli-fvm-smoke
```
sehz, line 1134: move HUB_REGISTRY_URL to env

EstebanBorai: Applying in next PR!

morenol:
bors retry

PR #3581, opened 2023-10-05
chore: use `surf` and `http-client` as workspace deps

EstebanBorai:
As these are also being used for FVM on #3576.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -100,6 +100,7 @@ hdrhistogram = "7.0"
 hex = "0.4"
 home = "0.5"
 http = { default-features = false, version = "0.2" }
+http-client = { version = "6.5.3", default-features = false, features = ["h1_client", "rustls"] }
 humantime = "2.0"
 humantime-serde = { version = "1.1.1", default-features = false }
 include_dir = "0.7.2"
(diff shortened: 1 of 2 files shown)


EstebanBorai:
> can we try to fix the unused dependency test on this PR?

Sure! Let me check...

EstebanBorai:
> > can we try to fix the unused dependency test on this PR?
> 
> Sure! Let me check...

Its related to `comfy-table` from `fluvio-version-manager`, removing it:

```
info: Loading depinfo from "/home/runner/work/fluvio/fluvio/target/debug/deps/smdk-35f5d453d440f31e.d"
info: Ignoring `fluvio` (Normal)
info: Ignoring `fluvio-socket` (Development)
unused dependencies:
`fluvio-version-manager v0.0.0 (/home/runner/work/fluvio/fluvio/crates/fluvio-version-manager)`
└─── dependencies
     └─── "comfy-table"
Note: They might be false-positive.
      For example, `cargo-udeps` cannot detect usage of crates that are only used in doc-tests.
      To ignore some dependencies, write `package.metadata.cargo-udeps.ignore` in Cargo.toml.
make: *** [makefiles/check.mk:25: check-udeps] Error 1
```

morenol (approved):

digikata:
We should keep an eye on especially some of the cross-platform builds after merging this.  It's not bad by itself but the surf features can be finicky and interact with the builds, and cargo workspace feature resolution can be finicky too.  

If a problem surfaces, we could consider moving surf to a common crate where we have simpler control of the feature configuration.  E.g. fluvio-cli-common also uses isahc (and maybe we could reduce the number http-client crates we use down by one).

EstebanBorai:
> We should keep an eye on especially some of the cross-platform builds after merging this.  It's not bad by itself but the surf features can be finicky and interact with the builds, and cargo workspace feature resolution can be finicky too.  
> 
> 
> 
> If a problem surfaces, we could consider moving surf to a common crate where we have simpler control of the feature configuration.  E.g. fluvio-cli-common also uses isahc (and maybe we could reduce the number http-client crates we use down by one). 

Gotcha! I will be using it with FVM PR, I will let you know any issues I found.

PR #3585, opened 2023-10-05
fix: change `surf` client to address security alert

matheus-consoli:
bot [alerted](https://github.com/infinyon/fluvio/security/dependabot/48) about a possible CPU denial of service in versions previous to `0.22.1` of the `webpki` crate.

This crate was in the dependency tree of `http-client` and `surf` when using the `h1-client-tls` feature.

I removed the `http-client` dependency, as we're not using it, and changed the `surf` client to `hyper-client` instead of `h1-client-tls` (the other option was to use `curl-client`).

Closes https://github.com/infinyon/roadmap/issues/182


digikata:
I don't mind the hyper-client per se, in the past there might have been some async compatibility issues that cropped up at runtime so we need to recheck that by doing a fluvio install to a k8 cluster.  The CI do that, but it probably worth doing locally.

sehz:
BTW, Hyper can work with std.  see: https://github.com/infinyon/k8-api/blob/master/src/k8-client/src/client/mod.rs

matheus-consoli:
Summarizing, the security alert is related to two crates we're using, `http-client` and `surf` (when using `h1-client-rustls`), both deps introduced in [#3581](https://github.com/infinyon/fluvio/pull/3581), which is related to `fvm`.
`surf` has 5 options to choose from for HTTP clients
- `h1-client-rustls`: depends on the vulnerable crate (`webpki`)
- `h1-client-no-tls`: doesn't depend on `webpki`, but doesn't support TLS
- `h1-client`: doesn't depend on `webpki`, but relies on native SSL, which brings some linking problems between musl and openssl
- `hyper-client`: no `webpki`, but depends on openssl
- `curl-client`: no `webpki`, also relies on openssl

Even more alarming, `surf` [notably](https://github.com/http-rs/surf/issues/352#issuecomment-1709404352) is unmaintained.

The current PR is using `http-client-no-tls`, but it's not acceptable, I say we should try to move out from `surf` and use a different HTTP client - `async-std` doesn't have many options here, so for the short-term using a blocking client + `task::spawn_blocking` may be viable (hypothesis: `fvm` is not heavily async)

sehz:
Please see my comment about `hyper,` which works with std using a custom executor.   It should also work with rustls.  We had pending ticket about moving to rustls as default instead of openssl

digikata:
Work will continue for this from a crate hosted at https://github.com/infinyon/fluvio-mini-http

digikata:
Good work
This preliminary work was incorporated with additional integration here https://github.com/infinyon/fluvio/pull/3761

PR #3591, opened 2023-10-09
Migrate from surf to reqwest

al-yisun:
Eliminating an undermaintained dependency on deadpool. After this PR we have

```shell
> cargo tree | rg deadpool
```

Fixes https://github.com/infinyon/fluvio/issues/3070


al-yisun:
There's a test failure in `issue_2168_regression_test`, but this seems to be broken off my branch as well.

Cargo.toml
```diff
@@ -119,14 +118,14 @@ proc-macro2 = "1.0"
 quote = "1.0"
 rand = "0.8.5"
 regex = "1.7"
+reqwest = { version = "0.11.22", features = ["rustls-tls"] }
```
al-yisun, line 121: I opted to keep `rustls` as an option, although I'm not 100% sure as to why.

crates/fluvio-hub-util/src/hubaccess.rs
```diff
@@ -64,26 +65,28 @@ impl HubAccess {
         let host = &self.remote;
         let api_url = format!("{host}/{HUB_API_HUBID}");
         debug!("Sending Action token {action_token}");
-        let req = surf::put(api_url)
-            .content_type(mime::JSON)
-            .body_bytes(msg_json)
-            .header("Authorization", &action_token);
+        let req = reqwest::Client::new()
+            .put(api_url)
+            .header(CONTENT_TYPE, "application/json")
```
al-yisun, line 70: I considered adding something like https://docs.rs/mime/latest/mime/ to provide a constant for `"application/json"`, I think that would be reasonable but in this case I limited myself to `reqwest`.

crates/fluvio-hub-util/src/utils.rs
```diff
@@ -4,13 +4,13 @@ use std::io::copy;
...
-use surf::http::mime;
-use surf::StatusCode;
 use tracing::{debug, info};
 
 use fluvio_hub_protocol::{PackageMeta, Result, HubError};
 use fluvio_hub_protocol::constants::HUB_PACKAGE_EXT;
-
+use reqwest;
+use reqwest::header::{AUTHORIZATION, CONTENT_TYPE};
+use reqwest::StatusCode;
```
al-yisun, line 13: Whoops, these should be moved to the block above. I think `rustfmt.toml` could be edited to add

```
group_imports = "StdExternalCrate"
imports_granularity = "Module"
```

as a more general solution here, but that's probably out of scope (and an extremely noisy change for me to make).

digikata:
Hi @al-yisun, thanks for the PR, we appreciate it!   We will review this PR, but not we also have to do some testing to check on some other interactions with upcoming work we have internally.

Cargo.toml
matheus-consoli: we use `rustls` because it's easier to target `musl` - the other option would be `native-tls`, which is a bit of a pain to set up


This is related to one of the reasons CI is failing - `reqwest`, by default, enables the `default-tls` which tries to use the `native-tls`, so in addition to `features = ["rustls-tls"]` you also need to disable the default feature:

```diff
- reqwest = { version = "0.11.22", features = ["rustls-tls"] }
+ reqwest = { version = "0.11.22", default-features = false, features = ["rustls-tls"] }
```

crates/fluvio-hub-util/src/hubaccess.rs
```diff
@@ -2,11 +2,12 @@ use std::path::{Path, PathBuf};
 
 use serde::{Deserialize, Serialize};
 use sha2::{Digest, Sha512};
-use surf::http::mime;
-use surf::StatusCode;
+use reqwest;
```
matheus-consoli, line 5: this line can be removed :)

crates/fluvio-hub-util/src/hubaccess.rs
matheus-consoli: I think it's fine the way it is :)

crates/fluvio-hub-util/src/hubaccess.rs
```diff
@@ -307,10 +312,9 @@ fn get_hubref() -> Option<String> {
         return None; // use default
     }
     let hubref_url = format!("{fcremote}/api/v1/hubref");
-    let reply: std::result::Result<String, surf::Error> = run_block_on(async {
-        let req = surf::get(hubref_url);
-        let mut res = req.await?;
-        let reply: ReplyHubref = res.body_json().await?;
+    let reply: std::result::Result<String, reqwest::Error> = run_block_on(async {
+        let res = reqwest::get(hubref_url).await?;
+        let reply: ReplyHubref = res.json().await?;
```
matheus-consoli, line 317: to use `.json()` here you need to enable the feature flag in the `Cargo.toml`

`reqwest = { ..., features = [..., "json"] }`

matheus-consoli (changes requested):
Hi, @al-yisun!

We appreciate your interest in our project!

I left some comments we need to address to get things to compile.

Please note that Fluvio uses `async-std` as the async runtime of choice, and `reqwest` doesn't support it. So, to get things properly running, we need to identify the places these changes affect and switch the async runtime to `tokio` accordingly.

By the way, feel welcome to join our Discord server and ask questions/help!
https://discord.gg/neVUKZ3T

crates/fluvio-hub-util/src/hubaccess.rs
al-yisun: Oddly this wasn't erroring locally, but will do.

crates/fluvio-hub-util/src/hubaccess.rs
al-yisun: Missed a clippy 🙃

digikata:
@matheus-consoli @al-yisun I appreciate the coordination here.   Let's adjust the direction a little here and maybe change this work to using the future crate here  https://github.com/infinyon/fluvio-mini-http.    It's empty right now, but @matheus-consoli you can post the draft of the work there and @al-yisun if you're inclined you can take a look at that interface.

crates/fluvio-benchmark/src/benchmark_driver.rs
```diff
@@ -121,7 +122,7 @@ impl BenchmarkDriver {
         // Close all worker tasks.
         send_control_message(&mut tx_controls, ControlMessage::Exit).await?;
         for jh in workers_jh {
-            timeout(config.worker_timeout, jh).await???;
+            timeout(config.worker_timeout, jh).await????;
         }
```
al-yisun, line 126: Side question, why are these for-looped over instead of joined?

crates/fluvio-benchmark/src/lib.rs
```diff
@@ -40,8 +39,8 @@ pub enum BenchmarkError {
     IoError(#[from] IoError),
     #[error("{0}")]
     ErrorWithExplanation(String),
-    #[error(transparent)]
-    TimeoutError(#[from] TimeoutError),
+    //#[error(transparent)]
+    //TimeoutError(#[from] tokio::timer::Error),
```
al-yisun, line 43: (todo)

crates/fluvio-channel-cli/src/bin/main.rs
```diff
@@ -186,7 +185,10 @@ fn main() -> Result<()> {
                     println!();
                     std::process::exit(0);
                 } else if let Some(subcmd) = &channel_opt.cmd {
-                    if let Err(e) = run_block_on(subcmd.process()) {
+                    let res = tokio::runtime::Runtime::new()
+                        .unwrap()
+                        .block_on(async { subcmd.process().await });
+                    if let Err(e) = res {
                         println!("{e}");
                         std::process::exit(1);
```
al-yisun, line 193: It seems like this pattern could just be some kind of `exec`.

examples/01-produce-key-value/Cargo.toml
```diff
@@ -9,7 +9,7 @@ publish = false
 
 [dependencies]
 anyhow = { workspace = true }
-async-std = { workspace = true,  features = ["attributes"] }
+tokio = { workspace = true }
```
digikata, line 12: .

al-yisun: Ah. I can probably close this PR.

PR #3597, opened 2023-10-12
[Merged by Bors] - feat: fvm switch command

EstebanBorai:
Provides support to switch between Fluvio Versions using `fvm switch` command.
The `fvm switch` command performs a side effect which involves updating `settings.toml`
file in `~/.fvm` which is used to keep track of the active channel and version.

Also updates the `~/.fvm/env` file to include `~/.fluvio/bin` in the host's PATH
when running `source ~/.fvm/env`.

## Demo

https://github.com/infinyon/fluvio/assets/34756077/f81960ee-b557-4720-8691-8f4afb79a84f


crates/fluvio-version-manager/src/command/itself/install.rs
```diff
@@ -17,7 +17,7 @@ case ":${PATH}:" in
     *:"$HOME/.fvm/bin":*)
         ;;
     *)
-        export PATH="$PATH:$HOME/.fvm/bin"
+        export PATH="$PATH:$HOME/.fvm/bin:$HOME/.fluvio/bin"
```
EstebanBorai, line 20: This allows the user to have both:

- `~/.fvm/bin`
- `~/.fluvio/bin`

In their PATH after executing `source ~/.fvm/env`

crates/fluvio-version-manager/src/command/switch.rs
```diff
@@ -0,0 +1,221 @@
...
+            }
+        }
+
+        if dir.manifest == PathBuf::new() {
+            return Err(anyhow::anyhow!(
+                "Failed to find manifest file in version directory"
+            ));
+        }
+
+        Ok(dir)
```
EstebanBorai, line 67: I was considering using `Option<PathBuf>` in the `VersionManifest` struct but it wasnt the actual purpose because I expect `manifest.json` to be present always.

crates/fluvio-version-manager/src/command/switch.rs
```diff
@@ -0,0 +1,221 @@
...
+    }
+}
+
+#[derive(Debug, Parser)]
+pub struct SwitchOpt {
+    #[command(flatten)]
+    global_opts: GlobalOptions,
+    /// Version to install
+    #[arg(index = 1, default_value_t = Channel::Stable)]
+    version: Channel,
```
EstebanBorai, line 93: If the user runs `fvm switch` with no params, it will switch to Stable by default. WDYT about this behavior?

digikata: I think `fvm switch` with no params should be an error.  Maybe it could print an error string and recommend running `fvm show` to see versions available to switch.

EstebanBorai: Agree!

sehz: Just remove default

EstebanBorai: I added the following behavior here: [3272fd5](https://github.com/infinyon/fluvio/pull/3597/commits/3272fd5aff1a751d16b21f0ed7203f681d3102bd)

- Shows a help message to recommend using the `fvm show` command
- Exits with code `1` and the message: `Error: No version provided`

Screenshot 2023-10-12 at 5 51 31 PM

crates/fluvio-version-manager/src/command/switch.rs
```diff
@@ -0,0 +1,230 @@
...
+        entries.copy_contents(&target_dir).unwrap();
+
+        let target_entries = read_dir(target_dir.path()).unwrap();
+        let target_entries: Vec<PathBuf> =
+            target_entries.map(|entry| entry.unwrap().path()).collect();
+
+        assert_eq!(target_entries.len(), 1);
+        assert_eq!(
+            target_entries.get(0).unwrap(),
+            &target_dir.path().join(random_file.file_name().unwrap())
```
sehz, line 227: Need to confirm copies are done by checking file contents and they are same

crates/fluvio-version-manager/src/command/switch.rs
```diff
@@ -0,0 +1,230 @@
...
+    #[command(flatten)]
+    global_opts: GlobalOptions,
+    /// Version to set as active
+    #[arg(index = 1)]
+    version: Option<Channel>,
+}
+
+impl SwitchOpt {
+    pub async fn process(&self) -> Result<()> {
+        let Some(version) = self.version.clone() else {
```
sehz, line 98: minor nit.  why cloning is necessary?

crates/fluvio-version-manager/src/command/switch.rs
```diff
@@ -0,0 +1,230 @@
...
+            self.notify_help(format!(
+                "You can use {} to see installed versions",
+                "fvm show".bold()
+            ));
+
+            return Err(anyhow::anyhow!("No version provided"));
+        };
+
+        let versions_path = fvm_versions_path()?;
```
sehz, line 107: Hmm. what does this code do? Shouldn't this find path for new version?

crates/fluvio-version-manager/src/command/switch.rs
```diff
@@ -0,0 +1,230 @@
...
+
+        // If the user runs this command before installing any versions, we
+        // should notify them and exit early.
+        if !versions_path.exists() {
```
sehz, line 111: ditto

crates/fluvio-version-manager/src/command/switch.rs
```diff
@@ -0,0 +1,230 @@
...
+            ))? == PACKAGE_SET_MANIFEST_FILENAME
+            {
+                dir.manifest = entry_path;
+            } else {
+                dir.contents.push(entry_path);
+            }
+        }
+
+        if dir.manifest == PathBuf::new() {
+            return Err(anyhow::anyhow!(
```
sehz, line 62: shouldn't this also check if manifest is correct?

crates/fluvio-version-manager/src/common/settings.rs
```diff
@@ -139,4 +163,26 @@ mod tests {
...
+            "the settings file should not exist at this point"
+        );
+
+        Settings::open().expect("Failed to create settings.toml file");
+
+        assert!(
+            settings_path.exists(),
+            "the settings file should exist at this point"
+        );
+
```
digikata, line 185: Should add a readback and asserts after the create

crates/fluvio-version-manager/src/command/switch.rs
```diff
@@ -0,0 +1,230 @@
...
+
+        // Opens the version directory and then copies the contents (but the manifest)
+        // to the Fluvio Binaries directory
+        let version_dir = VersionDirectory::open(pkgset_path)?;
+        version_dir.copy_contents(&fluvio_bin_dir)?;
+
+        // Update settings file to keep track of active Fluvio Version
+        let manifest = VersionManifest::open(version_dir.manifest)?;
+        let mut settings = Settings::open()?;
+        settings.update_version(&manifest.channel, &manifest.version)?;
```
sehz, line 156: Would prefer to move this code out of Cmd and into somewhere (maybe `VersionDirectory`) so it can be unit-tested indepently

crates/fluvio-version-manager/src/common/manifest.rs
```diff
@@ -55,4 +62,33 @@ mod test {
...
+    fn reads_manifest_from_json() {
+        let tempdir = TempDir::new().unwrap();
+        let version_manifest = VersionManifest::new(Channel::Stable, Version::new(0, 8, 0));
+        let json = serde_json::to_string_pretty(&version_manifest).unwrap();
+        let manifest = version_manifest
+            .write(tempdir.path().to_path_buf())
+            .unwrap();
+        let read_manifest = VersionManifest::open(manifest).unwrap();
+
+        assert_eq!(json, serde_json::to_string_pretty(&read_manifest).unwrap());
```
sehz, line 76: Prefer explicit literal.  Write down literal JSON as comparison

crates/fluvio-version-manager/src/common/manifest.rs
```diff
@@ -55,4 +62,33 @@ mod test {
...
+    }
+
+    #[test]
+    fn fails_to_read_manifest_from_invalid_json() {
+        let tempdir = TempDir::new().unwrap();
+        let manifest = tempdir.path().join(PACKAGE_SET_MANIFEST_FILENAME);
+
+        std::fs::write(manifest.clone(), "invalid json").unwrap();
```
sehz, line 84: Add invalid in the test file instead of TempDir.  That way it's easier to test and see

tests/cli/fvm_smoke_tests/fvm_basic.bats
```diff
@@ -184,3 +192,171 @@ setup_file() {
...
+    # Ensure `~/.fluvio` is not present
+    run bash -c '! test -d $FLUVIO_HOME_DIR'
+    assert_success
+
+    # Installs Stable Fluvio
+    run bash -c 'fvm install stable'
+    assert_success
+
+    # Switch version to use Stable Fluvio
+    run bash -c 'fvm switch stable'
```
sehz, line 215: Duplicate user's flow.  Install `stable` and `0.10.15` or some version then switch

tests/cli/fvm_smoke_tests/fvm_basic.bats
```diff
@@ -184,3 +192,171 @@ setup_file() {
...
+        fluvio
+        fluvio-run
+        fluvio-cloud
+        cdk
+        smdk
+    )
+
+    for binary in "${binaries[@]}"
+    do
+        export FVM_BIN_PATH="$VERSIONS_DIR/stable/$binary"
```
sehz, line 229: why check stable? it is source

tests/cli/fvm_smoke_tests/fvm_basic.bats
```diff
@@ -184,3 +192,171 @@ setup_file() {
...
+        echo "Checking binary: $BINARY_PATH"
+        run bash -c 'test -f $FVM_BIN_PATH'
+        assert_success
+
+        export FLV_BIN_PATH="$FLUVIO_BINARIES_DIR/$binary"
+        echo "Checking binary: $FLV_BIN_PATH"
+        run bash -c 'test -f $FLV_BIN_PATH'
```
sehz, line 236: check contents are same as stable version

tests/cli/fvm_smoke_tests/fvm_basic.bats
```diff
@@ -184,3 +192,171 @@ setup_file() {
...
+    # Sets `fvm` in the PATH using the "env" file included in the installation
+    source ~/.fvm/env
+
+    # Ensure `~/.fluvio` is not present
+    run bash -c '! test -d $FLUVIO_HOME_DIR'
+    assert_success
+
+    # Installs Fluvio at 0.10.15
+    run bash -c 'fvm install 0.10.15'
+    assert_success
```
sehz, line 262: Check if FVM installed 0.10.15.
Ensure setting is correct

tests/cli/fvm_smoke_tests/fvm_basic.bats
```diff
@@ -184,3 +192,171 @@ setup_file() {
...
+
+    # Installs Fluvio at 0.10.14
+    run bash -c 'fvm install 0.10.14'
+    assert_success
```
sehz, line 266: ditto

tests/cli/fvm_smoke_tests/fvm_basic.bats
```diff
@@ -184,3 +192,171 @@ setup_file() {
...
+
+    # Switch version to use Fluvio at 0.10.15
+    run bash -c 'fvm switch 0.10.15'
```
sehz, line 269: Make sure before this, that fluvio is running 0.10.14 by checking version when running

tests/cli/fvm_smoke_tests/fvm_basic.bats
```diff
@@ -184,3 +192,171 @@ setup_file() {
...
+    assert_success
+
+    # Checks version is set
+    run bash -c 'fluvio version > active_flv_ver.out && cat active_flv_ver.out | head -n 1 | grep "0.10.15"'
+    assert_output --partial "0.10.15"
+    assert_success
```
sehz, line 275: check fluvio version

sehz (changes requested):

crates/fluvio-version-manager/src/command/switch.rs
EstebanBorai: Because `Version' from `semver' is not `Copy`. I havent tried borrowing it, I can check how it goes!

crates/fluvio-version-manager/src/command/switch.rs
EstebanBorai: This function builds an absolute path to the `~/.fvm/versions` directory.

crates/fluvio-version-manager/src/command/switch.rs
EstebanBorai: Here I check if the `~/.fvm/versions` directory exists. This directory is where `fvm install` stores binaries pulled from the registry.

crates/fluvio-version-manager/src/command/switch.rs
EstebanBorai: I could check it as part of creating an instance of `VersionDirectory`. That would be a quick refactor.

crates/fluvio-version-manager/src/command/switch.rs
EstebanBorai: I think I could generate SHA256 checksums from each file and compare as part of the tests!

crates/fluvio-version-manager/src/common/settings.rs
EstebanBorai: Sure thing! I should mention that `Settings` will be an empty file after creation, due to the fact that no version nor channel is set, just in case both assertions on `is_none` looks weird!

EstebanBorai:
I just addressed suggestions here: https://github.com/infinyon/fluvio/pull/3597/commits/5d03acc1323f05ff638f00bd2eed26478c6ad0e9!

crates/fluvio-version-manager/src/common/manifest.rs
sehz: nice!

crates/fluvio-version-manager/src/common/settings.rs
```diff
@@ -118,7 +142,28 @@ mod tests {
...
+        settings.channel = Some(Channel::Stable);
+        settings.version = Some(Version::new(0, 11, 0));
+
+        settings.save().expect("Failed to save settings.toml file");
+
+        let settings_str =
+            read_to_string(settings_path).expect("Failed to read settings.toml file");
+
+        assert!(settings_str.contains("channel = \"stable\""));
```
sehz, line 160: do exact comparison

crates/fluvio-version-manager/src/common/settings.rs
```diff
@@ -136,7 +181,42 @@ mod tests {
...
 
-        remove_file(&settings_path).expect("Failed to remove settings.toml file");
+        let chan = Channel::Latest;
+        let vers = Version::new(0, 12, 0);
+        settings.update_version(&chan, &vers).unwrap();
+
+        let settings_str =
+            read_to_string(&settings_path).expect("Failed to read settings.toml file");
+
+        assert!(settings_str.contains("channel = \"latest\""));
```
sehz, line 191: Ditto.  perform exact comparison

sehz (changes requested):

crates/fluvio-version-manager/src/command/switch/version_directory.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    #[test]
+    fn copies_contents_to_target_directory() {
+        let source_dir = tempfile::tempdir().unwrap();
+        let target_dir = tempfile::tempdir().unwrap();
+        let random_file = source_dir.path().join("random_file");
+
+        // The manifest file is not copied but must be present for `VersionDirectory::open` to succeed
+        create_dummy_manifest(source_dir.path()).unwrap();
+        File::create(&random_file).unwrap();
```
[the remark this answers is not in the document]
EstebanBorai: Nice suggestion! I just added a copy of the PackageSet for version `0.10.14` and I use them to create tempdirs for testing. This way we test against a real set of files.

crates/fluvio-version-manager/src/command/switch/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+            ));
+
+            return Ok(());
+        }
+
+        let version_dir = VersionDirectory::open(pkgset_path)?;
+
+        version_dir.set_active()?;
+
+        Settings::open()?.update_from_manifest(&version_dir.manifest)?;
```
sehz, line 79: Why manifest files should be changed? This should be removed

crates/fluvio-version-manager/src/command/switch/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+use crate::common::settings::Settings;
+use crate::common::workdir::fvm_versions_path;
+use crate::common::notify::Notify;
+
+use self::version_directory::VersionDirectory;
+
+#[derive(Debug, Parser)]
+pub struct SwitchOpt {
+    #[command(flatten)]
+    global_opts: GlobalOptions,
```
sehz, line 23: Why global option is here? This should be moved to root command

crates/fluvio-version-manager/src/command/switch/version_directory.rs
```diff
@@ -0,0 +1,224 @@
...
+            remove_dir_all(&fluvio_bin_dir).unwrap();
+        }
+
+        create_dir_all(&fluvio_bin_dir).unwrap();
+
+        version_dir.set_active().unwrap();
+
+        // Verify every binary in the Version Directory is now in the `~/.fluvio/bin`
+        // directory
+        for (idx, name) in EXPECTED_BINARIES.iter().enumerate() {
```
sehz, line 207: I would prefer this to be explicit like this:
```
assert!(tempdir_join("cdk1").exists");
assert!(sa256_digest(tempdir_join("cdk"1)), "21be0d1f12fbf7c0fab86760b89e294f3a6b05ba0807c168bb2f493dd6fb7dfb");
```

crates/fluvio-version-manager/src/command/switch/version_directory.rs
```diff
@@ -0,0 +1,224 @@
...
+
+    use tempfile::TempDir;
+
+    use fluvio_hub_util::sha256_digest;
+
+    use super::*;
+
+    /// Binaries included in a Fluvio Version. These are mapped to the ones in
+    /// the `fixtures/version` directory
+    const EXPECTED_BINARIES: [&str; 5] = ["cdk", "fluvio", "fluvio-cloud", "fluvio-run", "smdk"];
```
sehz, line 97: This is a pretty file to check into git.  Instead, create a tiny executable file (a bash file would also work). As long as they have executable and checksum can be created

sehz (changes requested):
Much better.  A just couple of issues on testing active

crates/fluvio-version-manager/src/command/switch/mod.rs
EstebanBorai: Manifest file will not change. What this does is updating the `~/.fvm/settings.toml` file to point to the active Fluvio Version.

Here Im using the `manifest` from the Version Directory as a source to update `settings.toml`.

crates/fluvio-version-manager/src/command/switch/mod.rs
EstebanBorai: This is used by the `Notify` trait to access `is_quiet` global option, which supress output.

crates/fluvio-version-manager/src/command/switch/mod.rs
sehz: then VersionManager should update file since it is responsible for updating.

crates/fluvio-version-manager/src/command/switch/mod.rs
sehz: I assume this is needed to be repeated for every command.  either remove or move to root.  for example this could be passed from root cmd

crates/fluvio-version-manager/src/command/switch/version_directory.rs
```diff
@@ -0,0 +1,217 @@
...
+        assert!(
+            version_dir.contents.contains(&tempdir_path),
+            "version directory must include binary path as content"
+        );
+        assert_eq!(
+            TEST_BINARY_CHECKSUM,
+            sha256_digest(&tempdir_path).unwrap(),
+            "binary contents must be the same"
+        );
+
```
sehz, line 214: Need to verify setting file change

EstebanBorai: Gotcha! Thanks!

sehz (approved):
LGTM.  Thanks for nice sets of unit testing and CLI testing.
Copying based switch need more testing and fundamentally only works for very small binaries.  but this is good enough for now and resolve that question later

EstebanBorai:
> LGTM. Thanks for nice sets of unit testing and CLI testing.

🙌🏻

> Copying based switch need more testing and fundamentally only works for very small binaries. but this is good enough for now and resolve that question later

Looking forward into it!

EstebanBorai:
Errors doesn't look related to PR: https://github.com/infinyon/fluvio/actions/runs/6520022049/job/17707174387#step:5:1107

So I will rerun!

Issue #3537, opened 2023-09-14
Fluvio Version Manager


EstebanBorai:
Introduce Fluvio Version Manager (FVM) a CLI Tool used to switch between Fluvio Versions.
For our first release on the Fluvio Version Manager users will be able to download
Fluvio versions specifying the specific tag or update their Fluvio installation
directly from the FVM.

## Tasks (Integration)

The FVM integration is achieved command by command in the following order:

- [x] #3569 
- [x] #3573
- [x] #3596
- [x] #3644
- [x] Utilities and non-core features
  - [x] #3599
  - [x] #3600

### Improvements/Feedback

- [x] #3602
- [x] #3606
- [x] https://github.com/infinyon/fluvio/pull/3616
- [x] #3622
- [x] #3623
- [x] https://github.com/infinyon/fluvio/pull/3649
- [x] https://github.com/infinyon/fluvio/issues/3650

## Other Tasks

- [x] Update docs
  - [x] fluvio.io (docs)
  - [x] infinyon.com (docs & blogs)
- [ ] Remove `channels` and other operations covered by FVM from fluvio
- [ ] Add custom image for `Iot mirroring` & [update guide](https://infinyon.com/docs/guides/iot-mirroring-local/).
- [x] Ability to install stateful sets fluvio preview w/ added ssdk binary https://github.com/infinyon/fluvio-experiments/issues/187 
 
 ## Reference
[Prod RFD](https://github.com/infinyon/roadmap/blob/main/prod-rfd/fluvio-version-manager.md)

digikata:
Engineering detail https://github.com/infinyon/roadmap/issues/161

digikata:
New item: remove misc old version change and install commands
- fluvio install
- fluvio version switch

digikata:
consider printing to std err deprecation if being kept this iteration

PR #3601, opened 2023-10-16
[Merged by Bors] - feat: `fvm show` and `fvm current` command

EstebanBorai:
Introduces two utility commands for FVM used to check on the active version and
listing installed versions.

Use `fvm current` to print the active Fluvio Version and use `fvm show` to list
installed Fluvio versions.

## Demo

https://github.com/infinyon/fluvio/assets/34756077/a2327a63-1d9e-4ceb-86d3-02522de0aa42


crates/fluvio-hub-util/src/fvm/mod.rs
```diff
@@ -149,7 +149,7 @@ mod tests {
     #[test]
     fn performs_comparisons_between_tags() {
         let ver_a = Channel::parse("0.10.10").unwrap();
-        let ver_b = Channel::parse("0.10.13").unwrap();
+        let ver_b = Channel::parse("0.10.13-mirroring347239873+20231016").unwrap();
```
EstebanBorai, line 152: Update test to validate tags support.

crates/fluvio-version-manager/fixtures/version/0.10.15/manifest.json
```diff
@@ -0,0 +1,6 @@
+{
+  "channel": {
+    "tag": "0.10.15"
```
sehz, line 3: what's `tag` for?

crates/fluvio-version-manager/fixtures/version/stable/manifest.json
```diff
@@ -0,0 +1,4 @@
+{
+  "channel": "stable",
```
sehz, line 2: why channel is this needed here?

sehz (changes requested):
Overall look reasonable.  Couple of questions on why additional info for manifest is necessary.

ajhunyady:
@EstebanBorai after install, `fvm` should automatically move to the version installed. This should be true for the first version installed (stable). Not sure about subsequent version @sehz ?

sehz:
@ajhunyady not sure I understood question but `fvm` should install stable version and make it as default

crates/fluvio-version-manager/fixtures/version/stable/manifest.json
EstebanBorai: This help us understand which channel this package points to.
The `manifest.json` is metadata generated when running `fvm install`, so if one runs:
`fvm install stable`, FVM will fetch stable packages and write metadata to `manifest.json` that looks like this.

crates/fluvio-version-manager/src/command/show.rs
```diff
@@ -0,0 +1,171 @@
...
+        Ok(())
+    }
+
+    /// Retrieves the sorted list of installed versions [`VersionManifest`]
+    /// instances. In parallel, it also retrieves the active version if any.
+    ///
+    /// If theres any Active Version, the [`VersionManifest`] is not included
+    /// as part of the first tuple value, instead it is returned as the second
+    /// tuple value.
+    fn scan_versions_manifests(
```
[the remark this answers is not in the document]
EstebanBorai: I also thought about it but I as well, but I found the use case strongly tied to the purpose of `fvm show`.

Given that this function not only takes the manifest from `VersionDirectory` in the process of walking a directory, but it also puts apart the active version, so `fvm show` can render the `check` symbol later on.

I could move tho `scan` logic to be part of `VersionDirectory` and return the vector filled with all version directories, but then I would have to loop into this vector a second time to retrieve the active version and split it from the rest to provide it to `fn render_table(manifests: Vec<VersionManifest>, maybe_active: Option<VersionManifest>)`.

EstebanBorai:
Thanks for asking @ajhunyady and @sehz!

In my video Im showing commands with the build of FVM I had locally. In production users will have FVM installed via our cURL `install.sh` script which will also install Fluvio and set it up as active by default.

In practice, users running:

```bash
curl -fsS https://hub.infinyon.cloud/install/install.sh | bash
```

Will:

- Have `fvm` installed
- Have `fluvio` stable installed and set as active (Equivalent to `fvm install && fvm switch stable`)

With regards to setting as default after install, initially we designed a "no side effects" implementation which makes every command atomic.

But I think we could introduce `fvm install --switch/-s <VERSION>`? Which installs and sets that version as active? I could work a separate iteration for that as it needs some work to make `fvm switch` command logic reusable by `fvm install`.

ajhunyady:
install should perform switch under the hood. You can do `--no-switch` if required, but I don't see a use case for that. 

This feedback is on the usability not internal design decisions. Though usability comes first, and the appropriate design should follow.

crates/fluvio-version-manager/fixtures/version/0.10.15/manifest.json
sehz: Hmm.  But not show stopper. so let's come back to this later

crates/fluvio-version-manager/fixtures/version/stable/manifest.json
sehz: k

crates/fluvio-version-manager/src/command/show.rs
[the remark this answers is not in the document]
sehz: Issue is that this now has knowledge of internals of the versions, manifest.   This needs to move there

sehz (changes requested):

sehz:
```But I think we could introduce fvm install --switch/-s <VERSION>? Which installs and sets that version as active? I could work a separate iteration for that as it needs some work to make fvm switch command logic reusable by fvm install.```

Yes, that can be done in separate PR

crates/fluvio-version-manager/src/command/show.rs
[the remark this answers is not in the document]
EstebanBorai: Got it! Migrating to `VersionDirectory`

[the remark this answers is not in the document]
EstebanBorai: Here it is! [8c1db1c](https://github.com/infinyon/fluvio/pull/3601/commits/8c1db1cafb1a494aca8624eafca0d1fa3c31acd4)

sehz (approved):
K.   LGTM.   Please update install logic to switch to version in next PR

ajhunyady:
@EstebanBorai please file a new issue and add to tracking issue https://github.com/infinyon/fluvio/issues/3537

Issue #2271, opened 2022-03-25
Topic should show timestamp of last produced record


nacardin:

nacardin:
Approach could be to consume latest record and use it's timestamp

digikata:
Might be nice to do this with a "--verbose" flag

ajhunyady:
> Might be nice to do this with a "--verbose" flag

maybe: `timestamp`, `offset`, `key` and `value`.

BTW, this is all possible with `--format` option.

solidiquis:
@digikata I just submitted a draft PR that queries the latest timestamp and includes it in the output. I haven't yet included the `--verbose` flag yet I wanted to inquire further about what @ajhunyady just mentioned.

@ajhunyady, could you expand a bit on what you mentioned? Are you suggesting that instead of `--verbose`, we add support for `--timestamp`, `--offset`, `--key`, and `--value` for optional granular information about the latest record of a given topic?

Also could you tell me more about the `--format` option? Thank you!

ajhunyady:
@solidiquis, sorry for the confusion, and ignore my comment. I thought we were discussing the consumer.
The solution should be good enough.

BTW, we have a long standing bug in changing `STATUS` to a readable format:
`resolution::provisioned` should be `provisioned`
 ... same for the other states.

It would be great if we could fix this as well.

Also instead of `MODIFIED AT` -> `LAST MODIFIED`

solidiquis:
Awesome thanks for the input @ajhunyady! In that case I'll include `LAST MODIFIED` in the default output and carry on without adding `--verbose`. I'll also address the `STATUS` bug you just mentioned as well shortly.

PR #3609, opened 2023-10-19
Enhance `topic list` output

solidiquis:
Closes https://github.com/infinyon/fluvio/issues/2271

This PR enhances `$ fluvio topic list` by adding an additional column to the output table called `LAST MODIFIED` which surfaces information about the creation time of the most recent record for a given topic across all partitions. Additionally, `STATUS` was fixed to be more human readable: `resolution::provisioned` -> `provisioned`, etc..

**Note**: ~Right now the timestamp is included by default in the output but that will not be the case once this branch is ready to be merged. The mechanism whereby we optionally include `MODIFIED AT` in the output is currently being discussed in the aforementioned issue.~

**Example**:
Screen Shot 2023-10-19 at 9 19 53 AM

**Questions for reviewers**:
1. How do you folks prefer to log/handle errors in the CLI? Right now I am converting `Result`s to `Option`s when trying to query the latest record to retrieve the timestamp.
2. Do you folks have a preferred timestamp format? Right now It's using `rfc3339`.
3. ~How we feel about `MODIFIED AT` as the header name?~

**To-do before merge**:
- [ ] Assess correctness with reviewers.
- [ ] Ask team about best logging practices for the CLI.
- [x] Discuss with team preferred header name. It is currently `MODIFIED AT`.
- [ ] Discuss preferred timestamp format. Do we like `rfc3339`?
- [x] Hash out how to handle `--verbose`.
- [ ] Squash and rebase.

### DIFF
--- a/crates/fluvio-cli/src/client/topic/list.rs
+++ b/crates/fluvio-cli/src/client/topic/list.rs
@@ -32,41 +32,52 @@ impl ListTopicsOpt {
         let output_type = self.output.format;
         debug!("list topics {:#?} ", output_type);
         let admin = fluvio.admin().await;
-
         let topics = admin.all::<TopicSpec>().await?;
-        display::format_response_output(out, topics, output_type)?;
+        display::format_response_output(out, topics, output_type, fluvio)?;
         Ok(())
     }
 }
 
 mod display {
+    use std::time::{Duration, UNIX_EPOCH};
 
-    use std::time::Duration;
-
-    use humantime::{format_duration};
     use comfy_table::{Row, Cell, CellAlignment};
+    use humantime::{format_duration, format_rfc3339_millis};
     use serde::Serialize;
 
-    use fluvio::metadata::objects::Metadata;
-    use fluvio::metadata::topic::TopicSpec;
+    use fluvio::{
+        Fluvio,
+        metadata::{objects::Metadata, topic::TopicSpec},
+    };
+    use fluvio_future::task::run_block_on;
 
     use crate::common::output::{OutputType, TableOutputHandler, Terminal, OutputError};
     use crate::common::t_println;
+    use super::utils::latest_record_timestamp;
 
     #[derive(Serialize)]
-    struct ListTopics(Vec<Metadata<TopicSpec>>);
+    struct ListTopics<'a> {
+        inner: Vec<Metadata<TopicSpec>>,
+
+        #[serde(skip_serializing)]
+        fluvio: &'a Fluvio,
+    }
 
     /// Process server based on output type
     pub fn format_response_output<O>(
         out: std::sync::Arc<O>,
         list_topics: Vec<Metadata<TopicSpec>>,
         output_type: OutputType,
+        fluvio: &Fluvio,
     ) -> Result<(), OutputError>
     where
         O: Terminal,
     {
         if !list_topics.is_empty() {
-            let table_list = ListTopics(list_topics);
+            let table_list = ListTopics {
+                inner: list_topics,
+                fluvio,
+            };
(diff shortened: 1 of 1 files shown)


crates/fluvio-cli/src/client/topic/list.rs
```diff
@@ -120,11 +146,45 @@ mod display {
...
+        Fluvio, Offset,
+    };
+    use fluvio_types::Timestamp;
+    use futures_util::stream::StreamExt;
+
+    /// A utility function to get the timestamp of the `Offset::from_end(1)` record across all
+    /// partitions for a given `topic_name`.
+    pub async fn latest_record_timestamp(topic_name: String, fluvio: &Fluvio) -> Option<Timestamp> {
+        let consumer = fluvio
+            .consumer(PartitionSelectionStrategy::All(topic_name))
```
sehz, line 171: This only works for a single partition.  In order to get latest time, need to query all partitions and select latest

sehz (changes requested):
Thanks for contribution. 

Please see comments about getting information about partition.
Bigger issue is that it combines admin call and consumer which creates different functionality from API client.  

So order of preferences:
(1) add this metadata in the SC by extending partition status.  This is ideal since no client API changes are required (except adding new field)
(2) implement this in the Fluvio admin API.  Not ideal but at least hides implementation.  The additional metadata will be only exposed if additional config is set since normally, admin API should not call consumer API

solidiquis:
@sehz thanks a ton for the input! I'll implement the feedback shortly after lunch!

crates/fluvio-cli/src/client/topic/list.rs
solidiquis: Hi @sehz, could you clarify this point? [The documentation](https://docs.rs/fluvio/latest/src/fluvio/fluvio.rs.html#220) seems to suggest that this is indeed the way to query all records for a given topic across multiple partitions.

solidiquis: [This doc comment](https://docs.rs/fluvio/latest/src/fluvio/consumer.rs.html#591) also corroborates my aforementioned point.

sehz: But you are only a getting a single record.  You need to get all latest records from all partitions

solidiquis: Ahh I see. I just saw the footnote about the lack of ordering guarantees. Thank you!

sehz: Yes, ordering is guarantee only per partition.  

So if p0 has t1, p2 has t2 and p3 has t3.  you might have any random event with t

solidiquis: Hey @sehz, just wanted to make sure I'm on the right track to go about implementing your first suggestion.

As I understand it, we want to extend the [PartitionStatus](https://docs.rs/fluvio/latest/fluvio/metadata/partition/struct.PartitionStatus.html) behavior to allow us to get information about that particular partition's most recent record, is that correct? If that's the case would it be appropriate to give [PartitionStatus](https://docs.rs/fluvio/latest/fluvio/metadata/partition/struct.PartitionStatus.html) access to the [Fluvio](https://docs.rs/fluvio/latest/fluvio/struct.Fluvio.html#) streaming interface by way of something akin to:

```rust
impl PartitionStatus {
  ...
  pub async fn latest_record_timestamp(fluvio: &Fluvio) -> Option<Timestamp> { ... }
  ...
}
```

Also, are there more preferred alternatives to querying records via the consumer API?

Issue #3578, opened 2023-10-04
Move `fluvio hub connector download` to `ckd hub download`


ajhunyady:
### Context

Connectors used in the cloud are automatically download via `fluvio  cloud connector create`, so there is no need to download them first. Connectors used locally require `cdk` (connector developer kit), and should not require `fluvio` to download.

### Issue

Connectors only need to be downloaded when running locally via `cdk`, and they should be listed and downloaded via `cdk`.

### Change request - cdk

Implement the following commands in `cdk`:

1. List connectors in the Hub:

```
$ cdk hub list
  CONNECTOR                          Visibility 
  infinyon-labs/graphite-sink@0.1.2  public     
  infinyon/duckdb-sink@0.1.0         public     
  infinyon/http-sink@0.2.5           public     
  infinyon/http-source@0.2.5         public     
  infinyon/kafka-sink@0.2.7          public     
  infinyon/kafka-source@0.2.5        public     
  infinyon/mqtt-source@0.2.5         public     
  infinyon/sql-sink@0.3.3            public  
```

2. Download connectors from the Hub:

```
$ cdk hub download infinyon/http-sink@0.2.5 
downloading infinyon/http-sink@0.2.5 to infinyon-http-sink-0.2.5.ipkg
... downloading complete
```

3. Run connector (this command already exists)... adding for illustration purpose:

```
$ cdk deploy start -c config.yaml --ipkg infinyon-http-sink-0.2.5.ipkg 
... checking package
Log file: ~/http-sink.log
Connector runs with process id: 54384
Started connector `httpbin`
```

### Change Request - fluvio

Change existing command `fluvio hub connector download` to return a message:

```
$ fluvio hub connector download 
Connectors running on `InfinyOn Cloud` are automatically downloaded during `fluvio cloud connector create ...`. 
Connectors running locally require `cdk` to download and deploy:
1. Install cdk: `fluvio install cdk`
2. Download connector: `cdk hub download ...`
3. Deploy connector: `cdk deploy start ...`
```

PR #3611, opened 2023-10-20
[Merged by Bors] - chore: move hub connector commands to fluvio-hub-util

matheus-consoli:
This is the first half of a two-part feature related to #3578.

In this first part, we only move common code from `fluvio-cli` to `fluvio-hub-util` behind a feature flag. The second PR will reuse the command definitions in `cdk`.


digikata (approved):
LGTM, but we should hold off on merging w/ bors until the tests can pass on the other pr.  We're close to kicking off a fluvio release and don't want to be caught between these two PRs.

PR #3613, opened 2023-10-20
[Merged by Bors] - chore: move file iterators to fluvio_storage crate

morenol:


crates/fluvio-storage/Cargo.toml
```diff
@@ -56,6 +58,7 @@ fluvio-future = { workspace = true, features = ["fixture"] }
 flv-util = { workspace = true, features = ["fixture"] }
 fluvio-socket = { workspace = true,  features = ["file"] }
 fluvio-protocol = { workspace = true,  features = ["fixture"] }
+fluvio-compression = { workspace = true, features = ["compress"] }
```
sehz, line 61: is this only needed for iterator? if so, make it optional and add to iterators's dependency

sehz (changes requested):
Looks good except a minor nit about feature flag

crates/fluvio-storage/Cargo.toml
morenol: It is not needed for iterator, It is just used in tests. I added this in `dev-dependencies` only

sehz (approved):
LGTM

PR #3616, opened 2023-10-20
[Merged by Bors] - fix: replace `fvm` binary on `fvm self install`

EstebanBorai:
Ensures the FVM binary is replaced when running `fvm self install`

The approach involves making sure FVM where `self install` is executed is not
the same as `fvm` which is already installed. Otherwise `bus` errors will occur by
trying to replace the ongoing process binary.

### DIFF
--- a/crates/fluvio-version-manager/src/command/itself/install.rs
+++ b/crates/fluvio-version-manager/src/command/itself/install.rs
@@ -1,5 +1,5 @@
 use std::env::current_exe;
-use std::fs::{create_dir, copy, write};
+use std::fs::{copy, write, create_dir_all};
 use std::path::PathBuf;
 
 use clap::Parser;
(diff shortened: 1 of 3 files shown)


digikata (approved):
LGTM, good to add a test for it

crates/fluvio-version-manager/src/command/itself/install.rs
```diff
@@ -50,24 +40,46 @@ impl SelfInstallOpt {
...
-            tracing::debug!(?fvm_dir, "Created FVM home directory with success");
-        }
-
-        // Attempts to create the binary crate
-        let fvm_binary_dir = fvm_dir.join("bin");
-        create_dir(&fvm_binary_dir)?;
-        tracing::debug!(?fvm_binary_dir, "Created FVM bin directory with success");
+        // Creates the binaries directory
+        let bin_dir = fvm_dir.join("bin");
+        create_dir_all(bin_dir)?;
```
EstebanBorai, line 71: Instead of checking for directories before creating:

```rust
 if !fvm_dir.exists() {
   create_dir(&fvm_dir)?;
   tracing::debug!(?fvm_dir, "Created FVM home directory with success");
 }
```

We can use `create_dir_all` which not only create parent dirs in a single execution
but it also allows for no-op on already existent dirs.

crates/fluvio-version-manager/src/command/itself/install.rs
```diff
@@ -25,19 +25,9 @@ pub struct SelfInstallOpt;
...
-                "FVM is already installed at {}",
-                bin_path.display()
-            ));
-            return Ok(());
-        }
-
         let fvm_installation_path = self.install_fvm()?;
-        Settings::init()?;
+
+        Settings::open()?;
```
EstebanBorai, line 30: `Settings::init` cretes the `Settings` file if it doesnt exists, instead `Settings::open` will attempt to open the `settings.toml` file, but if it doesnt
exists it creates it (using `Settings::init`).

digikata (approved):

PR #3620, opened 2023-10-24
fix: restrict cluster commands based on profile

### LINKED ISSUE #3619: Restrict `cluster` commands based on current profile
Currently, fluvio cli may confuse users when operating on clusters because it's not always clear if they're running against a `cloud` or `local` profile. We should improve user experience by restricting which commands can be run depending on the current profile and providing a helpful error message.

---

We should restrict most of the `fluvio cluster` commands and display a helpful error message when the current profile is pointing to `cloud`:

```sh
$ fluvio profile
 cloud
$ fluvio cluster check
 Invalid command on `cloud`. Switch to `local` profile or start a `local` cluster and try again. 
```

the only exception is `fluvio cluster start`, which should work regardless of the current profile.

```sh
$ fluvio profile
 cloud
$ fluvio cluster start
 (start a local cluster)
$ fluvio profile
 local
```

matheus-consoli:
Closes #3619

Check the current profile to restrict which commands are available.

Also, improve the error message when there's no profile:

```diff
$ fluvio cluster check
-Config error: Config has no active profile
+no profile available, create a local cluster or log in to InfinyOn Cloud
```

### DIFF
--- a/crates/fluvio-cluster/src/cli/mod.rs
+++ b/crates/fluvio-cluster/src/cli/mod.rs
@@ -2,6 +2,9 @@ use std::sync::Arc;
 
 use clap::ValueEnum;
 use clap::Parser;
+use fluvio::config::CLOUD_PROFILE;
+use fluvio::config::ConfigFile;
+use fluvio::config::LOCAL_PROFILE;
 use semver::Version;
 use tracing::debug;
 
@@ -97,6 +100,7 @@ impl ClusterCmd {
         platform_version: Version,
         target: ClusterTarget,
     ) -> Result<()> {
+        self.ensure_command_is_available_for_current_profile()?;
         match self {
             Self::Start(mut start) => {
                 if let Ok(tag_strategy_value) = std::env::var(FLUVIO_IMAGE_TAG_STRATEGY) {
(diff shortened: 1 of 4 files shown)


matheus-consoli:
CI failed because the tests were creating a fake cloud profile pointing to an unreal address.

It seems that the tests weren't executing commands particularly related to these, so I changed the tests to create a fake cloud cluster pointing to the local cluster.

ajhunyady:
@digikata that's correct. I thought we captured that at some point when we discussed the PRD.

digikata:
I found this note on the cluster cleanup PRD, but it's a little obscure https://github.com/infinyon/roadmap/blob/c74519c8adb9ec049f679d977aba581601eb2396/prd/cluster-cleanup.md?plain=1#L168

ajhunyady:
> I found this note on the cluster cleanup PRD, but it's a little obscure https://github.com/infinyon/roadmap/blob/c74519c8adb9ec049f679d977aba581601eb2396/prd/cluster-cleanup.md?plain=1#L168

I think the summary above is a bit more thoughtful. We should debate that instead. I think multi-cluster cares more about naming conventions than anything else. 

From this requirement point of view, all those clusters would be of the same `infinyon-cloud` type.

sehz:
Yes, we decide to add `profile type`.   One is `infinyon`, other is `local`?

ajhunyady:
> Yes, we decide to add `profile type`. One is `infinyon`, other is `local`?

Yes, parts of that is captured in the doc (Yes) for INFINYON.

digikata (approved):

matheus-consoli:
@digikata, this is not ready for merging yet, I need to update it to use the cluster metadata.
I market it as a draft for now

PR #3628, opened 2023-10-26
[Merged by Bors] - feat: add support for cluster metadata

### LINKED ISSUE #3627: Allow annotations in fluvio profiles
The Fluvio profile should have an annotations section which allows other tools to "tag" or "label". I imagine this should work just like kubernetes object annotations.

Example 1:
```
[profile.<label>.annotations]
fluvio.io/cluster-type = "local-k8"
```

Example 2:
```
[profile.<label>.annotations]
infinyon.cloud/environment = "infinyon.cloud"
infinyon.cloud/cluster-name = "my-cloud-cluster-name"
```

matheus-consoli:
Add support for custom metadata in the cluster configuration.

```toml
[cluster.my_favorite_cluster]
endpoint = "127.0.0.1:9003"

[cluster.my_favorite_cluster.metadata.installation]
"fluvio.io/cluster-type" = "local-k8"
```

This will be important for other tasks, such as #3620.

These changes are retrocompatible.

Closes #3627

### DIFF
--- a/crates/fluvio/Cargo.toml
+++ b/crates/fluvio/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio"
-version = "0.21.1"
+version = "0.21.2"
 edition = "2021"
 license = "Apache-2.0"
 authors = ["Fluvio Contributors <team@fluvio.io>"]
--- a/crates/fluvio/Cargo.toml
+++ b/crates/fluvio/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio"
-version = "0.21.2"
+version = "0.21.1"
 edition = "2021"
 license = "Apache-2.0"
 authors = ["Fluvio Contributors <team@fluvio.io>"]
--- a/crates/fluvio/src/config/cluster.rs
+++ b/crates/fluvio/src/config/cluster.rs
@@ -62,6 +62,74 @@ impl FluvioConfig {
         self.tls = tls.into();
         self
     }
+
+    pub fn query_metadata_path<'de, T>(&self, path: &str) -> Option<T>
+    where
+        T: Deserialize<'de>,
+    {
+        let mut metadata = self.metadata.as_ref().unwrap();
+
+        let (path, key) = {
+            let mut split = path.split(&['[', ']']);
+            (split.next().unwrap_or(path), split.next())
+        };
+
+        for part in path.split('.') {
+            if let toml::Value::Table(table) = metadata {
+                metadata = table.get(part)?;
+            }
+        }
+
+        if let Some(key) = key {
+            metadata = metadata.as_table()?.get(key)?;
+        }
+
+        T::deserialize(metadata.clone()).ok()
+    }
+
+    pub fn update_metadata_path<S>(&mut self, path: &str, data: S) -> anyhow::Result<()>
+    where
+        S: Serialize,
+    {
+        use toml::{Value, map::Map};
+
+        let (path, key) = {
+            let mut split = path.split(&['[', ']']);
+            (split.next().unwrap_or(path), split.next())
+        };
+
+        if let Some(mut metadata) = self.metadata.as_mut() {
+            for part in path.split('.') {
+                let Value::Table(table) = metadata else {
+                    break;
+                };
+                let nested = table
+                    .entry(part)
+                    .or_insert_with(|| Value::Table(Map::new()));
+                metadata = nested;
+            }
+
+            if let Some(key) = key {
+                metadata = metadata
+                    .as_table_mut()
+                    .expect("metadata should be a table at this point")
+                    .get_mut(key)
+                    .ok_or_else(|| anyhow::anyhow!("key does not exist"))?;
+            }
+
+            *metadata = Value::try_from(data)?;
+        } else {
+            // insert new metadata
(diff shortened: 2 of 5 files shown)


sehz:
This just adds more noise.  

Instead add cluster type to config object so can avoid parsing error

sehz:
[cluster.my_favorite_cluster]
endpoint = "127.0.0.1:9003"
type = "local-k8"

matheus-consoli:
> Instead add cluster type to config object so can avoid parsing error

That would work for the cluster-type issue, but I think @galibey also has a use case for this same feature (cmiiw @galibey)

sehz:
then lets' add more properties as necessary.  annotations doesn't buy anything except adding more friction and increase complexity

sehz:
if there are more properties to add, then it probably make sense to move profile as separate crate to make it more reusable

digikata:
The eng team already discussed this in the engineering meeting and prefers to use annotations.  There is a tradeoff between hard coded properties and annotations. One benefit of annotation is that there is less coupling between different versions later. I think as long as Nick, Alex, and Matheus agree on the implementation detail, we can let them move forward.

nacardin (commented):
Why not add this under the `Profile` struct instead?  It would be easier to match to a profile if it was at the `Profile` level. Putting it at the `FluvioConfig` level requires querying the cluster reference while iterating through profiles.

crates/fluvio/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio"
-version = "0.21.1"
+version = "0.21.2"
```
sehz, line 3: 21.1 is not release, so no need to bump

sehz (changes requested):
Overall LGTM.  minor nit

galibey:
What is the point of having `toml::Value` field? It is not so much better than just vec of bytes, it is too loose type in my opinion. Plus, we couple our serialization format (toml) to public API.
If we choose to no go with strictly type fields in the cluster config, we could use some `MetadataAnnotation` enum that encapsulates different possible representations there and we expose `metadata: Vec<MetadataAnnotation>` instead of `pub metadata: Option<toml::Value>`

sehz:
This is similar to the approach cargo metadata: https://docs.rs/cargo_metadata/0.18.1/cargo_metadata/struct.Metadata.html.   The whole point of annotation is to add opinionated config to config without adding more dep to the Fluvio crate.  It's the same problem, even if this moves to a separate crate.   Then, it is the responsibility of the reader/producer of annotation to ensure the config is correct.  With serde `Value`,  they can choose to use strong type or not.

crates/fluvio/Cargo.toml
matheus-consoli: check crate-version [failed](https://github.com/infinyon/fluvio/actions/runs/6702295994/job/18211081175?pr=3628), I had to change it back to 0.21.2

sehz (approved):
LGTM

matheus-consoli:
the [first attempt failed](https://github.com/infinyon/fluvio/actions/runs/6712710226/job/18243007947) because of a socket timeout, trying again

bors r+

galibey:
@sehz Imagine you need access to some metadata item from 3 different crates: you have to have `serde_toml` in all these three places, and you have to do deserialization there, also, you need that target type dependency (or use code duplication). So, it will be something like `let my_type: MyType = serde_toml::from_value(config.metadata)` everywhere. It is not convenient at all.

matheus-consoli:
@galibey I think this is a valid concern, what do you think if we add a way for querying metadata?

I think it would look something like `pub fn query_metadata<'de, T: Deserialize<'de>>(&self, query: &str) -> Option<T>`, with a micro-dls to target specific fields from some metadata table.

for example, consider the table below and a query to get `"value"`:

```toml
[cluster.name.metadata.deep.nested]
key = "value"
```
```rust
assert_eq!(cluster.query_metadata::<&str>("deep.nested[key]"), Some("value"));
```

If we agree on this, it will target all your concerns to some extent:

- [x] no need to add `toml` as a dependency everywhere
- [x] (in some cases) we don't need the target type or code duplication
  - we can query specific fields, or deserialize them into some `map`-structure
- [x] is more convinient

some downsides:
- cost of search
- stringly-typed

galibey:
@matheus-consoli Yes, that would work. I like the idea.

galibey:
We will also need a method for inserting a metadata entry.

sehz:
Let's not add more complexity to this API and more work.   If some client wants to add strong typing, they can create a wrapper around it which at that point, can hide all dependency.

I agree with @galibey, inserting and updating is needed.

crates/fluvio/src/config/cluster.rs
```diff
@@ -58,6 +62,74 @@ impl FluvioConfig {
         self.tls = tls.into();
         self
     }
+
+    pub fn query_metadata_path<'de, T>(&self, path: &str) -> Option<T>
+    where
+        T: Deserialize<'de>,
+    {
+        let mut metadata = self.metadata.as_ref().unwrap();
```
galibey, line 70: This is not good. If there is no metadata, the querying here will shut my process.

matheus-consoli: ohno, you're right, I missed that one!

PR #3638, opened 2023-10-31
[Merged by Bors] - feat: add cascade deletion for local metadata

### LINKED ISSUE #3632: CI: add back `local` cluster to partitioning tests
Local partitioning test is still unstable: https://github.com/infinyon/fluvio/actions/runs/6685278450.  Add back to CI when it is more stable

galibey:
Added cascade deletion for local metadata.

Also, fixed a bug with deserialization failure for nested enums if the outer enum is tagged. More details here https://github.com/dtolnay/serde-yaml/issues/378. Fixed by removing external tagging for versioned spec and added unit-tests that cover this case.

Also fixes #3632

### DIFF
--- a/crates/fluvio-stream-dispatcher/src/metadata/local.rs
+++ b/crates/fluvio-stream-dispatcher/src/metadata/local.rs
@@ -1,6 +1,6 @@
 use std::{
     path::{Path, PathBuf},
-    collections::HashMap,
+    collections::{HashMap, hash_map::Entry},
     sync::{Arc, atomic::AtomicU64},
     any::Any,
     ffi::OsStr,
@@ -11,7 +11,7 @@ use async_channel::{Sender, Receiver, bounded};
 use async_lock::{RwLock, RwLockUpgradableReadGuard};
 use futures_util::{stream::BoxStream, StreamExt};
 use serde::{Serialize, Deserialize, de::DeserializeOwned};
-use tracing::{warn, debug, error, trace};
+use tracing::{warn, debug, trace};
 
 use fluvio_stream_model::{
     core::{MetadataItem, Spec, MetadataContext},
@@ -34,21 +34,13 @@ pub struct LocalMetadataStorage {
 pub struct LocalMetadataItem {
     id: String,
     revision: u64,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    parent: Option<Box<LocalMetadataItem>>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    children: Option<HashMap<String, Vec<LocalMetadataItem>>>,
 }
 
 pub type LocalStoreObject<S> = MetadataStoreObject<S, LocalMetadataItem>;
-
-impl MetadataItem for LocalMetadataItem {
-    type UId = String;
-
-    fn uid(&self) -> &Self::UId {
-        &self.id
-    }
-
-    fn is_newer(&self, another: &Self) -> bool {
-        self.revision > another.revision
-    }
-}
 
 #[async_trait::async_trait]
 impl MetadataClient<LocalMetadataItem> for LocalMetadataStorage {
@@ -70,14 +62,13 @@ impl MetadataClient<LocalMetadataItem> for LocalMetadataStorage {
         trace!(?metadata, "delete item");
         let store = self.get_store::<S>().await?;
         if let Some(item) = store.try_retrieve_item::<S>(&metadata).await? {
-            if let Some(owner) = item.ctx().owner() {
+            if let Some(owner) = item.ctx().item().owner() {
                 self.unlink_parent::<S>(owner, item.ctx().item()).await?;
             }
             self.delete_children(item).await?;
             store.delete_item(&metadata).await
-        } else {
-            Ok(())
-        }
+        };
+        Ok(())
     }
 
     async fn finalize_delete_item<S>(&self, metadata: LocalMetadataItem) -> Result<()>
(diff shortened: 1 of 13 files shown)


crates/fluvio-stream-model/src/core.rs
```diff
@@ -107,6 +117,12 @@ mod context {
         pub fn set_owner(&mut self, ctx: C) {
             self.owner = Some(ctx);
         }
+        pub fn children(&self) -> &HashMap<String, Vec<C>> {
+            &self.children
+        }
+        pub fn set_children(&mut self, children: HashMap<String, Vec<C>>) {
```
sehz, line 123: Since this is specific to `local` metadata, may be better to put it in local storage rather than put this into common.  This feels like it is leaking abstraction since parent-child relationship should be tracked by storage

crates/fluvio-stream-model/src/core.rs
```diff
@@ -117,13 +133,40 @@ mod context {
...
 
         pub fn set_labels<T: Into<String>>(self, labels: Vec<(T, T)>) -> Self {
             Self {
                 item: self.item.set_labels(labels),
                 owner: self.owner,
+                children: Default::default(),
+            }
+        }
+
+        pub fn put_child<S: Into<String>>(&mut self, kind: S, child: C) {
```
sehz, line 148: add unit test

crates/fluvio-stream-model/src/core.rs
```diff
@@ -117,13 +133,40 @@ mod context {
...
+                    if !vec.contains(&child) {
+                        vec.push(child);
+                    }
+                }
+                Entry::Vacant(entry) => {
+                    entry.insert(vec![child]);
+                }
+            }
+        }
+        pub fn remove_child<S: Into<String>>(&mut self, kind: S, child: &C) {
```
sehz, line 161: add unit test

sehz (changes requested):

crates/fluvio-stream-dispatcher/src/metadata/local.rs
```diff
@@ -214,6 +230,58 @@ impl LocalMetadataStorage {
...
+        for (kind, children) in item.ctx().children() {
+            let child_store = self.get_store_by_key(kind).await?;
+            for child in children {
+                trace!(?item, ?child, "delete child");
+                if let Err(err) = child_store.delete_item(child).await {
+                    error!(kind, ?child, "child deletion: {err}");
+                }
+            }
+        }
+        Ok(())
```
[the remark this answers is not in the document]
galibey: I don't understand what the is question here. Is the behavior not clear from the code?

crates/fluvio-stream-dispatcher/src/metadata/local.rs
```diff
@@ -1022,12 +1133,128 @@ spec:
...
+            .await
+            .expect("deleted parent");
+
+        let after = meta_store
+            .retrieve_items::<TestSpec>(&NameSpace::All)
+            .await
+            .expect("items");
+
+        //then
+        assert_eq!(before.items.len(), 2);
```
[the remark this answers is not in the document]
galibey: It does check that children's items are deleted. See the next line (1170)

crates/fluvio-stream-model/src/core.rs
galibey: It would be weird that Child -> Parent abstraction resides in `MetadataStoreObject` and the other way `Parent->Child` is in another place. It would not help for clarity and readability of the code.

crates/fluvio-stream-dispatcher/src/metadata/local.rs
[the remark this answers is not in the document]
sehz:
```
if let Err(err) = child_store.delete_item(child).await {
                    error!(kind, ?child, "child deletion: {err}");
                }
```

Can't just ignore this failure.  This should be propagated to the caller by raising an error or another way.  Or store error condition somewhere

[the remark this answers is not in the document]
sehz: Parent should not be allowed to delete unless all children are deleted successfully.

crates/fluvio-stream-model/src/core.rs
sehz: Currently, this call is only used by LocalStorage.  This leads to leak abstraction and inconsistency implementations.   Suggestions to work around this:
- Provide way to put "custom" object in the context so you can whatever you need do.  So something like: 
```
pub struct MetadataContext<C> {
        item: C,
        owner: Option<C>,
        custom: C::CustomContext
    }
```
- Or make metadata dynamic trait do downcast as necessary

-  Move all children relationship to Metadata and get rid of context
```
pub trait MetadataItem {
     ...
     owner()...
     setOwner(),
     getChildren()
     setChildren()
}
```

That way, you could implement

crates/fluvio-stream-dispatcher/src/metadata/local.rs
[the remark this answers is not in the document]
galibey: Yes, that's correct, the parent is not allowed to delete unless all children are deleted successfully. This method `child_store.delete_item(child)` is basically infallible, so this error handling is unnecessary. I removed `Result<()>` from its signature.

crates/fluvio-stream-model/src/core.rs
galibey: I chose the third option: moved all to `MetadataItem` and removed them from `MetadataContext`, but I kept `MetadataContext` itself so far to reduce the number of changes in one PR and the context might be beneficial in the future.

crates/fluvio-stream-model/src/core.rs
galibey: done

crates/fluvio-stream-model/src/core.rs
galibey: done

crates/fluvio-stream-model/src/core.rs
sehz: Great

sehz (approved):
LGTM.  Nice job. 

Only possible problem I see is that some of tests in the CI were not running reliably with `local` mode with previous changes.   So need to check if they are running ok once this is merged.

galibey:
bors retry

PR #3645, opened 2023-10-31
[Merged by Bors] - feat: fvm update command

EstebanBorai:
Introduces the `fvm update` command which checks current channel to be up to date.
Channels `stable` and `latest` are both "updateable", in the other hand the static
tag is not "updateable".

### How it works

When a user has either `stable` or `latest` as the active channel, FVM fetches the most
recent release for that channel.

- Stable: Will check if the upstream version is greater than the current using Semver.
- Latest: Will check if the upstream tag is different. We cant tell which is greater because we use SHAs to determine versions.
- Static Tag: When the user has an static tag as version using for instance `fvm install 0.10.14`, FVM will not attempt to update.

Then the same process as for `fvm install` is followed, this is why the logic is first
disengaged from `fvm install` into its own module on commit: [0a8b819](https://github.com/infinyon/fluvio/pull/3645/commits/0a8b8194e409ed3c6f967de6cf7a48b51b002bbb).

In order to determine the current version for the channel FVM uses the already implemented `settings.toml` file.

## Demo

CleanShot 2023-10-31 at 20 58 13

> By changing the version in the settings.toml, we can mimic an outdated version for the active channel.

### DIFF
--- a/crates/fluvio-version-manager/src/command/install.rs
+++ b/crates/fluvio-version-manager/src/command/install.rs
@@ -3,22 +3,18 @@
 //! Downloads and stores the sepecific Fluvio Version binaries in the local
 //! FVM cache.
 
-use std::fs::{File, create_dir, create_dir_all, rename};
-
-use std::path::PathBuf;
+use std::fs::create_dir_all;
 
 use anyhow::Result;
 use clap::Parser;
-use tempfile::TempDir;
 use url::Url;
 
 use fluvio_hub_util::HUB_REMOTE;
-use fluvio_hub_util::fvm::{Client, Download, PackageSet, DEFAULT_PKGSET, Channel};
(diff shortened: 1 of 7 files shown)


ajhunyady:
What happens if I run `fvm update` and there is not new image?

EstebanBorai:
> What happens if I run `fvm update` and there is no new image?
> 
> 

You will see:

```
You are up to date
```

EstebanBorai:
> What happens if I run `fvm update` and there is no new image?

Just added a test for this here: https://github.com/infinyon/fluvio/pull/3645/commits/b592395b2090dbc40acabdabeb2e73033d322b66#diff-9e90aaedcb1d6ed2a0448f1c8fb5306d2c9788f1ed82e0d22619efb9f2f277e2R745-R768

digikata (approved):
Looks good!

Issue #3577, opened 2023-10-04
Update syn@1 to syn@2 in fluvio-protocol-derive crate


digikata:
Dependabot identified the need for the dep update https://github.com/infinyon/fluvio/pull/3548

But, the update isn't a simple version change.

sehz:
@matheus-consoli ?

digikata:
We'll have him review if there is a candidate submission unless this is blocking work?

jsat97:
hi. new to fluvio here. need to start somewhere I guess.
would this be the change ?
jai@ubuntu:~/fluvio/crates/fluvio-protocol-derive$ git diff
diff --git a/crates/fluvio-protocol-derive/Cargo.toml b/crates/fluvio-protocol-derive/Cargo.toml
index db57e918..fde40195 100644
--- a/crates/fluvio-protocol-derive/Cargo.toml
+++ b/crates/fluvio-protocol-derive/Cargo.toml
@@ -18,7 +18,7 @@ quote = { workspace = true }
 tracing = { workspace = true }

 [dependencies.syn]
-version = "1.0.0"
+version = "2.0.0"
 features = ["full"]

Is yes, how do I test pls ?
--Jai

sehz:
API for syn has changed,  There are tests already https://github.com/infinyon/fluvio/tree/master/crates/fluvio-protocol/tests

morenol:
Similar to this PR https://github.com/infinyon/fluvio/pull/3513

jsat97:
@morenol  so change for this was already done with above PR ?

PR #3648, opened 2023-11-01
Updating fluvio-protocol-derive crate from syn1 to syn2

vrrashkov:
Hello, I took on updating the  `fluvio-protocol-derive` from `syn1` to `syn2` from #3577

I checked all the tests and of course are passing, but there is something I am not sure about.

Here it is expecting `Literal` value

https://github.com/infinyon/fluvio/blob/d0dd8595583c0dd4651713beb637fb0b5c48c7fb/crates/fluvio-protocol-derive/src/ast/prop.rs#L200C29-L200C29

but there is one place in the code in the `fluvio-spu-schema` crate, that `PRODUCER_TRANSFORMATION_API` is being passed as a value. Because this is not a `Literal` but a `Path` it actually never goes inside the if statement.  Also not sure but I couldn't find this constant anywhere. 

https://github.com/infinyon/fluvio/blob/d0dd8595583c0dd4651713beb637fb0b5c48c7fb/crates/fluvio-spu-schema/src/produce/request.rs#L44C16-L44C16

So because of that I have also set this place the same way, with an if statement and a literal, but it just will never go in there. Can you tell me how would you like me to handle that.  

Other than that please tell me if I have missed on something.

### DIFF
--- a/crates/fluvio-protocol-derive/src/util.rs
+++ b/crates/fluvio-protocol-derive/src/util.rs
@@ -67,12 +67,10 @@ pub(crate) fn find_name_value_from_meta_list(
 }
 
 pub(crate) fn find_name_value_from_meta(meta: &Meta, attr_name: &str) -> Option<MetaNameValue> {
-    let mut result_meta: Option<MetaNameValue> = None;
     if let Meta::List(list) = meta {
-        result_meta = find_name_value_from_meta_list(list, attr_name);
+        return find_name_value_from_meta_list(list, attr_name);
     }
-
-    result_meta
+    None
 }
 
 /// find name value with str value
--- a/crates/fluvio-protocol-derive/src/util.rs
+++ b/crates/fluvio-protocol-derive/src/util.rs
@@ -1,75 +1,113 @@
-use syn::{Attribute, Lit, LitStr, Meta, MetaNameValue, NestedMeta};
+use syn::{Attribute, Expr, Lit, LitStr, Meta, punctuated::Punctuated, Token, MetaNameValue, MetaList};
+pub fn get_lit_int(attr_name: String, value: &Expr) -> syn::Result<&syn::LitInt> {
+    match &value {
+        Expr::Lit(lit_expr) => {
+            if let Lit::Int(lit) = &lit_expr.lit {
+                Ok(lit)
+             } else {
+                 Err(syn::Error::new_spanned(
+                     value,
+                     format!("Expected {attr_name} attribute to be a int: `{attr_name} = \"...\"`"),
+                 ))
+             }
+        },
+        _ => {
+            Err(syn::Error::new_spanned(
+                value,
+                format!("Expected {attr_name} attribute to be Path or Lit: `{attr_name} = \"...\"`"),
+            ))
+        },
+    }
+}
+pub fn get_lit_str(attr_name: String, value: &Expr) -> syn::Result<&syn::LitStr> {
+    match &value {
+        Expr::Lit(lit_expr) => {
+            if let Lit::Str(lit) = &lit_expr.lit {
+                Ok(lit)
+             } else {
+                 Err(syn::Error::new_spanned(
+                     value,
+                     format!("Expected {attr_name} attribute to be a string: `{attr_name} = \"...\"`"),
+                 ))
+             }
+        },
+        _ => {
+            Err(syn::Error::new_spanned(
+                value,
+                format!("Expected {attr_name} attribute to be Path or string: `{attr_name} = \"...\"`"),
+            ))
+        },
+    }
+}
 
 pub(crate) fn find_attr(attrs: &[Attribute], name: &str) -> Option<Meta> {
     attrs.iter().find_map(|a| {
-        if let Ok(meta) = a.parse_meta() {
-            if meta.path().is_ident(name) {
-                Some(meta)
-            } else {
-                //println!("attr name: {}",meta.name());
-                None
-            }
+        if a.meta.path().is_ident(name) {
+            Some(a.meta.clone())
         } else {
-            //println!("unrecog attribute");
+            //println!("attr name: {}",meta.name());
             None
         }
     })
 }
-
-pub(crate) fn find_name_attribute<'a>(meta: &'a Meta, name: &str) -> Option<&'a MetaNameValue> {
-    find_meta(meta, name).map(|meta| match meta {
-        Meta::NameValue(name_value) => name_value,
-        _ => panic!("should not happen"),
-    })
-}
-
-pub(crate) fn find_meta<'a>(meta: &'a Meta, name: &str) -> Option<&'a Meta> {
-    if let Meta::List(list) = meta {
-        for attr in list.nested.iter() {
-            if let NestedMeta::Meta(named_meta) = attr {
-                match named_meta {
-                    Meta::NameValue(meta_name_value) => {
-                        if meta_name_value.path.is_ident(name) {
-                            return Some(named_meta);
-                        }
-                    }
-                    Meta::Path(path) => {
-                        if path.is_ident(name) {
-                            return Some(named_meta);
-                        }
-                    }
-                    Meta::List(_) => {}
+pub(crate) fn find_name_value_from_meta_list(meta_list: &MetaList, attr_name: &str) -> Option<MetaNameValue> {
+    let mut result_meta: Option<MetaNameValue> = None;
+    for item in meta_list
+        .parse_args_with(Punctuated::<syn::Meta, Token![,]>::parse_terminated)
+        .unwrap_or_else(|err| panic!("Invalid #[new] attribute: {}", err))
+    {
+        match &item {
+            Meta::NameValue(kv) => {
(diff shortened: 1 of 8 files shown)


jsat97:
wow @vrrashkov  that was pretty quick. I've been trying this for a while, but got stuck on util.rs (I'm new to Rust).

sehz:
@morenol  or @galibey respond do it?

vrrashkov:
> wow @vrrashkov that was pretty quick. I've been trying this for a while, but got stuck on util.rs (I'm new to Rust).

The main issue with proc-macros syn2 is the lack of documentation. In the future it will also be nice, if there are multiple proc-macro crates just to centralise the main utils and build some easy to use utils for cleaner proc-macro creation.

crates/fluvio-protocol-derive/src/util.rs
```diff
@@ -1,75 +1,110 @@
...
-pub(crate) fn find_int_name_value(version_meta: &Meta, attr_name: &str) -> Option<u64> {
-    if let Some(attr) = find_name_attribute(version_meta, attr_name) {
-        match &attr.lit {
-            Lit::Int(version_val) => {
-                //  println!("version value: {}",version_val.value());
-                version_val.base10_parse::<u64>().ok()
-            }
-            _ => unimplemented!(),
+/// find name value with str value
+pub(crate) fn find_string_name_value(meta: &Meta, attr_name: &str) -> Option<LitStr> {
```
galibey, line 86: It could be less verbose, something like this:
```rust
pub(crate) fn find_string_name_value(meta: &Meta, attr_name: &str) -> Option<LitStr> {
    let meta_name_value = find_name_value_from_meta(meta, attr_name)?;
    get_lit_str(String::from(attr_name), &meta_name_value.value).ok().cloned()
}
```

crates/fluvio-protocol-derive/src/util.rs
```diff
@@ -1,75 +1,110 @@
...
 }
 
-/// find name value with str value
-pub(crate) fn find_string_name_value(version_meta: &Meta, attr_name: &str) -> Option<LitStr> {
-    if let Some(attr) = find_name_attribute(version_meta, attr_name) {
-        match &attr.lit {
-            Lit::Str(val) => Some(val.clone()),
-            _ => unimplemented!(),
+pub(crate) fn find_int_name_value(meta: &Meta, attr_name: &str) -> Option<u64> {
+    let mut value: Option<u64> = None;
```
galibey, line 100: I would recommend avoiding mutable variables where it's possible. See some similar suggestions above.

galibey (changes requested):
Thanks for your contribution @vrrashkov. Overall looks great, just a couple of suggestions.

Regarding `PRODUCER_TRANSFORMATION_API`, it is a good catch! Ideally, we would want to support `Path` there along with `Lit` but this should be a subject for another PR. In this PR, I would ask you to replace that usage of `PRODUCER_TRANSFORMATION_API` with its actual literal value. So, we will have a git history with reasons for this change. And, add some compilation time warnings if that case repeats.

vrrashkov:
> galibey

Thank you for the awesome feedback, updated the code based on the above points. Please let me know if there is something else.

galibey:
@vrrashkov Could you please bump patch versions for `fluvio-spu-schema` and `fluvio-protocol-derive` crates to make ci checks happy?

Also, probably a rebase from master is needed.

crates/fluvio-protocol-derive/src/util.rs
```diff
@@ -1,75 +1,90 @@
...
-            Lit::Int(version_val) => {
-                //  println!("version value: {}",version_val.value());
-                version_val.base10_parse::<u64>().ok()
-            }
-            _ => unimplemented!(),
-        }
-    } else {
-        None
+pub(crate) fn find_name_value_from_meta(meta: &Meta, attr_name: &str) -> Option<MetaNameValue> {
+    let mut result_meta: Option<MetaNameValue> = None;
```
galibey, line 70: Not critical, but if you have time, remove this mut variable as well.

vrrashkov: Opps i missed that.

galibey:
@vrrashkov Seems like the rebase went not well, it is pulling other's commits as well now.

vrrashkov:
> @vrrashkov Seems like the rebase went not well, it is pulling other's commits as well now.

Sorry, should be fixed.

galibey (approved):
Good job!

crates/fluvio-spu-schema/src/produce/request.rs
```diff
@@ -41,7 +41,7 @@ pub struct ProduceRequest<R> {
     /// Each topic to produce to.
     pub topics: Vec<TopicProduceData<R>>,
 
-    #[fluvio(min_version = PRODUCER_TRANSFORMATION_API)]
```
sehz, line 44: This is regression.  Lets fix this in this PR

sehz (changes requested):

crates/fluvio-spu-schema/src/produce/request.rs
galibey: I suggest removing this line, which will be equivalent to what we have now. And deal with setting `min_version` to 8 in a separate issue.

vrrashkov: There are couple of things that need to be considered if both are going to be available. 
Currently the code is using the value of `min_version` for example in couple of places passing it as a `Literal` value to `struct`, doing some checks etc, for example here: 

https://github.com/infinyon/fluvio/blob/b0f846dfd618c3e7ece53918ca87391c217016fa/crates/fluvio-protocol-derive/src/ast/prop.rs#L38C25-L38C25

If this also accepts a `Path`, this means this code cannot work and this logic should change, because basically a `Path` is not a variable so we don't know what it's value is or if it actually has any value, the only thing we have is the name of the "variable". Even if you check the current crate, it's trying to use non existent variable without crashing, just because it doesn't know it's a variable because it's just a `Path`.

A way maybe this can work is if this is treated like an "env" variable and again we just have the Path name which will be the same as passing a literal with `"the_variable_name_as_literal"` ( similar to how `serde` does it for their `searlize_with `funcs for example). So it will only help with centralising the min/max versions for example. And again this is based on the idea that the current code should work where the Literals are used like currently in "validate_versions" for example.

That's what currently is on my mind but not really sure what the right direction would be.

galibey: Let's not worry about supporting `Path` in this PR

sehz: Problem is that this may not be addressed soon and we will lose consistent check.   @vrrashkov, Your suggestion using serde like approach seems to be good alternative as below:
```
fn min_transformation_api() -> i16 {
    8
}

   #[fluvio(min_version = "min_transformation_api")]
    pub smartmodules: Vec<SmartModuleInvocation>,
```

vrrashkov: I mentioned using something like a global "`env`" as a way of doing it, because we will need a value for the `proc-macro `and at this point it's too early for the compiler to know if certain functions/variables exist and we are back at the same problem. The code mentioned above tries to make checks during the proc-macro creation so we cannot just call a `function`, but we can create an `env global `variable or a `map` or something similar that will use the `literal` to get the real value at least that's what I am thinking can work with the current setup.

galibey: @sehz What you are suggesting (support for constants or functions as value) is not trivial. It is worth a separate issue.

sehz: Sure.  Lets create issue to ensure this gets addressed soon.

galibey: Created https://github.com/infinyon/fluvio/issues/3673

galibey:
@vrrashkov FYI, there are some clippy errors

sehz:
Some of the clipper error was from previous PRs.  So not sure if this is rebase or if clippy checks were not performed during merge.  @galibey check if clippy problem exists in the master

vrrashkov:
> Some of the clipper error was from previous PRs. So not sure if this is rebase or if clippy checks were not performed during merge. @galibey check if clippy problem exists in the master

I saw that there was a fix for those clippy error, so I pulled the changes and I think it should be fine now.

EstebanBorai:
Great job @vrrashkov!

digikata:
@sehz are you recommending we wait for #3673 to merge this?  Or merge this now and start in on 3673? I think @vrrashkov is amenable to working 3673 pretty much right away.

sehz:
Let's merge this.  Good job @vrrashkov

sehz (approved):

digikata:
Reverting due to minor CI failures (and a mismerge by the merge queue) https://github.com/infinyon/fluvio/actions/runs/6898837314

vrrashkov:
@galibey Not sure if this PR is needed anymore as I have introduced those changes in #3725 along with the new ones.

Issue #3657, opened 2023-11-02
Mock FVM Server on CI


EstebanBorai:
Use a mock server on FVM instead of hitting infra to remove dependency that may slow down CI process.

CI Workflows triggered on PRs, should run against this mock, while E2E Tests against staging servers could run in a periodic workflow.

PR #3659, opened 2023-11-02
chore: disable fvm tests

EstebanBorai:
Disable FVM tests while Mock Server is introduced. Refer to https://github.com/infinyon/fluvio/issues/3657 for Mock Server Tracking.


crates/fluvio-cli/src/error.rs
```diff
@@ -47,6 +47,8 @@ pub enum CliError {
```
sehz, line 47: remove other changes

digikata: Seconded, try to make this pr just be the disable

EstebanBorai:
Moved to #3660, local Git history was messed up

PR #3664, opened 2023-11-03
[Merged by Bors] - fix: use store version on metadata stream filter

galibey:
The metadata watch stream was mistakenly filtering by a spec version when it should filter by store version.


sehz (approved):
LGTM

PR #3666, opened 2023-11-03
[Merged by Bors] - fix: revert "feat: add support for cluster metadata (#3628)"

matheus-consoli:
This reverts commit b0f846dfd618c3e7ece53918ca87391c217016fa.

PR #3667, opened 2023-11-03
feat: add support for custom metadata in cluster configuration

matheus-consoli:
Add support for custom metadata in the cluster configuration.

```toml
[cluster.name.metadata.installation]
type = "local"

[cluster.name.metadata.dict]
key = "value"
```

Metadata is managed with two functions:

```rust
cluster.query_metadata_path("installation");

struct Example {
    key: String,
}

cluster.update_metadata_path("other_example", Example { key: "value".to_owned() });
```

Reopening https://github.com/infinyon/fluvio/pull/3628 after reverting at https://github.com/infinyon/fluvio/pull/3666


crates/fluvio/src/config/cluster.rs
```diff
@@ -71,3 +143,255 @@ impl TryFrom<FluvioConfig> for fluvio_socket::ClientConfig {
...
+    use serde::{Deserialize, Serialize};
+    use crate::config::Config;
+
+    #[test]
+    fn test_get_metadata_path() {
+        let toml = r#"version = "2"
+[profile.local]
+cluster = "name"
+
+[cluster.name]
```
sehz, line 158: This examples is not right.  All cluster should have specific name. So it would be something like
```
[cluster.local.metadata.custom]
```

crates/fluvio/src/config/cluster.rs
```diff
@@ -71,3 +143,255 @@ impl TryFrom<FluvioConfig> for fluvio_socket::ClientConfig {
...
+endpoint = "127.0.0.1:9003"
+
+[cluster.name.metadata.deep.nested]
+name = "foo"
+"#;
+        let profile: Config = toml::de::from_str(toml).unwrap();
+        let config = profile.cluster("name").unwrap();
+
+        let custom: Option<String> = config.query_metadata_path("deep.nested[name]");
```
sehz, line 197: there is no need to support nested.

crates/fluvio/src/config/cluster.rs
```diff
@@ -58,6 +62,74 @@ impl FluvioConfig {
         self.tls = tls.into();
         self
     }
+
+    pub fn query_metadata_path<'de, T>(&self, path: &str) -> Option<T>
+    where
```
sehz, line 67: Don't see use case for supporting nesting (not sure even that make sense).  In fact it would make supporting complex.  Just need to way to support multiple annotations.   With strong typing, any nesting would be taken care bout annotation structure

crates/fluvio/src/config/cluster.rs
```diff
@@ -29,6 +29,9 @@ pub struct FluvioConfig {
     #[serde(default)]
     pub tls: TlsPolicy,
 
+    /// Cluster custom metadata
+    pub metadata: Option<toml::Value>,
```
sehz, line 33: Since we are supporting multiple annotations, this should be `pub metadata: Vec<toml:Value>`

sehz (changes requested):

crates/fluvio/src/config/cluster.rs
matheus-consoli: but that's exactly what it's doing, right?

the name of the cluster in this example is "name": `[cluster.name.metadata.custom]`

crates/fluvio/src/config/cluster.rs
matheus-consoli: okay, when we discussed the feature we also talked about supporting nesting, but I can remove it, no problem.

crates/fluvio/src/config/cluster.rs
matheus-consoli: it already supports multiple annotations

`metadata` will map to a `toml::Value::Table` which is basically a `BTreeMap`, so this example:

```toml
[cluster.name.metadata.one]
key = "value"

[cluster.name.metadata.two]
foo = "bar"
baz = 2
```

will deserialize (roughly) into:

```rust
metadata: BTreeMap {
    "one": BTreeMap { "key": "value" },
    "two": BTreeMap { "foo": "bar", "baz": 2 },
}
```

crates/fluvio/src/config/cluster.rs
sehz: that's wasn't clear.  typically you don't have `name` as name 😉

crates/fluvio/src/config/cluster.rs
sehz: Then it should be `pub metadata: Option<Table>` , which make it more strict.  Or it can be `pub metadata: Table` since Table is collection so no need for option

crates/fluvio/src/config/cluster.rs
sehz: Nesting is not needed since you should be able to map arbitrary struct which can be any nested

matheus-consoli: so, do you mean we should remove support for nested queries?

`query_metadata_path("arbitrary.struct.with.nested.levels")`

if so, then comes the problem pointed out in the previous PR: it's less convenient because in order to update/query a specific value in some arbitrary level of metadata you need to duplicate code or add the crate that defines the type as a dependency to the crate manipulating the type.

sehz: Yes.  Can you give example why nesting queries are necessary?  

For example, let's say you have following annotation:
```
struct MyAnnotation {
     name: String,
     size: u64,
     field1: Field1,
     field2: Field2,
     items: Vec<Field2>
}

struct Field1 {
     width: u32,
     length: u16
}

struct Field2 {
      speed: u64
}
```

So any kind of nesting can be accomplished with this scheme

matheus-consoli: maybe we're only interested in `Field1.width` and don't want to deserialize everything else? or maybe `MyAnnotation` is defined at crate `A`, and we want to query `Field2.speed` in crate `B` but don't want to add crate `A` as a dependency?

I'm basing these assumptions on this comment: https://github.com/infinyon/fluvio/pull/3628#issuecomment-1788534087

sehz: They can always parse `Value` if don't want to add dependency but this will leads other issue due to lack of strong typing.  This is limitation with Annotation.   Its ok to push burden to Annotation consumer/producer.

nacardin (approved):

sehz (approved):
LGTM.  Nice job!

crates/fluvio/src/config/cluster.rs
matheus-consoli: I renamed the cluster name to "local" to make it more clear

crates/fluvio/src/config/cluster.rs
matheus-consoli: support removed

crates/fluvio/src/config/cluster.rs
matheus-consoli: I made the metadata private and made it a `Table`

matheus-consoli:
force-pushed a rebase with master

PR #3668, opened 2023-11-06
fix: check if custom spu exists on start

galibey:
During the local cluster start we check and do not create a new one if `CustomSpu` already exists. This allows to `resume` local after shutdown. So, one could do:
```
fluvio cluster start --local
//insert some data
fluvio cluster shutdown --local
//processes killed but metadata preserved
fluvio cluster start --local
//cluster started with existing metadata and data
```
Additionally, I changed the order of where to look up fluvio external command. Now, the same folder where the current executable is the first one. It is convenient for local development (you want fluvio-run from your workspace not in fluvio home dir) and for CI tasks where we use two versions of fluvio (stable and dev).

### DIFF
--- a/crates/fluvio-cli/src/lib.rs
+++ b/crates/fluvio-cli/src/lib.rs
@@ -315,15 +315,15 @@ mod root {
 
     /// Search for a Fluvio plugin in the following places:
     ///
-    /// - In the system PATH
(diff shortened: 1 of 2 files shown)


digikata (approved):
LGTM

morenol:
bors ping

galibey:
bors ping

PR #3671, opened 2023-11-07
fix: properly propagate sm errors on produce

galibey:
Currently, errors related to SM on produce requests are not properly propagated to the Fluvio client. It leads to the produce request just hanging and is killed by timeout. This includes when smartmodule does not exist but is used in transformation.

### DIFF
--- a/crates/fluvio-spu/src/services/public/produce_handler.rs
+++ b/crates/fluvio-spu/src/services/public/produce_handler.rs
@@ -1,8 +1,6 @@
-use std::io::{Error, ErrorKind};
 use std::time::Duration;
 
 use tokio::select;
-use tracing::warn;
 use tracing::{debug, trace, error};
 use tracing::instrument;
 use anyhow::{anyhow, Result};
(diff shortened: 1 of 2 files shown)


digikata (approved):
LGTM!

crates/fluvio-spu/src/services/public/produce_handler.rs
```diff
@@ -212,48 +212,48 @@ async fn handle_produce_partition(
...
     let sm_result = match process_batch(
-        sm_chain_instance,
+        sm_ctx.chain_mut(),
         &mut batches,
         std::usize::MAX,
         ctx.metrics().chain_metrics(),
     ) {
         Ok((result, sm_runtime_error)) => {
             if let Some(error) = sm_runtime_error {
-                return Err(anyhow!("smartmodule runtime error: {error}"));
```
sehz, line 247: Instead of trying to revert changes from moving to Anyhow, let's try to find out why this was not caught since we may similar problem elsewhere.

PR #3675, opened 2023-11-08
 feat: use installation type in cli commands

galibey:
Installation type (Local, LocalK8, K8, ReadOnly) will be stored in the profile after `fluvio cluster start --local/k8/local-k8/read-only` and will be used in other cluster operations, so, the user does need to specify it on `fluvio cluster delete/shutdown/diagnostics`

### DIFF
--- a/crates/fluvio-cluster/src/cli/shutdown.rs
+++ b/crates/fluvio-cluster/src/cli/shutdown.rs
@@ -2,6 +2,7 @@ use std::fs::remove_file;
 use std::process::Command;
 
 use clap::Parser;
+use fluvio::config::ConfigFile;
 use tracing::debug;
 use sysinfo::{ProcessExt, System, SystemExt};
 
@@ -11,17 +12,10 @@ use fluvio_command::CommandExt;
 use crate::render::ProgressRenderer;
 use crate::cli::ClusterCliError;
 use crate::progress::ProgressBarFactory;
-use crate::{ClusterError, UninstallError};
+use crate::{ClusterError, UninstallError, InstallationType};
 
 #[derive(Debug, Parser)]
-pub struct ShutdownOpt {
-    /// shutdown local spu/sc
-    #[arg(long)]
-    local: bool,
-
-    #[arg(long)]
-    no_k8: bool,
-}
+pub struct ShutdownOpt;
 
 impl ShutdownOpt {
     pub async fn process(self) -> Result<(), ClusterCliError> {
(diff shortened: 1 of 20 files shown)


matheus-consoli (approved):
LGTM!

crates/fluvio-cluster/src/cli/shutdown.rs
```diff
@@ -11,17 +12,10 @@ use fluvio_command::CommandExt;
...
 #[derive(Debug, Parser)]
-pub struct ShutdownOpt {
-    /// shutdown local spu/sc
-    #[arg(long)]
-    local: bool,
-
-    #[arg(long)]
-    no_k8: bool,
-}
+pub struct ShutdownOpt;
```
digikata, line 18: Are the any cases where a manual override of the recorded installation_type might be needed?

crates/fluvio-cluster/src/lib.rs
```diff
@@ -137,4 +143,89 @@ mod common {
...
+            }
+            if s.eq_ignore_ascii_case("read_only") {
+                return Ok(Self::ReadOnly);
+            }
+            if s.eq_ignore_ascii_case("read-only") {
+                return Ok(Self::ReadOnly);
+            }
+            if s.eq_ignore_ascii_case("readonly") {
+                return Ok(Self::ReadOnly);
+            }
```
digikata, line 184: Should  these different spelling variations be allowed in the installation type parsing from str?

digikata (approved):
LGTM, I added some notes (none intended as blocking).  

A lot of the changes were a nice exmple of much cleaner the test commands were in terms of fewer flags as a result of carrying the installaton_type in the profile.

crates/fluvio-cluster/src/cli/shutdown.rs
galibey: There should not be any cases when you install one type and want to override it for delete or shutdown.
But, if there is a need for a hack, it can be done by creating a profile `fluvio profile add some_profile localhost:1234 local-k8` with setting another installation type and then execute delete or shutdown.

crates/fluvio-cluster/src/lib.rs
galibey: This is involved in the only place where we pass the installation type as an input:
```
fluvio profile add <PROFILE_NAME> <CLUSTER_ADDRESS> [INSTALLATION_TYPE] 
```
and while using it I realized that I didn't want to remember if it should be `local-k8` or `local_k8` etc. So, I added support for all of them.

However, in start command:
```
fluvio cluster start --local-k8
```
it is a CLI argument with --, so it only supports one form.

PR #3676, opened 2023-11-09
chore: use individual cli test retries

galibey:
Currently, if we use `nick-fields/retry` retries for cli tests it restarts the entire test suite on a failure. In PR we leverage `BATS_TEST_RETRIES` which allows to configure retries for an individual test which is more effective.


digikata (approved):
Sounds reasonable to me, let's merge it!

tjtelan (approved):
That's a nice improvement. I didn't know about this setting in bats

PR #3679, opened 2023-11-09
feat: update fvm to use new hub endpoints on fetch

EstebanBorai:
Updates FVM internals, specifically the Hub Client, to use the new endpoint for fetching
Package Sets.

This change does not affect user facing state.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1056,24 +1056,24 @@ jobs:
         run: make CDK_BIN=~/bin/cdk cli-cdk-smoke
 
       # test fvm
-      # - name: Download artifact - fvm
-      #   if: matrix.test == 'fvm'
-      #   uses: actions/download-artifact@v3
-      #   with:
-      #     name: fvm-x86_64-unknown-linux-musl
-      #     path: ~/bin
+      - name: Download artifact - fvm
+        if: matrix.test == 'fvm'
+        uses: actions/download-artifact@v3
+        with:
+          name: fvm-x86_64-unknown-linux-musl
+          path: ~/bin
 
-      # - name: mark fvm
-      #   if: matrix.test == 'fvm'
-      #   run: |
-      #     chmod +x ~/bin/fvm
-      #     echo "~/bin" >> $GITHUB_PATH
+      - name: mark fvm
+        if: matrix.test == 'fvm'
+        run: |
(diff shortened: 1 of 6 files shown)


crates/fluvio-hub-util/src/fvm/api/client.rs
```diff
@@ -19,13 +19,8 @@ impl Client {
...
     /// Fetches a [`PackageSet`] from the Hub with the specific [`Channel`]
-    pub async fn fetch_package_set(
-        &self,
-        name: impl AsRef<str>,
-        channel: &Channel,
-        arch: &str,
-    ) -> Result<PackageSet> {
-        let url = self.make_fetch_package_set_url(name, channel, arch)?;
+    pub async fn fetch_package_set(&self, channel: &Channel, arch: &str) -> Result<PackageSet> {
+        let url = self.make_fetch_package_set_url(channel, arch)?;
```
EstebanBorai, line 23: `name` is no longer required as we only use one now.

crates/fluvio-hub-util/src/fvm/api/client.rs
```diff
@@ -43,16 +38,10 @@ impl Client {
...
-    fn make_fetch_package_set_url(
-        &self,
-        name: impl AsRef<str>,
-        channel: &Channel,
-        arch: &str,
-    ) -> Result<Url> {
+    fn make_fetch_package_set_url(&self, channel: &Channel, arch: &str) -> Result<Url> {
         let url = format!(
-            "{}hub/v1/fvm/pkgset/{name}/{channel}/{arch}",
+            "{}hub/v1/fvm/pkgset/{channel}?arch={arch}",
```
EstebanBorai, line 43: Our URL now holds the architecture as a query parameter.

sehz: That's good improvement but shouldn't that bump up version number since it's not compatible with V1?

sehz: Since hub V1 is not released, my concern is addressed

sehz (approved):
LGTM.  API looks much clean

.github/workflows/ci.yml
EstebanBorai, line 1: Reenabled tests to check on changes

digikata (approved):

PR #3681, opened 2023-11-10
fix: add feature guards to support minimal cli version

galibey:
Fix for compilation issue introduced in https://github.com/infinyon/fluvio/pull/3675 merge.

### DIFF
--- a/crates/fluvio-cli/src/profile/add.rs
+++ b/crates/fluvio-cli/src/profile/add.rs
@@ -2,7 +2,6 @@ use clap::Parser;
 use anyhow::Result;
 
 use fluvio::config::{ConfigFile, TlsPolicy};
-use fluvio_cluster::InstallationType;
 
 #[derive(Debug, Parser)]
 pub struct ManualAddOpt {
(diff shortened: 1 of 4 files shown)


crates/fluvio-cli/src/profile/add.rs
```diff
@@ -13,7 +12,8 @@ pub struct ManualAddOpt {
     cluster_address: String,
 
     /// Installation type of cluster, e.g. local, local-k8, k8
-    installation_type: Option<InstallationType>,
+    #[cfg(feature = "k8s")]
```
matheus-consoli, line 15: why does `installation_type` need to be behind the `k8s` feature flag? shouldn't it be always available?

tjtelan: @matheus-consoli This is my opinion, but in the big picture I can see this is reasonable for the local-first installation support we are working towards. This way we're not required to be tied to k8 on systems that might not support it.

The reason for this specific gate is because we include `fluvio-cluster` with the `k8s` feature, and we don't build windows fluvio cli with k8s support.

This is the failure we saw w/o this gate: https://github.com/infinyon/fluvio/actions/runs/6822041861/job/18553405148

galibey:
Closed in favor of https://github.com/infinyon/fluvio/pull/3691

PR #3684, opened 2023-11-10
chore: ci, patch up workflow sequencing

digikata:


tjtelan (approved):

.github/workflows/publish.yml
```diff
@@ -1,8 +1,10 @@
 name: Publish
 
 on:
-  push:
+  workflow_run:
```
sehz, line 4: had some issue with this.  push might be more reliable

digikata: This isn't tested yet due to other issues in the CI flow - so we'll have to see if this usage works or not.  Unfortunately,  leaving a push to master trigger also doesn't seem to work because publish seems to start too early (but that also might be getting obscured by some other of other flow failures).

PR #3685, opened 2023-11-10
ci: Misc fixes to failing tests in CI

tjtelan:
Fix paths and small details that was causing k8 and cli smoke tests to fail


digikata (approved):
Nice!

PR #3694, opened 2023-11-13
feat: publish PackageSet on Fluvio Publish

EstebanBorai:
Publishes PackageSets as part of bumping Fluvio version.

### DIFF
--- a/actions/publish-pkgset.sh
+++ b/actions/publish-pkgset.sh
@@ -1,0 +1,32 @@
+#!/usr/bin/env bash
+
+curl -v -X "POST" "https://hub.infinyon.cloud/hub/v1/fvm/pkgset" \
+     -H "Authorization: $BPKG_TOKEN" \
+     -H 'Content-Type: application/json; charset=utf-8' \
+     --data-binary @- << EOF
+{
+  "artifacts": [
+    {
+      "name": "fluvio",
+      "version": "$FLUVIO_VERSION"
+    },
+    {
+      "name": "fluvio-cloud",
+      "version": "$FLUVIO_CLOUD_VERSION"
+    },
+    {
+      "name": "fluvio-run",
+      "version": "$FLUVIO_VERSION"
+    },
+    {
+      "name": "cdk",
+      "version": "$FLUVIO_VERSION"
+    },
+    {
+      "name": "smdk",
+      "version": "$FLUVIO_VERSION"
+    }
+  ],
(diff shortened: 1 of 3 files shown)


digikata (approved):
reviewed a pair prog

actions/publish-pkgset.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+
```
tjtelan, line 2: It seems like we can hide some problems if not all of the env vars are set, but this script doesn't fail

```suggestion
set -u

```

makefiles/release.mk
```diff
@@ -213,10 +213,18 @@ bump-fluvio: install-fluvio-package
 bump-fluvio-stable: CHANNEL_TAG=stable
 bump-fluvio-stable: VERSION=$(REPO_VERSION)
 bump-fluvio-stable: bump-fluvio
+	export PKGSET_NAME=$(VERSION)
+	export FLUVIO_VERSION=$(VERSION)
+	export FLUVIO_CLOUD_VERSION="0.2.15"
```
tjtelan, line 218: If we think we'll realistically be maintaining static metadata values in the makefiles, we should move it up to the top of the file

makefiles/release.mk
```diff
@@ -213,10 +213,18 @@ bump-fluvio: install-fluvio-package
...
+	./actions/publish-pkgset.sh
 
 bump-fluvio-latest: CHANNEL_TAG=latest
 bump-fluvio-latest: VERSION=$(subst -$(GIT_COMMIT_SHA),+$(GIT_COMMIT_SHA),$(DEV_VERSION_TAG))
 bump-fluvio-latest: bump-fluvio
+	export PKGSET_NAME=$(VERSION)
+	export FLUVIO_VERSION=$(VERSION)
+	export FLUVIO_CLOUD_VERSION="0.2.15"
```
tjtelan, line 226: Same with previous, but if these values might differ, we should use different vars for the makefile.

Unless this is just a short-term fix, I would suggest having a default values set on the service-side. Otherwise we have deep coupling between the oss and cloud projects that we shouldn't have.

tjtelan (approved):
I have some reservations, but I don't feel strongly enough to block on it

makefiles/release.mk
EstebanBorai: Totally! This is temporal approach while we solve in the service side.

PR #3695, opened 2023-11-13
fix: skip already published instead of aborting

EstebanBorai:
Fallback to `true` in case the process of publishing a Package fails, this allows us to
bypass existing package re-publish error which aborts the publish process.


digikata (approved):

tjtelan (approved):

digikata:
https://github.com/infinyon/fluvio-packages/issues/49 - can remove the true after impl of this (post-release)

PR #3701, opened 2023-11-14
ci: Troubleshooting release CI

tjtelan:
- Copy debug output from https://github.com/infinyon/fluvio/pull/3700
- Testing makefile change for variable interpolation


tjtelan:
No longer needed

PR #3702, opened 2023-11-14
fix: properly pass owner references in conversions

galibey:
There were two places where `owner_references` were lost after conversions:
1. From `ObjectMeta` to `K8MetaItem` if `resource_version` is empty.
2. `InputObjectMeta` = `ObjectMeta::into()` does not pass `owner_references`, instead need to use `ObjectMeta::as_input()`


nacardin (approved):
Works in my test!

PR #3703, opened 2023-11-14
chore: release.mk, quotes cause pkgset publish to fail

digikata:


EstebanBorai (approved):

tjtelan (approved):
I'm not surprised that this would be the cause

PR #3705, opened 2023-11-15
chore: ci, switch ci->publish->cd_dev to repo_dispatch

digikata:
due to merge queue update, get really about sequencing this sets up a fix for publish running on failed CI.yml and CD_Dev running for a failed publish

### DIFF
--- a/.github/workflows/cd_dev.yaml
+++ b/.github/workflows/cd_dev.yaml
@@ -17,6 +17,15 @@ on:
     workflows: [Publish]
     branches: [master]
     types: [completed]
+
+  repository_dispatch:
+    types: [ publish-success ]
+
+  workflow_call:
+    inputs:
+      alt_version:
+        required: false
+
   workflow_dispatch:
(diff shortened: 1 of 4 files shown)


EstebanBorai (approved):

.github/workflows/cd_dev.yaml
```diff
@@ -13,10 +13,10 @@ on:
 #    branches: [master]
 #  pull_request:
 #    branches: [master]
-  workflow_run:
-    workflows: [Publish]
-    branches: [master]
-    types: [completed]
+
+  repository_dispatch:
+    types: [ publish-success ]
```
sehz, line 18: push event is more reliable since it will only be triggered on commit

digikata: I will try that next if this does not work, but I think we tried other methods because publish/cd_dev were misfiring - but maybe that was because of the merge_queue config

PR #3717, opened 2023-11-16
ci: fvm_basic.bats, smoke test adjustment

digikata:
Mixing artifact state w/ dev-hub status instead of hub objects was causing failures.


digikata:
Should probably move fvm smoke tests vs hub to hourly (or slower) tests

PR #3719, opened 2023-11-17
feat: support cluster upgrade/check/status for local cluster

galibey:
Cluster operations `upgrade`, `check`, and `status` should be aware of the installation type of the cluster.

Additionally, setting the profile during the installation is put before any actual actions to the system, so in case of failure, the delete command will clean up accordingly.


digikata (approved):
LGTM

PR #3720, opened 2023-11-17
chore: switch cli tests to local fluvio cluster

galibey:
1. Switch all cli tests to local fluvio cluster.
2. Run `fvm` test in a separate job. It needs neither fluvio cli nor cluster, so there were needless test matrix runs, plus, there were conflicts with `fvm` installed in previous steps.


digikata (approved):
Can you copy the old k8 cluster-delete test to cluster-delete-k8.bats and keep it around?  It might be good as a manual diagnostic tool - I sometimes run the bats scripts manually as testing.  Or we might later excercise it in an hourly type test.

EstebanBorai (approved):

galibey:
> Can you copy the old k8 cluster-delete test to cluster-delete-k8.bats and keep it around? It might be good as a manual diagnostic tool - I sometimes run the bats scripts manually as testing. Or we might later excercise it in an hourly type test.
> 
> LGTM, heres to faster and more reliable tests w/o random k8 install failures.

But there was not much of the testing. It did cluster deletion (I keep it), and the rest is just running `kubectl` to print if the entities exist, but there were no assertions, so there were no checks and no outputs.

digikata:
Ok, maybe it's more useful as a manual diagnostic script, and not so much for automated test.

Issue #3722, opened 2023-11-17
build.rs scripts are duplicated with small differences, consolidate them


al-yisun:
**Additional context**
This is mostly so I have an issue to tag in a PR.

PR #3723, opened 2023-11-17
chore: consolidate most `build.rs`s into a separate crate (#3722)

al-yisun:
Fixes https://github.com/infinyon/fluvio/issues/3722


crates/cdk/src/generate.rs
```diff
@@ -42,8 +44,7 @@ impl GenerateCmd {
             ..Default::default()
         };
 
-        let fluvio_dependency_version_hash =
-            format!("fluvio-cargo-dependency-hash={}", env!("GIT_HASH"));
+        let fluvio_dependency_version_hash = format!("fluvio-cargo-dependency-hash={}", GIT_HASH);
```
al-yisun, line 47: Ah, these lines should probably be changed to e.g.

```
let fluvio_dependency_version_hash = format!("fluvio-cargo-dependency-hash={GIT_HASH}");
```

crates/fluvio-cluster/build.rs
```diff
@@ -1,28 +1,21 @@
...
-    if let Ok(verpath) = std::fs::canonicalize("../../VERSION") {
-        if verpath.exists() {
-            println!("cargo:rerun-if-changed=../../VERSION");
+    println!("cargo:rerun-if-changed=build.rs");
+    let paths = [
+        "../../k8-util/helm/pkg_sys/fluvio-chart-sys.tgz",
+        "../../k8-util/helm/pkg_app/fluvio-chart-app.tgz",
+    ];
+    for path in paths {
+        // Otherwise when these files are missing, we always rebuild.
```
al-yisun, line 11: It's possible the build should be erroring when these files don't exist, I'm not sure. But locally for me this and dependent crates were just always rebuilding.

crates/fluvio-cluster/build.rs
```diff
@@ -1,28 +1,21 @@
...
     }
-    println!("cargo:rerun-if-changed=build.rs");
-    println!("cargo:rerun-if-changed=../../k8-util/helm/pkg_sys/fluvio-chart-sys.tgz");
-    println!("cargo:rerun-if-changed=../../k8-util/helm/pkg_app/fluvio-chart-app.tgz");
 
-    let _uname_output = Command::new("make")
+    let _ = Command::new("make")
         .arg("install")
         .output()
         .expect("package ");
```
al-yisun, line 20: Is this string unfinished?

crates/fluvio-version/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+/// The `git rev-parse HEAD` of the current build of Fluvio.
+pub const GIT_HASH: &str = env!("GIT_HASH");
+
+/// The current version of Fluvio.
+pub const VERSION: &str = include_str!("../../../VERSION");
```
al-yisun, line 5: This could be a `semver::Version`, enough dependents use it like that.

al-yisun: Also it's notable here, `include_str` adds a "build if changed" style condition on included files.

EstebanBorai: This would be nice!

It also would be nice to have a ENV VAR to solve workspace dir so we don't have to write 3 times the `../` bit, but AFAIK its not available.

crates/fluvio-version/src/lib.rs
```diff
@@ -0,0 +1,9 @@
...
+
+pub mod built_info {
```
EstebanBorai, line 7: Perhaps nit, but I would use `build` over `built_info` given that it is more conventional?

For instance, in version context, you tend to use `build` for the build number/identifier.

al-yisun: Agreed the module name is a little obtuse, the name is lifted directly from `fluvio-socket/src/lib.rs`, and comes from the fact that it is the info output by the `built` library. But I will change it.

crates/fluvio-version/src/lib.rs
al-yisun: I may have misremembered wrt `semver::Version`, it seems like only `fluvio-test-util` parses it and some of the others sort of need it anyway as a literal string.

I agree about the env var, although this line should refuse to compile if the file moves, and now we only have one reference in the rust code to this path instead of 15 or so.

crates/fluvio-cluster/build.rs
al-yisun: @EstebanBorai ^

al-yisun:
@EstebanBorai PTAL

PR #3724, opened 2023-11-17
feat: hide `fvm self install` from and help introduce `fvm self update`

EstebanBorai:
- Hide `fvm self install` as its only used for install script
- Provide update details via `fvm self update`
- Bumps patch version number to `0.1.1`

## Demo

fvm self update


digikata (approved):
LGTM

Issue #3673, opened 2023-11-08
Support constants for versions in `fluvio` proc macro


galibey:
Currently, `min_version` and `max_version` attributes of `fluvio` proc macro support only int literals. 
We need to add support for constants, e.g.:
```rust
const PRODUCER_TRANSFORMATION_API: i16 = 8;
...
#[fluvio(min_version = PRODUCER_TRANSFORMATION_API)]
pub smartmodules: Vec<SmartModuleInvocation>,

``` 
and for `const fn`:
```rust
const fn min_transformation_api() -> i16 {
    8
}
...
#[fluvio(min_version = "min_transformation_api")]
pub smartmodules: Vec<SmartModuleInvocation>,

``` 

It would also be nice to support any function, not only `const` but, if it requires a lot of work, it can be done afterward.

vrrashkov:
I would like to take on this, when the PR gets merged if possible. 

Also I noticed that in [fluvio-spu-schema/server/smartmodule.rs](https://github.com/infinyon/fluvio/blob/03806c2a4805d8df78e0084d9ba852b1e4b30409/crates/fluvio-spu-schema/src/server/smartmodule.rs#L88) it's also using constants but even if it was not there is no implementation for the `enum` attributes for `min_version`,`max_version` etc. there is only for `tag` you can see it [here](https://github.com/infinyon/fluvio/blob/03806c2a4805d8df78e0084d9ba852b1e4b30409/crates/fluvio-protocol-derive/src/ast/enum.rs#L75). Just wanted to mention it so this can be changed also.

galibey:
@vrrashkov yes, sure, go ahead. 

The PR is ready to merge, let's wait for release(planned for today) and then do the rebase, and hopefully, all checks will be green.

ajhunyady:
@galibey, this issue says it's ready to merge after the release (back in November). I assume it should be merged?

galibey:
> @galibey, this issue says it's ready to merge after the release (back in November). I assume it should be merged?

That was regarding another PR. The PR that corresponds to this issue is https://github.com/infinyon/fluvio/pull/3725. Closed as stale.

PR #3725, opened 2023-11-18
Support constants and functions for fluvio and some improvements to the derive macro

vrrashkov:
Hi, these are the new changes to the `fluvio-protocol-derive` and how it works. #3673

When passing values there are couple of ways you can do that: 

```rust
// Constant as a path
const TEST: i16 = 1;
#[fluvio(min_version = TEST)]

// Constant as a literal
const TEST: i16 = 1;
#[fluvio(min_version = "TEST")]

// Function as a literal
fn test() -> i16 { 1 }
#[fluvio(min_version = "test()")]

// Int
#[fluvio(min_version = 1)]
```

There are couple of improvements to how attributes are handled, which is with a new macro `parse_attributes!` this way is way more cleaner and will be much faster to add/update when needed.  I documented briefly some of the new functions/macros. 

Updated the enumerators and added min_version/max_version to it as previously they were not handled, again with the new macro. 

This PR also includes the changes from the 1st PR #3577. 

Here are some notes about couple of things i noticed:

The `ast/container.rs`: I have not updated the code here with the new macro, as I saw that currently those properties are not used anywhere in the project, so I kept it as is.

For the `ast/enum.rs`: Even though I added min_version/max_version, actually currently there is no logic behind that at all as previously it didn’t exist as well so not sure what should happen there, what kind of logic do we need to implement.

Connected to the above, because there is no logic for the enum variables, you can notice that inside [fluvio-spu-schema/server/smartmodule.rs](https://github.com/infinyon/fluvio/blob/0a4e03406db93c651fc3dbd1e0657a57281ab302/crates/fluvio-spu-schema/src/server/smartmodule.rs#L88) there are constants that are passed to `min_version/max_version` but they actually do not do anything as there is no logic for the enum properties. 

The [fluvio-spu-schema/produce/request.rs](https://github.com/infinyon/fluvio/blob/0a4e03406db93c651fc3dbd1e0657a57281ab302/crates/fluvio-spu-schema/src/produce/request.rs#L44C5-L44C5) has this field set with **min_version**:

```rust
#[fluvio(min_version = PRODUCER_TRANSFORMATION_API)]
pub smartmodules: Vec<SmartModuleInvocation>,
```

This is a non existent variable so in the 1st PR I changed this to use the real one, but because this was reverted and trough further research I noticed that this is actually not even used at all because we can see that in the file there is [Encoder/Decoder](https://github.com/infinyon/fluvio/blob/0a4e03406db93c651fc3dbd1e0657a57281ab302/crates/fluvio-spu-schema/src/produce/request.rs#L81) that are being implemented because of that it doesn’t even use the min_version in the proc-macro, so I decided to leave it like it is for now just so the changes are only for the `fluvio-protocol-derive` crate.


crates/fluvio-protocol-derive/src/util.rs
```diff
@@ -1,75 +1,256 @@
...
-                None
+use crate::ast::prop::PropAttrsType;
+
+/// Parses the specified attributes from a `syn::Attribute` iterator.
+/// # Arguments
+/// * `attrs` - `&[Attribute]` iterator
+/// * `attr_ident` - The path ident to search for
+/// * `meta` - a return variable that will give us the parse_nested_meta of each element
+/// * `field` - nested path to search for
+/// * `opt` - mutable variable to save data into
```
galibey, line 15: The `mutable variable to save data into` comment contradicts with the only occurrence `if $opt.is_none()`.
Handling duplicates of parsed attributes should be part of `parse_attributes` operation.

crates/fluvio-protocol-derive/src/util.rs
```diff
@@ -1,75 +1,256 @@
...
-                        }
+pub(crate) use parse_attributes;
+
+pub fn get_expr_value_from_meta(meta: &ParseNestedMeta) -> syn::Result<PropAttrsType> {
+    let (expr, attr_span, attr_name) = parse_attributes_data(meta);
+    get_expr_value(&attr_name, &expr, attr_span)
+}
+pub fn get_expr_value<'a>(
+    attr_name: &'a str,
+    field: &'a Option<Expr>,
```
galibey, line 75: Here and below `&'a Option<Expr>` should be replaced by `Option<&'a Expr>`. This will save from unnecessary cloning of `Exp`.

crates/fluvio-protocol-derive/src/util.rs
```diff
@@ -1,75 +1,256 @@
...
-    })
+// This piece of code can be included in the macro but there are some issues
+// For example if we return this from the parse_attributes! macro with closure function
+// It works but there is an issue with clippy giving a "warning: try not to call a closure in the expression where it is declared"
+// Here is how this can be fixed if used in macro: https://stackoverflow.com/questions/76989679/how-to-call-a-closure-from-a-macro
+// So to do this in a less work around way we use a separate function like so:
+// We return $meta variable like shown here: https://github.com/rust-lang/rust-clippy/issues/1553
+pub fn parse_attributes_data(meta: &ParseNestedMeta) -> (Option<syn::Expr>, Span, String) {
+    // we can safely unwarp as this is already checked from the parse_attributes macro
+    let ident = meta.path.get_ident().unwrap();
```
galibey, line 52: Even if it's true for now, it might change later. This is a public function in `utils` crate, so `unwrap` and ignoring `Err` results for functions should not be introduced.

crates/fluvio-protocol-derive/src/util.rs
```diff
@@ -1,75 +1,256 @@
...
-                            return Some(named_meta);
-                        }
-                    }
-                    Meta::Path(path) => {
-                        if path.is_ident(name) {
-                            return Some(named_meta);
-                        }
+pub(crate) use parse_attributes;
+
+pub fn get_expr_value_from_meta(meta: &ParseNestedMeta) -> syn::Result<PropAttrsType> {
```
galibey, line 69: Misleading naming `get_expr_value_from_meta` but return type is `PropAttrsType`

crates/fluvio-protocol-derive/src/util.rs
```diff
@@ -1,75 +1,256 @@
...
+            span,
+            format!(
+                "Expected {attr_name} attribute to be a Lit/Path/Unary: `{attr_name} = \"...\"`"
+            ),
+        )),
     }
+}
 
-    None
+fn parse_int_expr<'a>(attr_name: &'a str, expr: &'a Expr, span: Span) -> syn::Result<i16> {
```
galibey, line 131: Can we make it generic over different ints `i32`, `i8`, etc.?

crates/fluvio-protocol-derive/src/util.rs
```diff
@@ -1,75 +1,256 @@
...
+    if let Some(value_expr) = value {
+        parse_int_expr(attr_name, value_expr, span)
+    } else {
+        Err(syn::Error::new(
+            span,
+            format!("Expected {attr_name} to be valid Expr: `{attr_name} = \"...\"`"),
+        ))
+    }
+}
+pub fn get_lit_int<'a>(
```
galibey, line 162: We have two functions here `get_lit_int` and `get_lit_int_value` that do the same thing. Can we make one from them that can cover both cases?

crates/fluvio-protocol-derive/src/util.rs
```diff
@@ -1,75 +1,256 @@
...
-        None
+    })
+}
+pub(crate) fn find_name_value_from_meta_list(
+    meta_list: &MetaList,
+    attr_name: &str,
+) -> Option<MetaNameValue> {
+    for item in meta_list
+        .parse_args_with(Punctuated::<syn::Meta, Token![,]>::parse_terminated)
+        .unwrap_or_else(|err| panic!("Invalid #[new] attribute: {}", err))
```
galibey, line 224: We should use error with the proper span instead of panicking

crates/fluvio-protocol-derive/src/util.rs
```diff
@@ -1,75 +1,256 @@
...
+
+/// find name value with str value
+pub(crate) fn find_string_name_value(meta: &Meta, attr_name: &str) -> Option<LitStr> {
+    let meta_name_value = find_name_value_from_meta(meta, attr_name)?;
+    get_lit_str(attr_name, &Some(meta_name_value.value), Span::call_site())
+        .ok()
+        .cloned()
+}
+
+pub(crate) fn find_int_name_value(meta: &Meta, attr_name: &str) -> Option<u64> {
```
galibey, line 250: Misleading function name

galibey (changes requested):
Thanks again for your contribution @vrrashkov 

Apart from the comments below, since this is a new feature we need to also write unit-tests that cover the new functionality.

crates/fluvio-protocol-derive/src/ast/enum.rs
```diff
@@ -60,28 +64,31 @@ pub(crate) struct EnumProp {
     pub tag: Option<String>,
     pub discriminant: Option<DiscrimantExpr>,
     pub kind: FieldKind,
+    pub min_version: Option<PropAttrsType>,
```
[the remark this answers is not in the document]
vrrashkov: I added them because I saw that here it's trying to use them but they were not existent. Should I just remove them even though this Enum is trying to use them?

https://github.com/infinyon/fluvio/blob/0a4e03406db93c651fc3dbd1e0657a57281ab302/crates/fluvio-spu-schema/src/server/smartmodule.rs#L88-L104

crates/fluvio-protocol-derive/src/ast/prop.rs
```diff
@@ -151,76 +143,128 @@ impl UnnamedProp {
         }
     }
 }
-
-pub fn validate_versions(min: i16, max: Option<i16>, field: Option<&str>) -> Option<String> {
```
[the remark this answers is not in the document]
vrrashkov: This place was used to validate on compile time. Because min_version and max_version support constants and functions they can only be evaluated at run time, so we can't actually use the validate_versions as it was before, that's why I removed it. 

There is validation for the runtime which check the token stream directly. 

https://github.com/infinyon/fluvio/blob/4b9b32db623f39af18a0a8f95c75d66277d1e786/crates/fluvio-protocol-derive/src/ast/prop.rs#L47-L63

Even if we change that to throw errors with to_compile_error in the token stream it will not work again because the TokenStream that loads the values for min/max versions won't be evaluated.

crates/fluvio-protocol-derive/src/util.rs
vrrashkov: Ah yes.. the option is actually just used as a way to check if the field is already used. Updated the docs.

crates/fluvio-protocol-derive/src/util.rs
vrrashkov: I kept them both but improved a bit on them. Also it was incorrect to call it `get_lit_int_value`. I renamed it to `get_expr_int_value` because even though we are returning Int value we need to check for the Unary operator as well and returns just the int value, that's also why I kept them both because the get_lit_int only does checks for int without the Unary and returns Lit.

crates/fluvio-protocol-derive/src/util.rs
vrrashkov: I changed it so it returns **Result** instead of **Option** on most of the places and improved on the error handling.

vrrashkov:
@galibey  I added tests for the TokenStream how it should handle the different cases when passing values with the new type **PropAttrsType**. Let me know if you think more tests are needed for that functionality.

crates/fluvio-protocol-derive/src/ast/enum.rs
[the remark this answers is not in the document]
vrrashkov: It won't change anything if I remove them, it will still work and still won't do anything of course :D

sehz:
Can you add to tests here?  https://github.com/infinyon/fluvio/blob/master/crates/fluvio-protocol/tests/version.rs

vrrashkov:
> Can you add to tests here? https://github.com/infinyon/fluvio/blob/master/crates/fluvio-protocol/tests/version.rs

Added couple of more cases covering the different types that can be passed and updated `test_encode_version`/`test_decode_version` accordingly. 

Also found an issue with the Unary operator so fixed that as well and updated the test.

crates/fluvio-protocol-derive/src/ast/enum.rs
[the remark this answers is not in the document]
galibey: Yes, so there are a couple of options here: 
1. make it work
2. leave as is and create a follow-up issue
3. remove wrong occurrences from the existing code, add compilation warning and create a follow-up ticket
Pick any.
Adding the unworking code to the unworking code is not a proper solution :)

crates/fluvio-protocol-derive/src/util.rs
galibey: The comment is still about the old version.

And, the result of `value.parse()` below is ignored.

crates/fluvio-protocol-derive/src/ast/prop.rs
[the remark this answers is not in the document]
galibey: Ok, let's do it in runtime then. If users mess up with versions, at least they will receive a proper message during encoding/decoding, it's better than nothing.

crates/fluvio-protocol-derive/src/ast/prop.rs
```diff
@@ -151,76 +143,322 @@ impl UnnamedProp {
...
+/// ```
+///
+/// None has a default Int value of 0 which is set in prop_attrs_type_value
+#[derive(Default, Clone)]
+pub enum PropAttrsType {
+    Lit(Ident),
+    Fn(Ident),
+    Int(i16),
+    #[default]
+    None,
```
galibey, line 222: Why do we need `None` variant? Property attributes can't be without type or with undefined type, right?

crates/fluvio-protocol-derive/src/ast/prop.rs
```diff
@@ -151,76 +143,322 @@ impl UnnamedProp {
...
+/// ````
+/// To set a specific type you can do this:
+/// ````ignore
+/// let ident_type = Ident::new("u8", Span::call_site());
+/// let func_value = prop_attrs_type_value(prop_attr_type, Some(&ident_type))
+/// ````
+///
+pub fn prop_attrs_type_value(
+    attrs_type: Option<&PropAttrsType>,
+    ident_type: Option<&Ident>,
```
galibey, line 166: This is always `None` except one occurrence which is in tests.

crates/fluvio-protocol-derive/src/ast/prop.rs
```diff
@@ -151,76 +143,322 @@ impl UnnamedProp {
...
+) -> TokenStream {
+    if let Some(attr) = attrs_type {
+        match &attr {
+            PropAttrsType::Lit(data) => parse_quote!(#data),
+            PropAttrsType::Fn(data) => TokenStream::from_str(&format!("{}()", data)).unwrap(),
```
galibey, line 171:
```suggestion
            PropAttrsType::Fn(data) => parse_quote!(#data()),
```

crates/fluvio-protocol-derive/src/ast/prop.rs
```diff
@@ -151,76 +143,322 @@ impl UnnamedProp {
...
+            PropAttrsType::Int(data) => {
+                if let Some(itype) = ident_type {
+                    TokenStream::from_str(&format!("{}_{}", data, itype)).unwrap()
```
galibey, line 174: Need to find a way without a need for `unwrap` or properly propagate the error

crates/fluvio-protocol-derive/src/ast/prop.rs
```diff
@@ -151,76 +143,322 @@ impl UnnamedProp {
...
+    pub min_version: Option<PropAttrsType>,
     /// Optional max version.
     /// The field won't be decoded from the buffer if it has a larger version than what is specified here.
-    /// Note: `None` is encoded as "-1" so it's i16.
-    pub max_version: Option<i16>,
+    pub max_version: Option<PropAttrsType>,
     /// Sets this value to the field when it isn't present in the buffer.
     /// Example: `#[fluvio(default = "-1")]`
     pub default_value: Option<String>,
+    pub ignorable: Option<bool>,
```
galibey, line 235: If this is not used in code generation, we should not add it

crates/fluvio-protocol-derive/src/util.rs
```diff
@@ -1,75 +1,273 @@
...
+    <T as FromStr>::Err: std::fmt::Display,
+{
+    if let Some(value_expr) = value {
+        parse_int_expr::<T>(attr_name, value_expr, span)
+    } else {
+        Err(syn::Error::new(
+            span,
+            format!("Expected {attr_name} to be valid Expr: `{attr_name} = \"...\"`"),
+        ))
+    }
```
galibey, line 179:
```suggestion
    let value_expr = value.ok_or_else(|| {
        syn::Error::new(
            span,
            format!("Expected {attr_name} to be valid Expr: `{attr_name} = \"...\"`"),
        )
    })?;
    parse_int_expr::<T>(attr_name, value_expr, span)
```

crates/fluvio-protocol-derive/src/util.rs
```diff
@@ -1,75 +1,273 @@
...
+
+pub(crate) fn find_int_from_meta(meta: &Meta, attr_name: &str) -> syn::Result<u64> {
+    match find_expr_from_meta(meta, attr_name) {
+        Ok(ref value) => {
+            let value = get_lit_int(attr_name, Some(value), meta.span())?;
+
+            value.base10_parse::<u64>()
+        }
+        Err(err) => Err(err),
     }
```
galibey, line 272:
```suggestion
    let value = find_expr_from_meta(meta, attr_name)?;
    let value = get_lit_int(attr_name, Some(&value), meta.span())?;
    value.base10_parse::<u64>()
```

crates/fluvio-protocol-derive/src/util.rs
```diff
@@ -1,75 +1,273 @@
...
-pub(crate) fn find_string_name_value(version_meta: &Meta, attr_name: &str) -> Option<LitStr> {
-    if let Some(attr) = find_name_attribute(version_meta, attr_name) {
-        match &attr.lit {
-            Lit::Str(val) => Some(val.clone()),
-            _ => unimplemented!(),
+pub(crate) fn find_string_from_meta<'a>(meta: &'a Meta, attr_name: &'a str) -> syn::Result<LitStr> {
+    match find_expr_from_meta(meta, attr_name) {
+        Ok(ref value) => {
+            let lit_str = get_lit_str(attr_name, Some(value), meta.span())?;
+            Ok(lit_str.clone())
```
galibey, line 258:
```suggestion
    let value = find_expr_from_meta(meta, attr_name)?;
    let lit_str = get_lit_str(attr_name, Some(&value), meta.span())?;
    Ok(lit_str.clone())
```

crates/fluvio-protocol-derive/src/util.rs
```diff
@@ -1,75 +1,273 @@
...
-                        if path.is_ident(name) {
-                            return Some(named_meta);
-                        }
+pub(crate) use parse_attributes;
+
+pub fn get_attr_type_from_meta(meta: &ParseNestedMeta) -> syn::Result<PropAttrsType> {
+    let (expr, attr_span, attr_name) = parse_attributes_data(meta)?;
+    get_expr_value(&attr_name, expr.as_ref(), attr_span)
+}
+pub fn get_expr_value<'a>(
```
galibey, line 84: Misleading naming: we are not getting expression value here, we detecting `PropAttrsType`

crates/fluvio-protocol-derive/src/util.rs
```diff
@@ -1,75 +1,273 @@
...
+    attr_name: &'a str,
+    field: Option<&'a Expr>,
```
galibey, line 86: In almost all cases we pass here `Some()` and in only one case `None`. Maybe we could consider changing this value to `field: &'a Expr` and do the unwrapping in that one case, wdyt? This will make the code of this function more clear.

crates/fluvio-protocol-derive/src/util.rs
```diff
@@ -1,75 +1,273 @@
...
+                    format!("Expected {attr_name} attribute to be a LitStr or LitInt: `{attr_name} = \"...\"`"),
+                ))
             }
         }
+        Some(Expr::Unary(ExprUnary { expr, .. })) => {
+            // When passing -1 as a value it is returned as type Unary
+            // So to handle that we are checking if it's Unary Lit and continue as usual
+            // If needed this can be extended to handle the Unary operators
+            // But it doesn't seem that is necessary currently
+            Ok(PropAttrsType::Int(-parse_int_expr(attr_name, expr, span)?))
```
galibey, line 117: `parse_int_expr` is also handling `unary` expr inside. We should decide on which level we want to deal with it and do it only once.

crates/fluvio-protocol-derive/src/util.rs
vrrashkov: Removed the function because we don't need it anymore, because we have no more `Option<&Expr>` so no need for a separate function to unwrap, using directly the `parse_int_expr`.

crates/fluvio-protocol-derive/src/ast/prop.rs
vrrashkov: I thought it might be cleaner to do it like that so all the defaults are set in the `prop_attrs_type_value`, but it might be cleaner to do it without the `None` and just set the default as `Int(0)` so will change it with the new changes.

crates/fluvio-protocol-derive/src/util.rs
vrrashkov: They do both handle Unary, but in this because we are in the Unary case the type of `expr` is `Expr::Lit` so just using the `parse_int_expr` to get the Literal from there and convert it so we save some lines of code and set it as a negative because we know this is Unary. Of course in the future if needed we can use the other operators as well and the logic might change but i don't see a reason to do it for now as the only place Unary is used if -1 is being passed.

crates/fluvio-protocol-derive/src/ast/prop.rs
vrrashkov: I am not really sure if this is an issue to be unwrap here as this is just a TokenStream and if anything happens it will be in compile time. Maybe we can just use `expect` insteead of `unwrap`?

crates/fluvio-protocol-derive/src/ast/enum.rs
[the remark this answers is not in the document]
vrrashkov: I guess I lied a bit here :D Because of the new changes we are checking if attribute actually exists, so currently because there are some places where there are enums with min_version/max_version then if we remove it from here it throws similar to this error "`attribute min_version unrecognized`". So maybe lets leave it like that at the moment and with a new issue add the functionality for the enum versioning?

crates/fluvio-protocol-derive/src/ast/prop.rs
sehz: https://docs.rs/syn/latest/syn/struct.Error.html

vrrashkov: I don't think we can convert LexError with compile_error. From what I understand LexError already emits any errors but if we just want to remove the `unwrap` maybe we can do it like this.

`TokenStream::from_str(&format!("{}_i16", data)).map_err(|_| TokenStream::new())`

This way we don't have unwrap and still will be having all the errors.

crates/fluvio-protocol-derive/src/ast/prop.rs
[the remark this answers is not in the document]
vrrashkov: We already have the runtime checks, so I haven't added anything more to it.

crates/fluvio-protocol-derive/src/ast/prop.rs
vrrashkov: Currently yes but we might need this at any point in the future as currently the default is i16, but if there is even a single value that needs to be different than that, we would need this as I can see that there are places in the the proc-macro that are specifically set as `u8` even though they don't affect this specific case for now.

crates/fluvio-protocol-derive/src/ast/prop.rs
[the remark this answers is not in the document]
galibey: What runtime check do you mean? We used to check that min is not greater than max, min is not negative, etc. We don't have it now, don't we?

crates/fluvio-protocol-derive/src/ast/prop.rs
galibey: In any case, this is hard to read. You are passing `ident_type: Option<&Ident>` which itself is only relevant for only one variant of `PropAttrsType`. 
And if there is no `attrs_type` (None), why do we return `0_i16`

crates/fluvio-protocol-derive/src/ast/prop.rs
```diff
@@ -151,76 +143,324 @@ impl UnnamedProp {
...
+/// fn test() -> i16 { 1 }
+/// #[fluvio(min_version = "test()")]
+/// ```
+///
+/// ```ignore
+/// // Int
+/// #[fluvio(min_version = 1)]
+/// ```
+///
+/// None has a default Int value of 0 which is set in prop_attrs_type_value
```
galibey, line 214:
```suggestion

```

crates/fluvio-protocol-derive/src/ast/prop.rs
```diff
@@ -151,76 +143,324 @@ impl UnnamedProp {
...
+#[derive(Clone)]
+pub enum PropAttrsType {
+    Lit(Ident),
+    Fn(Ident),
+    Int(i16),
+}
+impl Default for PropAttrsType {
```
galibey, line 221: Why do we need to impl `Default` for attr type? If the type is not identifiable, we should just throw an error.

crates/fluvio-protocol-derive/src/ast/enum.rs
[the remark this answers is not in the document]
galibey: Ok

crates/fluvio-protocol-derive/src/ast/prop.rs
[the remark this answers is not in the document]
vrrashkov: There are checks here that check the version.

https://github.com/infinyon/fluvio/blob/4b9b32db623f39af18a0a8f95c75d66277d1e786/crates/fluvio-protocol-derive/src/ast/prop.rs#L47-L63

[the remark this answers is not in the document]
vrrashkov: Previously the `validate_versions `only did checks and threw error of something was returned. We cannot throw syn::Error in this case. We need to have a code generation in our case in order to have the values evaluated before using them and then we throw an error. Not sure how you want the error to be thrown but currently the old code as I see prints the data with tracing is that not enough?

crates/fluvio-protocol-derive/src/ast/prop.rs
vrrashkov: I was just following how it worked previously that it defaults to 0 so decided to set a default value as well.

crates/fluvio-protocol-derive/src/ast/prop.rs
vrrashkov: I was just following how it worked previously that it defaults to 0 so I set it to return 0 always.

crates/fluvio-protocol-derive/src/ast/prop.rs
galibey: If you mean that previously default value of `min_version: i16` is 0, that's fine. The equivalent of that for new `min_version: Option<PropAttrsType>` is None. This doesn't mean that the default needs for `PropAttrsType ` type.

crates/fluvio-protocol-derive/src/ast/prop.rs
vrrashkov: About how the type is parsed, let me know if you like this better. I added a new case IntParse, so if it is needed we can use that now instead of passing ident_type.

https://github.com/infinyon/fluvio/blob/f6d081c81a29c03d495aef8fde94a412f9c30f03/crates/fluvio-protocol-derive/src/ast/prop.rs#L164-L173

crates/fluvio-protocol-derive/src/ast/prop.rs
vrrashkov: I removed the default with the new changes.

crates/fluvio-protocol-derive/src/ast/prop.rs
galibey: Now clippy check fails. Just remove it for now because it is not used. We will add it later once it's needed.

vrrashkov: Yep I saw that will delete for now but actually I was thinking we can even make it just Int(data, typ) without the IntParse, so we will always be parsing the type and this should be even better.

crates/fluvio-protocol-derive/src/ast/prop.rs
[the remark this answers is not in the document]
vrrashkov: I made it so we make checks with the generated code and panic if it finds an issue. I used the original checks and errors as text. Does that look okay?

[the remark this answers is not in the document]
galibey: I found a way of doing it in compile time for literals, consts and const fn. `assert` is stable in const context already. We can leverage it. See the example here https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a58bfe08d0bb0ea8f579e488d124b728

We can drop support for non-const function for now.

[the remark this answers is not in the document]
vrrashkov: That is a pretty cool hack doing it like this: `const _: ()` . Currently I replaced all the panics with the const asserts and pretty much didn't change anything else, because even if you try to use a normal function it will still throw an error so no need to handle that differently than how it works.

crates/fluvio-protocol-derive/src/ast/prop.rs
```diff
@@ -16,6 +21,38 @@ pub(crate) struct UnnamedProp {
...
 
+pub fn validate_versions_tokens(
+    min_prop: Option<&PropAttrsType>,
+    max_props: Option<&PropAttrsType>,
+    field: Option<&str>,
+) -> TokenStream {
+    let min = prop_attrs_type_value(min_prop);
+    let max = prop_attrs_type_value(max_props);
+
+    match (max_props, field) {
```
galibey, line 32: I don't understand the conditions here: we check `#min>#max` twice, we check `#min < 0` only if max_props is Some(_), why? We use `concat!` but don't do actual concatenation.

crates/fluvio-protocol-derive/Cargo.toml
```diff
@@ -13,12 +13,8 @@ proc-macro = true
...
 proc-macro2 = { workspace = true }
 quote = { workspace = true }
 tracing = { workspace = true }
-
-[dependencies.syn]
-version = "1.0.0"
-features = ["full"]
-
-
+anyhow = { workspace = true }
```
galibey, line 20: We use `anyhow` only once in tests, and the usage is no different from `core::result::Result`, so we can drop it.

PR #3727, opened 2023-11-20
chore: update local clusters setup on CD and Hourly

galibey:
1. `CD_DEV` and `CD_DEV_Mac` - do not install k8s for local clusters, explicitly set installation type so it won't be broken when defaults change.
2. `Hourly` - added local cluster to longevity tests, set it to every 2 hours instead of every hour to compensate for increased CI time.

### DIFF
--- a/.github/workflows/cd_dev.yaml
+++ b/.github/workflows/cd_dev.yaml
@@ -77,20 +77,12 @@ jobs:
         with:
           ref: ${{ needs.setup.pub_sha }}
       - name: Set up K3d for Ubuntu
-        if: ${{ matrix.os == 'ubuntu-latest' }}
+        if: ${{ matrix.cluster_type == 'k8' }}
         run: |
           curl -s https://raw.githubusercontent.com/rancher/k3d/main/install.sh | TAG=${{ env.K3D_VERSION }} bash
       - name: Set up K8 for ubuntu(kind)
-        if: ${{ matrix.os == 'ubuntu-latest' }}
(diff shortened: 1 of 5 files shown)


.github/workflows/cd_dev.yaml
```diff
@@ -103,13 +95,8 @@ jobs:
         timeout-minutes: 3
         if: ${{ matrix.cluster_type == 'k8' }}
         run: |
-          if [[ ${{ matrix.os }} == 'macos-12' ]]; then
-              export PROXY="--proxy-addr  127.0.0.1"
-            else
-              export PROXY=""
-          fi
```
digikata, line 110: Does the macos-12 k8 not need the PROXY variable anymore? The other removal of PROXY below sets proxy always, so it's not clear if macos uniquely needs a PROXY or if all os combos will use it on some tests?

digikata (approved):
Approved but up to you if the macos-12 PROXY question needs further updates

.github/workflows/cd_dev.yaml
galibey: In `CD_DEV` we no longer have `macos-12` as os, only `ubuntu-latest`, so I removed dead code.

in both `CD_DEV` and `CD_DEV_Mac` I removed the condition with `PROXY` and added the actual proxy argument to the start command on `CD_DEV_Mac`.

`--proxy-addr  127.0.0.1` is always needed for macos, and is not needed for ubuntu.

tjtelan (approved):
Changes look fine. Curious if things will just work

PR #3728, opened 2023-11-20
feat: restrict `fluvio cluster` commands based on current profile

matheus-consoli:
Check the current profile to restrict which commands are available.

Restrict which commands are available based on the current profile metadata.

For old profiles with unknown installation types, we try to infer which installation it is.

Closes https://github.com/infinyon/fluvio/issues/3619


galibey (changes requested):
I know it is tempting to put `Cloud` to installation type enum, but it is not correct. It is a different dimension. Strictly speaking, Cloud is K8 installation type(although it may change in the future).  For `Cloud/No cloud` property there should be a separate metadata item in the config. Besides the logical part, the code will be cleaner: no need to for those `unreachable` statements.

PR #3730, opened 2023-11-20
feat: do not allow version switching when process are running

EstebanBorai:
When either `fluvio` or `fluvio-run` are running, switching versions via FVM,
will break pipes thus causing `fluvio` process to be killed on every execution.

To avoid this, check if theres any processes on these two binaries running
before switching versions with FVM.

## Demo

https://github.com/infinyon/fluvio/assets/34756077/6917d96a-b5ae-4d5b-8844-afedacb67261

### DIFF
--- a/crates/fluvio-version-manager/src/common/version_directory.rs
+++ b/crates/fluvio-version-manager/src/common/version_directory.rs
@@ -2,6 +2,7 @@ use std::fs::{read_dir, copy, create_dir_all};
 use std::path::PathBuf;
 
 use anyhow::Result;
+use sysinfo::{System, SystemExt};
 
 use fluvio_hub_util::fvm::Channel;
 
@@ -59,6 +60,8 @@ impl VersionDirectory {
 
     /// Sets this version as the active Fluvio Version
     pub fn set_active(&self) -> Result<()> {
+        Self::check_running_processes()?;
+
         // Verify `~/.fluvio/bin` exists and create it if it doesn't
         let fluvio_bin_dir = fluvio_binaries_path()?;
 
@@ -116,6 +119,20 @@ impl VersionDirectory {
         }
 
         Ok((manifests, active_version))
+    }
+
+    pub fn check_running_processes() -> Result<()> {
+        let system = System::new_all();
+        let fluvio_running = system.processes_by_exact_name("fluvio").count();
+        let fluvio_run_running = system.processes_by_exact_name("fluvio-run").count();
+
+        if fluvio_running > 0 || fluvio_run_running > 0 {
+            return Err(anyhow::anyhow!(
+                "Cannot uninstall fluvio while fluvio or fluvio-run are running"
+            ));
+        }
+
+        Ok(())
     }
 }
 
--- a/crates/fluvio-version-manager/src/common/version_directory.rs
+++ b/crates/fluvio-version-manager/src/common/version_directory.rs
@@ -1,7 +1,7 @@
 use std::fs::{read_dir, copy, create_dir_all};
 use std::path::PathBuf;
 
-use anyhow::Result;
+use anyhow::{Result, bail};
 use sysinfo::{System, SystemExt};
 
 use fluvio_hub_util::fvm::Channel;
@@ -127,9 +127,7 @@ impl VersionDirectory {
         let fluvio_run_running = system.processes_by_exact_name("fluvio-run").count();
 
         if fluvio_running > 0 || fluvio_run_running > 0 {
-            return Err(anyhow::anyhow!(
-                "Cannot uninstall fluvio while fluvio or fluvio-run are running"
-            ));
+            bail!("Cannot switch versions while `fluvio` or `fluvio-run` are running");
         }
 
         Ok(())
--- a/tests/cli/fvm_smoke_tests/fvm_basic.bats
+++ b/tests/cli/fvm_smoke_tests/fvm_basic.bats
@@ -818,7 +818,7 @@ setup_file() {
     source ~/.fvm/env
 
     # Starts Fluvio Cluster
-    fluvio cluster delete || true
+    fluvio cluster delete --local || true
 
     # Installs Fluvio Stable
     run bash -c 'fvm install'
--- a/tests/cli/fvm_smoke_tests/fvm_basic.bats
+++ b/tests/cli/fvm_smoke_tests/fvm_basic.bats
@@ -817,6 +817,9 @@ setup_file() {
     # Sets `fvm` in the PATH using the "env" file included in the installation
     source ~/.fvm/env
 
+    # Starts Fluvio Cluster
+    fluvio cluster delete || true
+
     # Installs Fluvio Stable
     run bash -c 'fvm install'
     assert_success
--- a/tests/cli/fvm_smoke_tests/fvm_basic.bats
+++ b/tests/cli/fvm_smoke_tests/fvm_basic.bats
@@ -809,3 +809,41 @@ setup_file() {
     rm -rf $FLUVIO_HOME_DIR
     assert_success
 }
+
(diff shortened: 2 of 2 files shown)


crates/fluvio-version-manager/src/common/version_directory.rs
```diff
@@ -117,6 +120,20 @@ impl VersionDirectory {
 
         Ok((manifests, active_version))
     }
+
+    pub fn check_running_processes() -> Result<()> {
```
digikata, line 124: local bats test?

EstebanBorai: I have been testing locally but tests hang for this scenario, will push and see how it goes on CI!

EstebanBorai: Okay! Disowning the `run` process and adding some sleep did the trick!

```
run bash -c 'fluvio cluster start --local' &
```

tests/cli/fvm_smoke_tests/fvm_basic.bats
```diff
@@ -809,3 +809,41 @@ setup_file() {
...
+    source ~/.fvm/env
+
+    # Installs Fluvio Stable
+    run bash -c 'fvm install'
+    assert_success
+
+    # Starts Fluvio Cluster
+    run bash -c 'fluvio cluster start --local' &
+    sleep 10
+    assert_success
```
EstebanBorai, line 827: The `&` at the end is required so the process is spawned and the test continues in parallel. Then add some sleep because we are not hooked to the process anymore and starting the cluster takes some time.

galibey: Why can't we wait until the process 'fluvio cluster start --local' finishes? It takes a couple of seconds to start and then this process ends. If it hangs we should find out why. Relying on `sleep 10` is very flaky, especially on slow CI execution sometimes.

crates/fluvio-version-manager/src/common/version_directory.rs
```diff
@@ -59,6 +60,8 @@ impl VersionDirectory {
 
     /// Sets this version as the active Fluvio Version
     pub fn set_active(&self) -> Result<()> {
+        Self::check_running_processes()?;
```
galibey, line 63: We should have done this check earlier, not after we spent some time and network traffic downloading the artifacts.

tests/cli/fvm_smoke_tests/fvm_basic.bats
EstebanBorai: Yeah it hangen in my local but I can try here!

I agree with the `sleep` comment!

EstebanBorai:
> Preventing switching the fluvio version will be limiting. We should allow multiple versions of fluvio on the same machine.

Hi @sehz!

The goal would be aborting the update when a Local Fluvio Cluster is actively running.
We could add a `--force` option to kill those processes and then proceed.

But allowing to replace the binaries, while Fluvio is running, will continue to produce the
error. So the goal is to avoid breaking Fluvio process by protecting the binaries while in use.

tests/cli/fvm_smoke_tests/fvm_basic.bats
digikata: Another way to do it instead of a long sleep is to loop on an operation to verify response (e.g. `fluvio topic list`) with a short sleep

EstebanBorai: Thats a good one! Im not sure if you already saw the cluster deletion error in CI?

Im investigating on it, because for this test suite we shouldnt have any Clusters running, but turns out, that it seems to be a Cluster and when I run:

```bash
fluvio cluster delete
```

It is failing with _Failed to delete Helm Charts_.

Perhaps the loop you mention could be helpful, to "force" deletion. But I want to make sure no cluster is present first.

tests/cli/fvm_smoke_tests/fvm_basic.bats
```diff
@@ -809,3 +809,43 @@ setup_file() {
...
+    run bash -c 'fluvio cluster start --local'
+    assert_success
+
+    # Attempts to switch version
+    run bash -c 'fvm install latest'
+    assert_line --index 6 "Error: Cannot switch versions while \`fluvio\` or \`fluvio-run\` are running"
+    assert_failure
+
+    # Deletes Fluvio Cluster
+    run bash -c 'fluvio cluster delete'
```
digikata, line 837: I'd also make this '--local'

tests/cli/fvm_smoke_tests/fvm_basic.bats
```diff
@@ -809,3 +809,43 @@ setup_file() {
...
+
+@test "Disallows switching versions when there are active proccesses" {
+    run bash -c '$FVM_BIN self install'
+    assert_success
+
+    # Sets `fvm` in the PATH using the "env" file included in the installation
+    source ~/.fvm/env
+
+    # Starts Fluvio Cluster
+    fluvio cluster delete --local || true
```
galibey, line 821: 1. `fluvio cluster delete --local` is no longer valid in the latest.
2. Why do you stop the cluster in the test if you didn't start it? It might be started on CI earlier, and other tests might expect it. This is not the case here, but, in general, we should make tests isolated and order-independent. 
3. If the fluvio cluster is running at that place, need to investigate and find why the cluster is running in that place. Recently, I fixed similar magical cluster runs in this commit https://github.com/infinyon/fluvio/commit/ad005a5e53a647a1ce3d3e59ac98f94b7a26bb73. Maybe there is another place.

tests/cli/fvm_smoke_tests/fvm_basic.bats
galibey: The `delete` no longer accepts such parameter. It reads everything it needs from the profile. If it tries to delete the k8s cluster when you expect it deletes local, it means somewhere was an attempt to start k8s cluster and the current profile is not the one you expect

galibey (changes requested):
Need to figure out what is going on with Fluvio cluster(where it started, where it is deleted) or run/kill `fluvio-run` binary manually that will be enough to test

crates/fluvio-version-manager/src/common/version_directory.rs
digikata: I second it that it would be nicer to run the check earlier in the process, either first or maybe after the pkgset metadata download (and before the actual downloads).

EstebanBorai:
Surpassed by #3735

EstebanBorai:
> The delete no longer accepts such parameter. It reads everything it needs from the profile. If it tries to delete the k8s cluster when you expect it deletes local, it means somewhere was an attempt to start k8s cluster and the current profile is not the one you expect

@galibey thanks for sharing!

tests/cli/fvm_smoke_tests/fvm_basic.bats
EstebanBorai: > Why do you stop the cluster in the test if you didn't start it? It might be started on CI earlier, and other tests might expect it. This is not the case here, but, in general, we should make tests isolated and order-independent.

Yeah I thought no cluster was present, but turns out that the process detection for `fluvio` and/or `fluvio-run` determined there was one of those two processes.

> If the fluvio cluster is running at that place, need to investigate and find why the cluster is running in that place. Recently, I fixed similar magical cluster runs in this commit https://github.com/infinyon/fluvio/commit/ad005a5e53a647a1ce3d3e59ac98f94b7a26bb73. Maybe there is another place.

Agree! I also saw your PR: https://github.com/infinyon/fluvio/pull/3720 and I thought there wasn't any clusters on FVM test because of this. But perhaps I still have to dig in on this.

PR #3732, opened 2023-11-21
chore: explicitly set cluster type in CI/CD

galibey:
Set explicitly `k8` cluster type where it is required. This is preparation for changing local cluster type as default on cluster start.


digikata (approved):
LGTM

PR #3752, opened 2023-11-29
fix: topic reducer should wait for partitions sync

galibey:
`TopicReducer` can start acting while partitions are not loaded to the store yet. We are waiting for the initial change of partitions store before we process the new topic state.


sehz (approved):
LGTM

PR #3755, opened 2023-11-30
feat: added minimal support for docker inst type

galibey:
Added new installation type `Docker` with minimal support: it is only possible to add it manually and it won't allow to do commands that do not support it. It prevents errors if fluvio cli runs against clusters in docker.


digikata (approved):
LGTM

PR #3767, opened 2023-12-08
chore: use `ctx` ci for installations

EstebanBorai:
Provide the `ctx` query param for CI related installations


digikata (approved):
LGTM

PR #3775, opened 2023-12-12
feat: setup fluvio gh action

EstebanBorai:
Setups a GitHub Action to install Fluvio using the installation script and FVM.

## Usage

```diff
-     - name: Install stable CLI and start Fluvio cluster
-        run: |
-          curl -fsS https://hub.infinyon.cloud/install/install.sh?ctx=ci | bash
-          echo "~/.fluvio/bin" >> $GITHUB_PATH
+      - name: Setup Fluvio
+        uses: infinyon/fluvio/.github/actions/setup-fluvio
```

> Open Question: Perhaps we can also host the action file on root dir, which would result in a slightly change from: `infinyon/fluvio/.github/actions/setup-fluvio`  to: `infinyon/fluvio/actions/setup-fluvio`

## Example

The following is an example from tho repository fork, where the option `version` is supported and defaults to `stable`.

Screenshot 2023-12-12 at 4 02 15 PM

Refer to workflow run here: https://github.com/infinyon/fluvio/actions/runs/7185354341/job/19570442003?pr=3776#step:5:1

### DIFF
--- a/.github/actions/setup-fluvio/action.yml
+++ b/.github/actions/setup-fluvio/action.yml
@@ -11,6 +11,7 @@ runs:
   steps:
     - name: Check Platform Compatibility
       if: runner.os != 'Linux' && runner.os != 'macOS'
+      shell: bash
       run: |
         echo "Current platform $env:RUNNER_OS, is not supported by Fluvio CLI Installer."
         echo "Please use a Linux or macOS runner."
@@ -18,6 +19,7 @@ runs:
         exit 1
 
     - name: Install Fluvio
+      shell: bash
       run: |
         curl -fsS https://hub.infinyon.cloud/install/install.sh | FLUVIO_VERSION=${{ inputs.version }} bash
         echo "$HOME/.fluvio/bin" >> $GITHUB_PATH
--- a/.github/actions/setup-fluvio/action.yml
+++ b/.github/actions/setup-fluvio/action.yml
@@ -13,6 +13,8 @@ runs:
       if: runner.os != 'Linux' && runner.os != 'macOS'
       run: |
         echo "Current platform $env:RUNNER_OS, is not supported by Fluvio CLI Installer."
+        echo "Please use a Linux or macOS runner."
+        echo "For further details on Fluvio Installation, please visit https://fluvio.io/docs/install/"
         exit 1
 
     - name: Install Fluvio
--- a/.github/actions/setup-fluvio/action.yml
+++ b/.github/actions/setup-fluvio/action.yml
@@ -13,10 +13,9 @@ runs:
       if: runner.os != 'Linux' && runner.os != 'macOS'
       shell: bash
       run: |
-        echo "Current platform $env:RUNNER_OS, is not supported by Fluvio CLI Installer."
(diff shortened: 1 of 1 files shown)


.github/actions/setup-fluvio/action.yml
```diff
@@ -0,0 +1,25 @@
...
+      run: |
+        echo "Current platform $env:RUNNER_OS, is not supported by Fluvio CLI Installer."
+        echo "Please use a Linux or macOS runner."
+        echo "For further details on Fluvio Installation, please visit https://fluvio.io/docs/install/"
+        exit 1
+
+    - name: Install Fluvio
+      shell: bash
+      run: |
+        curl -fsS https://hub.infinyon.cloud/install/install.sh | FLUVIO_VERSION=${{ inputs.version }} bash
```
digikata, line 24: The url needs ctx-ci

.github/actions/setup-fluvio/action.yml
```diff
@@ -0,0 +1,25 @@
...
+    default: "stable"
+
+runs:
+  using: "composite"
+  steps:
+    - name: Check Platform Compatibility
+      if: runner.os != 'Linux' && runner.os != 'macOS'
+      shell: bash
+      run: |
+        echo "Current platform $env:RUNNER_OS, is not supported by Fluvio CLI Installer."
```
digikata, line 16: I suggest making this a warning instead of a failing step.

As a warning, always report success for the step and let the install fail in the next step.  This way the log gets the warning to help diagnosis, but we don't end up with the step unintentionally blocking if a different runner.os string comes along.

digikata (changes requested):
Looks pretty reasonable, minor changes to make.

.github/actions/setup-fluvio/action.yml
EstebanBorai: Sounds good!

.github/actions/setup-fluvio/action.yml
EstebanBorai: Here its added: [3f9d78c](https://github.com/infinyon/fluvio/pull/3775/commits/3f9d78c8f55043ce3350ad890c6acdb3a9519fa2)! Thanks!

EstebanBorai:
Added a warning message based on: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-a-warning-message

EstebanBorai:
Rerunning jobs here: https://github.com/infinyon/fluvio/pull/3776 to ensure changes don't break!

.github/actions/setup-fluvio/action.yml
EstebanBorai: Please check this for details: https://github.com/infinyon/fluvio/pull/3775/commits/52a56563a403875de684c393376b9fd737c40847

digikata: I think to make it successful, the run script needs to end in `exit 0` or `true`

digikata: nvm I see it just echos

digikata (approved):
LGTM, thanks for the quick update!

PR #3779, opened 2023-12-14
chore: use `setup-fluvio` action

EstebanBorai:
Uses `setup-fluvio` instead of in-situ cURL call to install Fluvio.


sehz (approved):
LGTM.
Nice!

Issue #3787, opened 2023-12-19
[Bug]: fluvio-socket v0.14.6 doesn't compile in Docker unless git is installed. 


marvin-hansen:
**What happened**

I am trying to compile fluvio in Docker while building  a release container. 

However, the compile in Docker fails UNLESS git is installed. 

**FIX / Workaround:** Install git in the build container. 

See relevant Dockerfile at the bottom. 

**Expected behavior**

I would expect fluvio to compile in Docker without a hard 
dependency on git. I don't think I ever had to install git in a build container to compile any other published crate.

The error originates from the Build script in the fluvio-socket crate that uses a git checkout. 

What puzzles me is that the build script in question is pinned to HEAD of the remote repo instead of a specific commit hash, 
which makes the entire build non-hermetic and non-deterministic. This is problematic because you never know
 with which commit fluvio was actually build so this really is a problem that only waits to bite you one day. 

The way I see the situation, this might be relatively easy to fix with a proper versioning of dependencies 
instead of using git and pinning to head. 

However, if a proper solution is infeasible for any reason, please update the fluvio-socket  build script 
in such a way that it tests whether git is installed and if not, issues a meaningful error message
that makes it clear that git must be installed for the build to succeed. This would have saved me a bit of time
tracing down the root cause of the build fail. 

**Describe the setup**
- Local install.
- Rust Fluvio: 0.21
- What version of Fluvio are you using? `fluvio version`. 
Fluvio CLI           : 0.11.2
Fluvio CLI Arch      : aarch64-apple-darwin
Fluvio CLI SHA256    : 2db084880a36e9aeed5b28e67f78fa2df3531558e6d8bd9e36a355ca970654ac
Fluvio channel frontend SHA256 : 2db084880a36e9aeed5b28e67f78fa2df3531558e6d8bd9e36a355ca970654ac
Fluvio Platform      : 0.11.2 (local)
Git Commit           : cddc05388b7a1278837adc506430f339f311212d
OS Details           : Darwin 14.2 (kernel 23.2.0)

**How to reproduce it (as minimally and precisely as possible)**
1. Make a minimal fluvion app or copy over a code example. 
2. Copy the Docker file from below
3. Build the app in Docker with the Dockerfile 

The build script that causes the error:

Lines 15 - 21
https://github.com/infinyon/fluvio/blob/master/crates/fluvio-socket/build.rs

**Log output**
```
106.2 error: failed to run custom build command for `fluvio-socket v0.14.6`
[...]

PR #3789, opened 2023-12-19
build: don't require git to build fluvio

morenol:
Attempt to fix: https://github.com/infinyon/fluvio/issues/3787

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -342,7 +342,7 @@ dependencies = [
  "futures-lite 2.0.1",
  "parking",
  "polling 3.3.1",
- "rustix 0.38.25",
+ "rustix 0.38.28",
  "slab",
  "tracing",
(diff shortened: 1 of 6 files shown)


sehz (commented):
this is not actually needed.  it is left over while back.  probably can remove this entire file

morenol:
> this is not actually needed. it is left over while back. probably can remove this entire file

But we also have that in `fluvio` crate:

https://github.com/infinyon/fluvio/blob/682930c10480a4065d25507c7a64220638e9886b/crates/fluvio/build.rs#L17

sehz:
should get rid of too.  all published crate should not have ref to git.   We moved that to non published crate while back.

sehz (approved):

EstebanBorai (approved):

PR #3792, opened 2023-12-21
feat:rename `transform` and `transforms-file` args

### LINKED ISSUE #3773: Usability Issue: move `--tranforms-file` to `-t --tranforms` and remove inline option.
### Problem

In both `fluvio` and `smdk` we have 2 options to test transformations:

```
     --transforms-file <TRANSFORMS_FILE>  (Optional) File path to transformation specification
  -t, --transform <TRANSFORM>              (Optional) Pass transformation specification as JSON formatted string. E.g. smdk test --text '{}'
                                           --transform='{"uses":"infinyon/jolt@0.1.0","with":{"spec":"[{\"operation\":\"default\",\"spec\":{\"source\":\"te...
```

I found myself using the wrong keyword multiple times. That coupled with the error message, leads into confusion and base use experience.

```
$ fluvio consume quotes -B -t ./transforms.yaml
Invalid argument: unable to parse `transform` argument: expected value at line 1 column 1
```

### Solution

In both `smdk` and `fluvio` change transforms flag to use a file:

```
  -t, --transform <TRANSFORM_FILE>          (Optional) File path to transformation specification
```

Remove inline variant:

 <s> -t, --transform <TRANSFORM>              (Optional) Pass transformation specification as JSON formatted string. E.g. smdk test --text '{}'
                                           --transform='{"uses":"infinyon/jolt@0.1.0","with":{"spec":"[{\"operation\":\"default\",\"spec\":{\"source\":\"te...</s>

galibey:
In the following commands: `smdk test`, `fluvio sm test`, `fluvio produce`, `fluvio consume`
Argument `--transform` renamed to `--transforms_line` and `--transforms_file` to `--transforms` with short version `-t`.

Old names will also be supported (but hidden from help) for at least one version to support current CLI tests for cross-version combinations. 

Fixes #3773

### DIFF
--- a/crates/fluvio-cli-common/src/smartmodule.rs
+++ b/crates/fluvio-cli-common/src/smartmodule.rs
@@ -55,18 +55,18 @@ pub struct BaseTestCmd {
         long= "params",
         value_parser=parse_key_val,
         num_args = 1,
-        conflicts_with_all = ["transforms_file", "transform"]
+        conflicts_with_all = ["transforms", "transforms_line"]
     )]
     pub params: Vec<(String, String)>,
 
     /// (Optional) File path to transformation specification.
-    #[arg(long, group = "TestSmartModule")]
-    pub transforms_file: Option<PathBuf>,
+    #[arg(short, long, group = "TestSmartModule", alias = "transforms_file")]
+    pub transforms: Option<PathBuf>,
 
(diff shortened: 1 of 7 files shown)


digikata (approved):
LGTM

Issue #3798, opened 2023-12-27
CD Fails on Fluvio Installer


EstebanBorai:
When attempting to check version.

https://github.com/infinyon/fluvio/actions/runs/7341014988/job/19987980066#step:6:18

PR #3811, opened 2024-01-12
feat(cd): `fluvio_version` output

EstebanBorai:
Provides an output for the `setup-fluvio` action which holds the installed Fluvio version.
Then uses such output in the CD workflow.


digikata:
This should address https://github.com/infinyon/fluvio/issues/3798 which is a CI blocker for fluvio release

EstebanBorai:
Surpassed by #3812

PR #3852, opened 2024-02-04
chore: exit when there is no unzip cmd

fraidev:
I installed fluvio on my two machines this weekend with: 

```bash
curl -fsS https://hub.infinyon.cloud/install/install.sh | bash
```

Worked well on my MacBook, but the installation crashed on my Debian which didn't have `unzip` command installed. 

I am not sure if this is the correct installation bash script.


EstebanBorai:
Hi @fraidev! Thanks a lot for opening this PR!

Yeah, our install script relies on having `unzip` installed in your machine, this is a known dependency and we plan to move from that approach in the future.

The install script you used is the correct one.

EstebanBorai (approved):

digikata (approved):
lgtm

PR #3856, opened 2024-02-07
rfc: add offset management proposal

digikata:


rfc/offset-management.md
```diff
@@ -0,0 +1,224 @@
+## Scope
+
+Define a minimal offset management functionality for integration into fluvio.
+
+The initial implementation eg would default for a store of "memory", with a retention of 24 hours, and a count of 100 consumer groups of single consumers. Later expansion could allow for more durable storage (for example, internal partition), higher consumer group counts, multiple member consumer groups, and longer retention times.
```
ajhunyady, line 5: @digikata - I suggest we remove the default to `memory`, as the whole point of offset management is to make them durable. Also I don't think that a solution that does not recover from durable storage is viable.

rfc/offset-management.md
```diff
@@ -0,0 +1,224 @@
...
+
+Use Case 1:
```
ajhunyady, line 7: #### Use Case 1:

rfc/offset-management.md
```diff
@@ -0,0 +1,224 @@
...
+An simple offset retention kind and time should be settable for a given duration and a maximum number of consumers.
+The offsets are managed manually with client side API calls.
+
+This can be used internally for more automatic offset management, but also where the user may need more complex consumption patterns. e.g. consume multiple records at a time before committing an offset.
+
+Use Case 2:
```
ajhunyady, line 13: #### Use Case 2:

ajhunyady (changes requested):
Other than that, it looks good to me.

rfc/offset-management.md
sehz:
```suggestion
Provide a simple minimum way to save and recover offset for the consumer.   We are not considering consumer group use case as part of this RFC
```

rfc/offset-management.md
```diff
@@ -0,0 +1,224 @@
+## Scope
+
+Define a minimal offset management functionality for integration into fluvio.
+
+The initial implementation eg would default for a store of "memory", with a retention of 24 hours, and a count of 100 consumer groups of single consumers. Later expansion could allow for more durable storage (for example, internal partition), higher consumer group counts, multiple member consumer groups, and longer retention times.
+
+Use Case 1:
+An simple offset retention kind and time should be settable for a given duration and a maximum number of consumers.
```
sehz, line 6:
```suggestion
Offset mgt facility would provide way to set TTL for offset lifetime
```

ajhunyady (approved):

rfc/offset-management.md
```diff
@@ -0,0 +1,220 @@
+## Scope
+
+Define a minimal offset management functionality for integration into fluvio.
+
+Use Case 1:
+An simple offset retention kind and time should be settable for a given duration and a maximum number of consumers.
+The offsets are managed manually with client side API calls.
```
sehz, line 7:
```suggestion
Offsets should be managed as default.   The API for manual control would be provided.
```

rfc/offset-management.md
```diff
@@ -0,0 +1,220 @@
...
+
+Use Case 2:
+Automatic offset commit for single member consumer groups and a default pattern of
```
sehz, line 10: Remove this since this implies consumer group

rfc/offset-management.md
```diff
@@ -0,0 +1,220 @@
...
+
+    // synchronously flush for shutdown (or none if intentionally ending processing)
+    stream.offset_flush().await?;
+
+    Ok(())
+}
+```
+
+### Use Case 2: Automatic Offset API
+
```
sehz, line 56: this should go first since automatic should be default

sehz (changes requested):
Couple of nits

ajhunyady:
@digikata it doesn't look like formatting for `use case1` and `use case 2` came trough.

rfc/offset-management.md
digikata: Goog suggestion, updated!

sehz (approved):
LGTM.  Nice update

PR #3861, opened 2024-02-12
Exploratory impl of disconnecting on topic delete

jdafont:
Here's the beginning of an implementation to disconnect consumers when a topic is deleted. Of course, I'll need to add tests. Looking for feedback to get this PR up to standards :-)

### DIFF
--- a/crates/fluvio-spu/src/replication/leader/replica_state.rs
+++ b/crates/fluvio-spu/src/replication/leader/replica_state.rs
@@ -41,6 +41,8 @@ use super::FollowerNotifier;
 pub type SharedLeaderState<S> = LeaderReplicaState<S>;
 pub type SharedFileLeaderState = LeaderReplicaState<FileReplica>;
 
+pub const CLEANUP_FREQUENCY: usize = 10;
+
 #[derive(Debug)]
 pub struct LeaderReplicaState<S> {
     replica: Replica,
@@ -392,17 +394,18 @@ where
         self.followers.read().await.clone()
     }
 
+    #[cfg(test)]
+    pub fn consumer_offset_publishers(&self) -> Arc<Mutex<Vec<WeakSharedOffsetPublisher>>> {
+        self.consumer_offset_publishers.clone()
+    }
+
     pub async fn register_offset_publisher(&self, offset_publisher: &SharedOffsetPublisher) {
-        let cleanup_frequency = 10;
-        let mut publishers = 
-            self.consumer_offset_publishers
-            .lock()
-            .await;
+        let mut publishers = self.consumer_offset_publishers.lock().await;
 
         // Filter out any dead weak pointers every so often
-        if publishers.len() % cleanup_frequency == 0 {
-            let cleaned_publishers: Vec<WeakSharedOffsetPublisher> = 
-                publishers.iter()
+        if publishers.len() % CLEANUP_FREQUENCY == 0 {
+            let cleaned_publishers: Vec<WeakSharedOffsetPublisher> = publishers
+                .iter()
                 .filter_map(|p| p.upgrade())
                 .map(|p| Arc::downgrade(&p))
                 .collect();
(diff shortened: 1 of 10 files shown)


digikata:
Thanks for the PR @jdafont   it's very much appreciated!  We'll be reviewing this more closely soon.

crates/fluvio-spu/src/replication/leader/replica_state.rs
```diff
@@ -46,6 +50,7 @@ pub struct LeaderReplicaState<S> {
     followers: Arc<RwLock<BTreeMap<SpuId, OffsetInfo>>>,
     status_update: SharedStatusUpdate,
     sm_ctx: Option<SharedSmartModuleContext>,
+    consumer_offset_publishers: Arc<Mutex<Vec<Arc<OffsetPublisher>>>>,
```
sehz, line 53: alias `Arc<OffsetPublisher>` so it's easier to to read

crates/fluvio-spu/src/replication/leader/replica_state.rs
```diff
@@ -384,6 +391,20 @@ where
...
+            .push(offset_publisher.clone());
+    }
+
+    pub async fn signal_topic_deleted(&self) {
+        let offset_publishers = self.consumer_offset_publishers.lock().await;
+        for publisher in offset_publishers.iter() {
+            publisher.update(TOPIC_DELETED);
+        }
+    }
 }
```
sehz, line 408: would need some periodic clean up to ensure stale publishers are removed.

sehz:
This seems a viable approach (with cleanup).  In addition, add integration tests ( they can be found in the `services/public/tests`)

crates/fluvio-spu/src/replication/leader/replica_state.rs
jdafont: Added `SharedOffsetPublisher` and `WeakSharedOffsetPublisher` aliases

crates/fluvio-spu/src/replication/leader/replica_state.rs
jdafont: Downgraded the pointers to be weak pointers (to allow dropping of values) and changed the registration logic to check/filter the list for dead pointers every ~10 registrations. Ideally this means that you're limited to, at most, 10 wasted allocations. Maybe this isn't so great if we expect extreme numbers of consumers since this is just a simple vector, but still O(n) * (time it takes to upgrade/downgrade pointers).

jdafont:
> This seems a viable approach (with cleanup). In addition, add integration tests ( they can be found in the `services/public/tests`)

Went ahead and added a test to check for the receipt of a TopicDeleted error code when the topic delete signal is triggered

crates/fluvio-spu/src/replication/leader/replica_state.rs
```diff
@@ -384,6 +391,39 @@ where
     pub async fn followers_info(&self) -> BTreeMap<SpuId, OffsetInfo> {
         self.followers.read().await.clone()
     }
+
+    pub async fn register_offset_publisher(&self, offset_publisher: &SharedOffsetPublisher) {
+        let cleanup_frequency = 10;
```
sehz, line 396: Can you make this constant?

crates/fluvio-spu/src/services/public/tests/stream_fetch.rs
```diff
@@ -3025,3 +3025,68 @@ async fn read_records(
...
+        debug!("received first message");
+        assert_eq!(response.topic, topic);
+
+        let partition = &response.partition;
+        assert_eq!(partition.error_code, ErrorCode::TopicDeleted);
+    }
+
+    server_end_event.notify();
+    debug!("terminated controller");
+}
```
sehz, line 3092: can you add one more test to ensure cleanup occurred?

sehz (changes requested):
Looks great. 

To finish this, just need one more test to ensure cleanup happens properly

crates/fluvio-spu/src/replication/leader/replica_state.rs
jdafont: Made it const!

sehz (approved):
Nice job.  Thanks for contribution

Issue #3859, opened 2024-02-09
Python client async and admin upgrade


digikata:
Map missing fluvio rust crate async and admin interfaces to python.

Details in Fluvio-client-python repo https://github.com/infinyon/fluvio-client-python/issues/367

urbit-pilled:
I was trying to implement async without changing the fluvio repo upstream with the Clone modifications, but I think this is a good solution @Shylock-Hg. You can takeover the bounty.

digikata:
Hi @Shylock-Hg , not sure if this helps, but in the core rust fluvio crate, when you generate an async stream, (e.g. https://docs.rs/fluvio/latest/fluvio/consumer/struct.MultiplePartitionConsumer.html#method.stream)  the future delivers a stream that doesn't keep a reference to self of the MultiplePartitionConsumer.

drc-infinyon:
Hey @Shylock-Hg are you done with this. Can you claim the bounty if you are please. :)

drc-infinyon:
@Shylock-Hg - This ticket is resolved.

You need to claim the bounty based on this workflow: https://docs.algora.io/bounties/workflow - 

To claim a bounty on Github, navigate to your pull request that solves the bountied issue and add `/claim <ISSUE_NUMBER>` somewhere in the body of the PR.

drc-infinyon:
@Shylock-Hg it is fixed now. I am paying the bounty.

weirdo-github:
I'd like to work on this. I've submitted a PR here: https://github.com/fluvio-community/fluvio/pull/4672

Let me know if there's anything you'd like me to adjust.

PR #3868, opened 2024-02-16
Impl Clone for TopicProducer, ParitionConsumer and MultiplePartitionC…

Shylock-Hg:
Sub job of #3859 
To support pass `&self` to async block by clone.


sehz:
bump crate version

sehz (approved):
LGTM

PR #3869, opened 2024-02-16
Manually impl Clone for PartitionConsumer.

Shylock-Hg:
Derive Clone require type parameter be bound by `Clone` too, here `P` in `Arc<P>`, so refactor it by manually.


digikata (approved):
LGTM, thanks @Shylock-Hg

sehz (approved):
LGTM

PR #3882, opened 2024-02-29
Impl Clone for WatchResponse.

Shylock-Hg:
Required by https://github.com/infinyon/fluvio-client-python/pull/375, to pass inner data to Python environment.


sehz (approved):
LGTM

PR #3895, opened 2024-03-11
fix: improve kubectl access errors

digikata:


crates/fluvio-cluster/src/cli/start/mod.rs
```diff
@@ -107,11 +107,14 @@ pub struct K8Install {
...
-    /// Uses port forwarding for connecting to SC during install
+    /// Uses port forwarding for connecting to SC (only during install)
+    ///
+    /// For connecting to a remote cluster during install and after
+    /// --proxy-addr <IP or DNS> may work better
     #[arg(long)]
     use_k8_port_forwarding: bool,
 
-    /// Uses port forwarding for connecting to SC during install
+    /// Uses ip for connecting to SC (only during install)
```
nacardin, line 117: This is for the case where this crate is running inside the same k8 cluster as the installation target

digikata: Seems pretty narrow.  Should I just remove the text between parens and hide it by default?

nacardin (approved):

PR #3902, opened 2024-03-14
fix: use specialized error for evicted offset case

### LINKED ISSUE #3898: Report evicted batch back to consumer
If there is an attempt to read records with starting offsets that have been evicted before, the consumer gets a general `ErrorCode::Other` error with `String` message: `Segment not found for start_offset: N`.

We need to introduce a more specialized variant of error that will also bring the first available offset ready to be consumed. This will get all the needed information for the Consumer to decide how to handle the data gap, and, if needed, the Consumer will know from which offset to proceed reading. 

So, instead of returning 
```rust
pub enum ErrorCode {
...
Other(String),
...
}
```
we would want something like this:
```rust
...
EvictedOffset {
    next_available: Offset,
}
...
```

galibey:
Fixes #3898

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2719,7 +2719,7 @@ dependencies = [
 
 [[package]]
 name = "fluvio-protocol"
-version = "0.10.10"
+version = "0.10.11"
 dependencies = [
  "bytes",
  "content_inspector",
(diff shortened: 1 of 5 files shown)


digikata (approved):
LGTM

PR #3905, opened 2024-03-18
feat: added fluvio kv storage crate

galibey:
Added `fluvio-kv-storage` crate that contains traits and types for general KV storage and log-based storage impl. The `Log` type and traits are also defined there.


digikata (approved):
LGTM

PR #3915, opened 2024-03-22
feat: add consumer storage to spu

galibey:
Added SPU private API (between SPUs) and public (except fetch list, will be added later) for Offset Mng.


crates/fluvio-controlplane/src/lib.rs
```diff
@@ -10,3 +10,5 @@ mod alias {
 
     pub type PartitionMetadata<C> = MetadataStoreObject<PartitionSpec, C>;
 }
+
+pub const CONSUMER_STORAGE_TOPIC: &str = "fluvio-consumers";
```
sehz, line 14: `fluvio` prefix is kind of redundant.   Something like: `consumer-offset` ?

crates/fluvio-spu-schema/src/server/api.rs
```diff
@@ -42,6 +43,12 @@ pub enum SpuServerRequest {
     FileStreamFetchRequest(RequestMessage<FileStreamFetchRequest>),
     #[fluvio(tag = 5)]
     UpdateOffsetsRequest(RequestMessage<UpdateOffsetsRequest>),
```
sehz, line 45: Big question is this intended between client and SPU or just between SPU to SPU.   If it is only for SPU to SPU then this should be moved to SPU private protocol

crates/fluvio-spu/src/core/global_context.rs
```diff
@@ -153,6 +156,10 @@ where
     pub(crate) fn metrics(&self) -> Arc<SpuMetrics> {
         self.metrics.clone()
     }
+
+    pub(crate) fn consumers(&self) -> &SharedConsumerStorages {
```
sehz, line 160: renamed this to `consumer_offset` to make it clear

crates/fluvio-spu/src/kv/consumer.rs
```diff
@@ -0,0 +1,295 @@
...
+    ops::AddAssign,
+};
+
+use anyhow::Result;
+use async_rwlock::RwLock;
+use fluvio_kv_storage::KVStorage;
+use fluvio_protocol::{record::ReplicaKey, Encoder, Decoder};
+use fluvio_storage::FileReplica;
+use fluvio_types::PartitionId;
+use tracing::trace;
```
sehz, line 14: can separate out third party and fluvio dep?

crates/fluvio-spu/src/kv/consumer.rs
```diff
@@ -0,0 +1,295 @@
...
+pub(crate) struct SharableConsumerStorage(Arc<RwLock<ConsumerStorage>>);
+
+#[derive(Debug, Hash, Clone, Default, PartialEq, Eq, PartialOrd, Ord, Encoder, Decoder)]
+pub(crate) struct Key {
+    replica_id: ReplicaKey,
+    consumer_id: String,
+}
+
+#[derive(Debug, Default, Clone, PartialEq, Eq, PartialOrd, Ord, Encoder, Decoder)]
+pub(crate) struct Consumer {
```
sehz, line 37: Can you rename this to ConsumerOffset or ConsumerSession.  Not sure what this tracks

crates/fluvio-spu/src/kv/consumer.rs
```diff
@@ -0,0 +1,295 @@
...
+    pub offset: i64,
+    pub ttl: Duration,
```
sehz, line 39: and maybe add more description

crates/fluvio-spu/src/kv/consumer.rs
```diff
@@ -0,0 +1,295 @@
...
+    pub expire_time: TimestampSecs,
+}
+
+#[derive(Debug)]
+pub(crate) struct ConsumerStorage {
```
sehz, line 44: renamed to `ConsumreOffsetStorage` or something like that

crates/fluvio-spu/src/kv/consumer.rs
```diff
@@ -0,0 +1,295 @@
...
+            self.changes_since_flush = Default::default();
+        }
+        Ok(())
+    }
+}
+
+impl KVStorage<Key, Consumer> for ConsumerStorage {
+    async fn get(&self, key: &Key) -> Result<Option<Consumer>> {
+        trace!(?key, "get");
+        self.kv.get(key).await.map(|op| op.filter(is_not_expired))
```
sehz, line 102: why need to filter for expired?  this is assuming that expiration will be evicted automatically

crates/fluvio-spu/src/kv/consumer.rs
```diff
@@ -0,0 +1,295 @@
...
+    }
+
+    async fn delete(&mut self, key: &Key) -> Result<()> {
```
sehz, line 105: we should not be deleting from key. we should add that to offset and rebuild

crates/fluvio-spu/src/kv/consumer.rs
```diff
@@ -0,0 +1,295 @@
...
+        self.changes_since_flush.add_assign(1);
+        self.maybe_flush().await?;
+        result
+    }
+
+    async fn put(&mut self, key: impl Into<Key>, value: impl Into<Consumer>) -> Result<()> {
+        let key = key.into();
+        let value = value.into();
+        trace!(?key, ?value, "put");
+        let result = self.kv.put(key, value).await;
```
sehz, line 117: same comment.  cache should be rebuild from events

sehz (changes requested):
first set of comments

sehz:
It would be better also break this into smaller PR

crates/fluvio-spu/src/kv/consumer.rs
```diff
@@ -0,0 +1,295 @@
...
+const DEFAULT_FLUSH_THRESHOLD: usize = 100;
+
+#[derive(Debug, Default)]
+pub(crate) struct SharedConsumerStorages(Arc<RwLock<HashMap<ReplicaKey, SharableConsumerStorage>>>);
+
+#[derive(Debug, Clone)]
+pub(crate) struct SharableConsumerStorage(Arc<RwLock<ConsumerStorage>>);
+
+#[derive(Debug, Hash, Clone, Default, PartialEq, Eq, PartialOrd, Ord, Encoder, Decoder)]
+pub(crate) struct Key {
```
sehz, line 31: should change this into more meaningful like `ConsumerOffsetKey`.   `Key` is too generic

crates/fluvio-spu/src/kv/consumer.rs
```diff
@@ -0,0 +1,295 @@
...
+impl KVStorage<Key, Consumer> for ConsumerStorage {
+    async fn get(&self, key: &Key) -> Result<Option<Consumer>> {
+        trace!(?key, "get");
+        self.kv.get(key).await.map(|op| op.filter(is_not_expired))
+    }
+
+    async fn delete(&mut self, key: &Key) -> Result<()> {
+        trace!(?key, "delete");
+        let result = self.kv.delete(key).await;
+        self.changes_since_flush.add_assign(1);
```
sehz, line 108: why need to have flush logic? Shouldn't that be built into KV?

crates/fluvio-spu/src/kv/consumer.rs
```diff
@@ -0,0 +1,295 @@
...
+        self.maybe_flush().await?;
+        result
+    }
+
+    async fn put(&mut self, key: impl Into<Key>, value: impl Into<Consumer>) -> Result<()> {
+        let key = key.into();
```
sehz, line 114: same comment.  as delete

crates/fluvio-spu/src/kv/consumer.rs
sehz, line 39: why is TTL needed?  Should that be derived from event timestamp?

crates/fluvio-spu/src/kv/consumer.rs
```diff
@@ -0,0 +1,295 @@
...
+        let value = value.into();
```
sehz, line 115: consumer offset cache must ensure it's not going backward.  so that means, once we have offset value 100, shouldn't accept offset 95.

crates/fluvio-spu/src/kv/consumer.rs
```diff
@@ -0,0 +1,295 @@
...
+    use super::*;
+
+    #[fluvio_future::test]
+    async fn test_expiration_checked() {
+        //given
+        let leader = create_replica("test_expiration_checked").await;
+        let notifier = FollowerNotifier::shared();
+        let mut storage = ConsumerStorage::new(leader.clone(), notifier);
+        let key1 = Key::new(("topic1", 0), "consumer1");
+        let value1 = Consumer::with(1, Duration::from_secs(2), now_timestamp() - 100);
```
sehz, line 213: why do we need TTL?

crates/fluvio-spu/src/kv/consumer.rs
```diff
@@ -0,0 +1,295 @@
...
+            .put(key1.clone(), value1)
+            .await
+            .expect("put record1");
+        storage
+            .put(key2.clone(), value2.clone())
+            .await
+            .expect("put record2");
+
+        //when
+        let record1 = storage.get(&key1).await.expect("record1");
```
sehz, line 226: confused. is this getting record or offset?

crates/fluvio-spu/src/kv/consumer.rs
```diff
@@ -0,0 +1,295 @@
...
+        config::ReplicationConfig, storage::SharableReplicaStorage,
+        control_plane::StatusMessageSink,
+    };
+
+    use super::*;
+
+    #[fluvio_future::test]
+    async fn test_expiration_checked() {
+        //given
+        let leader = create_replica("test_expiration_checked").await;
```
sehz, line 209: change name to something like `offset_replica`

crates/fluvio-spu/src/kv/consumer.rs
```diff
@@ -0,0 +1,295 @@
...
+            .put(key2.clone(), value2.clone())
+            .await
+            .expect("put record2");
+
+        //when
+        let record1 = storage.get(&key1).await.expect("record1");
+        let record2 = storage.get(&key2).await.expect("record2");
+
+        //then
+        leader.remove().await.expect("removed");
```
sehz, line 230: confused on what this test.  probably start with simpler

crates/fluvio-spu/src/kv/consumer.rs
```diff
@@ -0,0 +1,295 @@
...
+        let notifier = FollowerNotifier::shared();
+        let mut storage = ConsumerStorage::with(leader.clone(), notifier, 1);
+        let key1 = Key::new(("topic1", 0), "consumer1");
+
+        //when
+        assert_eq!(leader.hw(), 0);
+        storage.delete(&key1).await.expect("put record1");
+        assert_eq!(leader.hw(), 1);
+        storage.delete(&key1).await.expect("put record1");
+        assert_eq!(leader.hw(), 3);
```
sehz, line 249: this doesn't seem right.  deleting existing key should not add more record

crates/fluvio-spu/src/kv/consumer.rs
```diff
@@ -0,0 +1,295 @@
...
+
+        //when
+        assert_eq!(leader.hw(), 0);
+        storage
+            .put(key1.clone(), value1.clone())
+            .await
+            .expect("put record1");
+        assert_eq!(leader.hw(), 1);
+        storage.put(key1, value1).await.expect("put record1");
+        assert_eq!(leader.hw(), 4);
```
sehz, line 272: why would this be 4?

crates/fluvio-spu/src/replication/leader/kv.rs
```diff
@@ -0,0 +1,364 @@
...
+        ));
+
+        //when
+        kv.put("key1", "value1").await.unwrap();
+        kv.put("key2", "value2").await.unwrap();
+        kv.put("key3", "value3").await.unwrap();
+
+        kv.delete(&"key2".to_string()).await.unwrap();
+
+        drop(kv);
```
sehz, line 180: separate out into simpler test cases.  For example, check if log contains all the values

crates/fluvio-spu/src/replication/leader/kv.rs
```diff
@@ -0,0 +1,364 @@
...
+use tracing::debug;
+
+use super::{LeaderReplicaState, FollowerNotifier};
+
+const RECORDS_SERIALIZATION_VERSION: Version = 0;
+
+pub type LeaderKVStorage<K, V, S> = LogBasedKVStorage<K, V, LeaderReplicaLog<S>>;
+
+#[derive(Debug)]
+pub struct LeaderReplicaLog<S> {
```
sehz, line 23: Not sure I understand purpose of this.  Can you add more description?

crates/fluvio-spu/src/replication/leader/kv.rs
```diff
@@ -0,0 +1,364 @@
...
+
+    #[fluvio_future::test]
+    async fn test_log() {
+        //given
+        let leader = create_replica("test_log").await;
+        let notifier = FollowerNotifier::shared();
+
+        //when
+        let mut log = LeaderReplicaLog::new(leader.clone(), notifier);
+        log.append_batch(vec![b"1".to_vec()]).await.unwrap();
```
sehz, line 213: why need this since basic leader writer exists?

crates/fluvio-spu/src/services/internal/fetch_consumer_handler.rs
```diff
@@ -0,0 +1,63 @@
...
+
+use crate::{
+    core::DefaultSharedGlobalContext, replication::leader::LeaderReplicaState,
+    kv::consumer::Consumer,
+};
+
+use super::fetch_consumer_request::{FetchConsumerRequest, FetchConsumerResponse};
+
+#[instrument(skip(req_msg, ctx))]
+pub async fn handle_fetch_consumer_request(
```
sehz, line 22: pub(crate)

crates/fluvio-spu/src/services/internal/fetch_consumer_request.rs
```diff
@@ -0,0 +1,75 @@
...
+use fluvio_protocol::{Encoder, Decoder};
+
+use fluvio_spu_schema::COMMON_VERSION;
+use fluvio_types::PartitionId;
+
+use super::SPUPeerApiEnum;
+
+#[derive(Decoder, Encoder, Default, Debug)]
+pub struct FetchConsumerRequest {
+    pub topic: String,
```
sehz, line 15: use `ReplicaKey`

sehz (changes requested):
more comments

crates/fluvio-controlplane/src/lib.rs
galibey: Renamed

crates/fluvio-spu-schema/src/server/api.rs
galibey: between client and SPU

crates/fluvio-spu/src/kv/consumer.rs
galibey: Sure, done

crates/fluvio-spu/src/kv/consumer.rs
galibey: Done

crates/fluvio-spu/src/kv/consumer.rs
galibey: What do you mean automatically? There is an eviction mechanism on topic, but we have to support TTL per customer, hence this logic.

crates/fluvio-spu/src/kv/consumer.rs
galibey: Renamed

crates/fluvio-spu/src/kv/consumer.rs
galibey: Flush logic is on KV, but since it is not a cheap operation (memory and latency), the control over when it is used is put on a higher level to give control to the business logic level. For example, for consumers offset we might want to flush with one frequency, but for other entities with different. That's why the inner KV storage does not do it internally.

crates/fluvio-spu/src/kv/consumer.rs
galibey: We need to show this information (TTL column) on CLI by requirements.

crates/fluvio-spu/src/kv/consumer.rs
galibey: This is going to be in another place. On the client side.

crates/fluvio-spu/src/kv/consumer.rs
galibey: It is getting a value from the storage. The type of the value is `ConsumerOffset`, `record1` is a variable name

crates/fluvio-spu/src/kv/consumer.rs
galibey: Renamed

crates/fluvio-spu/src/kv/consumer.rs
galibey: It checks if expiration works as the name suggests.

crates/fluvio-spu/src/kv/consumer.rs
galibey: Delete operation is another record on the underlying topic.

crates/fluvio-spu/src/replication/leader/kv.rs
galibey: List all values is not implemented yet. It will be coming in next PRs.

crates/fluvio-spu/src/services/internal/fetch_consumer_request.rs
galibey: Done

crates/fluvio-spu/src/services/internal/fetch_consumer_handler.rs
galibey: Done

crates/fluvio-spu/src/replication/leader/kv.rs
galibey: Added

crates/fluvio-spu/src/replication/leader/kv.rs
galibey: I don't understand the question, could you please re-phrase?

crates/fluvio-spu/src/kv/consumer.rs
galibey: Added

crates/fluvio-spu-schema/src/server/api.rs
sehz: Go it

crates/fluvio-spu/src/kv/consumer.rs
sehz: fair enough

crates/fluvio-spu/src/kv/consumer.rs
sehz: This should be put into server logic.   Can add client logic as well but ultimately, server should be responsible for data integrity is enforced

crates/fluvio-spu/src/kv/consumer.rs
sehz: Still confused about need to have TTL in the offset value.  It doesn't make sense.   We should provide minimum API, which in case just need incremental offset value.   
In either case, this test case is misleading because there is no testing of TTL anyway.   So either test should case reflect TTL or provide API just with offset so only need to check if offset is persist

crates/fluvio-spu/src/kv/consumer.rs
sehz: the k/v API should make sure cache is consistent.  In this case,  we only have one record.  so deleting no existent key should not add more records.  it should lead to error since the API contract is not satisfied.

crates/fluvio-spu/src/kv/consumer.rs
sehz: It's hard to read this.   Can you ad explicit checkpoint or some kind of configuration.

sehz:
Should separate out SPUSchema related changes into it's PR so can start merging that

crates/fluvio-spu/src/replication/leader/kv.rs
sehz: What's purpose of `LeaderReplicaLog`?  It seems like just wrapper around Leader.

crates/fluvio-spu/src/replication/leader/kv.rs
sehz: Can you change this to something like `LeaderReplicaaLogReader` ?  Still confused why this is necessary since that should be in the KV.

sehz: Why is notifier needed?  Since that would be already done by Leader.

crates/fluvio-spu/src/replication/leader/kv.rs
```diff
@@ -0,0 +1,367 @@
...
+        }
+    }
+}
+
+impl<S: ReplicaStorage + Send + Sync + 'static> Log for LeaderReplicaLog<S> {
+    async fn read_from_end(&self) -> Result<impl futures_util::Stream<Item = Result<Vec<u8>>>> {
+        Ok(ReplicaRevStream::new(self.replica.clone()))
+    }
+
+    async fn append_batch(&mut self, entries: Vec<Vec<u8>>) -> Result<()> {
```
sehz, line 45: Can we crate some abstraction for for key value which I assume this is for.   Just adding bytes seems to be bit raw

crates/fluvio-spu/src/replication/leader/kv.rs
```diff
@@ -0,0 +1,367 @@
...
+        //when
+        kv.put("key1", "value1").await.unwrap();
+        kv.put("key2", "value2").await.unwrap();
+        kv.put("key3", "value3").await.unwrap();
+
+        kv.delete(&"key2".to_string()).await.unwrap();
+
+        drop(kv);
+
+        let mut kv = LeaderKVStorage::<String, String, _>::new(LeaderReplicaLog::new(
```
sehz, line 185: why is using LeaderKvStorage here.  That seems like integration testing.  Or least move to different folder.  Here, we just want to test `LeaderReplicaLog`

crates/fluvio-spu/src/replication/leader/kv.rs
sehz: Why we are testing append batching? that's already done at the Leader.  It seems like this is creating unnecessary abstraction.

crates/fluvio-spu/src/replication/leader/kv.rs
```diff
@@ -0,0 +1,367 @@
...
+        let mut log = LeaderReplicaLog::new(leader.clone(), notifier);
+        log.append_batch(vec![b"1".to_vec()]).await.unwrap();
+        log.append_batch(vec![b"2".to_vec(), b"3".to_vec()])
+            .await
+            .unwrap();
+        log.append_batch(vec![b"4".to_vec()]).await.unwrap();
+        let output: Vec<Result<Vec<u8>>> = log.read_from_end().await.expect("read").collect().await;
+
+        //then
+        leader.remove().await.expect("removed");
```
sehz, line 224: dont' understand this here

crates/fluvio-spu/src/services/internal/update_consumer_offset_handler.rs
```diff
@@ -0,0 +1,71 @@
...
+        consumer_id,
+        offset,
+        ttl,
+        replica_id,
+    } = req_msg.request;
+
+    let consumers_replica_id =
+        ReplicaKey::new(CONSUMER_STORAGE_TOPIC, <PartitionId as Default>::default());
+
+    let error_code = if let Some(ref replica) = ctx.leaders_state().get(&consumers_replica_id).await
```
sehz, line 37: Shouldn't this get offset kv rather than trying to get leaders state?  Why bother creating abstraction if this is not being used?

crates/fluvio-spu/src/services/public/mod.rs
```diff
@@ -145,3 +175,33 @@ impl FluvioService for PublicService {
...
+    };
+    let Some(spu_spec) = ctx.spu_localstore().spec(&spu) else {
+        return Err(ErrorCode::SpuNotFound);
+    };
+    let leader_endpoint = spu_spec.private_endpoint.to_string();
+    debug!(
+        spu,
+        leader_endpoint, "send private request to replica leader"
+    );
+    let mut socket = FluvioSocket::connect(&leader_endpoint)
```
sehz, line 196: this should be using private API to SPU, otherwise this will incur connection latency

crates/fluvio-spu/src/kv/consumer.rs
galibey: Ok, I am going to add this in next PRs. This one is already too big.

crates/fluvio-spu/src/kv/consumer.rs
galibey: Again, TTL is needed to show on CLI. This test does not test TTl, it test that the storage does not return expired values. Expired means `expiration_time` < `now`

crates/fluvio-spu/src/replication/leader/kv.rs
galibey: `KVStorage` operates on some generic `Log` abstraction. This `LeaderReplicaLog` is an implementation of this abstraction for `LeaderReplicaState`

crates/fluvio-spu/src/replication/leader/kv.rs
galibey: `KVStorage` operates on some generic `Log` abstraction. This `LeaderReplicaLog` is an implementation of this abstraction for `LeaderReplicaState`

galibey: We test `append_batch` method on the impl of `Log` trait. There is a code, and there is a test for that code.

crates/fluvio-spu/src/replication/leader/kv.rs
galibey: I think we have enough abstractions here. Not sure I understand your questions.

crates/fluvio-spu/src/replication/leader/kv.rs
galibey: We wrote `LeaderReplicaLog` for using it within KvStorage. So, this one test to check it works together.

crates/fluvio-spu/src/replication/leader/kv.rs
galibey: To remove files from disk after test is done

crates/fluvio-spu/src/services/internal/update_consumer_offset_handler.rs
galibey: In this way we guarantee, that offset storage is used only on SPU that is a replica leader.

crates/fluvio-spu/src/services/public/mod.rs
galibey: This is using private api:
```
let leader_endpoint = spu_spec.private_endpoint.to_string();
```

galibey:
> Should separate out SPUSchema related changes into it's PR so can start merging that

Created https://github.com/infinyon/fluvio/pull/3918

galibey:
TTL has been removed

crates/fluvio-spu/src/services/internal/fetch_consumer_offset_handler.rs
```diff
@@ -0,0 +1,67 @@
...
+    ctx: DefaultSharedGlobalContext,
+) -> Result<ResponseMessage<FetchConsumerOffsetResponse>, IoError> {
+    let FetchConsumerOffsetRequest {
+        consumer_id,
+        replica_id,
+    } = req_msg.request;
+
+    let consumers_replica_id =
+        ReplicaKey::new(CONSUMER_STORAGE_TOPIC, <PartitionId as Default>::default());
+
```
sehz, line 34: add more description on what's being done here

sehz (approved):
Couple of minor clarification but no show stopper.   LGTM to proceed

crates/fluvio-spu/src/replication/leader/kv.rs
sehz: Try to add more description which hope to clarify purpose

Issue #3810, opened 2024-01-10
Add  new command `fluvio cluster resume`


digikata:
On some systems `fluvio cluster start` can fail to recognize started spus and timeout despite the spus reporting status.  (MacOs Sonoma 14.1.2)

### Problem

Starting up with multiple spus, creating a topic with replicas higher than 1, then restarting with a different number of spus causes this problem.  This can be fixed by starting the cluster with the original start options or a `fluvio cluster delete` and a fresh `fluvio cluster start`

```
$ fluvio cluster start --spus 3
$ fluvio topic create --replication 3 repltopic
$ fluvio produce repltopic
Ok!
> b
Ok!
> c
Ok!
> ^C
$ fluvio cluster shutdown

$ fluvio cluster start        
📝 Running pre-flight checks
    ✅ Local Fluvio is not installed
🎉 All checks passed!
✅ Local Cluster initialized
👤 Profile set
✅ SC Launched
🤖 Starting SPU: (1/1) \
🖥️ 3/1 SPU confirmed, 15 seconds elapsed 
```

## Solution
disallow `fluvio cluster start` after a `fluvio cluster shutdown`.  Instead add new command for `fluvio cluster resume`

## Community Note
This issue is of intermediate to expert difficulty, but Infinyon mentoring is available if you are interested in taking it on.

digikata:
This can cause timeouts on startup that look like:
```
$ fluvio cluster start               
📝 Running pre-flight checks
    ✅ Local Fluvio is not installed
🎉 All checks passed!
✅ Local Cluster initialized
👤 Profile set
✅ SC Launched
🤖 Starting SPU: (1/1) \
🖥️ 3/1 SPU confirmed, 15 seconds elapsed 
```

avikam:
i think the problem is coming from `confirm_spu()`. It actually confirmed more SPUs (based on the admin interface) than needed and therefore [doesn't leave](https://github.com/infinyon/fluvio/blob/master/crates/fluvio-cluster/src/start/local.rs#L666). 

```rust
if ready_spu /* ~admin.all().filter() */ == spu /* cli request */ as usize {
    ...
    return Ok(()); // never happens 
}
```
In fact, the bug reproduces whenever `start` is called with _less_ SPUs and doesn't with greater or equal SPUs, than the previously specified under `start`.

[...]

PR #3923, opened 2024-03-30
Scaffold resume - local config persisted between runs

avikam:
This is a draft first change for [issue 3810](https://github.com/infinyon/fluvio/issues/3810)
If we agree that we need to store the local cluster configuration in a file - this commit is the first step before adding more edge cases and error handling


digikata (approved):
LGTM, thanks @avikam for all the work on the contribution!

ajhunyady:
Nice work @avikam, and thank you for your contribution. We should add this valuable 
feature to the documentation once it's checked-in.

PR #3928, opened 2024-04-02
feat: added offset management to consumer api(unstable)

galibey:
Added offset management to consumer API. The feature is marked as unstable until all components support it, and the feature is fully tested and documented.

### DIFF
--- a/crates/fluvio/src/consumer/mod.rs
+++ b/crates/fluvio/src/consumer/mod.rs
@@ -1,6 +1,14 @@
+#![allow(dead_code)]
+
+mod config;
+mod stream;
+mod offset;
+
 use std::sync::Arc;
 
 use anyhow::Result;
+use async_channel::Sender;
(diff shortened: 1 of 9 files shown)


crates/fluvio/src/consumer/mod.rs
```diff
@@ -348,12 +366,20 @@ where
         if stream_fetch_version < CHAIN_SMARTMODULE_API {
             warn!("SPU does not support SmartModule chaining. SmartModules will not be applied to the stream");
         }
+        if stream_fetch_version < OFFSET_MANAGEMENT_API {
```
sehz, line 369: This warning should be only applied if consumer id is not null

sehz (changes requested):
Overall nice work.  A minor nit about warning

crates/fluvio/src/consumer/mod.rs
galibey: Yes, indeed. Fixed.

sehz (approved):

PR #3930, opened 2024-04-03
feat: added hidden topics and partitions

galibey:
Added an ability to create hidden topics, which by default will not be shown on cli.
To see them, need to use `--system` cli arg. This is needed for "special" offsets topic.

### DIFF
--- a/crates/fluvio-cli/src/client/partition/list.rs
+++ b/crates/fluvio-cli/src/client/partition/list.rs
@@ -18,6 +18,9 @@ use crate::common::OutputFormat;
 pub struct ListPartitionOpt {
     #[clap(flatten)]
     output: OutputFormat,
+    /// Show hidden partitions
+    #[arg(long, required = false)]
+    show_hidden: bool,
 }
 
 impl ListPartitionOpt {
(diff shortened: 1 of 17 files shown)


crates/fluvio-cli/src/client/partition/list.rs
```diff
@@ -30,6 +33,14 @@ impl ListPartitionOpt {
         let admin = fluvio.admin().await;
 
         let partitions = admin.all::<PartitionSpec>().await?;
+        let partitions = if !self.show_hidden {
```
sehz, line 36: Issue with this approach is that it add more complexity to client and make less flexible to add more policy in the future.  This should be pushed to SC.   There is already `summary` flag in the `ListRequest`.

sehz (changes requested):

crates/fluvio-cli/src/client/partition/list.rs
galibey: Moved filtration to SC, and renamed to `system`

sehz (approved):
LGTM.  Nice and clean

PR #3935, opened 2024-04-05
feat: added auto-creation of offsets topic

galibey:
Added a control of Offset Topic existence on SC. The check is periodic, once in a minute.

The Offset topic has the following characteristics: 
 - system=true
 - retention 7days
 - segment size 512 Mb
 - partition size 2 Gb

### DIFF
--- a/crates/fluvio-sc/src/controllers/topics/controller.rs
+++ b/crates/fluvio-sc/src/controllers/topics/controller.rs
@@ -4,11 +4,16 @@
 //! Reconcile Topics
 
 use fluvio_controlplane_metadata::spu::SpuSpec;
+use fluvio_controlplane::CONSUMER_STORAGE_TOPIC;
+use fluvio_controlplane_metadata::topic::CleanupPolicy;
+use fluvio_controlplane_metadata::topic::SegmentBasedPolicy;
+use fluvio_controlplane_metadata::topic::TopicStorageConfig;
+use fluvio_stream_dispatcher::actions::WSAction;
 use fluvio_stream_model::core::MetadataItem;
 use fluvio_stream_model::store::ChangeListener;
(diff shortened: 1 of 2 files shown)


crates/fluvio-sc/src/controllers/topics/controller.rs
```diff
@@ -152,4 +164,34 @@ impl<C: MetadataItem> TopicController<C> {
             }
         }
     }
+
+    async fn ensure_offsets_topic_exists(&mut self) {
```
sehz, line 168: Hmm.  This is not correct usage of TopicController.   The purpose of TopicController is handle topic lifecycle such as topic->partitioning mapping.  This assumes Topic object is created by external entity not the controller.    So to follow similar pattern, we should have something like `SystemTopicController` which will ensure system topic exists. 

However, not sure use of controller is right approach.   We could instead
- Create/update system topic when SC is started.   
- Don't allow system topic to be deleted.

sehz (changes requested):

crates/fluvio-sc/src/controllers/topics/controller.rs
galibey: I created `SystemTopicController` and moved the logic into it.

Regarding the topic deletion, it is not a problem to forbid deletion, but what if it does need to be deleted for some reason? Troubleshooting, dev purposes, memory problems, etc. We need to provide at least some way of deleting, if not the topic but the content maybe. So, I would propose postponing that thing for later, until we figure out a convenient strategy for system topics.

sehz (approved):
LGTM.  Not showstopper but need confirm this in the E2E in the CI

digikata (approved):

PR #3941, opened 2024-04-09
feat: added consumer offset to cli

galibey:
Added consumer offsets to Fluvio CLI (under `unstable` feature):
List:
```bash
fluvio consumer list
``` 
Delete:
```bash
fluvio consumer delete c1
```
Consume:
```bash
fluvio consume my-topic --consumer c1
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2322,6 +2322,7 @@ dependencies = [
  "async-trait",
  "atty",
  "bytesize",
+ "cfg-if",
  "clap",
  "clap_complete",
  "colored",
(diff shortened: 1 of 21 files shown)


digikata (approved):
LGTM

galibey:
> Code looks good but we should add Bats Test whenever we have new CLI commands

All tests will come in a separate PR. So far, all new functionalities are under the `unstable` feature which is not enabled by default.  Moving it out of `unstable` will happen along with adding integration and cli tests.

sehz:
fair enough

sehz (approved):

PR #3942, opened 2024-04-10
feat: disallow deletion system topic unless forced

galibey:
If user tries to delete a system topic:
```bash
$ fluvio topic delete consumer-offset
system topic 'consumer-offset' can only be deleted forcibly
```

with the `--system` argument:
```bash
$ fluvio topic delete consumer-offset --system
You are trying to delete a system topic 'consumer-offset'. It can affect the functioning of the cluster.

Are you sure you want to proceed? (y/n)
y
system topic "consumer-offset" deleted
```

with `--system` and `--force`:
```bash
$ fluvio topic delete consumer-offset --system --force
system topic "consumer-offset" deleted
```

### DIFF
--- a/crates/fluvio/src/admin.rs
+++ b/crates/fluvio/src/admin.rs
@@ -214,6 +214,38 @@ impl FluvioAdmin {
         Ok(())
     }
 
+    /// Forcibly delete object by key
+    /// key is dependent on spec, most are string but some allow multiple types.
+    ///
+    /// This method allows to delete objects marked as 'system'.
+    ///
+    /// For example, to delete a system topic:
+    ///
+    /// ```edition2021
+    /// use fluvio::Fluvio;
+    /// use fluvio::metadata::topic::TopicSpec;
+    ///
+    /// async fn delete_system_topic(name: String) -> anyhow::Result<()> {
+    ///     let fluvio = Fluvio::connect().await?;
+    ///     let admin = fluvio.admin().await;
+    ///     admin.force_delete::<TopicSpec>(name).await?;
+    ///     Ok(())
+    /// }
+    /// ```
+    #[instrument(skip(self, key))]
+    pub async fn force_delete<S>(&self, key: impl Into<S::DeleteKey>) -> Result<()>
+    where
+        S: DeletableAdminSpec + Sync + Send,
+    {
+        let delete_request: DeleteRequest<S> = DeleteRequest::with(key.into(), true);
+        debug!("sending force delete request: {:#?}", delete_request);
+
+        self.send_receive_admin::<ObjectApiDeleteRequest, _>(delete_request)
+            .await?
+            .as_result()?;
+        Ok(())
+    }
+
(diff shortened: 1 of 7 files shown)


ajhunyady:
Nice job @galibey! Looks good from the usability perspective.

crates/fluvio/src/admin.rs
```diff
@@ -214,6 +214,38 @@ impl FluvioAdmin {
...
+    ///
+    /// async fn delete_system_topic(name: String) -> anyhow::Result<()> {
+    ///     let fluvio = Fluvio::connect().await?;
+    ///     let admin = fluvio.admin().await;
+    ///     admin.force_delete::<TopicSpec>(name).await?;
+    ///     Ok(())
+    /// }
+    /// ```
+    #[instrument(skip(self, key))]
+    pub async fn force_delete<S>(&self, key: impl Into<S::DeleteKey>) -> Result<()>
```
sehz, line 236: we should not try to create different variation of API for every new parameter.  Instead, let's try to remapped existing API as like this:
```
pub async fn delete<S>(&self, key: impl Into<S::DeleteRequest>)
```

And them add default mapping for string and so forth.

ajhunyady (approved):
Usability looks good (I did not review the code).

sehz (changes requested):
Overall looks good except issue about delete API

crates/fluvio/src/admin.rs
galibey: But this is breaking change. It will come in a patch version bump,  users won't be happy

sehz: This is not breaking change since old API should still work

sehz: And it would be ok to bump up minor version if needed to be rather than create another API

galibey: This won't work. `DeleteRequest` is generic over `<S: DeletableAdminSpec>` and to construct it you need to provide `S::DeleteKey`. Yes, mostly `DeleteKey=String` but not always. There is `CustomSpuKey` also. Anyway, the API needs the associated type, not concrete `String` or `CustomSpuKey`, I tried to do something like:
```rust
impl<S: DeletableAdminSpec> From<S::DeleteKey> for DeleteRequest<S> {
    fn from(value: S::DeleteKey) -> Self {
        todo!()
    }
}
```
or 
```rust
impl<T, S> From<T> for DeleteRequest<S>
where
    S: DeletableAdminSpec,
    T: Into<S::DeleteKey>,
{
    fn from(value: S::DeleteKey) -> Self {
        todo!()
    }
}
```
but it doesn't work because the `DeleteRequest` type also fits into `DeleteKey` trait bounds  and the compiler fails it with `conflicting trait impl` error.

sehz: K.  Lets keep that as is and revisit this later when new WIT base is reformulated

sehz (approved):

PR #3945, opened 2024-04-15
feat: mark consumers offset feature as stable

galibey:
Added tests for consumers offset feature on Fluvio client, and on Fluvio CLI. 

Also, marked the feature as stable there.


sehz (approved):
LGTM!

digikata (approved):
LGTM

PR #3947, opened 2024-04-16
feat: set old consumer API deprecated

galibey:
Set the old consumer API as deprecated and replace its usages with the new one.

Fixes infinyon/roadmap#234

### DIFF
--- a/crates/fluvio-benchmark/src/consumer_worker.rs
+++ b/crates/fluvio-benchmark/src/consumer_worker.rs
@@ -4,7 +4,9 @@ use std::time::{Duration, Instant};
 use async_channel::{Sender, Receiver};
 use anyhow::Result;
 
-use fluvio::{consumer::ConsumerConfigBuilder, Offset, dataplane::link::ErrorCode};
+use fluvio::consumer::ConsumerConfigExtBuilder;
+use fluvio::Fluvio;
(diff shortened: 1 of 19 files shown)


tests/cli/cdk_smoke_tests/cdk-multi-partition-consumer.bats
```diff
@@ -113,55 +113,3 @@ EOF
     assert_success
 }
 
-@test "Topic with 3 partitions. Consumer reads 2 partitions" {
```
sehz, line 116: Shouldn't  this be migrated rather than delete?

sehz (changes requested):

tests/cli/cdk_smoke_tests/cdk-multi-partition-consumer.bats
galibey: The reason this test is being deleted is that the new consumer API doesn't support several partitions, only one or all. But, actually, supporting it is relatively cheap. I am going to add support for it, and we won't need to delete tests.

galibey: this test is kept without changes

sehz (approved):
LGTM

digikata (approved):

PR #3957, opened 2024-04-22
feat: use `ConsumerStream` trait in connectors

galibey:
Use the `ConsumerStream` trait with offset management capabilities in the common connectors dependencies.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2196,7 +2196,7 @@ dependencies = [
 
 [[package]]
 name = "fluvio"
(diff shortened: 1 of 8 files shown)


digikata (approved):
LGTM for manual config.

Is there some config/testing for auto strategy?

galibey:
> LGTM for manual config.
> 
> Is there some config/testing for auto strategy?

In the connectors - no, but it is used and tested in Fluvio CLI.

PR #3958, opened 2024-04-22
chore: rename SpuSocket to StreamSocket

fraidev:
New function to create streaming connection. 
Will be necessary for mirroring sc-to-sc connection.

Usage: https://github.com/infinyon/fluvio/blob/1c0022c1d1db2fd11197334e59a49f8eac959f53/crates/fluvio-sc/src/controllers/mirroring/controller.rs#L91

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2196,7 +2196,7 @@ dependencies = [
 
 [[package]]
 name = "fluvio"
-version = "0.21.9"
+version = "0.22.0"
 dependencies = [
  "anyhow",
  "async-channel 1.9.0",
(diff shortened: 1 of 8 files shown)


sehz:
There is already `WatchRequest`.  Is there need to create different one?

fraidev:
> There is already `WatchRequest`. Is there need to create different one?

I guess so, `watch` only works for `Spec` objects, for mirroring I need by `Request`.

sehz:
same code exists here: https://github.com/infinyon/fluvio/blob/master/crates/fluvio/src/spu.rs.   Let's consolidate them to make it common.

fraidev:
> same code exists here: https://github.com/infinyon/fluvio/blob/master/crates/fluvio/src/spu.rs. Let's consolidate them to make it common.

Indeed! It's the same code. 

I propose to rename `SpuSocket` to something more “generic” as we want it for SC too now.
What do you think?

sehz (approved):
LGTM

PR #3959, opened 2024-04-22
feat: rename mirroring cmds to home and remote

fraidev:
What is changed:
- [x] `core` commands are `remote` commands now
- [x] `edge` commands are `home` commands now


crates/fluvio-cli/src/client/home/status.rs
```diff
@@ -27,12 +28,26 @@ impl StatusOpt {
             .filter_map(|item| {
                 match item.spec.remote_type {
                     RemoteType::Core(core) => {
+                        let now = SystemTime::now()
+                            .duration_since(SystemTime::UNIX_EPOCH)
+                            .unwrap_or_default()
```
sehz, line 33: add unit test and also seems duplicated twice

sehz (changes requested):

crates/fluvio-cli/src/client/home/status.rs
fraidev: Right! About "duplicated twice" is about the home/status.rs and remote/list.rs code?

fraidev: I added a new method `last_seen` in RemoteStatus with tests

sehz (approved):
LGTM

PR #3961, opened 2024-04-22
feat(hub): enable `repository_url`

EstebanBorai:


crates/fluvio-hub-protocol/src/package_meta.rs
```diff
@@ -18,16 +19,15 @@ pub struct PackageMeta {
     pub name: String,
     pub version: String, // SemVer?, package version
     pub group: String,
-    // author: Option<String>,
     pub description: String,
     pub license: String,
+    pub manifest: Vec<String>, // Files in package, package-meta is implied, signature is omitted
+    pub repository_url: Option<Url>,
+    pub repository_commit_sha: Option<String>,
```
digikata, line 26: Lets leave out repository_commit_sha for now.  We can re-introduce it later as a tag if needed.

digikata (approved):
LGTM (just the nit of the sha field comment)

crates/fluvio-hub-protocol/src/package_meta.rs
EstebanBorai: SGTM! Here is the change: https://github.com/infinyon/fluvio/pull/3961/commits/9babcd76defc24cef2ed30db061acce04d984b8d

PR #3966, opened 2024-04-24
chore: rename remote, edge and core to mirror, remote and home

fraidev:
RemoteSpec is MirrorSpec now.

Which could be a Remote (edge) or Home (core)


sehz (approved):
LGTM

PR #3969, opened 2024-04-25
feat(cli): graceful period for stream interruption

### LINKED ISSUE #3963: [Bug]: in some conditions, `fluvio consumer -c <id> <topic>` does not record offsets
**What happened**
In some cases, `fluvio consumer -c <id> <topic>` does not record offsets

**Expected behavior**
A consumer should record offsets when a consumer id is defined at the cli.

**Describe the setup**
- fluvio 0.11.6 stable
- tested bug on Linux and Macos

1. `fluvio cluster start`   start a local cluster
2. `fluvio topic create test`
3. `fluvio produce test` # create 3-4 records
4. `fluvio consume test -c c_works -Bd`  this creates an offset
5. `fluvio consume test -c c_notcreated` this consumes offsets from the end, leave running for next steps
6. `flvuio produce test` # create 3-4 records
7. `fluvio consumer list` shows a `c_works` offset, but no `c_notcreated` offset

galibey:
Added graceful shutdown for the `consume` CLI operation.

Fixes #3963

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2318,6 +2318,7 @@ name = "fluvio-cli"
 version = "0.0.0"
 dependencies = [
  "anyhow",
+ "async-channel 1.9.0",
  "async-net",
  "async-trait",
  "atty",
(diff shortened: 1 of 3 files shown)


digikata (approved):
LGTM, nice!

PR #3979, opened 2024-04-30
chore: add endpoint field in mirror

fraidev:
Better to send the spu_endpoint from home instead try to get it later from remote.

### DIFF
--- a/crates/fluvio-controlplane-metadata/src/mirroring/mod.rs
+++ b/crates/fluvio-controlplane-metadata/src/mirroring/mod.rs
@@ -1,11 +1,11 @@
 use std::fmt::Debug;
 
 use fluvio_stream_model::core::{Spec, Status};
-use fluvio_types::ReplicaMap;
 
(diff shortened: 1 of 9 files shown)


crates/fluvio-controlplane-metadata/src/mirroring/mod.rs
```diff
@@ -104,15 +103,17 @@ impl MirroringStatusResponse {
 pub struct MirroringSpecWrapper<S> {
     pub key: String,
     pub spec: S,
-    pub replica_map: ReplicaMap,
+    pub spu_id: i32,
```
galibey, line 106: i32 -> SpuId

galibey (approved):

digikata (approved):
LGTM

crates/fluvio-controlplane-metadata/src/mirroring/mod.rs
fraidev: Good catch!

PR #3980, opened 2024-04-30
fix: cdk/smdk publish, fluvio-hub-util

digikata:
- **fix(fuvio-hub-util): upadate http client, better client isolation**
- **fix: fluvio-hub-util, replace htclient w/ ureq**


EstebanBorai (approved):
Great! Nice that we are now using ureq!

EstebanBorai:
@digikata seems like a 404 response is being handled by ureq instead of mapped into our string in a package set fetch

digikata:
Yup that's what the latest commit should fix

PR #3988, opened 2024-05-04
fix: explicitly enable wasi on smartengine uses

digikata:


Cargo.toml
```diff
@@ -165,28 +165,28 @@ k8-diff = { version = "0.1.2" }
 trybuild = { branch = "check_option", git = "https://github.com/infinyon/trybuild" }
 
 # Internal fluvio dependencies
-fluvio = { version = "0.22.0", path = "crates/fluvio" }
```
nacardin, line 168: Removing versions doesn't cause problems with crate publishing anymore?

digikata: Good catch, it seems like it's still an issue: https://github.com/rust-lang/cargo/issues/11133

The odd part is that it seems like the version at the workspace Cargo.toml doesnt seem to be used for publish - the crate version for fluvio was published at 0.22.1 last stable release.  I wonder if we can fill in version =  "0.0.0"?

digikata: Added a commit to restore version property on those deps

makefiles/check.mk
```diff
@@ -51,7 +51,7 @@ run-integration-test: build_smartmodules install_rustup_target
 
 	cargo test -p fluvio-smartengine -- --ignored --test-threads=1
 	rustup target add wasm32-wasi
-	cargo test  --features wasi -p fluvio-smartengine -- --ignored --test-threads=1
+	cargo test -p fluvio-smartengine -- --ignored --test-threads=1
```
morenol, line 54: this is repeated with line 52

EstebanBorai (approved):

morenol (approved):

nacardin (approved):

PR #3997, opened 2024-05-09
chore: update infinyon/regex-filter@0.1.1 for wasi

digikata:


digikata:
Fixed for this narrow ci test - note: we should do a follow on to download the built smdk and use it to build all the ci smartmodules.

smartmodule/regex-filter/Makefile
```diff
@@ -1,5 +1,11 @@
 default: build
 
-build:
-	rustup target add wasm32-unknown-unknown
-	cargo build --profile release-lto --target wasm32-unknown-unknown
+setup:
```
sehz, line 3: not show stopper but should reuse common targets and constants so we have consistent and repeatable build

digikata: yes, here is our reminder issue https://github.com/infinyon/fluvio/issues/3998

sehz (approved):
LGTM

digikata:
`infinyon/regex-filter@0.2.0` is published

PR #4001, opened 2024-05-10
fix(smdk): post-pone group ownership check on `generate`

EstebanBorai:
Moves group validation from generation phase to publish phase.


digikata (approved):
LGTM

PR #4002, opened 2024-05-10
feat: add auth and tls for mirroring connections

fraidev:
In this PR I used the same logic for auth on SC for SPU. 
Also added methods to validate by CN of the certificate, not only by scopes. Allowing to check if the remote has the same name as the CN of the certificate.

# Testing

## Happy path

On home side:
```bash
$ flvd cluster start --k8 --develop --tls --domain fluvio.local --server-key ./tls/certs/server.key --server-cert ./tls/certs/server.crt --ca-cert ./tls/certs/ca.crt --client-cert ./tls/certs/client-root.crt --client-key ./tls/certs/client-
root.key
$ flvd remote register user1
$ echo '[                                                                                                                                                                                                               
    "user1"
]' > devices.json 
$ flvd topic create mirror-topic --mirror-apply devices.json
$ flvd remote export user1 --file user1.json --key tls/certs/client-user1.key --cert tls/certs/client-user1.crt 
```

On remote side:

```bash
$ flvd cluster start
$ flvd home connect --file user1.json 
```


Then check:
- the topics and partitions are on both sides with: `flvd topic list` and `flvd partition list`
- only topics that are mirror topics and assigned to the remote1 are sync on remote side.
- try to produce on the topic on remote side and consume on home side.
- check status on remote side with: `flvd home status`
- check status on home side with: `flvd remote list`

## Unhappy path

Do the same using invalid certs or valid certs with another CN (e.g. tls/certs/client-root.crt).


k8-util/helm/fluvio-sys/templates/crd_mirror.yaml
```diff
@@ -46,14 +46,17 @@ spec:
...
                           type: string
-                keyPair:
-                  type: object
-                  required: ["privateKey", "publicKey"]
-                  properties:
-                    publicKey:
-                      type: string
-                    privateKey:
-                      type: string
+                        tls:
```
fraidev, line 49: Should this be a secret?

crates/fluvio-sc/Cargo.toml
```diff
@@ -47,6 +47,7 @@ tracing = { workspace = true }
 
 
 # Fluvio dependencies
+fluvio = { workspace = true }
```
fraidev, line 50: Is too bad to add this dependence? 
We need `DomainConnector::try_from`, I even tried to move it to fluvio-auth for example, but this method needs the build features of fluvio package.

crates/fluvio-auth/src/x509/authenticator.rs
```diff
@@ -36,15 +36,15 @@ impl ScopeBindings {
 
 #[derive(Debug)]
 pub struct X509Authenticator {
```
sehz, line 38: Initial scope of the authenticator was for only for SC.   Should not overload these with two different purpose. Probably simpler rename original `X509Authenticator` as `X509AuthenticatorSC` and move to `SC`.   And use this crate only for common auth utilities and libraries.

crates/fluvio-controlplane-metadata/src/mirror/spec.rs
```diff
@@ -75,4 +74,25 @@ pub struct Home {
     pub id: String,
     pub remote_id: String,
     pub public_endpoint: String,
+    #[cfg_attr(feature = "use_serde", serde(skip_serializing_if = "Option::is_none"))]
+    pub tls: Option<ClientTls>,
```
sehz, line 78: TLS should be mandatory

crates/fluvio-sc/Cargo.toml
sehz: If just need `DomainConnector` this can be imported from `fluvio-future`

crates/fluvio-sc/src/config/sc_config.rs
```diff
@@ -29,6 +29,7 @@ pub struct ScConfig {
     pub namespace: String,
     pub x509_auth_scopes: Option<PathBuf>,
     pub white_list: HashSet<String>,
+    pub tls: bool,
```
sehz, line 32: x509 requires TLS, so it redundant configuration

crates/fluvio-sc/src/controllers/mirroring/controller.rs
```diff
@@ -195,3 +223,23 @@ impl<C: MetadataItem> RemoteMirrorController<C> {
         }
     }
 }
+
+fn option_tlspolicy(home: &Home) -> Option<TlsPolicy> {
```
sehz, line 227: TLS should mandatory

crates/fluvio-sc/src/init.rs
```diff
@@ -142,6 +143,12 @@ where
                     ctx,
                     Arc::new(BasicAuthorization::new(policy)),
                 ));
+            } else if ctx.config().tls {
+                info!("using spu remote authorization");
+                start_public_server(AuthGlobalContext::new(
+                    ctx,
+                    Arc::new(RemoteAuthorization::new()),
```
sehz, line 150: Hmm.  Can't have two different policy.  Instead this should be unified

crates/fluvio-sc/src/services/auth/remote.rs
```diff
@@ -0,0 +1,65 @@
+use tracing::instrument;
+use async_trait::async_trait;
+
+use fluvio_auth::{AuthContext, Authorization, TypeAction, InstanceAction, AuthError};
+use fluvio_controlplane_metadata::extended::ObjectType;
+use fluvio_auth::x509::X509Identity;
+
+#[derive(Debug, Clone)]
+pub struct RemoteAuthorization {}
```
sehz, line 9: This should be unified

crates/fluvio-spu/src/services/auth/root.rs
```diff
@@ -0,0 +1,56 @@
...
+use fluvio_controlplane_metadata::extended::ObjectType;
+use fluvio_socket::FluvioSocket;
+use tracing::instrument;
+
+/// Authorization that allows by remote Id
+#[derive(Debug, Clone)]
+pub struct SpuRootAuthorization {}
+
+#[async_trait]
+impl Authorization for SpuRootAuthorization {
```
sehz, line 12: Why do you need SPU Authorization since it allows for everything?  Let's not try to re-map same semantics as SC which may not be same meaning

sehz (changes requested):

crates/fluvio-controlplane-metadata/src/mirror/spec.rs
ajhunyady: @sehz we had this conversation on Friday. TLS should not be mandatory, as there are 3 use cases.

crates/fluvio-sc/src/controllers/mirroring/controller.rs
ajhunyady: Not mandatory.

crates/fluvio-controlplane-metadata/src/mirror/spec.rs
sehz: It always been mandatory to have TLS for these connections:
- Edge to Home
- Home

Edge doesn't not need to have TLS

crates/fluvio-sc/src/config/sc_config.rs
fraidev: Do you mean x509_auth_scopes? I can run a cluster with TLS without scopes.

sehz: No I meant, why do need another `tis` config as TLS is already implied in the existing config.

crates/fluvio-controlplane-metadata/src/mirror/spec.rs
fraidev: I understand that run a mirroring in production without TLS is super risky, but maybe the user wants to test it locally first without certificate overhead.

Maybe we should just print warns that this is not a secure connection?

ajhunyady: As we discussed in Friday, when I test this locally, I don't need TLS connection.

fraidev: I don't get it. 

no certs = no authentication
with certs = we have authentication

This is already working like it.

Do you mean create something to encrypt it even without authentication?

sehz: authentication is mandatory as without it there is no way to ensure data integrity which lead to corruption.

ajhunyady: This is not going to be solved here. Let's keep optional until we discuss.

sehz: So there are two different questions:
(1) What strategy for Authentication (TLS, None, etc).
(2) Whether this flag is needed which will be depend on answering first question.  If this is needed, name should be change to so intent is clear.

crates/fluvio-sc/Cargo.toml
fraidev: True, but we need `fluvio::config::TlsPolicy` there too. 
Same problem, this is hard to move because use build features of fluvio package.

crates/fluvio-sc/src/init.rs
fraidev: True, make sense I just make the policy optional for BasicAuthorization?

crates/fluvio-spu/src/services/auth/root.rs
fraidev: Because I am calling an `AuthContext` method, the `allow_remote_id`.

Should I rename this default authorization?

crates/fluvio-sc/src/config/sc_config.rs
fraidev: where?

crates/fluvio-spu/src/services/auth/root.rs
fraidev: I don't undestand. 

should I not check authorization like it:

```rust
// authorization check
if !auth_ctx
    .auth
    .allow_remote_id(&req_msg.request.remote_cluster_id)
{
    warn!(
        "identity mismatch for remote_id: {}",
        req_msg.request.remote_cluster_id
    );
    return;
}
```

sehz: Something like this:
```
if let Ok(authorized) = auth_ctx
        .auth
        .allow_instance_action(RemoteSpec::Edge, InstanceAction::Cluster, cluster_id)
        .await
    {
```

sehz: Auth context is always there.  It will be configured differently on whether we configure with TLS or not.
If no TLS then it will be set to Root level authorization ( means you can do whatever you want).  If TLS then authorization is derived from TLS certs

crates/fluvio-sc/src/init.rs
sehz: use whatever as default.  no need for change

crates/fluvio-sc/src/controllers/mirroring/controller.rs
sehz: This is not needed anyway since SCConfig and Remote config are two different things.

crates/fluvio-controlplane-metadata/src/mirror/spec.rs
sehz: Decision:  Edge authentication  is  optional for mirroring,.  However TLS is mandatory if auth is required

sehz: But in this case, TLS should be attached to Edge not to home.

k8-util/helm/fluvio-sys/templates/crd_mirror.yaml
nacardin: Is this metadata used by the edge or home? Why is `tls` config needed here?

crates/fluvio-auth/src/x509/authenticator.rs
sehz: based on other conversation, this should need to be changed since `remote` can be added to scope

nacardin (commented):
We don't need to store the Edge's client certs as metadata right?

k8-util/helm/fluvio-sys/templates/crd_mirror.yaml
fraidev: > Is this metadata used by the edge or home? 

Edge is storing it as Home. 


> Why is tls config needed here?

To connect to home with TLS.

crates/fluvio-controlplane-metadata/src/mirror/spec.rs
fraidev: Edge cluster is storing TLS in record/struct called Home

crates/fluvio-sc/src/init.rs
sehz: scope should be part of auth context

fraidev:
This PR it was divided into:
https://github.com/infinyon/fluvio/pull/4016
https://github.com/infinyon/fluvio/pull/4017
https://github.com/infinyon/fluvio/pull/4021
https://github.com/infinyon/fluvio/pull/4022
https://github.com/infinyon/fluvio/pull/4024

PR #4005, opened 2024-05-11
fix(cdk): `sink` generation for conns fails to build

EstebanBorai:


morenol:
We should have a CI test to catch this

sehz (changes requested):
Yes need to add test to build this

EstebanBorai:
@sehz @morenol currently all our CLI solutions are tested using [BATS](https://bats-core.readthedocs.io/en/stable/index.html) which execute commands
and assert output on it.

The limitation on this approach is that interfaces like the one in `cdk` relies on an interactive terminal which BATS doesn't support, resulting in the following error:

```shell
 Error: IO error: not a terminal
```

As a way to test `generate` integrity, an approach like the one in `smdk generate` could be used:

```bash
smdk generate \
        --no-params \
        --template-path $(pwd)/smartmodule \
        --sm-type filter \
        --sm-public public \
        --silent \
        my-new-sm
```

This approach would still have to be implemented to CDK, this approach limits interactive
prompts to the limit of not showing any of them if all the options are passed as CLI arguments.

sehz:
k.  let's put that as separate issue

sehz (approved):
Good enough for now.   CI testing as separate issue

sehz:
K8 test are failing

PR #4013, opened 2024-05-16
feat(fvm): partial download pkgset updates

EstebanBorai:


crates/fluvio-version-manager/src/common/version_installer.rs
```diff
@@ -74,22 +59,106 @@ impl VersionInstaller {
         Ok(())
     }
 
+    pub async fn update(&self, upstream_artifacts: &[Artifact]) -> Result<()> {
+        let tmp_dir = self.download(upstream_artifacts).await?;
```
digikata, line 63: If update downloads all the new artifacts, this doesn't really save much time or bandwidth?  Is there a way to download just the upstream manifest?

EstebanBorai: Here `upstream_artifacts` holds the new artifacts only, as result from this step:

https://github.com/infinyon/fluvio/pull/4013/files#diff-049cead8b5c7cdafcc18670f14a9e44ff2606747a438bf15ddb6d839de27446aR70

So only new artifacts are being downloaded.

digikata (approved):
LGTM, nice!

PR #4017, opened 2024-05-22
feat: tls and authorization on SC mirroring

fraidev:
Needs: 
- [x] https://github.com/infinyon/fluvio/pull/4016


sehz (commented):
Looks good.  Is there E2E test? If not, it should be added

sehz (approved):
Let's add E2E as next step

PR #4021, opened 2024-05-22
feat: instance level authorization for basic authorization

fraidev:


crates/fluvio-sc/src/services/auth/basic.rs
```diff
@@ -203,17 +283,61 @@ mod test {
...
 
+    #[test]
+    fn test_action_urn_serialization() {
+        let action_urn = ActionUrn::new(
+            Action::Read,
+            Some(vec!["user1".to_string(), "user2".to_string()]),
+        );
+        let serialized =
+            serde_json::to_string(&action_urn).expect("failed to serialize action urn");
+        assert_eq!(serialized, r#""Read:user1:user2""#);
```
sehz, line 294: URN should be for individual objects not compounding this way.  So in this case should be serialized to `["Read:user1","Read:user2"]`.   

It would be good to add new policy file to see examples with instance level.  That would be make clear

fraidev: True! I forgot it. It's implemented as individual objects now.

I also added crates/fluvio-sc/test-data/auth_config/policy-instance.json

sehz: much better!

sehz:
rebase

sehz (approved):
Nice job!

PR #4026, opened 2024-05-24
feat(cdk): generate command params

EstebanBorai:
Provides params on CDK CLI generation command to be executed in not I/O environments, similar to how SMDK does.

This allow to run CDK's `generate` command for testing environments where interactive I/O
is not supported.

As part of this PR tests on the same command are also introduced, which ensures the integrity of the generated project for Sink and Source connectors.

For reference in SMDK: https://github.com/infinyon/fluvio/blob/ce2bcb5303257fbd7126e22365c38d40cfe893ee/tests/cli/smdk_smoke_tests/smdk-basic.bats#L97-L106


connector/cargo_template/Cargo.toml.liquid
```diff
@@ -4,6 +4,8 @@ version = "0.1.0"
 authors = ["{{authors}}"]
 edition = "2021"
 
+[workspace]
+
```
EstebanBorai, line 8:
```suggestion
```

sehz:
Can you add more description on what this does?

EstebanBorai:
> Can you add more description on what this does?

Of course! I just updated the PR description.

crates/cdk/src/generate.rs
```diff
@@ -17,9 +20,17 @@ static CONNECTOR_TEMPLATE: Dir<'static> =
 /// Generate new SmartConnector project
 #[derive(Debug, Parser)]
 pub struct GenerateCmd {
-    /// SmartConnector Project Name
+    /// Connector Name
     name: Option<String>,
 
+    #[arg(long, value_name = "GROUP")]
+    /// Connector developer group
+    group: Option<String>,
```
sehz, line 28: Isn't this already available in the connector metadata? why not just pass connector config rater than duplicate here

EstebanBorai: This would be the value used on generation to set the `group` field in `Connector.toml`. This is equivalent to replying the prompt on:

```
Please set a group name
```

In the following output:

```
cdk generate
🤷   Project Name: test
🤷   Please set a group name: infinyon
✔ 🤷   Which type of Connector would you like [source/sink]? · source
✔ 🤷   Will your Connector be public? · true
[1/8]   Done: .gitignore
[2/8]   Done: Cargo.toml
[3/8]   Done: Connector.toml
[4/8]   Done: README.md
[5/8]   Done: sample-config.yaml                                                [6/8]   Done: src/config.rs                                                     [7/8]   Done: src/main.rs                                                       [8/8]   Done: src
```

sehz: Got it

sehz (approved):

PR #4028, opened 2024-05-24
test: add mirroring e2e test

fraidev:
This PR should cover an initial structure to write e2e tests for mirroring.

The CI should start a local cluster and a k8s cluster.

The test should use k8s as home and local as remote.


.github/workflows/ci.yml
```diff
@@ -1199,6 +1199,96 @@ jobs:
...
+        run: fluvio version && fluvio-test -h
+
+      - name: Load Fluvio Docker Image
+        run: |
+          ls -la /tmp
+          docker image load --input /tmp/infinyon-fluvio-${{ matrix.rust-target }}.tar
+          docker image tag infinyon/fluvio:${{ github.sha }}-${{ matrix.rust-target }} infinyon/fluvio:${{ github.sha }}
+          docker image ls -a
+
+      - name: Export Docker Image to tarball
```
sehz, line 1262: why need to save image to tar again?

.github/workflows/ci.yml
```diff
@@ -1199,6 +1199,96 @@ jobs:
...
+          ./k8-util/cluster/reset-k3d.sh
+      - name: Load image to K3d
+        run: |
+          k3d image import -k /tmp/infinyon-fluvio.tar -c fluvio
+
+      - name: Create k8 namespace
+        run: |
+          kubectl create ns fluvio-test
+
+      - name: Start cluster locally
```
sehz, line 1276: Should name this as `start edge cluster`.
Also, there should be at least 2 edge cluster

sehz (changes requested):

.github/workflows/ci.yml
fraidev: Indeed I will change it.
About the edges, it seems that fluvio cluster gets SC and SPU by process, so I think it's impossible to have two clusters running in an environment. (Even with -skip-check option).

sehz: Should be able to have to local cluster.  They just need to use different ports.  Try to to use env variable to override SC and SPU port

fraidev: Based on my tests, Fluvio uses too much const variables for local clusters, I don't think it's only SC and SPU ports the problem. 

I think we need to have these const vars changeable by cluster arguments:
LOCAL_PROFILE or implement skip_profile_creation for local clusters (works for k8s).
LOCAL_CONFIG_PATH
DEFAULT_DATA_DIR
LOCAL_SC_ADDRESS
LOCAL_SC_PORT
SC_PUBLIC_PORT
SC_PRIVATE_PORT

I would like to sync about it this week if this is a priority right now.

sehz: There are alternatives
- Use K8 but with different name space. 
- Add option to cluster start with option to specify different ports and different SPU.

.github/workflows/ci.yml
fraidev: is doing it twice?

sehz: yes, no need to save image since it's already exported

fraidev: I got this from others test cases.
 https://github.com/infinyon/fluvio/blob/dcd3df0e4a85db4ef47ecbf646e0d78bb758e0ff/.github/workflows/ci.yml#L898

At line 898 it's saying that it's loading and exporting again to remove the arch from the tag.

This makes sense? Do we need this for this case?

.github/workflows/ci.yml
fraidev: using two edge nodes now

.github/workflows/ci.yml
sehz: ah ok.  we should consolidate these as script but it's out of scope of this.

tests/cli/mirroring_smoke_tests/e2e/fluvio-core.bats
```diff
@@ -0,0 +1,177 @@
...
+    run jq -r .home.id remote.json 
+    assert_output "$HOME_NAME"
+    assert_success
+
+    run jq -r .home.remoteId remote.json 
+    assert_output "$REMOTE_NAME"
+    assert_success
+
+    run jq -r .home.publicEndpoint remote.json 
+    assert_output --partial ":30003"
```
sehz, line 76: where is this come from? maybe this should be specified in the set up?

tests/cli/mirroring_smoke_tests/e2e/fluvio-core.bats
```diff
@@ -0,0 +1,177 @@
...
+    assert_success
+}
+
+@test "Can consume message from mirror topic produced from remote 2" {
+    sleep 5
+    run timeout 15s "$FLUVIO_BIN" consume "$TOPIC_NAME" -p 1 -B -d
+
+    assert_output 9$'\n'z$'\n'8$'\n'y
+    assert_success
+}
```
sehz, line 177: need couple of more test cases:
- produce message on remote 1 before connecting.  
- disconnect home from remote 2.  then produce 2nd message on remote 2

sehz (changes requested):
Nice Job!  Just need couple of more things to address

.github/workflows/ci.yml
```diff
@@ -1199,6 +1199,106 @@ jobs:
...
+          kubectl create ns fluvio-test
+
+      - name: Start edge cluster remote 1
+        timeout-minutes: 10
+        run: fluvio cluster start --local
+
+      - name: Start edge cluster remote 2
+        timeout-minutes: 10
+        run: | 
+          fluvio profile add local2 127.0.0.1:9103 local
```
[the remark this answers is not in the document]
fraidev: Sure! I added these envs:
      REMOTE_2_SC_PUB_ADDR: "127.0.0.1:9103"
      REMOTE_2_SC_PRIV_ADDR: "0.0.0.0:9104"
      REMOTE_2_SPU_ID: 6001
      REMOTE_2_SPU_PUB_ADDR: "0.0.0.0:9020"
      REMOTE_2_SPU_PRIV_ADDR: "0.0.0.0:9021"
      REMOTE_2_BASE_DIR: "~/.fluvio_2"

tests/cli/mirroring_smoke_tests/e2e/fluvio-core.bats
fraidev: It's the default public port for Kubernetes clusters. I change to use a env var

tests/cli/mirroring_smoke_tests/e2e/fluvio-core.bats
fraidev: I added a test for the first one, but we don't have disconnect command to have the second. 
But I added a TODO comment 😅

tests/cli/mirroring_smoke_tests/e2e/fluvio-core.bats
```diff
@@ -0,0 +1,189 @@
...
+
+@test "Can switch to remote cluster 1" {
+    run timeout 15s "$FLUVIO_BIN" profile switch "$REMOTE_PROFILE"
+    assert_output ""
+    assert_success
+}
+
+@test "Can't produce message to mirror topic from remote 1 before connect to home" {
+    sleep 5
+    run bash -c 'echo 1 | timeout 15s "$FLUVIO_BIN" produce "$TOPIC_NAME"'
```
sehz, line 118: this should work.  initial json spec should contain mirror topic config

tests/cli/mirroring_smoke_tests/e2e/fluvio-core.bats
sehz: you can simulate same condition by:
- shutdown remote cluster
- restart remote cluster
- produce message on remote.

probably need command on remote cluster to shutdown connection in the future.

tests/cli/mirroring_smoke_tests/e2e/fluvio-core.bats
fraidev: Really? The main idea of SC-to-SC is not sync the topics?

sehz: Yes but initial json should contain mirror topics.   This means remote cluster should be able to produce to local topic without connecting to home first.

fraidev: I thought that it was a limitation of using read_only mode. Make sense. It should be covered on this PR?

fraidev: I put a task in the roadmap to fix it https://github.com/infinyon/roadmap/issues/277

sehz: Ok. this seems at least cover most of core use cases

tests/cli/mirroring_smoke_tests/e2e/fluvio-core.bats
fraidev: I tried, unfortunately, `shutdown` seems to have the same problem as `cluster delete` of deleting all executables by name. Also, resume does not work after it when it's running two clusters for some reason, with one works.

sehz: Ok. let's create issue for this.  this is critical path to cover.  thanks for trying

sehz (approved):
Good enough coverage

PR #4029, opened 2024-05-27
fix: disallow produce mirror topic from home

fraidev:


crates/fluvio-controlplane-metadata/src/partition/spec.rs
```diff
@@ -87,6 +87,14 @@ impl PartitionSpec {
             "".to_owned()
         }
     }
+
+    /// check if this partition is home mirror
+    pub fn is_home_mirror(&self) -> bool {
```
sehz, line 92: add unit test

crates/fluvio/src/fluvio.rs
```diff
@@ -180,11 +180,16 @@ impl Fluvio {
         debug!(topic = &*topic, "Creating producer");
 
         let spu_pool = self.spu_pool().await?;
-        if !spu_pool.topic_exists(&topic).await? {
-            return Err(FluvioError::TopicNotFound(topic).into());
-        }
 
-        TopicProducer::new(topic, spu_pool, config, self.metric.clone()).await
+        if let Some(partition) = spu_pool.get_partition_by_topic(&topic).await? {
```
sehz, line 184: the enforcement logic should be moved to SPU.

tests/cli/mirroring_smoke_tests/mirror-topic.bats
```diff
@@ -0,0 +1,49 @@
...
+
+@test "Can create a mirror topic" {
+    echo "[\"$REMOTE_NAME\"]" > remotes.json
+    run timeout 15s "$FLUVIO_BIN" topic create "$TOPIC_NAME" --mirror-apply remotes.json
+
+    assert_output "topic \"$TOPIC_NAME\" created"
+    assert_success
+}
+
+@test "Can't produce to a mirror topic from home" {
```
sehz, line 43: also need positive test as well

sehz (changes requested):

tests/cli/mirroring_smoke_tests/mirror-topic.bats
fraidev: Do you mean a test that can produce? Will have one in the PR of add e2e tests https://github.com/infinyon/fluvio/pull/4028

crates/fluvio/src/fluvio.rs
fraidev: Do you mean that get_partition_by_topic should not return an option and should return a not found err instead of none?

crates/fluvio-controlplane-metadata/src/partition/spec.rs
fraidev: Sure, added unit tests

crates/fluvio/src/fluvio.rs
sehz: no I meant, SPU should enforce business logic not in the client.  Client can do validation but at the end of day, SPU need to make sure data is valid and consistent

fraidev: Something like this? 

```rust
    pub async fn get_produceble_partition_by_topic(
        &self,
        topic: impl Into<String>,
    ) -> Result<PartitionSpec, FluvioError> {
        let topic: String = topic.into();
        let replica = ReplicaKey::new(topic.clone(), 0u32);
        let partition = self.metadata.partitions().lookup_by_key(&replica).await?;

        match partition {
            Some(partition) => {
                if partition.spec.is_home_mirror() {
                    return Err(FluvioError::CannotProduceToMirrorFromHome);
                }

                Ok(partition.spec)
            }
            None => Err(FluvioError::TopicNotFound(topic.into())),
        }
    }
```

fraidev: I changed to `is_topic_produceble` and put it on spu. What do you think?

fraidev: Before the `create_serial_socket_from_leader`? But we don't pass the topic for it, so how to check it?

Flush calling create_serial_socket_from_leader as example: 
```rust
    /// Flush all the batches that are full or have reached the linger time.
    /// If force is set to true, flush all batches regardless of linger time.
    pub(crate) async fn flush(&self, force: bool) -> Result<()> {
        let leader = self.current_leader().await?;

        let spu_socket = self
            .spu_pool
            .create_serial_socket_from_leader(leader)
            .await?;

        let mut batches_ready = vec![];

```

sehz: https://github.com/infinyon/fluvio/blob/master/crates/fluvio-spu/src/services/public/produce_handler.rs#L116

fraidev: got it, added there:

```rust
        if let Some(mirror) = &leader_state.get_replica().mirror {
            if mirror.is_home_mirror() {
                debug!(%replica_id, "Mirror replica is not supported for produce");
                topic_result.partitions.push(PartitionWriteResult::error(
                    replica_id,
                    ErrorCode::MirrorProduceFromHome,
                ));
                continue;
            }
        }
```

what do you think?

sehz (approved):
LGTM

PR #4056, opened 2024-06-04
feat: add sni prefix on spu to spu

fraidev:
Needed for cloud.


digikata (approved):

crates/fluvio-controlplane-metadata/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "fluvio-controlplane-metadata"
 edition = "2021"
-version = "0.28.2"
+version = "0.28.3"
```
sehz, line 4: this doesn't need to be bump up since `0.28.2` is not released

fraidev: Make sense! I thought that It was based from master version. Fixed

sehz (approved):

PR #4068, opened 2024-06-17
Feature guard compression

crajcan:
Keep compression on by default, but allow it to be turned off with `--no-default-features`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -465,7 +465,7 @@ jobs:
         run: make -C smartmodule/regex-filter
 
   check_wasm:
-    name: Build WASM crates (${{ matrix.wasm-crate }})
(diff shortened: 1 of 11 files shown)


.github/workflows/ci.yml
```diff
@@ -488,7 +494,11 @@ jobs:
       - name: install wasm target
         run: rustup target add wasm32-unknown-unknown
       - name: Build WASM for ${{ matrix.wasm-crate }}
```
digikata, line 496: Some of the titles are partially blank, I think this needs"
`${{ matrix.wasm-crate }}` -> `${{matrix.context.crate}}`

digikata (approved):
Looks good overall, see comments for fixups around `matrix.wasm-crate` references

PR #4076, opened 2024-06-21
chore: add update commands for public api requests

fraidev:
This PR is adding Update commands for the public api requests. 

This PR is needed for the add partition feature: https://github.com/infinyon/fluvio/pull/4065


crates/fluvio-sc-schema/src/objects/classic.rs
```diff
@@ -412,6 +413,197 @@ mod delete_macro {
     pub(crate) use ClassicDecodingDelete;
 }
 
+/// carry from prev version for compatibility test
+mod update_macro {
+
```
sehz, line 418: should not be needed.  only for supporting old protocol

sehz (changes requested):
looks good.  just remove compatibility macro

crates/fluvio-sc-schema/src/objects/classic.rs
fraidev: Make sense! We don't need this. 

I removed that.

sehz (approved):
LGTM

PR #4059, opened 2024-06-09
feat(fluvio-cli): add base offset to partition list command

PanGan21:
Adds `base offset` to `fluvio partition list` command.
Related: #3975

PR #4077, opened 2024-06-24
try: changes based on #4059 for support

EstebanBorai:


makefiles/check.mk
EstebanBorai, line 1: Ignore this please.

Issue #4079, opened 2024-06-24
Produce connections are not identifying new partitions


fraidev:
Producer must identify new partitions on the same connection. 
This already works for consumers.

Related:
- https://github.com/infinyon/fluvio/issues/3836
- https://github.com/infinyon/fluvio/pull/4065

fraidev:
https://github.com/infinyon/fluvio/pull/4090 solves it

PR #4090, opened 2024-06-30
fix: produce recognize new partitions

fraidev:
Needs:
- [x] https://github.com/infinyon/fluvio/pull/4092

This PR must solve: https://github.com/infinyon/fluvio/issues/4079

## Testing case

```bash
fluvio topic create my-topic1
fluvio produce my-topic1
```

In another terminal:

```bash
fluvio topic add-partition my-topic1
```

back to the produce:

```bash
> 1
Ok!
> 2
Ok!
```

Then the new partition must have records, you can check it with:
```bash
fluvio partition list
```

## Implementation

The fluvio CLI is using the context `LocalMetadataItem` now. Because it also stores changes, not only new specs.
Some changes in `RecordAccumulator` and `ProducerPool` to include new partitions.


crates/fluvio-stream-dispatcher/src/metadata/local.rs
```diff
@@ -256,460 +91,630 @@ impl LocalMetadataItem {
...
 
-#[derive(Debug)]
-struct SpecStore {
-    version: AtomicU64,
-    data: RwLock<HashMap<String, SpecPointer>>,
-    sender: Sender<SpecUpdate>,
-    receiver: Receiver<SpecUpdate>,
-    path: PathBuf,
-}
+cfg_if::cfg_if! {
```
fraidev, line 94: The only change here is that removed the feature flag k8 for the whole metadata mod and put this if for the things that really need this flag. Then it's possible to import `LocalMetadataItem` without the k8 flag.

I can move it to another PR if it's not very clear.

sehz: Yes, it would be better made as separate PR

crates/fluvio/src/producer/accumulator.rs
```diff
@@ -66,27 +67,40 @@ impl RecordAccumulator {
...
         Self {
-            batches: Arc::new(batches),
+            batches: Arc::new(RwLock::new(batches)),
             batch_size,
             compression,
             queue_size,
         }
     }
 
+    pub async fn add_partition(
```
sehz, line 82: is this need to be made public?

sehz: unit test?

crates/fluvio/src/producer/mod.rs
```diff
@@ -90,24 +94,72 @@ impl ProducerPool {
         }
     }
 
-    fn shared(
+    #[allow(clippy::too_many_arguments)]
+    pub async fn add_producer_if_needed(
```
sehz, line 98: is this need to be public?

crates/fluvio/src/producer/mod.rs
```diff
@@ -90,24 +94,72 @@ impl ProducerPool {
...
         topic: String,
         spu_pool: Arc<SpuPool>,
-        batches: Arc<Vec<BatchHandler>>,
         client_metric: Arc<ClientMetrics>,
-    ) -> Arc<Self> {
-        Arc::new(ProducerPool::new(
+        partition_id: PartitionId,
+        batchs: Arc<BatchesDeque>,
+        batch_events: Arc<BatchEvents>,
+        record_accumulator: Arc<RecordAccumulator>,
```
sehz, line 107: having too many arguments may indicate need to put them in a struct?

crates/fluvio/src/producer/mod.rs
```diff
@@ -90,24 +94,72 @@ impl ProducerPool {
...
-        ))
+            partition_id,
+            batchs.clone(),
+            batch_events.clone(),
+        );
+        Some((batch_events.clone(), batchs.clone()))
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    pub fn add_producer(
```
sehz, line 130: probably not made public.  seems to share same arg as other

crates/fluvio/src/producer/mod.rs
```diff
@@ -352,8 +419,9 @@ impl TopicProducer {
                 config,
                 topic,
                 spu_pool,
-                producer_pool,
-                record_accumulator,
+                producer_pool: Arc::new(RwLock::new(producer_pool)),
+                record_accumulator: Arc::new(record_accumulator),
+                metrics: metrics.clone(),
             }),
             #[cfg(feature = "smartengine")]
```
sehz, line 426: seems need unit/integration to tests producers

crates/fluvio-stream-dispatcher/src/metadata/local.rs
fraidev: Done with: https://github.com/infinyon/fluvio/pull/4092

crates/fluvio/src/producer/accumulator.rs
fraidev: I think so, but only for crate, I change it.

crates/fluvio/src/producer/mod.rs
fraidev: No it's not, good catch! Removed!

crates/fluvio/src/producer/mod.rs
fraidev: Yeah, should be private! Removed!

crates/fluvio/src/producer/accumulator.rs
fraidev: unit test added

crates/fluvio/src/producer/mod.rs
fraidev: Fixed!

crates/fluvio/src/producer/mod.rs
fraidev: How to handle SpuPool in this case? Create a mock?

sehz: Could build a fake SPU handler that serve respond to socket similar to integration tests in SPU

fraidev: As talked in Company Meeting: it's better to merge it now for a new release and create a new pull request with unit tests.

fraidev: Issue: https://github.com/infinyon/fluvio/issues/4093

sehz: Why is better to merge now? Since this significant change and effect core path,  there should be enough test coverage.  Simple e2e is not sufficient.

fraidev: I don't remember the reason. 

Anyway, I tried to write integration tests on fluvio package, but it's needed to import `fluvio-spu`, and seems that `fluvio-spu` package is not very friendly to be imported.

Also tried to write them on `fluvio-spu` but `ProducerTopic` and other structs needs are private or protected (pub(crate)).

What do you think? Which side should we test it?

By the way, I noticed that there are some tests in `fluvio-test` using TopicProducer: 
- longevity 
- producer 
- producer_fail

sehz: Probably don't want to to use SPU directly in the client.  All this is way to test behavior changes.  This could be mocking SPU response (may need to change internal interface to make it easier).

fraidev: Hmm. The only solution I came to mind is to use traits + generics. 
But this will be a big break change for connectors! As `TopicProduce` will be changed to `TopicProduce<S>`.

I added a new commit these changes + a test, what do you think?

connector/json-test-connector/src/main.rs
```diff
@@ -1,13 +1,13 @@
...
-use fluvio::{TopicProducer, RecordKey};
+use fluvio::{spu::SpuSocketPool, RecordKey, TopicProducer};
 use fluvio_connector_common::{Source, connector, Result};
 use futures::StreamExt;
 
 use crate::source::TestJsonSource;
 
 #[connector(source)]
-async fn start(config: CustomConfig, producer: TopicProducer) -> Result<()> {
+async fn start(config: CustomConfig, producer: TopicProducer<SpuSocketPool>) -> Result<()> {
```
sehz, line 10: can you create type alias like: 
```
type DefaultTopicProducer = TopicProducer<SpuSocketPool>
```
so it will be be easier to use by rest of client.

sehz (changes requested):
Nice change.   This is probably good enough for now.   Just have minor suggestion to simplify usage

connector/json-test-connector/src/main.rs
fraidev: Much better! I changed to use `DefaultTopicProducer`.

digikata: I'd maybe suggest TopicProduceProol ?

digikata (approved):
LGTM, nitpick on the naming of DefaultTopicProducer

connector/json-test-connector/src/main.rs
fraidev: Sure! I rename it to `TopicProducePool`

sehz (approved):
LGTM

PR #4094, opened 2024-07-04
Fix delete topic partitions

fraidev:
Solves: https://github.com/infinyon/fluvio/issues/4095

### DIFF
--- a/crates/fluvio-sc/src/stores/topic/store.rs
+++ b/crates/fluvio-sc/src/stores/topic/store.rs
@@ -19,7 +19,7 @@ pub type DefaultTopicLocalStore = TopicLocalStore<u32>;
 
 #[async_trait]
 pub trait TopicMd<C: MetadataItem> {
-    async fn create_new_partitions(
+    async fn partitions_from_replicas(
         &self,
         partition_store: &PartitionLocalStore<C>,
     ) -> Vec<PartitionMetadata<C>>;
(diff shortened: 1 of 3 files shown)


crates/fluvio-sc/src/stores/topic/store.rs
```diff
@@ -19,42 +17,28 @@ pub type DefaultTopicLocalStore = TopicLocalStore<u32>;
...
-                    MetadataStoreObject::with_spec(replica_key, partition_spec)
-                        .with_context(self.ctx.create_child()),
-                )
-            } else {
-                debug!(?replica_key, "partition already exists");
-            }
-        }
-        partitions
+    /// get partitions from replica map
+    async fn partitions_from_replicas(&self) -> Vec<PartitionMetadata<C>> {
```
sehz, line 29: this need unit test

sehz (changes requested):

crates/fluvio-sc/src/stores/topic/store.rs
fraidev: Sure! I added one test for `partitions_from_replicas`

fraidev:
> LGTM. We need to start adding tests for controllers .

Just to let you, I put back the partition store because it's an easy way to guarantee the same SPU.  
The election tests got this problem. But I also added leader checks in the unit test.

PR #4103, opened 2024-07-15
chore: default to zigbuild for cross-compilation

EstebanBorai:
While working on building the Fluvio image to run in Kubernetes on macOS I found that
cross-compilation of Fluvio might have some incompatibility issues.

In order to speed up development I ended up using `zigbuild` as an alternative to cross-compile Fluvio.

In the future, perhaps it would be nice to detect host's OS and conditionally use `zigbuild` if running on `macOS` for such step.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -257,12 +257,11 @@ dependencies = [
 
 [[package]]
 name = "async-channel"
-version = "2.1.1"
+version = "2.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


sehz:
maybe we should convert default build to use `zigbuild` instead?

EstebanBorai:
> maybe we should convert default build to use `zigbuild` instead?

Lets try that!

digikata (approved):

sehz (commented):
can remove all zig related flags

nacardin:
> can remove all zig related flags

I don't think so, I believe zigbuild only replaces linker, not CC.

sehz:
got it

PR #4115, opened 2024-07-23
chore: bump fluvio-smartmodule

fraidev:


morenol (commented):
Not sure if we should upgrade to 0.8

morenol:
Could you update https://github.com/infinyon/fluvio/blob/171fba00b4ae70094bfd54a4a32427baef9ecf02/Cargo.toml#L181 too?

fraidev:
> Could you update
> 
> https://github.com/infinyon/fluvio/blob/171fba00b4ae70094bfd54a4a32427baef9ecf02/Cargo.toml#L181
> 
> too?
Fixed!

morenol (approved):

crates/fluvio-smartmodule/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio-smartmodule"
-version = "0.7.3"
+version = "0.8.0"
```
digikata, line 3: maybe 0.7.4 instead of 0.8

digikata: Or is there any reason to prefer 0.8 instead?

fraidev: Changed to 0.7.4. 

Maybe because it's a breaking change?

morenol: Not sure if this change on fluvio-protocol will be an issue for next releases

digikata: It will be if we maintain the current compatibility handshake, but we will just have to keep in it sync for sdf too until we can improve that.

Issue #4102, opened 2024-07-12
Mirroring: Update mirror status when find errors with SPU-to-SPU


fraidev:
The mirror status should be changed if the SPU is down or up, for example.
This is already working for SC.

ajhunyady:
is this done?

fraidev:
> is this done?

Yeah!

PR #4120, opened 2024-07-28
feat: update mirror status when find errors on spu

fraidev:
This PR aims to improve the feedback of the SPU-to-SPU connection on mirroring. Also improving the feedback of the failures showing errors.

# Changes
- Add two new columns (SPU STATUS and ERRORS) for `fluvio remote list` and `fluvio home status` commands
- Mirror Failed Status Enum now stores an error.
- New private SC API to receive the status from SPU
- A dispatcher on SC to apply these new status

# Error column

The new error column will show the error from SPU or SC or both. 
When both will be "SC:ERROR SPU:ERROR".
When none will be "-".

# Outputs
image
Screenshot 2024-07-31 at 00 06 58
image
image
image

Related:
- https://github.com/infinyon/fluvio/issues/4102

### DIFF
--- a/crates/fluvio-cli/src/client/home/status.rs
+++ b/crates/fluvio-cli/src/client/home/status.rs
@@ -7,6 +7,8 @@ use fluvio_extension_common::target::ClusterTarget;
 use fluvio_extension_common::{OutputFormat, Terminal};
 use fluvio_sc_schema::mirror::{MirrorSpec, MirrorType};
 
+use self::output::HomeStatusRow;
+
 use super::get_admin;
 
 #[derive(Debug, Parser)]
@@ -14,6 +16,7 @@ pub struct StatusOpt {
     #[clap(flatten)]
     output: OutputFormat,
 }
+
 impl StatusOpt {
     pub async fn execute<T: Terminal>(
         self,
@@ -24,17 +27,19 @@ impl StatusOpt {
         let list = admin.all::<MirrorSpec>().await?;
         let now = SystemTime::now().duration_since(SystemTime::UNIX_EPOCH)?;
 
-        let outlist: Vec<(String, String, String, String)> = list
+        let outlist: Vec<HomeStatusRow> = list
             .into_iter()
             .filter_map(|item| {
                 match item.spec.mirror_type {
                     MirrorType::Home(home) => {
-                        Some((
-                            home.id.to_string(),        // Source ID
-                            home.public_endpoint,       // Route
-                            item.status.to_string(),    // Status
-                            item.status.last_seen(now), // Last-Seen
-                        ))
+                        Some(HomeStatusRow {
+                            home: home.id.to_string(),                       // Source ID
+                            route: home.public_endpoint,                     // Route
+                            sc_status: item.status.pairing_sc.to_string(),   // SC Status
+                            spu_status: item.status.pairing_spu.to_string(), // SPU Status
+                            last_seen: item.status.last_seen(now),           // Last-Seen
+                            errors: item.status.pair_errors(),               // Errors
+                        })
                     }
                     _ => None,
                 }
@@ -51,7 +56,6 @@ mod output {
     //!
     use comfy_table::{Cell, Row};
     use comfy_table::CellAlignment;
-    use tracing::debug;
     use serde::Serialize;
     use anyhow::Result;
 
@@ -60,10 +64,18 @@ mod output {
     use fluvio_extension_common::output::TableOutputHandler;
     use fluvio_extension_common::t_println;
 
-    type ListVec = Vec<(String, String, String, String)>;
+    #[derive(Serialize)]
+    pub struct HomeStatusRow {
+        pub home: String,
+        pub route: String,
+        pub sc_status: String,
+        pub spu_status: String,
+        pub last_seen: String,
+        pub errors: String,
+    }
 
     #[derive(Serialize)]
-    struct TableList(ListVec);
+    struct TableList(Vec<HomeStatusRow>);
 
     // -----------------------------------
     // Format Output
@@ -71,11 +83,9 @@ mod output {
 
     pub fn format<O: Terminal>(
         out: std::sync::Arc<O>,
-        listvec: ListVec,
+        listvec: Vec<HomeStatusRow>,
         output_type: OutputType,
     ) -> Result<()> {
-        debug!("listvec: {:#?}", listvec);
-
         if !listvec.is_empty() {
             let rlist = TableList(listvec);
             out.render_list(&rlist, output_type)?;
@@ -92,7 +102,17 @@ mod output {
     impl TableOutputHandler for TableList {
         /// table header implementation
         fn header(&self) -> Row {
-            Row::from(["HOME", "ROUTE", "STATUS", "LAST SEEN"])
+            Row::from(
+                [
+                    "HOME",
+                    "ROUTE",
+                    "SC STATUS",
+                    "SPU STATUS",
+                    "LAST SEEN",
+                    "ERRORS",
+                ]
+                .iter(),
+            )
         }
 
         /// return errors in string format
@@ -106,10 +126,12 @@ mod output {
                 .iter()
                 .map(|e| {
                     Row::from([
-                        Cell::new(&e.0).set_alignment(CellAlignment::Left),
-                        Cell::new(&e.1).set_alignment(CellAlignment::Left),
-                        Cell::new(&e.2).set_alignment(CellAlignment::Left),
-                        Cell::new(&e.3).set_alignment(CellAlignment::Left),
+                        Cell::new(&e.home).set_alignment(CellAlignment::Left),
+                        Cell::new(&e.route).set_alignment(CellAlignment::Left),
+                        Cell::new(&e.sc_status).set_alignment(CellAlignment::Left),
+                        Cell::new(&e.spu_status).set_alignment(CellAlignment::Left),
+                        Cell::new(&e.last_seen).set_alignment(CellAlignment::Left),
+                        Cell::new(&e.errors).set_alignment(CellAlignment::Left),
                     ])
                 })
                 .collect()
--- a/crates/fluvio-cli/src/client/home/status.rs
+++ b/crates/fluvio-cli/src/client/home/status.rs
@@ -1,13 +1,13 @@
-pub use std::sync::Arc;
+use std::sync::Arc;
 use std::time::SystemTime;
 
 use anyhow::Result;
 use clap::Parser;
+use serde::Serialize;
+
 use fluvio_extension_common::target::ClusterTarget;
 use fluvio_extension_common::{OutputFormat, Terminal};
 use fluvio_sc_schema::mirror::{MirrorSpec, MirrorType};
-
-use self::output::HomeStatusRow;
 
 use super::get_admin;
 
@@ -49,6 +49,16 @@ impl StatusOpt {
     }
 }
 
+#[derive(Serialize)]
+struct HomeStatusRow {
+    home: String,
+    route: String,
+    sc_status: String,
+    spu_status: String,
+    last_seen: String,
+    errors: String,
+}
+
 mod output {
 
     //!
@@ -64,15 +74,7 @@ mod output {
     use fluvio_extension_common::output::TableOutputHandler;
     use fluvio_extension_common::t_println;
 
-    #[derive(Serialize)]
-    pub struct HomeStatusRow {
-        pub home: String,
-        pub route: String,
-        pub sc_status: String,
-        pub spu_status: String,
-        pub last_seen: String,
-        pub errors: String,
-    }
+    use super::HomeStatusRow;
 
     #[derive(Serialize)]
     struct TableList(Vec<HomeStatusRow>);
--- a/crates/fluvio-cli/src/client/remote/list.rs
+++ b/crates/fluvio-cli/src/client/remote/list.rs
@@ -1,13 +1,13 @@
-pub use std::sync::Arc;
+use std::sync::Arc;
 use std::time::SystemTime;
 
 use anyhow::Result;
 use clap::Parser;
+use serde::Serialize;
+
 use fluvio_extension_common::target::ClusterTarget;
 use fluvio_extension_common::{OutputFormat, Terminal};
 use fluvio_sc_schema::mirror::{MirrorSpec, MirrorType};
-
-use self::output::RemoteStatusRow;
 
 use super::get_admin;
 
@@ -16,6 +16,7 @@ pub struct ListOpt {
     #[clap(flatten)]
     output: OutputFormat,
 }
+
 impl ListOpt {
     pub async fn execute<T: Terminal>(
         self,
@@ -46,6 +47,15 @@ impl ListOpt {
     }
 }
 
+#[derive(Serialize)]
+struct RemoteStatusRow {
+    remote: String,
+    sc_status: String,
+    spu_status: String,
+    last_seen: String,
+    errors: String,
+}
+
 mod output {
 
     //!
@@ -61,14 +71,7 @@ mod output {
     use fluvio_extension_common::output::TableOutputHandler;
     use fluvio_extension_common::t_println;
 
-    #[derive(Serialize)]
-    pub struct RemoteStatusRow {
-        pub remote: String,
-        pub sc_status: String,
-        pub spu_status: String,
-        pub last_seen: String,
-        pub errors: String,
-    }
+    use super::RemoteStatusRow;
 
     #[derive(Serialize)]
     struct TableList(Vec<RemoteStatusRow>);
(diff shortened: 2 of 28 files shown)


crates/fluvio-cli/src/client/home/status.rs
```diff
@@ -24,16 +24,18 @@ impl StatusOpt {
         let list = admin.all::<MirrorSpec>().await?;
         let now = SystemTime::now().duration_since(SystemTime::UNIX_EPOCH)?;
 
-        let outlist: Vec<(String, String, String, String)> = list
+        let outlist: Vec<(String, String, String, String, String, String)> = list
```
sehz, line 27: this is bit too long of tuple.  should make it to struct

crates/fluvio-cli/src/client/remote/list.rs
```diff
@@ -24,13 +24,19 @@ impl ListOpt {
         let list = admin.all::<MirrorSpec>().await?;
         let now = SystemTime::now().duration_since(SystemTime::UNIX_EPOCH)?;
 
-        let outlist: Vec<(String, String, String)> = list
+        let outlist: Vec<(String, String, String, String, String)> = list
```
sehz, line 27: similar comment. make into struct

crates/fluvio-controlplane-metadata/src/mirror/status.rs
```diff
@@ -8,8 +8,9 @@ use fluvio_protocol::{Encoder, Decoder};
     serde(rename_all = "camelCase")
 )]
 pub struct MirrorStatus {
```
sehz, line 10: is there reason why this is public?

crates/fluvio-controlplane-metadata/src/mirror/status.rs
```diff
@@ -36,9 +68,11 @@ pub enum MirrorPairStatus {
     #[fluvio(tag = 1)]
     Succesful,
     #[fluvio(tag = 2)]
-    Failed,
+    Failed(String),
```
sehz, line 71: since this is protocol change, should bump up fluvio version

crates/fluvio-sc/src/controllers/mirroring/controller.rs
```diff
@@ -174,12 +96,81 @@ impl<C: MetadataItem> RemoteMirrorController<C> {
 
             debug!(
                 "sleeping {} seconds before next iteration",
-                MIRRORING_CONTROLLER_INTERVAL
+                MIRRORING_CONTROLLER_INTERVAL_MS
             );
-            sleep(Duration::from_secs(MIRRORING_CONTROLLER_INTERVAL)).await;
+            sleep(Duration::from_millis(MIRRORING_CONTROLLER_INTERVAL_MS)).await;
```
sehz, line 101: is there reason why this is so small?

crates/fluvio-sc/src/controllers/mirroring/controller.rs
```diff
@@ -174,12 +96,81 @@ impl<C: MetadataItem> RemoteMirrorController<C> {
...
     }
 
+    async fn request_stream(
+        &self,
+        home: &Home,
+        home_config: ClientConfig,
+    ) -> Result<AsyncResponse<ObjectMirroringRequest>> {
+        let versioned_socket = home_config.connect().await?;
+        let (socket, config, versions) = versioned_socket.split();
+        info!("connecting to home: {}", home.public_endpoint);
```
sehz, line 112: make it structured.
```
info!(endpoint = home.public_endpoint, "connection to home")
```

crates/fluvio-sc/src/controllers/mirroring/controller.rs
```diff
@@ -174,12 +96,81 @@ impl<C: MetadataItem> RemoteMirrorController<C> {
...
+
+        let request = MirrorConnect {
+            remote_id: home.remote_id.clone(),
+        };
+        debug!("sending connect request: {:#?}", request);
```
sehz, line 117: make it structured

crates/fluvio-spu-schema/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio-spu-schema"
-version = "0.15.0"
+version = "0.15.1"
 edition = "2021"
```
sehz, line 4: this probably needs to be bump up to `0.16.0`

crates/fluvio-spu/src/control_plane/message_sink.rs
```diff
@@ -3,26 +3,34 @@ use std::sync::Arc;
...
 use fluvio_controlplane::sc_api::update_lrs::LrsRequest;
+use fluvio_controlplane::sc_api::update_mirror::MirrorStatRequest;
 
 pub type SharedStatusUpdate = Arc<StatusMessageSink>;
+pub type SharedMirrorStatusUpdate = Arc<StatusMirrorMessageSink>;
 
 /// channel used to send message to sc
 #[derive(Debug)]
-pub struct StatusMessageSink(Mutex<HashSet<LrsRequest>>);
+pub struct MessageSink<R>(Mutex<HashSet<R>>);
```
sehz, line 13: is reason why this is generic instead of enum.  Since this is dynamic, how would receiver would distinguish?

crates/fluvio-spu/src/mirroring/remote/controller.rs
```diff
@@ -280,12 +286,26 @@ where
         Ok(())
     }
 
+    async fn update_status(&self, pair_status: MirrorPairStatus) -> Result<()> {
+        let now = SystemTime::now()
```
sehz, line 290: this is repeating.  probably show move to common util

crates/fluvio-spu/src/mirroring/remote/controller.rs
```diff
@@ -492,8 +518,8 @@ where
 
 fn create_backoff() -> ExponentialBackoff {
     ExponentialBackoffBuilder::default()
-        .min(Duration::from_secs(1))
-        .max(Duration::from_secs(300))
+        .min(Duration::from_millis(200))
```
sehz, line 521: this is probably too small.

sehz (changes requested):

crates/fluvio-cli/src/client/home/status.rs
fraidev: Sure! Using a struct now!

crates/fluvio-cli/src/client/remote/list.rs
fraidev: Sure! Using a struct now!

crates/fluvio-controlplane-metadata/src/mirror/status.rs
fraidev: It's been used in other crates. For example, we're using it on SPU crate now to send to SC.

crates/fluvio-controlplane-metadata/src/mirror/status.rs
fraidev: Sure, minor right? done

crates/fluvio-sc/src/controllers/mirroring/controller.rs
fraidev: This is the interval of remote SC connecting to home SC. It's a live forever connection that home SC response. So, it's very interesting this have a minor interval to connect as soon as possible as the home SC is handling the real interval, which now is 5 seconds.

Also, there is another interval of 10 seconds in the outside loop when there is an error (connection close, for example) or there is no remote registered.

I also notice a lot of `err: Socket io Broken pipe (os error 32)` on the home SC for the first requests, and for some reason this returns None instead of an error for the stream. Maybe a bug on our stream?

So with a small interval here, I noticed the first connection went quickly.

crates/fluvio-sc/src/controllers/mirroring/controller.rs
fraidev: Sure!

crates/fluvio-sc/src/controllers/mirroring/controller.rs
fraidev: Sure!

crates/fluvio-spu-schema/Cargo.toml
fraidev: Sure!

crates/fluvio-spu/src/mirroring/remote/controller.rs
fraidev: for min or max?

crates/fluvio-spu/src/control_plane/message_sink.rs
fraidev: Hmm, I think that we could use an enum if we have only one store and use only one function to dispatcher them. 

I added a new store for the mirror status and a new function (`send_mirror_status_back_to_sc`) slightly different from the partition (`LrsRequest`) status (`send_status_back_to_sc`).

what do you think?

crates/fluvio-spu/src/mirroring/remote/controller.rs
fraidev: Maybe as a method for `SharedMirrorStatusUpdate`?

crates/fluvio-spu/src/mirroring/home/connection.rs
```diff
@@ -102,13 +91,12 @@ impl MirrorHomeHandler {
                 "remote cluster not found: {}",
                 req_msg.request.remote_cluster_id
             );
-            return;
+            return Err(anyhow!("remote cluster not found"));
```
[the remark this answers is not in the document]
fraidev: Sure!

crates/fluvio-spu/src/mirroring/remote/controller.rs
fraidev: Increased min to 1s and max to 30s. Wdyt?

crates/fluvio-spu/src/mirroring/remote/controller.rs
fraidev: Added this to a new method in `StatusMirrorMessageSink`. what do you think?

crates/fluvio-spu/src/control_plane/message_sink.rs
fraidev: I think that one good reason to separate it into different stores is to respect the locks by context. wdyt?

fraidev: I renamed `StatusMessageSink` to `StatusLrsMessageSink` to make it easy to understand that this is not a generic store.

crates/fluvio-cli/src/client/remote/list.rs
```diff
@@ -55,22 +61,27 @@ mod output {
     use fluvio_extension_common::output::TableOutputHandler;
     use fluvio_extension_common::t_println;
 
-    type ListVec = Vec<(String, String, String)>;
+    #[derive(Serialize)]
+    pub struct RemoteStatusRow {
```
sehz, line 65: is this need to be public?

crates/fluvio-controlplane-metadata/src/mirror/status.rs
```diff
@@ -20,11 +21,42 @@ impl MirrorStatus {
...
+        self.connection_stat = other.connection_stat;
+    }
+
+    pub fn merge_from_spu(&mut self, other: Self) {
+        self.pairing_spu = other.pairing_spu;
+        self.connection_stat = other.connection_stat;
+    }
+
+    pub fn pair_errors(&self) -> String {
+        match (self.pairing_sc.clone(), self.pairing_spu.clone()) {
```
sehz, line 51: this is bit awkward.  clone should be done after match

crates/fluvio-controlplane-metadata/src/mirror/status.rs
sehz, line 71: This will break existing API.  Instead maybe
```
#[fluvio(tag = 2)]
Failed,
#[fluvio(tage=4]
DetailFailure(String)
```

And provide compatibility with previous version.  Need to increment protocol version as well

crates/fluvio-sc/src/controllers/mirroring/controller.rs
```diff
@@ -43,97 +44,29 @@ impl<C: MetadataItem> RemoteMirrorController<C> {
...
-                    "sleeping {} seconds try again",
-                    MIRRORING_CONTROLLER_RETRY_INTERVAL
-                );
+                if let Err(err) = self
+                    .update_status(MirrorPairStatus::Failed(err.to_string()))
+                    .await
+                {
+                    error!("error updating status: {:#?}", err);
+                }
                 sleep(Duration::from_secs(MIRRORING_CONTROLLER_RETRY_INTERVAL)).await;
```
sehz, line 56: should exponential retry

crates/fluvio-sc/src/controllers/mirroring/controller.rs
```diff
@@ -147,39 +80,97 @@ impl<C: MetadataItem> RemoteMirrorController<C> {
...
+                                self.sync_topic(&home, topic).await?;
                             }
 
-                            let status = MirrorStatus::new(
-                                MirrorPairStatus::Succesful,
-                                ConnectionStatus::Online,
-                                now as u64,
-                            );
-                            self.mirrors.update_status(home.id.clone(), status).await?;
+                            debug!("synced topics from home");
```
sehz, line 86: make it info

crates/fluvio-sc/src/services/public_api/mirroring/connect.rs
```diff
@@ -100,19 +99,35 @@ impl<AC: AuthContext, C: MetadataItem> RemoteFetchingFromHomeController<AC, C> {
...
+                        error!("error updating status: {}", err);
+                    }
+                }
                 break;
             }
 
-            trace!("{}: waiting for changes", self.req.remote_id);
             select! {
                 _ = self.end_event.listen() => {
                     debug!("connection has been terminated");
```
sehz, line 121: make int info

crates/fluvio-spu/src/control_plane/dispatcher.rs
```diff
@@ -131,6 +136,7 @@ impl ScDispatcher<FileReplica> {
 
                 _ = status_timer.next() =>  {
                     self.send_status_back_to_sc(&mut sink).await?;
+                    self.send_mirror_status_back_to_sc(&mut sink).await?;
```
sehz, line 139: why sending status twice?  Should consolidate into a single status

crates/fluvio-spu/src/mirroring/remote/controller.rs
```diff
@@ -242,34 +243,37 @@ where
             // update home if flag is set and we know what home leo is
             if home_updated_needed && home_leo >= 0 {
                 self.update_home(&mut home_sink, home_leo).await?;
+                self.update_status(MirrorPairStatus::Succesful).await?;
                 home_updated_needed = false;
             }
 
             select! {
+                _ = leader_offset_listner.listen() => {
+                    debug!("leader offset has changed, home cluster needs to be updated");
```
sehz, line 252: make this info

sehz (changes requested):

crates/fluvio-cli/src/client/remote/list.rs
fraidev: I moved it to outside of this module, so it's needed. I did the same for `home/status.rs` too.

crates/fluvio-controlplane-metadata/src/mirror/status.rs
fraidev: Changing `&self` to just `self` does not even require clones, fixed!

crates/fluvio-sc/src/controllers/mirroring/controller.rs
fraidev: Fixed!

crates/fluvio-sc/src/controllers/mirroring/controller.rs
fraidev: Done!

crates/fluvio-sc/src/services/public_api/mirroring/connect.rs
fraidev: the debug one, right? Fixed

crates/fluvio-spu/src/mirroring/remote/controller.rs
fraidev: Fixed!

crates/fluvio-spu/src/control_plane/dispatcher.rs
fraidev: `send_status_back_to_sc` is only sending lrs status. 
I renamed this function to `send_lrs_status_back_to_sc` to make more sense.

crates/fluvio-controlplane-metadata/src/mirror/status.rs
fraidev: Sure! COMMON_VERSION is the protocol version, right?

So would be something like it?

`pub(crate) const COMMON_VERSION: i16 = 17;`

and

```
#[fluvio(tag = 5, min_version = 17)]
DetailFailure(String),
```

sehz (approved):

PR #4118, opened 2024-07-25
chore: use Rust `1.80.0`

EstebanBorai:
Updates Rust Toolchain to `1.80.0` and applies clippy improvements

PR #4121, opened 2024-07-28
fix: enable tracing subscribe again

fraidev:
I think that had a refactoring removing not used features flags, but I noticed that it was removed a block of code enabling tracing subscribe when `telemetry` feature was disabled.

I noticed this because only prints were working, all the tracings functions stopped working.

Related: https://github.com/infinyon/fluvio/pull/4118.


EstebanBorai (approved):
Good catch @fraidev!

PR #4122, opened 2024-07-29
feat: allow setting local addr on custom spu

nacardin:


crates/fluvio-controlplane-metadata/src/spu/spec.rs
```diff
@@ -137,6 +137,8 @@ pub struct CustomSpuSpec {
     pub private_endpoint: Endpoint,
     #[cfg_attr(feature = "use_serde", serde(skip_serializing_if = "Option::is_none"))]
     pub rack: Option<String>,
+    #[cfg_attr(feature = "use_serde", serde(skip_serializing_if = "Option::is_none"))]
+    pub public_endpoint_local: Option<Endpoint>,
```
sehz, line 141: why need to add here?

nacardin: This is used in the command to register the custom SPU

sehz: It is using CustomSpuSpec,  don't think there is need to add to `SpuSpec`

nacardin: `SpuSpec` already has the `public_endpoint_local` field.
This PR adds the field to `CustomSpuSpec`.

sehz: Got it. got confused about diff

sehz (approved):

EstebanBorai (approved):

PR #4135, opened 2024-08-13
chore: update wasmtime and wasi-common to 23.0.2

fraidev:
This version fixed the cross-compilation bug. 

Related:
- https://github.com/bytecodealliance/wasmtime/issues/8898
- https://github.com/bytecodealliance/wasmtime/pull/9100

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1335,21 +1335,32 @@ dependencies = [
 
 [[package]]
 name = "cranelift-bforest"
-version = "0.109.0"
+version = "0.110.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


digikata:
We should probably stay in sync w/ SDF on these dep versions

morenol (approved):

fraidev:
> We should probably stay in sync w/ SDF on these dep versions, but maybe we can merge when sdf goes to 23 too?

I **think** that is ok to merge because SDF is already using two diferent wasmtime versions. 
It's using the wasmtime 21 and Fluvio 0.11.11 that uses wasmtime 22.

When we update SDF to Fluvio 0.11.12 we could update wasmtime to 23 too.

digikata (approved):

PR #4146, opened 2024-08-21
chore: clean up not used dependencies

fraidev:
Using [cargo-machete](https://github.com/bnjbvr/cargo-machete) I notice some dependencies that we don't use.

I also added a CI step to check them in that PR: https://github.com/infinyon/fluvio/pull/4148/commits

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -826,7 +826,6 @@ dependencies = [
  "anyhow",
  "cargo_metadata",
  "derive_builder",
- "thiserror",
 ]
 
 [[package]]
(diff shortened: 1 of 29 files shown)


sehz (changes requested):
should separate out PR into two:
1) change cargo dep version
2) adding dep check.

fraidev:
> should separate out PR into two:
> 
> 1. change cargo dep version
> 2. adding dep check.

Sure! The CI part is here: https://github.com/infinyon/fluvio/pull/4148

sehz (approved):
nice cleanup

Issue #3836, opened 2024-01-25
When a topic is deleted, connected clients should have their connection closed


digikata:
Both producers and consumers should have their connections closed.

Preferably with a specific error (TopicDeleted or TopicNotAvailable).

This is an task of intermediate difficulty, mentoring from Infinyon is available if you are interested in working on this issue.

jdafont:
Howdy, thanks for the assistance so far in Discord for helping me figure my way through this. So far, a viable solution to me is looking like the following, but would love to get some thoughts/gut checks before implementing:

In `crates/fluvio-spu/src/services/public/stream_fetch.rs` `async fn process (...) -> Result<(), StreamFetchError>`, we do a `select!` in a loop. Here, we could listen to a new `ReplicaDeleted` event. When this event is selected, we check if the replica deleted matches our own, and if so we can simply return a `StreamFetchError::Fetch(...)` with a new appropriate error code like `ReplicaDeleted` with the appropriate error message. When the error is returned, `send_back_error` is called. I think this might be enough to terminate the connection, but it's not totally obvious to me because we return `Ok(())` here, which bubbles up to a loop at `crates/fluvio-spu/src/services/public/mod.rs` `async fn respond(...)`.

The next question is how to emit the event whenever the SC sends over a deleted topic notification. Both the consumer connection and SC dispatcher connections have access to the shared global context `ctx`. The dispatcher processes replica changes and eventually works it's way down into `crates/fluvio-spu/src/core/global_context.rs` `async fn apply_replica_actions(...)`. Here, we process replica deletions for both followers and leaders. Currently, it seems like only leaders handle file stream fetch requests. However, in either case, we can emit the replica deleted signal whenever we process the replica deletion typically. 

Sorry for the wall of text! Let me know if that sounds like a good plan and if I should continue with an implementation.

sehz:
Just to clarify,  there are two I/O to fluvio cluster.   First is to SC which manages metadata and second to SPU which handles actual consumer/producer.   SC should not be handling SPU connection.  It is responsibility of SPU to manage its own connection.   This is fundamentally how Fluvio works.

Here's suggestion;  just add new `ErrorCode` variant.    In the SPU,`ScDispatcher` can expire OffsetChangeLister which is used by `StreamFetchHandler`.    Then `StreamFetchHandler` can detect then and send appropriate ErrorCode.    This should work because consumer always talks to leader who manages replica.   

Thanks for working on this, this is not trivial excercise.

[...]

PR #4147, opened 2024-08-21
feat: close producer when topic is deleted

fraidev:
Related: https://github.com/infinyon/fluvio/issues/3836

We already close the consumer when the topic is deleted, now we will close the producer too.

Given:
```bash
fluvio produce my-topic   
```

When:
```bash
fluvio topic delete my-topic   
```

Then:
```bash
Topic "my-topic" was deleted
```

### DIFF
--- a/crates/fluvio-cli/Cargo.toml
+++ b/crates/fluvio-cli/Cargo.toml
@@ -39,6 +39,7 @@ producer-file-io = ["fluvio-cli-common/file-records"]
 [dependencies]
 async-channel = { workspace = true }
 async-trait = { workspace = true }
+async-std = { workspace = true }
 anyhow = { workspace = true }
 bytesize = { workspace = true, features = ['serde'] }
 clap = { workspace = true, features = ["std", "derive", "string", "help", "usage", "env", "error-context"] }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2440,6 +2440,7 @@ version = "0.0.0"
 dependencies = [
  "anyhow",
  "async-channel 1.9.0",
+ "async-std",
  "async-trait",
  "bytesize",
  "clap",
--- a/crates/fluvio-cli/src/client/produce/mod.rs
+++ b/crates/fluvio-cli/src/client/produce/mod.rs
@@ -11,17 +11,23 @@ mod cmd {
     #[cfg(feature = "producer-file-io")]
     use std::path::PathBuf;
 
+    use async_std::io::stdin;
     use async_trait::async_trait;
+    use fluvio_future::io::StreamExt;
+    use fluvio_sc_schema::message::MsgType;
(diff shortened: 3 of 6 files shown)


crates/fluvio-cli/Cargo.toml
```diff
@@ -39,6 +39,7 @@ producer-file-io = ["fluvio-cli-common/file-records"]
 [dependencies]
 async-channel = { workspace = true }
 async-trait = { workspace = true }
+async-std = { workspace = true }
```
sehz, line 42: don't bring in `async-std`.

sehz (changes requested):
I don't think this is right approach.  This should work for all client not just CLI.  client should detect if topic is closed then close connection

fraidev:
> I don't think this is right approach. This should work for all client not just CLI. client should detect if topic is closed then close connection

Make sense! But the producer is never what is blocking. 
In this case, the blocker is the stdin and for connectors the source stream that is blocking:

```rust
#[connector(source)]
async fn start(config: CustomConfig, producer: TopicProducerPool) -> Result<()> {
    let source = TestJsonSource::new(&config)?;
    let mut stream = source.connect(None).await?;
    while let Some(item) = stream.next().await {
        println!("producing a value: {}", &item);
        producer.send(RecordKey::NULL, item).await?;
    }
    Ok(())
}
```

Maybe we can change TopicProducerPool to detect an error when the topic is deleted and also add a method to detect it, but we will also need to add something like `select!` and change uses of the producer.

Another solution that I had is have a method that receives a stream and returns a stream that handles it.

```bash    
async fn from_stream(stream: GenericStream) -> ProducerStream
```

But we will need change implementations too:

```rust
#[connector(source)]
async fn start(config: CustomConfig, producer: TopicProducerPool) -> Result<()> {
    let source = TestJsonSource::new(&config)?;
    let mut stream = source.connect(None).await?;
    while let Some(item) = producer.from_stream(stream).await {
        println!("producing a value: {}", &item);
        producer.send(RecordKey::NULL, item).await?;
    }
    Ok(())
}
```

What do you think?

digikata:
Is cluster side socket already closed by the SPU?  I think that's the first thing to handle for this case.  In addition, if possible it would be nice for the SPU to send a message that the produce stream is being closed due to topic delete, and then close the spu side socket.

fraidev:
> Is cluster side socket already closed by the SPU? I think that's the first thing to handle for this case. In addition, if possible it would be nice for the SPU to send a message that the produce stream is being closed due to topic delete, and then close the spu side socket.

Yeah, I think so. The SPU already send a `TopicNotFound` error when does not find it.

But I think that we will have the same problem. All usage of producer seems to be:

- Await an input
- Produce it

The `Produce it` part will return the SPU error, but it will be trigger only after the input.

I think that this issue aims to be notified before the input.

PR #4149, opened 2024-08-22
fix: better error when watch api versions are not compatible

fraidev:
Related: https://github.com/infinyon/fluvio/issues/3441

### DIFF
--- a/crates/fluvio/src/lib.rs
+++ b/crates/fluvio/src/lib.rs
@@ -46,6 +46,7 @@ use tracing::instrument;
 
 /// The minimum VERSION of the Fluvio Platform that this client is compatible with.
 const MINIMUM_PLATFORM_VERSION: &str = "0.9.0";
+pub(crate) const VERSION: &str = include_str!("../../../VERSION");
 
 /// Creates a producer that sends records to the named topic
 ///
--- a/crates/fluvio/src/admin.rs
+++ b/crates/fluvio/src/admin.rs
@@ -135,7 +135,9 @@ impl FluvioAdmin {
                 metadata,
             })
         } else {
-            Err(anyhow!("WatchApi version not found"))
+            let platform_version = versions.platform_version();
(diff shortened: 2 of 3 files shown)


crates/fluvio/src/lib.rs
```diff
@@ -46,6 +46,7 @@ use tracing::instrument;
 
 /// The minimum VERSION of the Fluvio Platform that this client is compatible with.
 const MINIMUM_PLATFORM_VERSION: &str = "0.9.0";
+pub(crate) const VERSION: &str = include_str!("../../../VERSION");
```
digikata, line 49: did we have this somewhere already?   Seems like we should name this PLATFORM_VERSION

fraidev: Yeah, we have this same const in many places: 
image

fraidev: Sure, rename only in this case?

fraidev: About `MINIMUM_PLATFORM_VERSION` it's been checked after the lookup_version of the watch API. And will return it: 

```bash
    #[error("Cluster (with platform version {cluster_version}) is older than the minimum required version {client_minimum_version}
To interact with this cluster, please install the matching CLI version using the following command:
    curl -fsS https://hub.infinyon.cloud/install/install.sh | VERSION={cluster_version} bash
    ")]
```

But as we're not updating this const does not seem so useful.

fraidev: Should I remove it?

digikata: Hmm, let's just leave it all alone until we get a improved wider compatability check.  

Lets leave leaving the constant as VERSION.   Forget renaming it  PLATFORM_VERSION as it would be out of place with all the other crates.

digikata (approved):
LGTM

PR #4151, opened 2024-08-28
feat(hub): provide `wasm32-unknown-unknown` support for `hub-util`

EstebanBorai:
Provides `fluvio-hub-util` support for target `wasm32-unknown-unknown`.
Using the feature `wasm` certain `HubAccess` utilities wont be available.

### DIFF
--- a/crates/fluvio-hub-util/src/fvm/api/download.rs
+++ b/crates/fluvio-hub-util/src/fvm/api/download.rs
@@ -43,6 +43,12 @@ pub trait Download {
 
 #[async_trait]
 impl Download for Artifact {
+    #[cfg(target_arch = "wasm32")]
+    async fn download(&self, target_dir: PathBuf) -> Result<PathBuf> {
+        unimplemented!("The download support is not implemented for wasm32 architecture");
+    }
+
+    #[cfg(not(target_arch = "wasm32"))]
     #[instrument(skip(self, target_dir))]
     async fn download(&self, target_dir: PathBuf) -> Result<PathBuf> {
         tracing::info!(
--- a/crates/fluvio-hub-util/src/fvm/api/download.rs
+++ b/crates/fluvio-hub-util/src/fvm/api/download.rs
@@ -43,12 +43,6 @@ pub trait Download {
 
 #[async_trait]
 impl Download for Artifact {
-    #[cfg(target_arch = "wasm32")]
-    async fn download(&self, target_dir: PathBuf) -> Result<PathBuf> {
-        unimplemented!("The download support is not implemented for wasm32 architecture");
-    }
-
-    #[cfg(not(target_arch = "wasm32"))]
     #[instrument(skip(self, target_dir))]
     async fn download(&self, target_dir: PathBuf) -> Result<PathBuf> {
         tracing::info!(
--- a/crates/fluvio-hub-util/src/hubaccess.rs
+++ b/crates/fluvio-hub-util/src/hubaccess.rs
@@ -1,26 +1,41 @@
 use std::path::{Path, PathBuf};
 
+#[cfg(not(target_arch = "wasm32"))]
 use anyhow::anyhow;
+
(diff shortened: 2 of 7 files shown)


EstebanBorai:
@digikata seems to me that `--all-features` CI test is failing due to the fact that `wasm` feature conflicts with others given that some functions are missing for the `wasm` variant.

I think we could use a matrix based test? And ensure that each feature compiles?

digikata:
> @digikata seems to me that `--all-features` CI test is failing due to the fact that `wasm` feature conflicts with others given that some functions are missing for the `wasm` variant.
> 
> I think we could use a matrix based test? And ensure that each feature compiles?

Can we setup the regular build w/o wasm and a side build for wasm.  Maybe a  ci-wasm workflow to build a fluvio wasm client?

EstebanBorai:
> > @digikata seems to me that `--all-features` CI test is failing due to the fact that `wasm` feature conflicts with others given that some functions are missing for the `wasm` variant.
> 
> > 
> 
> > I think we could use a matrix based test? And ensure that each feature compiles?
> 
> 
> 
> Can we setup the regular build w/o wasm and a side build for wasm.  Maybe a  ci-wasm workflow to build a fluvio wasm client? (trying to avoid doubling the ci time)

Thats a nice alternative, will do an update using the `target` instead of the feature.

EstebanBorai:
@digikata now we use the `target_arch` as a conditional to disable certain non-compatible
WASM32 features.

Reqwest is used in place of Ureq for WASM32 to perform HTTP requests but we still return `http::Response` instances, maintaining same signatures as with remaining architectures.

crates/fluvio-hub-util/src/hubaccess.rs
```diff
@@ -50,12 +65,18 @@ impl HubAccess {
...
+    #[cfg(not(target_arch = "wasm32"))]
     pub fn default_load(remote: &Option<String>) -> Result<Self> {
         let cfgpath = default_cfg_path()?;
         let profileopt = std::env::var(FLUVIO_HUB_PROFILE_ENV).ok();
         HubAccess::load_path(cfgpath, profileopt, remote)
     }
 
+    #[cfg(target_arch = "wasm32")]
+    pub fn default_load(_remote: &Option<String>) -> Result<Self> {
+        unimplemented!("Support for `HubAccess::default_load` is not available for `wasm32`");
```
digikata, line 77: should this panic, or  just return an error?

crates/fluvio-hub-util/src/fvm/api/download.rs
```diff
@@ -43,6 +43,12 @@ pub trait Download {
 
 #[async_trait]
 impl Download for Artifact {
+    #[cfg(target_arch = "wasm32")]
+    async fn download(&self, target_dir: PathBuf) -> Result<PathBuf> {
+        unimplemented!("The download support is not implemented for wasm32 architecture");
+    }
+
+    #[cfg(not(target_arch = "wasm32"))]
```
digikata, line 51: Since the entire fvm module is omitted for targeting wasm32, are these cfg lines needed?

digikata (approved):
LGTM other than a minor question

crates/fluvio-hub-util/src/hubaccess.rs
EstebanBorai: I think returning error its fine as well!

crates/fluvio-hub-util/src/fvm/api/download.rs
EstebanBorai: I dont think so, It could be left behind while testing on it against a WASM target locally. I could give a second check.

EstebanBorai:
@digikata suggestions addressed here: https://github.com/infinyon/fluvio/pull/4151/commits/f918ed2e884518e4e14787ce895c7f5058f64dbe

PR #4163, opened 2024-09-02
fix: update sysinfo and fix fd leak on linux

fraidev:
Update sysinfo to 0.31.4 and fix file descriptor leak on Linux. 
`sysinfo` leaks file descriptors on Linux if the limit is not set.

Spu Before:

```
▷ ls /proc/570230/fd/ | wc -l
1114
```

Spu After:
```
▷ ls /proc/574099/fd/ | wc -l
19
```

### DIFF
--- a/crates/fluvio-cluster/src/cli/diagnostics.rs
+++ b/crates/fluvio-cluster/src/cli/diagnostics.rs
@@ -262,6 +262,7 @@ impl DiagnosticsOpt {
             Ok(())
         };
 
+        sysinfo::set_open_files_limit(0);
         let mut sys = System::new_all();
         let mut net = Networks::new();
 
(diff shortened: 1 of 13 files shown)


crates/fluvio-cluster/src/cli/diagnostics.rs
```diff
@@ -452,13 +453,17 @@ impl ProcessInfo {
...
-                processes.push(ProcessInfo {
-                    pid: pid.as_u32(),
-                    name: process.name().to_string(),
-                    disk_usage: format!("{:?}", process.disk_usage()),
-                    cmd: format!("{:?}", process.cmd()),
-                });
+            let process_name = process.name().to_str();
+
+            if let Some(process_name) = process_name {
+                if process_name.contains("fluvio") {
```
sehz, line 459: can you make `fluvio` as constant?

crates/fluvio-cluster/src/delete.rs
```diff
@@ -149,16 +150,23 @@ impl ClusterUninstaller {
         pb.set_message("Uninstalling fluvio local components");
 
         let kill_proc = |name: &str, command_args: Option<&[String]>| {
+            sysinfo::set_open_files_limit(0);
             let mut sys = System::new();
-            sys.refresh_processes(); // Only load what we need.
-            for process in sys.processes_by_exact_name(name) {
+            sys.refresh_processes(sysinfo::ProcessesToUpdate::All); // Only load what we need.
```
sehz, line 155: this seems like repeating. can you change into common util?

sehz (changes requested):
LGTM.  Couple of minor nits

crates/fluvio-cluster/src/delete.rs
fraidev: Sure! I created a `process` mod.

crates/fluvio-cluster/src/cli/diagnostics.rs
fraidev: Done!

sehz (approved):

Issue #4166, opened 2024-09-04
[Bug]: `fluvio profile export` should export config format by default


digikata:
While fluvio profile export supports a number of output formats, it currently does not export in a format readable by a fluvio client as a regular config file, for example with environment variable FLV_PROFILE_PATH set.

The desired workflow would be:
```
fluvio profile export <PROFILE_NAME> > config-exported

# in a new client environment
export FMV_PROFILE_PATH=/path/to/config-exported
# run a fluvio client or cli
fluvio topic list
```

This is useful for cases where a user wants to deploy a fluvio client and wants to export local development profiles: e.g.
https://github.com/infinyon/fluvio/issues/3736

PanGan21:
Hi looking at this issue. From what I understand the goal is to export to a format that is supported by fluvio clients.
Is it yml format? Also is the new env variable `FLV_PROFILE_PATH` where the location of the export?
Thanks in advance.

digikata:
Hi @PanGan21, nice to see you again.  It's actually a toml format.  Here is how the config files are normally written: https://github.com/infinyon/fluvio/blob/acf2f94e1f3a269bde4ab915457dba87571420a6/crates/fluvio-types/src/config_file.rs#L38   

The file itself is normally found in `~/.fluvio/config`, but that often contains an accumulation of different profiles - the `fluvio profile export` command selects one and exports just that one profile.  It would be nice to get the default to be  output config toml, while keeping the other output formats (which are sometimes useful for other deployment automation).

PanGan21:
Thanks, I think it is clear. Can I give it a try?

PanGan21:
In any case, I prepared a pr for this in #4172
Please let me know if it covers the issue!

digikata:
@PanGan21 there was a little modification needed when we added a test, but your pr was key to getting this all working.  Closed as completed. Thanks @PanGan21!

PanGan21:
> @PanGan21 there was a little modification needed when we added a test, but your pr was key to getting this all working. Closed as completed. Thanks @PanGan21!

Thank you too for completing the ticket!

PR #4172, opened 2024-09-07
fix: implement profile export toml and make it default option

PanGan21:
Related: #4166 
- Implements exporting profile in toml format
- Makes toml format default when exporting


digikata (approved):
LGTM, Thanks @PanGan21!

PR #4173, opened 2024-09-09
chore: enable again local_upgrade_test

fraidev:


sehz:
Let's create `upgrade-test-local.sh` instead of using `upgrade-local.sh` since that's designed for k8 and and we need more asserting

fraidev:
> Let's create `upgrade-test-local.sh` by cloning `upgrade-local.sh` since that's designed for k8 and and we need more asserting

Sure! Done.

fraidev:
https://github.com/infinyon/fluvio/pull/4175 should cover it better

PR #4179, opened 2024-09-12
fix: profile export, full config, add cli test

digikata:
Fixes https://github.com/infinyon/fluvio/issues/4166

Add test and fix export format for toml to full config file


fraidev (approved):
LGTM!

PR #4189, opened 2024-09-17
feat(fvm): support for infinyon ci context

EstebanBorai:
Makes FVM context aware by using the `INFINYON_CI_CONTEXT` env var.

### DIFF
--- a/crates/fluvio-hub-util/src/fvm/api/client.rs
+++ b/crates/fluvio-hub-util/src/fvm/api/client.rs
@@ -1,10 +1,15 @@
 //! Hub FVM API Client
 
+use std::env::var;
+
 use anyhow::{Error, Result};
 use serde::{Deserialize, Serialize};
 use url::Url;
 
 use crate::fvm::{Channel, PackageSet, PackageSetRecord};
+
+const INFINYON_CI_CONTEXT: &str = "INFINYON_CI_CONTEXT";
+const CONTEXT_QUERY_PARAM: &str = "ctx";
 
 #[derive(Debug, Deserialize, Serialize)]
 pub struct ApiError {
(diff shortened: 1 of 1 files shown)


crates/fluvio-hub-util/src/fvm/api/client.rs
```diff
@@ -89,20 +108,36 @@ mod tests {
...
             url.as_str(),
-            "https://hub.infinyon.cloud/hub/v1/fvm/pkgset/stable?arch=arm-unknown-linux-gnueabihf"
+            "https://hub.infinyon.cloud/hub/v1/fvm/pkgset/stable?arch=arm-unknown-linux-gnueabihf",
+            "failed on Scenario Using Stable Channel"
         );
-    }
 
-    #[test]
-    fn builds_url_for_fetching_pkgsets_on_version() {
+        // Scenario: Using Tag
```
EstebanBorai, line 125: Centralized `builds_url_...` tests because using `set_var` is not thread safe, so each scenario is centralized in a single test with multiple assertions.

crates/fluvio-hub-util/src/fvm/api/download.rs
```diff
@@ -32,26 +34,49 @@ async fn checksum(artf: &Artifact, path: &PathBuf) -> Result<()> {
...
 pub trait Download {
     /// Downloads the artifact to the specified directory
     ///
     /// Internally validates the checksum of the downloaded artifact
     /// and returns the path to the downloaded artifact
-    async fn download(&self, target_dir: PathBuf) -> Result<PathBuf>;
+    async fn download(
+        &self,
+        target_dir: PathBuf,
+        url_query_params: Option<UrlQueryParams>,
```
digikata, line 59: I don't think we should add query_params to the artifact downloads.  These are served directly by public S3 buckets not the hub, and the service may not allow arbitrary query params.  We basically capture all we need from just the find pkgset event.

crates/fluvio-hub-util/src/htclient.rs
```diff
@@ -1,13 +1,85 @@
...
+        Self(Vec::new())
+    }
+
+    pub fn insert<T: ToString>(&mut self, key: T, value: T) {
+        self.0.push((key.to_string(), value.to_string()));
+    }
+}
+
+impl Display for UrlQueryParams {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
digikata, line 63: we already have the url crate, could we use the url. form_urlencoded serializer/deserializer instead of reimplmenting this?  Though if we don't inject them to the artifact download, none of this may be needed?

crates/fluvio-hub-util/src/fvm/api/download.rs
EstebanBorai: Thats right! I will aim this to PkgSet endpoint instead.

crates/fluvio-hub-util/src/htclient.rs
EstebanBorai: Of course! Will take a look on form_urlencoded!

digikata (approved):
LGTM

PR #4191, opened 2024-09-25
experiment: fix compressed batch size

digikata:
- **add: bats test for batch size w/ compression**
- **fix: use estimated compress ratio for batch size est**
- **add: record header attribute check**
- **fix: fmt**


digikata:
fixed in other pr

PR #4193, opened 2024-09-25
fix: use `Compression` as type and introduce `CompressionExt` with algorithms impl

EstebanBorai:
Disables compression for Connector Package as its using types for conn meta and not
compression algorithms themselves.

Fluvio Compression will now use `compress` and `types` features by default, but if you need only types for use cases like configurations, then you can turn off default features and use `types` feature only instead.

### DIFF
--- a/crates/fluvio-compression/Cargo.toml
+++ b/crates/fluvio-compression/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio-compression"
-version = "0.3.3"
+version = "0.3.4"
 edition = "2021"
 license = "Apache-2.0"
 authors = ["Fluvio Contributors <team@fluvio.io>"]
@@ -10,7 +10,7 @@ repository = "https://github.com/infinyon/fluvio"
 description = "Fluvio Compression library"
 
 [features]
-default = ["compress"]
+default = ["compress", "types"]
 compress = [
     "gzip",
     "lz4",
@@ -21,6 +21,7 @@ gzip = ["dep:flate2"]
 lz4 = ["dep:lz4_flex"]
 snap = ["dep:snap"]
 zstd = ["dep:zstd"]
+types = []
 
 [dependencies]
 serde = { workspace = true,  features = ['derive'] }
--- a/crates/fluvio-compression/Cargo.toml
+++ b/crates/fluvio-compression/Cargo.toml
@@ -10,7 +10,7 @@ repository = "https://github.com/infinyon/fluvio"
 description = "Fluvio Compression library"
 
 [features]
-default = ["compress", "types"]
+default = ["compress"]
 compress = [
     "gzip",
     "lz4",
@@ -21,7 +21,6 @@ gzip = ["dep:flate2"]
 lz4 = ["dep:lz4_flex"]
 snap = ["dep:snap"]
 zstd = ["dep:zstd"]
-types = []
 
 [dependencies]
 serde = { workspace = true,  features = ['derive'] }
--- a/crates/fluvio-compression/Cargo.toml
+++ b/crates/fluvio-compression/Cargo.toml
@@ -28,6 +28,8 @@ serde = { workspace = true,  features = ['derive'] }
 thiserror = { workspace = true }
 bytes = { workspace = true }
 
+fluvio-types = { workspace = true }
+
 # Optional Dependencies
 flate2 = { workspace = true, optional = true }
 lz4_flex = { version = "0.11.1", default-features = false, features = ["safe-decode", "safe-encode", "frame"], optional = true }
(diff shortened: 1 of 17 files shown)


digikata (approved):
Looking at the cargo-tree deps it seems reasonable to me.

sehz:
maybe more explanation?

crajcan:
> maybe more explanation?

This crate only uses types from fluvio-compression, but is not intended to invoke the compression algorithms themselves. 

By not including the compression algorithms, we can compile this crate for wasm targets successfully.

sehz (approved):

crates/fluvio-types/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio-types"
-version = "0.5.0"
+version = "0.6.0"
```
sehz, line 3: no need to bump to `0.6.0` since this doesn't change existing api.   Can bump to `0.5.1`

crates/fluvio-compression/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "fluvio-compression"
-version = "0.3.3"
+version = "0.3.4"
```
sehz, line 3: since API changes, this should be bump into `0.4.0`

crates/fluvio-protocol/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "fluvio-protocol"
 edition = "2021"
-version = "0.11.0"
+version = "0.11.1"
```
sehz, line 4: bump to 0.12.0

crates/fluvio-types/Cargo.toml
EstebanBorai: Of course!

crates/fluvio-protocol/Cargo.toml
```diff
@@ -61,7 +61,7 @@ tracing = { workspace = true }
 fluvio-protocol-derive = { version = "0.5.0", path = "../fluvio-protocol-derive", optional = true }
 fluvio-future = { workspace = true, optional = true }
 flv-util = { workspace = true,  optional = true }
-fluvio-compression = { path = "../fluvio-compression", version = "0.3.0", default-features = false, optional = true }
+fluvio-compression = { path = "../fluvio-compression", version = "0.3.0", default-features = false, features = ["compress"], optional = true }
```
sehz, line 64: seems like unnecessary feature flag.  compression need to compress

EstebanBorai: I was keeping it to reduce amount of changes, will give a shot on removing it!

crates/fluvio-storage/src/iterators.rs
```diff
@@ -5,9 +5,9 @@ use std::io::{Error as IoError, ErrorKind, Cursor};
 
 use nix::sys::uio::pread;
 
+use fluvio_compression::CompressionExt;
```
sehz, line 8: why needed here?

EstebanBorai: Because we have some compression logic going on here: https://github.com/infinyon/fluvio/pull/4193/files/7441509eae14aa0900cfaf68d81afd07e9dc4350#diff-eedcb248ce7c084703163bcd1f43eb4595ed0af4bdaf068a2f5bc7b9228cc7e4R146

crates/fluvio-types/src/compression.rs
```diff
@@ -0,0 +1,79 @@
...
+pub const COMPRESSION_SNAPPY: &str = "snappy";
+pub const COMPRESSION_LZ4: &str = "lz4";
+pub const COMPRESSION_ZSTD: &str = "zstd";
+
+/// The compression algorithm used to compress and decompress records in fluvio batches
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Hash)]
+#[serde(rename_all = "lowercase")]
+#[repr(i8)]
+#[derive(Default)]
+pub enum Compression {
```
sehz, line 28: Seems like this shouldn't belong here.  This is compression coding.

crajcan: I thought this was what @digikata suggested this morning in slack.

We need a way for crates to include these types without including the compression algorithms.

This PR first attempted to accomplished this with the `types` feature on `fluvio-compression`. 

Then I thought we agreed to instead move `enum Compression` to `fluvio-types`.

EstebanBorai:
Alternate approach provided in #4194

EstebanBorai:
Surpassed by #4194

PR #4196, opened 2024-09-26
chore: decrease exponential timeout of mirroring

fraidev:
That should improve to not sleeping a lot at first seconds trying to connect. 
Also helping our mirroring tests be more reliable.


digikata (commented):
LGTM, but should we leave the max exp backoff, it seems like especially with an IoT network there could be fairly slow responses or dropped packets?

sehz:
please add description on what this PR solves

digikata (approved):

PR #4199, opened 2024-09-27
chore: update `bytes` to `1.7.2`

EstebanBorai:


sehz (commented):
should bump up fluvio version if dep changed

sehz (approved):

PR #4201, opened 2024-10-01
feat(hub): `published at` method for `PackageMeta`

EstebanBorai:
Provides a `published_at` method for `PackageMeta` to retrieve the `PkgTag` for metadata
on publish date.

### DIFF
--- a/crates/fluvio-hub-protocol/src/package_meta.rs
+++ b/crates/fluvio-hub-protocol/src/package_meta.rs
@@ -1,6 +1,5 @@
 use std::default::Default;
 
-use chrono::{DateTime, Utc};
 use serde::{Deserialize, Serialize};
 use tracing::{info, error};
 use url::Url;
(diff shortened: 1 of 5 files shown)


crates/fluvio-hub-protocol/src/package_meta.rs
```diff
@@ -195,6 +196,23 @@ impl PackageMeta {
             None
         }
     }
+
+    pub fn published_at(&self) -> Result<DateTime<Utc>> {
```
digikata, line 200: Instead of adding this to `fluvio-hub-protocol` as an impl fn of PackageMeta, can it be added to `fluvio-hub-util` to the `PackageMetaExt`  This keeps the deps in the protocol crate lower and less complex.

digikata (changes requested):

crates/fluvio-hub-protocol/src/package_meta.rs
EstebanBorai: Makes sense!

digikata (approved):
LGTM

Issue #4161, opened 2024-08-31
Checking for max batch size  ignores compression


sehz:
Steps to reproduce:

1) create a topic:  `fluvio topic create   test`
2) try to send large json greater than 128k but reduces to 4k.
`fluvio produce test --file large.json --raw  --linger 0ms --compression  zstd`

This result in error:
```
the given record is larger than the buffer max_size (16384 bytes). Try increasing the producer batch size or reducing the record size enabling a compression algorithm
```

even thought compressed record is small 4k

digikata:
It might be a good to also modify the batching flow logic a little too in correcting this bug?

If the compressed record is larger than the batch size, send the accumulated batch (w/o the record), and then if the compressed record is still larger than the batch size, send it individually, and start an new batch.

PR #4202, opened 2024-10-02
docs: produce message size rfc

fraidev:
# TL;DR

`batch_size` producer config must not reject large records, just send them directly.

Create a new `max_request_size` producer config that must reject large messages.
I am using `max_request_size` because Kafka uses `max.request.size` but we can change it to other config name.

Compression sizes should not be used for these producer configs.

Related: 
- https://github.com/infinyon/fluvio/issues/4161
- https://github.com/infinyon/fluvio/pull/4195

### DIFF
--- a/rfc/produce-message-size.md
+++ b/rfc/produce-message-size.md
@@ -1,0 +1,99 @@
+# Enhancing Fluvio to Message Sizes
+
+This RFC proposes modifications to Fluvio's handling of message sizes. 
+
+## Proposed Enhancements
+
+1. Handling Larger Messages than Batch Size
+
+If a single record exceeds the defined `batch_size`, Fluvio will process the record as a standalone request, ensuring that larger messages are not discarded or delayed. If the record does not exceed the `batch_size`, Fluvio will process the record as part of an already existing batch or create a new one if the batch is full.
+
+2. Handling Larger Messages than the Max Request Size
+
+Fluvio will have a new configuration parameter, max_request_size, that will define the maximum size of a request that can be sent by the producer. This configuration will make Fluvio display errors when a message exceeds the defined `max_request_size`, even if it's a message with only one record or a batch of records.
+
+3. Compression Behavior
+
+Fluvio will ensure that configuration limits that use size constraints, such as `batch_size` and `max_request_size` will only use the uncompressed message size.
+It affects the size of messages in transit but doesn't change the maximum request size constraints.
+
+
+## Fluvio CLI
+
+Preparing the environment, with a topic and a large data file:
+
+```bash
+fluvio topic create large-data-topic
+printf 'This is a sample line. ' | awk -v b=500000 '{while(length($0) < b) $0 = $0 $0}1' | cut -c1-500000 > large-data-file.txt
+```
+
+### Batch Size
+
+`batch_size` will define the maximum size of a batch of records that can be sent by the producer. If a record exceeds this size, Fluvio will process the record as a standalone message.
+
+```bash
+fluvio produce large-data-topic --batch-size 16536 --file large-data-file.txt --raw
+```
+
+There will not be any errors displayed, even if the message exceeds the batch size. But the record will be processed as a standalone message.
+
+### Max Request Size
+
+`max_request_size` will define the maximum size of a message that can be sent by the producer. If a message exceeds this size, Fluvio will throw an error. Even if it's a message with only one record or a batch of them.
+
+```bash
+fluvio produce large-data-topic --max-request-size 16384 --file large-data-file.txt --raw
+```
+
+Will be displayed the following error:
+
+```bash
+the given record is larger than the max_request_size (16384 bytes).
+```
+
+### Compression
+
+`batch_size` and `max_request_size` will only use the uncompressed message size.
+
+```bash
+fluvio produce large-data-topic --batch-size 16536 --compression gzip --file large-data-file.txt --raw
+fluvio produce large-data-topic --max-request-size 16384 --compression gzip --file large-data-file.txt --raw
+```
+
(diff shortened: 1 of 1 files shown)


rfc/produce-message-size.md
```diff
@@ -0,0 +1,99 @@
+# Enhancing Fluvio to Message Sizes
+
+This RFC proposes modifications to Fluvio's handling of message sizes. 
+
+## Proposed Enhancements
+
+1. Handling Larger Messages than Batch Size
+
+If a single record exceeds the defined `batch_size`, Fluvio will process the record as a standalone request, ensuring that larger messages are not discarded or delayed. If the record does not exceed the `batch_size`, Fluvio will process the record as part of an already existing batch or create a new one if the batch is full.
```
ajhunyady, line 9: Hmmm, is this true? Customers are complaining of packets being dropped if they are larger than the batch size.

rfc/produce-message-size.md
```diff
@@ -0,0 +1,99 @@
...
+
+2. Handling Larger Messages than the Max Request Size
+
+Fluvio will have a new configuration parameter, max_request_size, that will define the maximum size of a request that can be sent by the producer. This configuration will make Fluvio display errors when a message exceeds the defined `max_request_size`, even if it's a message with only one record or a batch of records.
+
+3. Compression Behavior
+
+Fluvio will ensure that configuration limits that use size constraints, such as `batch_size` and `max_request_size` will only use the uncompressed message size.
```
ajhunyady, line 17: That's the problem we need to solve. The algorithm needs to compute the result `after` compression, as that's what's carried in the payload.

rfc/produce-message-size.md
fraidev: I wrote the expectation, I'll rewrite it with what is doing now and what should be changed.

rfc/produce-message-size.md
fraidev: As batch is a concept before the message, I really think that we should not use or try to guess the compression size for that. 

Right now, Fluvio is just "tring to guess" the compression, multiplying every uncompressed size to half (0.5) when it's a compressed batch.

About `max_request_size` I don't see problem to use compression size for that, but Kafka does not use the compression size for `max_request_size`. So if we want that, it's better use another name to not generate confusion for the clients.

rfc/produce-message-size.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Proposed Enhancements
+
+1. Handling Larger Messages than Batch Size
+
+If a single record exceeds the defined `batch_size`, Fluvio will process the record as a standalone request, ensuring that larger messages are not discarded or delayed. If the record does not exceed the `batch_size`, Fluvio will process the record as part of an already existing batch or create a new one if the batch is full.
+
+2. Handling Larger Messages than the Max Request Size
+
+Fluvio will have a new configuration parameter, max_request_size, that will define the maximum size of a request that can be sent by the producer. This configuration will make Fluvio display errors when a message exceeds the defined `max_request_size`, even if it's a message with only one record or a batch of records.
```
fraidev, line 13: do we already have producer parameter to config the max size of the message sent? We're using batch_size for this for now

sehz:
Never mind previous comments.  Here is clarifying semantics.

`max_request_size` applies to pre-compression record size so can be applied to per record constrain.    `batch-size` applies to post-compression limit (or underlying raw records limit).   We can set `max-request-size` high for default case and bump up `batch-size` as well.

rfc/produce-message-size.md
sehz: not sure why this behavior is needed.  should move to enhancement if this doesn't applies to original problem

rfc/produce-message-size.md
```diff
@@ -0,0 +1,99 @@
...
+Preparing the environment, with a topic and a large data file:
+
+```bash
+fluvio topic create large-data-topic
+printf 'This is a sample line. ' | awk -v b=500000 '{while(length($0) < b) $0 = $0 $0}1' | cut -c1-500000 > large-data-file.txt
+```
+
+### Batch Size
+
+`batch_size` will define the maximum size of a batch of records that can be sent by the producer. If a record exceeds this size, Fluvio will process the record as a standalone message.
```
sehz, line 32: `batch_size` should be ultimate raw limit.   again not sure what this address

fraidev:
> max_request_size applies to pre-compression record size so can be applied to per record constrain. batch-size applies to post-compression limit (or underlying raw records limit). 

I think that we should use the compressed size for `batch-size` as we only compress the whole batch. Not each record of the batch.

> We can set max-request-size high for default case and bump up batch-size as well.

Sure, Kafka for example uses 16kb for `batch_size` like us and 1mb for `max.request.size`

rfc/produce-message-size.md
```diff
@@ -0,0 +1,99 @@
...
+
+Will be displayed the following error:
+
+```bash
+the given record is larger than the max_request_size (16384 bytes).
+```
+
+### Compression
+
+`batch_size` and `max_request_size` will only use the uncompressed message size.
```
sehz, line 56: see suggestion.  Having `batch_size` applies to final raw batch should simplify calculation.

rfc/produce-message-size.md
fraidev: I think this behavior will be confused, other products has the same parameter config `batch_size`. In other products, Its behavior is to just check if the record will be inside a batch or not.

fraidev: not reject them

rfc/produce-message-size.md
sehz: this can be subsumed into exiting batching behavior.  if `batch_size` is exceed, then batch is flushed

ajhunyady (approved):
Let's check this in. @sehz you had some comments are they resolved?

digikata (approved):
LGTM

PR #4203, opened 2024-10-02
feat(cli):  release date column for package listings

EstebanBorai:
Hub Clients such as:

- [x] `cdk`
- [x] `fluvio hub connector list` (uses same impl as from `cdk`)
- [x] `fluvio hub smartmodule list`

Should include the release date as part of their listing.

## Demo

https://github.com/user-attachments/assets/27d74324-ee8f-4678-b6cc-1570df88a0b0


crates/fluvio-hub-util/src/cmd/list.rs
```diff
@@ -88,15 +104,54 @@ mod output {
...
 }
+
+fn format_duration(date: DateTime<Utc>, exact: bool) -> String {
+    let duration = Utc::now().signed_duration_since(date);
+
+    if exact {
+        return date.format("%Y-%m-%d %H:%M:%S %Z").to_string();
+    }
+
+    if duration.num_weeks() >= 1 {
```
sehz, line 132: This ia already available in humantime crate.  No need to reinvent

EstebanBorai: Yes that was my first attempt, using humantime, but I didn't found a format specifier,[`FormatDuration`](https://docs.rs/humantime/latest/humantime/fn.format_duration.html) doesn't let you pick which fields to use.

And given that release dates are from more that a year in some cases we have outputs like `2years 2min 12us` (as shown here: https://docs.rs/humantime/latest/humantime/)

EstebanBorai: Here is an example of how many packages looked like: `2years 12h 4m 37s 47ms 997us 519ns`

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=933296e2dc7dcd9f93dc91fc75849617

It would be nice to use from `years` to `minutes` but thing is that theres no way to specify that.

crates/fluvio-extension-common/src/common/time.rs
```diff
@@ -0,0 +1,35 @@
+use chrono::{DateTime, Utc};
+
+pub fn format_duration(date: DateTime<Utc>, exact: bool) -> String {
+    let duration = Utc::now().signed_duration_since(date);
```
sehz, line 4: Lets add unit test

EstebanBorai: Sure!

EstebanBorai: Tests provided here: [231efc9](https://github.com/infinyon/fluvio/pull/4203/commits/231efc91f651adb6b52378bb0cd5d6e995f74181)

crates/fluvio-extension-common/src/common/time.rs
```diff
@@ -0,0 +1,110 @@
...
+    #[test]
+    fn renders_minutes() {
+        let next_min = Utc::now().minute() - 30;
+        let date = Utc::now().with_minute(next_min).expect("failed to set min");
+        let formatted = format_duration(date, false);
+
+        assert_eq!(formatted, "30 minutes ago");
+    }
+
+    #[test]
```
sehz, line 104: what happens if this is 2 month, 3 days, 23 sec and 1/10 ms?

ajhunyady:
@EstebanBorai can we default to `--exact-time`? The data alone is not sufficient in most cases.

A nice improvement would be to have human readable values `3 days`, or `24 minutes`, etc.
And maybe default to that.

crates/fluvio-hub-util/src/cmd/list.rs
ajhunyady: That's too much. It should default to the largest value `2 years`, or `4 days`, or `26 min`. 
Maybe anything in hours would show `3 hours 4 min`, but nothing more granular than that.

ajhunyady: BTW, you can use Kubernetes pods output for hints.

EstebanBorai: Thats correct @ajhunyady! Thank you so much for the suggestions!

I have updated tests to use a static past date and use that for calculations, so dynamic current time doest break them.

Yeah it would be nice to have humantime format support but seems that its not being actively maintained, last commit was 2 years ago: https://github.com/tailhook/humantime

EstebanBorai:
> @EstebanBorai can we default to `--exact-time`? The data alone is not sufficient in most cases.
> 
> A nice improvement would be to have human readable values `3 days`, or `24 minutes`, etc. And maybe default to that.

I could make it the default and perhaps later implement the duration since

EstebanBorai:
@sehz @ajhunyady I have updated the formatter so we can inject a comparison date to improve testing. Now tests will not break because we use a fixed date.

With that I think we are good to go?

EstebanBorai:
> @EstebanBorai can we default to `--exact-time`? The data alone is not sufficient in most cases.
> 
> A nice improvement would be to have human readable values `3 days`, or `24 minutes`, etc. And maybe default to that.

Yeah thats the default, in the video we just see the date string output due to the fact that these packages are from more than a month ago

EstebanBorai:
> > @EstebanBorai can we default to `--exact-time`? The data alone is not sufficient in most cases.
> > A nice improvement would be to have human readable values `3 days`, or `24 minutes`, etc. And maybe default to that.
> 
> Yeah thats the default, in the video we just see the date string output due to the fact that these packages are from more than a month ago

Here are tested all variants: https://github.com/infinyon/fluvio/pull/4203/files#diff-779c45328cff39a63747a2e73d809fa7c2e7d8fda7e0e88e5a90fb07b0b5af6aR57-R140

ajhunyady:
Is there a video?

EstebanBorai:
> Is there a video?

Yeah! On the top in the PR description Im showing both use cases (exact and not exact) and also for connectors and smart modules.

crates/fluvio-cli/src/client/hub/smartmodule/list.rs
```diff
@@ -81,7 +98,7 @@ mod output {
     impl TableOutputHandler for ListSmartModules {
         /// table header implementation
         fn header(&self) -> Row {
-            Row::from(["SMARTMODULE", "Visibility"])
+            Row::from(["SmartModule", "Visibility", "Released"])
```
sehz, line 101: should be `published` since there there is no concept of `released` concept in hub

crates/fluvio-cli/src/client/hub/smartmodule/list.rs
```diff
@@ -91,12 +108,20 @@ mod output {
...
                 .iter()
                 .map(|e| {
                     Row::from([
                         Cell::new(e.pkg_name()).set_alignment(CellAlignment::Left),
                         Cell::new(&e.visibility).set_alignment(CellAlignment::Left),
+                        Cell::new(
+                            e.published_at()
+                                .map(|date| {
+                                    TimeElapsedFormatter::default()
+                                        .time_since_str(date, self.exact_time)
```
sehz, line 121: prefer human time as default since that's how typically most tools shows

crates/fluvio-extension-common/src/common/time.rs
```diff
@@ -0,0 +1,141 @@
...
+    fn determines_exact_date() {
+        let next_min = december_17_1989_10_30_00().minute() - 30;
+        let date = december_17_1989_10_30_00()
+            .with_minute(next_min)
+            .expect("failed to set min");
+        let formatted = make_fixed_time_elapsed_fmtr().time_since_str(date, true);
+
+        assert_eq!(formatted, date.format(EXACT_FORMAT).to_string());
```
sehz, line 80: test should have assertion like this: 
````
assert_eq!(date,"December 1989");
````

File: crates/fluvio-extension-common/src/common/time.rs [hunk 7]
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    #[test]
+    fn simple_date_format_for_exceeding_a_week() {
+        let date = december_17_1989_10_30_00()
+            .checked_sub_days(Days::new(7))
+            .expect("failed to sub days");
+        let formatted = make_fixed_time_elapsed_fmtr().time_since_str(date, false);
+
+        assert_eq!(formatted, date.format(SIMPLE_DATE_FORMAT).to_string());
```
sehz, line 90: see comments above

crates/fluvio-hub-util/src/cmd/list.rs
```diff
@@ -78,7 +91,7 @@ mod output {
     impl TableOutputHandler for ListConnectors {
         /// table header implementation
         fn header(&self) -> Row {
-            Row::from(["CONNECTOR", "Visibility"])
+            Row::from(["Connector", "Visibility", "Released"])
```
sehz, line 94: `Published`

sehz: Probably `Connector` -> `Versions`

crates/fluvio-cli/src/client/hub/smartmodule/list.rs
EstebanBorai: Sounds good! Applied here: [ba6348a](https://github.com/infinyon/fluvio/pull/4203/commits/ba6348ad15e02909fbd3aa381f87f2df0ce26413)

crates/fluvio-cli/src/client/hub/smartmodule/list.rs
EstebanBorai: Yep! `exact_time` is default to `false` in CLI's

https://github.com/infinyon/fluvio/blob/ba6348ad15e02909fbd3aa381f87f2df0ce26413/crates/fluvio-cli/src/client/hub/smartmodule/list.rs#L19-L21

https://github.com/infinyon/fluvio/blob/ba6348ad15e02909fbd3aa381f87f2df0ce26413/crates/fluvio-hub-util/src/cmd/list.rs#L19-L21

crates/fluvio-extension-common/src/common/time.rs
EstebanBorai: Of course! Here is the update: [0a4d54a](https://github.com/infinyon/fluvio/pull/4203/commits/0a4d54ac3df22a05cc769e0504f477f0d3e3a2a3)

crates/fluvio-hub-util/src/cmd/list.rs
EstebanBorai: Updated the Published already! About `Versions` I think that would belong to a new column doesn't it? Perhaps as an iteration from this we could split name from version and use a column for versions only?

crates/fluvio-extension-common/src/common/time.rs
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    #[test]
+    fn simple_date_format_for_exceeding_a_week() {
+        let date = december_17_1989_10_30_00()
+            .checked_sub_days(Days::new(7))
+            .expect("failed to sub days");
+        let formatted = make_fixed_time_elapsed_fmtr().time_since_str(date, false);
+
+        assert_eq!(formatted, "1989/12/10");
```
sehz, line 90: why it is different from `determines_exact_date` ?

EstebanBorai: Its just simpler to keep it compact, an output including time and thus timezone (e.g. `1989/12/10 00:00::00 UTC`) would only appear if user uses `exact-date`.

Do you rather using exact date as fallback for longer periods?

sehz: I thought it automatically converts to exact if exceed certain time.  This is only 7 days, so shouldn't be `a week` or something like that?

crates/fluvio-extension-common/src/common/time.rs
```diff
@@ -0,0 +1,141 @@
...
+        Self { bot: Utc::now() }
+    }
+}
+
+impl TimeElapsedFormatter {
+    pub fn new(bot: DateTime<Utc>) -> Self {
+        Self { bot }
+    }
+
+    pub fn time_since_str(&self, date: DateTime<Utc>, exact: bool) -> String {
```
sehz, line 21: should add docs on rule

EstebanBorai: Here I have added docs: [e96128c](https://github.com/infinyon/fluvio/pull/4203/commits/e96128c58f5e514fae653130fb9eb4c5486f06ad), I will use such statements in the docs as guide for free-form expressions.

crates/fluvio-extension-common/src/common/time.rs
sehz: There is well established library like: https://www.npmjs.com/package/humanize-duration that specifies behavior

EstebanBorai:
Surpassed by #4290

PR #4204, opened 2024-10-07
fix: update version mismatch error text

digikata:


fraidev (approved):
LGTM! 


But what do you think to create an enum similar to `MinimumPlatformVersion` or `MaximumPlatformVersion` for it? Or maybe reuse them.

digikata:
@fraidev I appreciate the comment & it's good to consider these aspects - I had looked over Min/Max errors too and they're more appropriate for when the cluster emits errors (the cluster has the final word on what client versions are accepted).  I think these error paths are generally where the fluvio client will output errors and the anyhow generated errors.  We can revisit this when we expand the version compatibility.

PR #4207, opened 2024-10-08
feat: high level api for connecting with a specific profile

morenol:


digikata:
Is connect with profile name what we want for this or am I thinking of a different use case?   It might be easier to connect with config file path (with a config file generated via `fluvio profile export  PROFILE`)

sehz:
conf file path is low level, it expose client to physical mapping.

digikata (approved):

PR #4212, opened 2024-10-15
add: optional v4 token access

digikata:


crates/fluvio-hub-protocol/src/infinyon_tok.rs
```diff
@@ -20,20 +22,133 @@ type InfinyonRemote = String;
...
+            println!(
+                "Using org access: {}",
+                cli_access_tokens.get_current_org_name()?
+            );
+            return Ok(AccessToken::V4(cli_access_tokens));
+        }
+        Err(_e) => {
+            // fallback to old token logic
+        }
+    };
```
EstebanBorai, line 94: More abou code styling/read than logic but I think this could be more compact with a `let/else`.

crates/fluvio-hub-util/src/hubaccess.rs
```diff
@@ -167,8 +169,24 @@ impl HubAccess {
...
-        if !authn_token.is_empty() {
-            builder = builder.header("Authorization", &authn_token);
+        match token {
+            Some(AccessToken::V4(cli_access_tokens)) => {
+                let org = cli_access_tokens.get_current_org_name()?;
+                let tok = cli_access_tokens
+                    .org_access_tokens
+                    .get(&org)
+                    .ok_or(HubError::HubAccess("Missing org token".to_string()))?;
+                let authn_token = format!("bearer {tok}");
```
nacardin, line 179: Does "bearer" need to be "Bearer"?

nacardin (changes requested):

EstebanBorai (approved):
LGTM! Comment is not blocker and I think they could be applied as iteration!

crates/fluvio-hub-protocol/src/infinyon_tok.rs
```diff
@@ -20,20 +22,133 @@ type InfinyonRemote = String;
...
+    pub fn is_v4(&self) -> bool {
+        matches!(self, AccessToken::V4(_))
+    }
+}
+
+// multi-org access token output
+#[derive(Debug, Serialize, Deserialize)]
+pub struct CliAccessTokens {
+    pub remote: String,
+    pub user_access_token: String,
```
nacardin, line 57: I'm changing `user_access_token` this to be optional.

crates/fluvio-hub-util/src/hubaccess.rs
digikata: The hub server accepts both, but updating to "Bearer" as it may be more standard?

crates/fluvio-hub-protocol/src/infinyon_tok.rs
digikata: added commit to make the fielld optional

nacardin (approved):

fraidev (approved):

digikata:
rebased

PR #4215, opened 2024-10-17
feat(fvm): override installed binaries target

EstebanBorai:
Provides support to override the target triple used when fetching binaries from package registry.

## Demo

`fvm install` now accepts the `--target` option which takes a [Rust Target Triple](https://doc.rust-lang.org/nightly/rustc/platform-support.html#tier-1-with-host-tools) thats used to replace the default target which is host's OS target.

Alternately a user could specify `FVM_BINARY_ARCH_TRIPLE` environment variable, used for the same purpose. This is useful for CI environments.

help

If the target triple is not available, FVM will notify the user before performing any further tasks.

```bash
fvm install latest --target aarch64-unknown-linux-gnu
Error: PackageSet "latest" is not available for architecture: "aarch64-unknown-linux-gnu"
```

bad-target

### DIFF
--- a/crates/fluvio-version-manager/src/command/install.rs
+++ b/crates/fluvio-version-manager/src/command/install.rs
@@ -20,6 +20,9 @@ use crate::common::workdir::fvm_versions_path;
 /// The `install` command is responsible of installing the desired Package Set
 #[derive(Debug, Parser)]
 pub struct InstallOpt {
+    /// Binaries architecture triple to use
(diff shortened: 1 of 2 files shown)


sehz (approved):
LGTM

digikata (approved):
LGTM,  could you also set the INFINYON_CI_CONTEXT for fvm tests?

EstebanBorai:
> LGTM, could you also set the INFINYON_CI_CONTEXT for fvm tests?

Of course!

PR #4222, opened 2024-10-29
chore: remove unneeded encoding mark for api protocol

sehz:
The enum encoding of API was not needed since each api is encoded directly.  Enum is used only for decoding only


fraidev (approved):
Pretty nice this encoder attribute! 
LGTM

PR #4239, opened 2024-11-06
fix: mirror compatibility with old versions

fraidev:


sehz (approved):

sehz (approved):

sehz:
Post commit.

May want to use common version for all protocol version instead of hardcoding.  That may need to update fluvio macro

PR #4243, opened 2024-11-07
feat: support producing by mirror

fraidev:
Similar to https://github.com/infinyon/fluvio/pull/4241 but for mirrors.

`$ fluvio produce --mirror <mirror_name> my__topic`
or 
`$ fluvio produce -m <mirror_name> my__topic`


sehz (approved):
Nice

sehz (approved):

PR #4247, opened 2024-11-08
fix: compatibility issues with old versions

fraidev:
This PR aims to create tests to check backward compatibility since a specific version, for now, v0.11.10.

Also, it was fixed the compatibility with v0.11.11 and v0.11.10.


digikata (approved):
LGTM, I like the differientiation between fluvio client bin and cluster bin

Issue #4216, opened 2024-10-22
Show actual record size when throw maximum request size error


fraidev:
Now we're only showing the max request limit, we should also show the record size that is larger than the limit.

Actually
```
the given record is larger than the max_request_size (1048576 bytes).
```

Expected:
```
Error: Record dropped: record size (xyz bytes), exceeded maximum request size (1048576 bytes)
```

fraidev:
Solved by https://github.com/infinyon/fluvio/pull/4251

PR #4251, opened 2024-11-11
feat: show actual record size when throw maximum request size error

fraidev:
Related: https://github.com/infinyon/fluvio/issues/4216

### DIFF
--- a/crates/fluvio/src/producer/error.rs
+++ b/crates/fluvio/src/producer/error.rs
@@ -9,8 +9,8 @@ use crate::producer::PartitionId;
 #[derive(thiserror::Error, Debug, Clone)]
 #[non_exhaustive]
 pub enum ProducerError {
-    #[error("the given record is larger than the max_request_size ({0} bytes)")]
-    RecordTooLarge(usize),
+    #[error("record size ({0} bytes), exceeded maximum request size ({0} bytes)")]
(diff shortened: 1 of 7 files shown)


digikata (approved):
LGTM, nice improvement on the err message

crates/fluvio/src/producer/error.rs
```diff
@@ -9,8 +9,8 @@ use crate::producer::PartitionId;
 #[derive(thiserror::Error, Debug, Clone)]
 #[non_exhaustive]
 pub enum ProducerError {
-    #[error("the given record is larger than the max_request_size ({0} bytes)")]
-    RecordTooLarge(usize),
+    #[error("record size ({0} bytes), exceeded maximum request size ({0} bytes)")]
```
sehz, line 12: since api is changed, need to bump major version

sehz (changes requested):

crates/fluvio/src/producer/error.rs
fraidev: sure! I bumped the fluvio major version

sehz (approved):

sehz:
@fraidev ,  fluvio `0.24` was not released so there was no need to bump to `0.25`.  can you revert back version number

PR #4264, opened 2024-11-19
fix: do not allow connectors offset without id

fraidev:


crates/fluvio-connector-common/src/consumer.rs
```diff
@@ -47,6 +49,16 @@ pub async fn consumer_stream_from_config(
...
+
+        if consumer_offset.strategy != OffsetStrategyConfig::None
+            && config
+                .meta()
+                .consumer()
+                .and_then(|c| c.id.as_ref())
+                .is_none()
+        {
+            bail!("Consumer id is required when using offset strategy");
+        }
```
digikata, line 61: I think it might be more general for this logic be placed into a custom ConsumerConfigExitBuilde::build() method? (you may have to define one like the older ConsumerConfigBuilder::build() impl)

fraidev: Make sense! I moved it to the build method.

digikata (approved):
LGTM

PR #4268, opened 2024-11-20
feat: add graceful shutdown for consume connectors

fraidev:
When a consumer connector disconnects it is not dropping the offset management, then it was losing some record in the offset storage.

I added a graceful shutdown similar to what we have in `fluvio-cli`.


sehz (approved):
LGTM.  it would be good to add integration test

digikata (approved):

fraidev:
> LGTM. it would be good to add integration test

E2E test added!

PR #4276, opened 2024-11-27
fix: shutdown of the cdk consumer by future not stream

fraidev:
Related: https://github.com/infinyon/fluvio/pull/4273

This PR fixes an issue where streams could fail to drop in certain states, leading to resource leaks. The solution ensures the future managing the stream is explicitly shut down, resolving the problem.

Changes:

- Replaced the unreliable `TakeUntil` wrapper with explicit stream management logic.
- Uses the same E2E test of https://github.com/infinyon/fluvio/pull/4273, but with a custom config to replicate these problematic states

### DIFF
--- a/crates/fluvio-connector-derive/src/generator.rs
+++ b/crates/fluvio-connector-derive/src/generator.rs
@@ -35,7 +35,7 @@ fn generate_source(func: &ConnectorFn) -> TokenStream {
                         match user_fn_result {
                             Ok(_) => ::fluvio_connector_common::tracing::info!("Connector arrived at end of stream"),
                             Err(e) => {
-                                ::fluvio_connector_common::tracing::error!(%e, "Connector failed");
+                                ::fluvio_connector_common::tracing::error!(%e, "Error encountered producing records in source connector");
                                 return Err(e.into());
                             },
                         }
@@ -78,7 +78,7 @@ fn generate_sink(func: &ConnectorFn) -> TokenStream {
                         match user_fn_result {
                             Ok(_) => ::fluvio_connector_common::tracing::info!("Connector arrived at end of stream"),
                             Err(e) => {
-                                ::fluvio_connector_common::tracing::error!(%e, "Connector failed");
+                                ::fluvio_connector_common::tracing::error!(%e, "Error encountered processing records in sink connector");
                                 return Err(e.into());
                             },
                         }
--- a/crates/fluvio-connector-derive/src/generator.rs
+++ b/crates/fluvio-connector-derive/src/generator.rs
@@ -20,6 +20,7 @@ fn generate_source(func: &ConnectorFn) -> TokenStream {
 
         fn main() -> ::fluvio_connector_common::Result<()> {
             #init_and_parse_config
+            let stop_signal = ::fluvio_connector_common::consumer::init_ctrlc()?;
 
             ::fluvio_connector_common::future::run_block_on(async {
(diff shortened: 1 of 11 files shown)


sehz:
Let's add E2E test

sehz (commented):
Check if version needed to be bump?

crates/fluvio-connector-derive/src/generator.rs
```diff
@@ -45,17 +62,32 @@ fn generate_sink(func: &ConnectorFn) -> TokenStream {
...
-                let mut stream = stream.take_until(stop_signal.recv());
-                #user_fn(user_config, stream).await
+                ::fluvio_connector_common::future::select! {
+                    user_fn_result = async {
+                        #user_fn(user_config, stream).await
+                    } => {
+                        match user_fn_result {
+                            Ok(_) => ::fluvio_connector_common::tracing::info!("Connector arrived at end of stream"),
+                            Err(e) => {
+                                ::fluvio_connector_common::tracing::error!(%e, "Connector failed");
```
sehz, line 81: instead of `connector failed`,  say something like: `error encountered processing records in sink`. it would be good to add connector type?
Also try to add offset and other metadata that could be help to diagnose problem

crates/fluvio-connector-derive/src/generator.rs
```diff
@@ -45,17 +62,32 @@ fn generate_sink(func: &ConnectorFn) -> TokenStream {
...
+                                return Err(e.into());
```
sehz, line 82: Is there retry logic or how would error would be hadled

crates/fluvio/src/consumer/stream.rs
```diff
@@ -242,6 +242,7 @@ impl Drop for OffsetManagement {
             if let Err(err) = offset_store.try_flush() {
                 warn!("flush on drop failed: {err:?}");
             }
+            debug!("offsets flushed on drop");
```
sehz, line 245: let's make info and add other information to figure out what's going on

crates/fluvio-connector-derive/src/generator.rs
fraidev: sure, wdyt "Error encountered producing records in source" for source?

fraidev: offset? 
I changed this part of code just to have sure that we're printing the error form connector.

crates/fluvio-connector-derive/src/generator.rs
fraidev: just error handled, it was handled with `?` before, I am printing the error that came from the connector function.

crates/fluvio/src/consumer/stream.rs
fraidev: sure, I am printing the local offset store now

crates/fluvio-connector-derive/src/generator.rs
fraidev: also, I can't use `?` again, because of the new async block without result inside select!

fraidev:
> Let's add E2E test

As we didn't publish Fluvio 0.24.1, I think it's fine.

crates/fluvio-connector-derive/src/generator.rs
sehz: K

sehz (approved):

PR #4301, opened 2024-12-14
chore: update packages needed for release

fraidev:
as we updated `fluvio-controlplane-metadata` we also need update `fluvio-sc-schema`

https://github.com/infinyon/fluvio/actions/runs/12332023979/job/34419349906

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3004,7 +3004,7 @@ dependencies = [
 
 [[package]]
 name = "fluvio-sc-schema"
-version = "0.25.0"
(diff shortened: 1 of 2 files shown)


sehz (commented):
Instead, use stable API and just update lock file if it's patch.  Otherwise do major bump

fraidev:
> Instead, use stable API and just update lock file if it's patch. Otherwise do major bump

Sure, but closing because I think this is not the problem.

https://github.com/infinyon/fluvio/pull/4302

Issue #4304, opened 2024-12-18
[Bug]: Fluvio without `openssl` but `rustls` does not compile


TheSamsa:
**What happened**
Using fluvio as dependency like this: `fluvio = { version = "0.24.0", features = ["compress", "rustls"], default-features = false }`
Fails to build complaining `IoError` not existent in (fluvio) `src/config/tls.rs:247`

**Expected behavior**
I expect it to build the same as I would use the `openssl` feature

**Describe the setup**
- I am using a local fluvio running in a docker container

**How to reproduce it (as minimally and precisely as possible)**
1. `cargo init fluvio-rustls`
2. `cd fluvio-rustls`
3. `cargo add fluvio -F rustls,compress --no-default-features`
4. `cargo build`

**Log output**
It is rather an error output while building:
```
error[E0412]: cannot find type `IoError` in this scope
   --> ../.cargo/registry/src/index.crates.io-6f17d22bba15001f/fluvio-0.24.0/src/config/tls.rs:247:26
    |
247 |             type Error = IoError;
    |                          ^^^^^^^ not found in this scope
    |
help: there is an enum variant `crate::dataplane::types::config_file::LoadConfigError::IoError` and 6 others; try using the variant's enum
    |
247 |             type Error = crate::dataplane::types::config_file::LoadConfigError;
    |                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
247 |             type Error = fluvio_compression::CompressionError;
    |                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
247 |             type Error = fluvio_protocol::types::config_file::LoadConfigError;
    |                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
247 |             type Error = fluvio_smartmodule::dataplane::core::types::config_file::LoadConfigError;
    |                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      and 2 other candidates
```

**Environment (please complete the following information):**
output of `fluvio version`:
```
 Fluvio CLI                     : 0.13.0
 Fluvio CLI Arch                : x86_64-unknown-linux-musl
[...]

PR #4305, opened 2024-12-18
Fixes compilation error if openssl feature is deactivated but rustls activated

TheSamsa:
Fixes: #4304

If openssl default feature is deactivated but rustls is active fluvio does not compile.
It complains about a missing `IoError` in `crates/fluvio/src/config/tls.rs:247`


digikata (approved):
Thanks!   I performed a test build w/ the `rustls` feature on and default features off and it built great

PR #4306, opened 2024-12-29
chore: wrap offset_flush return in pin<box<_>>

fraidev:
I need that changes to be able to store dynamic `ConsumerStream` in a struct for https://github.com/infinyon/fluvio-client-python/pull/541

Remembering that we can't use generics with pyo3 structures.

### DIFF
--- a/crates/fluvio/src/consumer/stream.rs
+++ b/crates/fluvio/src/consumer/stream.rs
@@ -1,3 +1,4 @@
+use std::pin::Pin;
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
 use std::time::{Duration, SystemTime};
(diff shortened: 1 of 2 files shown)


crates/fluvio/src/consumer/stream.rs
```diff
@@ -19,7 +20,7 @@ pub trait ConsumerStream: Stream<Item = Result<Record, ErrorCode>> + Unpin {
     fn offset_commit(&mut self) -> Result<(), ErrorCode>;
 
     /// Send the committed offset to the server. The method waits for the server's acknowledgment before it finishes.
-    fn offset_flush(&mut self) -> impl Future<Output = Result<(), ErrorCode>> + Send;
+    fn offset_flush(&mut self) -> Pin<Box<dyn Future<Output = Result<(), ErrorCode>> + Send + '_>>;
```
sehz, line 23: can you try to alias this type?

fraidev: sure! what do you think:

```rust
type OffsetFlushFuture<'a> = Pin<Box<dyn Future<Output = Result<(), ErrorCode>> + Send + 'a>>;
```

sehz (approved):

PR #4313, opened 2025-01-09
chore: update Rust to `1.84.0`

EstebanBorai:
Updates Rust te `1.84.0`

### DIFF
--- a/crates/cargo-builder/src/package.rs
+++ b/crates/cargo-builder/src/package.rs
@@ -123,7 +123,7 @@ impl PackageInfo {
         Ok(path)
     }
 
-    /// path to package's wasm32-wasi target
+    /// path to package's wasm32-wasip1 target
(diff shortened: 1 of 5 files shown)


EstebanBorai:
Hey @sehz! I see that since Rust 1.84 the target: `wasm32-wasi` wont be available anymore.

```
Starting Rust 1.78 (May 2nd, 2024), users of WASI 0.1 are encouraged to begin migrating to the new wasm32-wasip1 target before the existing wasm32-wasi target is removed in Rust 1.84 (January 5th, 2025).
```

I think we could now target SmartModules to use `wasm32-wasip2` perhaps?

More on the topic here: https://blog.rust-lang.org/2024/04/09/updates-to-rusts-wasi-targets.html

sehz (approved):

PR #4315, opened 2025-01-13
cleanup fbm

sehz:
focus fbm to only on producer performance test.  Remove code related to consumer testing which will be brought back later.


crates/fluvio-benchmark/src/benchmark_driver.rs
```diff
@@ -1,246 +1,124 @@
...
 
+        stat.end.store(true, Ordering::Relaxed);
+
         Ok(())
     }
-    pub async fn run_benchmark(config: BenchmarkConfig, all_stats: AllStatsSync) -> Result<()> {
+    pub async fn run_benchmark(config: BenchmarkConfig) -> Result<()> {
         // Create topic for this run
 
         let new_topic = TopicSpec::new_computed(config.num_partitions as u32, 1, None);
```
fraidev, line 81: Maybe we create the topic inside the producer, but no block, I can do this in the next PR

fraidev (approved):
LGTM

PR #4320, opened 2025-01-14
feat: improve hw perfomance

fraidev:


fraidev:
Already implemented with https://github.com/infinyon/fluvio/pull/4331

PR #4330, opened 2025-01-15
connector: provide sample Dockerfile

sehz:
Dockerfile to run a connector

### DIFF
--- a/connector/docker/README.md
+++ b/connector/docker/README.md
@@ -1,0 +1,27 @@
+# Introduction
+
+This is sample project to demonstrate how to use the docker to run the connector.
+This examples is for http connector but can be used for any connector.
+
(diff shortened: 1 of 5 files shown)


digikata (commented):
LGTM

digikata (approved):
LGTM

connector/docker/README.md
```diff
@@ -0,0 +1,27 @@
...
+To build docker image run:
+```
+make build
+```
+
+This will copy following files from your local file system:
+- your current fluvio profile
+- connector.yaml
+
+The connector.yaml is connector config for http source
```
ajhunyady, line 17: We should add a link to `connector.yaml` or inline it.

I ran without it... and got an error.

PR #4331, opened 2025-01-16
perf:  Improve checkpoint updating performance

### LINKED ISSUE #4322: improvement checkpoint performance
Without writing to checkpoint file, we get:

### 1MB batch
total message send: 216832
message: per second: 108416, bytes per  sec:     542.1mb/s, 

### 16K batch
total bytes send: 392280000
total message send: 78456
message: per second: 39228, bytes per  sec:     196.1mb/s, 

The checkpoint is used for hw which may not be need to fsync. We need to add following business logic if we remove fsync in write path.

- [x] confirm hw is fully synchronized with multiple followers. This should be already in the code but confirm again.

---  these two are unnecessary after confirming hw offset will be fully reconciled upon start up

- [ ] add flush api to checkpoint which will perform hard fsync.all. This will be called upon cluster shutdown or delete.
- [ ] sync checkpoint upon replica initialization. if no followers then hw=Leo, if there are follower, then hw should be updated as in next step

sehz:
resolves #4322 

Disable `fsync`  in the hw checkpoint updating.  It turns out that hw values will be auto reconciled anyway so even if hw values are behind actual values, it will be updated when cluster is restarted.

Clean up `storage` crate by moving dep as workspace and update its CLI to read hw values.


fraidev (approved):
Awesome! 🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

PR #4335, opened 2025-01-17
perf improvement: implement async write for hw checkpoint

sehz:
Since HW write doesn't need to be sync with file, we can separate into  asynchronous task.

Writer task is spawned to write to file which doesn't block checkpoint update.


fraidev (approved):
LGTM!

PR #4348, opened 2025-01-21
Fix benchmark many producers

fraidev:


crates/fluvio-benchmark/src/stats_collector.rs
```diff
@@ -41,98 +37,33 @@ impl ProducerStat {
...
     ) {
         spawn(async move {
-            let hist = histogram.clone();
+            let latency_histogram = latencies.clone();
             while let Ok(event) = event_receiver.recv().await {
-                let hist = hist.clone();
-                let stats = stats.clone();
+                let latencies = latency_histogram.clone();
+                let stats = total_stats.clone();
                 spawn(async move {
```
sehz, line 67: double spawn loop is inefficient.   it may contribute to some latency.  not show stopper for this PR, but should flatten this.  it's probably better perform accumulator with atomic counter callback then perform analytics in the stat collector

sehz (approved):

PR #4363, opened 2025-01-25
feat: add benchmark table again

fraidev:
image


crates/fluvio/src/producer/accumulator.rs
```diff
@@ -126,7 +128,7 @@ impl RecordAccumulator {
 
                     // Create and push a new batch if needed
                     let push_record = self
-                        .create_and_new_batch(batch_events, &mut batches, record, 1)
+                        .create_and_new_batch(batch_events, &mut batches, record, 1, created_at)
```
sehz, line 131: bit unclear whether latency is account for but should be better than now

sehz (approved):

PR #4367, opened 2025-01-28
fix: fluvio time panicking on wasm

fraidev:


crates/fluvio/Cargo.toml
```diff
@@ -40,6 +40,7 @@ thiserror = { workspace = true }
 semver = { workspace = true }
 pin-project = { workspace = true }
 siphasher = { workspace = true }
+web-time = { workspace = true }
```
sehz, line 43: Since this rely on wasmbingen, need to configure this target for wasm32 only

sehz (changes requested):

crates/fluvio/Cargo.toml
morenol: this is supposed to do the magic of using `std::time` and not bring additional dependencies when we are not in wasm32-unknown-unknown, but yes we could explictly avoid relying on that an only installing this dep in case of wasm32

fraidev: @morenol yeah, the seems that lib already does that for us. But I changed to be explicit to have sure.

sehz (approved):

PR #4373, opened 2025-01-29
chore: change checkpoint error to info

sehz:
this is recoverable, no need to make it error.
updated expired certs


fraidev (approved):

sehz:
@fraidev, updated certs here to fix CI

PR #4375, opened 2025-02-01
chore: produce config builder as mutable

fraidev:
TopicProducerConfigBuilder was created as [owner](https://docs.rs/derive_builder/latest/derive_builder/#owned-aka-consuming), making it very difficult to handle and create wrappers for python client.

I changed this builder to the default behavior (mutable).

### DIFF
--- a/crates/fluvio-cli/src/client/produce/mod.rs
+++ b/crates/fluvio-cli/src/client/produce/mod.rs
@@ -199,47 +199,30 @@ mod cmd {
             fluvio: &Fluvio,
         ) -> Result<()> {
             init_monitoring(fluvio.metrics());
-            let config_builder = if self.interactive_mode() {
-                TopicProducerConfigBuilder::default().linger(std::time::Duration::from_millis(10))
-            } else {
-                Default::default()
+            let mut config_builder = TopicProducerConfigBuilder::default();
+            if self.interactive_mode() {
+                config_builder.linger(std::time::Duration::from_millis(10));
             };
-
             // Compression
-            let config_builder = if let Some(compression) = self.compression {
-                config_builder.compression(compression)
-            } else {
-                config_builder
-            };
-
+            if let Some(compression) = self.compression {
+                config_builder.compression(compression);
+            }
             // Linger
-            let config_builder = if let Some(linger) = self.linger {
-                config_builder.linger(linger)
-            } else {
-                config_builder
-            };
-
+            if let Some(linger) = self.linger {
+                config_builder.linger(linger);
+            }
             // Batch size
-            let config_builder = if let Some(batch_size) = self.batch_size {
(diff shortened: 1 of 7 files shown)


crates/fluvio-cli/src/client/produce/mod.rs
```diff
@@ -199,47 +199,30 @@ mod cmd {
             fluvio: &Fluvio,
         ) -> Result<()> {
             init_monitoring(fluvio.metrics());
-            let config_builder = if self.interactive_mode() {
-                TopicProducerConfigBuilder::default().linger(std::time::Duration::from_millis(10))
-            } else {
-                Default::default()
+            let mut config_builder = TopicProducerConfigBuilder::default();
+            if self.interactive_mode() {
```
sehz, line 203: do we still need interactive mode?

crates/fluvio/src/producer/config.rs
```diff
@@ -88,7 +88,7 @@ pub struct TopicProducerConfig {
     pub(crate) linger: Duration,
     /// Partitioner assigns the partition to each record that needs to be send
     #[builder(default = "default_partitioner()")]
-    pub(crate) partitioner: Box<dyn Partitioner + Send + Sync>,
+    pub(crate) partitioner: Arc<dyn Partitioner + Send + Sync>,
```
digikata, line 91: Perhaps we should make this Arc Box?  It may be important in long running or high concurrency uses

fraidev: I don't think so.
In our usecase that we will only clone once, in the builders, I think that we will not have a overhead here.

Maybe we can use [dyn_clone](https://docs.rs/dyn-clone/latest/dyn_clone/) and keep using box. That allows us to have clone and use box.
But using Arc here does not seem a problem.

crates/fluvio-cli/src/client/produce/mod.rs
fraidev: Do you mean have different linger on interactive mode? Or about interactive mode itself?

sehz: I meat interactive mode.   Doesn't seem like nobody uses and not sure why it set to linger non zero

morenol: probably for interactive mode we should use the default now that it is 0.

fraidev: Ok, I removed this condition as the default is 0 now

crates/fluvio/src/producer/config.rs
digikata: Given that arc uses heap memory from the internals that you researches, the current Arc return is fine!

PR #4376, opened 2025-02-01
fix: record too large error

fraidev:
Got this error testing max_request_size on python client: 

```
mymodule.PyFluvioError: record size (17 bytes), exceeded maximum request size (17 bytes)
```

Should be

```
mymodule.PyFluvioError: record size (17 bytes), exceeded maximum request size ("max_request_size" bytes)
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2333,7 +2333,7 @@ dependencies = [
 
 [[package]]
 name = "fluvio"
-version = "0.24.4"
+version = "0.24.5"
 dependencies = [
  "anyhow",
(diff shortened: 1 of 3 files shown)


sehz (commented):
Can you check fluvio need to be bumped?

fraidev:
> Can you check fluvio need to be bumped?

Sure, I didn't update because I updated already at https://github.com/infinyon/fluvio/pull/4375.

But I also updated here.

sehz (approved):

Issue #2167, opened 2022-02-08
Consumer stream should be able to reconnect


morenol:
Currently, producer and consumer APIs are able to reconnect in case of failure, but a consumer stream does not

nacardin:
Should handle either SPU or consumer failure

fraidev:
Solved by https://github.com/infinyon/fluvio/pull/4394

PR #4394, opened 2025-02-14
feat: consumer stream should be able to reconnect

fraidev:
Required:
- [x] https://github.com/infinyon/fluvio/pull/4402

Solves https://github.com/infinyon/fluvio/issues/2167

TryForever(default):
```rust
let config = ConsumerConfigExtBuilder::default()
    .topic(topic)
    .retry_mode(RetryMode::TryForever)
    .build()
    .expect("config");
let stream = fluvio_client.consumer_with_config(config).await?
```
or just:

```rust
let config = ConsumerConfigExtBuilder::default()
    .topic(topic)
    .build()
    .expect("config");
let stream = fluvio_client.consumer_with_config(config).await?
```

Disabled:
```rust
let config = ConsumerConfigExtBuilder::default()
    .topic(topic)
    .retry_mode(RetryMode::Disabled)
    .build()
    .expect("config");
let stream = fluvio_client.consumer_with_config(config).await?
```

Try until 10 attempts:
```rust
let config = ConsumerConfigExtBuilder::default()
    .topic(topic)
    .retry_mode(RetryMode::TryUntil(10))
    .build()
    .expect("config");
let stream = fluvio_client.consumer_with_config(config).await?
```

### DIFF
--- a/crates/fluvio/src/consumer/retry.rs
+++ b/crates/fluvio/src/consumer/retry.rs
@@ -1,18 +1,22 @@
 use std::future::Future;
+use std::pin::Pin;
+use std::sync::Arc;
 use std::task::{Context, Poll};
-use std::{pin::Pin, sync::Arc, time::Duration};
+use std::time::{Duration, Instant};
 
 use anyhow::Result;
 use adaptive_backoff::prelude::{
-    ExponentialBackoff, ExponentialBackoffBuilder, Backoff, BackoffBuilder,
+    Backoff, BackoffBuilder, ExponentialBackoff, ExponentialBackoffBuilder,
 };
-use futures_util::stream::Stream;
+use futures_util::future::BoxFuture;
+use futures_util::Stream;
 use futures_util::StreamExt;
 use tracing::{debug, info, warn};
 
 use fluvio_future::timer::sleep;
 use fluvio_protocol::record::ConsumerRecord;
 use fluvio_sc_schema::errors::ErrorCode;
+
 use crate::consumer::RetryMode;
 use crate::{Fluvio, FluvioConfig, Offset};
 use super::{ConsumerConfigExt, ConsumerStream};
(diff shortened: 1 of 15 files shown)


crates/fluvio/src/consumer/retry.rs
```diff
@@ -0,0 +1,279 @@
...
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let self_mut = self.get_mut();
+
+        // If there is no pending future, create one using the current stream.
+        if self_mut.pending.is_none() {
+            let current_stream = self_mut
+                .current_stream
+                .take()
+                .expect("current_stream must be available");
```
sehz, line 81: should avoid expect

crates/fluvio/src/consumer/retry.rs
```diff
@@ -0,0 +1,279 @@
...
+        }
+    }
+}
+
+impl ConsumerStream for ConsumerWithRetry {
+    fn offset_commit(&mut self) -> std::result::Result<(), ErrorCode> {
+        if let Some(ref mut stream) = self.current_stream {
+            stream.offset_commit()
+        } else {
+            warn!("Can't commit, no current stream");
```
sehz, line 120: probably this return error since caller should no offset commit should not work.  or this could be made async.   ignoring is not good option

crates/fluvio/src/consumer/retry.rs
```diff
@@ -0,0 +1,279 @@
...
+            Ok(())
+        }
+    }
+
+    fn offset_flush(
```
sehz, line 125: same comment

crates/fluvio/src/consumer/retry.rs
```diff
@@ -0,0 +1,279 @@
...
+                        if attempts > 0 {
+                            debug!(
+                                target: SPAN_RETRY,
+                                "Record produced successfully after reconnect"
+                            );
+                        }
+                        return Some(Ok((record, current_stream, new_offset)));
+                    }
+                    Err(e) => {
+                        warn!(target: SPAN_RETRY, "Error consuming record: {}", e);
```
sehz, line 183: return error?

crates/fluvio/src/consumer/retry.rs
```diff
@@ -0,0 +1,279 @@
...
+            // Wait before retrying.
+            backoff_and_wait(&mut backoff).await;
+            attempts += 1;
+
+            // Determine the offset for reconnection.
+            let offset = if let Some(next) = inner.next_offset_to_read {
+                match Offset::absolute(next) {
+                    Ok(off) => off,
+                    Err(e) => {
+                        warn!(target: SPAN_RETRY, "Error creating offset: {}", e);
```
sehz, line 205: return error?
or alternatively, have option to ignore error

crates/fluvio/src/consumer/retry.rs
```diff
@@ -0,0 +1,279 @@
...
+}
+
+/// Creates an exponential backoff configuration.
+fn create_backoff() -> ExponentialBackoff {
+    ExponentialBackoffBuilder::default()
+        .factor(BACKOFF_FACTOR)
+        .min(BACKOFF_MIN_DURATION)
+        .max(BACKOFF_MAX_DURATION)
+        .build()
+        .expect("Failed to build backoff configuration")
```
sehz, line 266: remove expect

crates/fluvio/src/consumer/retry.rs
```diff
@@ -0,0 +1,279 @@
...
+}
+
+/// Waits for the duration determined by the exponential backoff.
+async fn backoff_and_wait(backoff: &mut ExponentialBackoff) {
+    let wait_duration = backoff.wait();
+    debug!(
```
sehz, line 272: let's change to info so this is important log

crates/fluvio/src/consumer/retry.rs
fraidev: we can't return error here, only if retry mode is Disabled. We need to keep trying on next lines of code.

crates/fluvio/src/consumer/retry.rs
fraidev: fixed!

crates/fluvio/src/consumer/retry.rs
fraidev: fixed!

crates/fluvio/src/consumer/retry.rs
fraidev: fixed

crates/fluvio/src/consumer/retry.rs
fraidev: Fixed! returning a error now

crates/fluvio/src/consumer/retry.rs
fraidev: Fixed! returning an error now

crates/fluvio/src/consumer/retry.rs
fraidev: We're returning error at next line, no?

sehz: k

sehz (approved):
Nice job.

fraidev:
@sehz I added a new commit to:
- move state and pending to enum instead two options
- wait commit and flush offset

crates/fluvio/src/consumer/retry.rs
```diff
@@ -0,0 +1,314 @@
...
+                    warn!("offset_commit called but stream is terminated");
+                    return Ok(());
+                }
+                _ => {
+                    if start.elapsed() > MAX_COMMIT_DELAY {
+                        warn!("offset_commit timed out");
+                        return Err(ErrorCode::MaxCommitDuration);
+                    }
+                    // Yield a bit; in a busy loop this is not ideal, but it's a workaround.
+                    std::thread::yield_now();
```
sehz, line 154: Hmm.  We should avoid doing this.  Can we convert this to async?

sehz: Alternatively, return error and let client handle this.

fraidev: I changed to wait a notification instead, what do you think? 

```rust
impl ConsumerStream for ConsumerWithRetry {
    fn offset_commit(&mut self) -> Result<(), ErrorCode> {
        run_block_on(async {
            loop {
                match &mut self.state {
                    ConsumerState::Current(stream) => {
                        return stream.offset_commit();
                    }
                    ConsumerState::Terminated => {
                        warn!("offset_commit called but stream is terminated");
                        return Ok(());
                    }
                    _ => {
                        self.notify.notified().await;
                    }
                }
            }
        })
    }

    fn offset_flush(&mut self) -> BoxFuture<'_, Result<(), ErrorCode>> {
        async move {
            loop {
                match &mut self.state {
                    ConsumerState::Current(stream) => {
                        return stream.offset_flush().await;
                    }
                    ConsumerState::Terminated => {
                        warn!("offset_flush called but stream is terminated");
                        return Ok(());
                    }
                    _ => {
                        self.notify.notified().await;
                    }
                }
            }
        }
        .boxed()
    }
}

```

crates/fluvio/src/consumer/retry.rs
```diff
@@ -0,0 +1,321 @@
...
+        }
+    }
+}
+
+const MAX_COMMIT_DELAY: Duration = Duration::from_secs(5);
+
+impl ConsumerStream for ConsumerWithRetry {
+    /// TODO: rework this as an async method.
+    fn offset_commit(&mut self) -> Result<(), ErrorCode> {
+        run_block_on(async {
```
sehz, line 151: let's  move to async instead of trying to work around this sync api

sehz: Alternatively,  not expose this API to public.

fraidev: I am using block_on only on Pending state now, what do you think? 
This is a public trait for the users already, it's the Stream trait, for that reason, I didn't change to async.
Also, changing all of this to async will be a problem because WASM does not support Send here.

fraidev: hmm actually I can do all async for non-wasm and this workarround for wasm: 
```
    #[cfg(not(target_arch = "wasm32"))]
    fn offset_flush(&mut self) -> BoxFuture<'_, Result<(), ErrorCode>> {
        let notify = self.notify.clone();
        Box::pin(async move {
            loop {
                match self.state {
                    ConsumerState::Current(ref mut stream) => return stream.offset_flush().await,
                    ConsumerState::Terminated => {
                        warn!("offset_flush called but stream is terminated");
                        return Ok(());
                    }
                    _ => {
                        notify.notified().await;
                    }
                }
            }
        })
    }

    #[cfg(target_arch = "wasm32")]
    fn offset_flush(&mut self) -> BoxFuture<'_, Result<(), ErrorCode>> {
        loop {
            match self.state {
                ConsumerState::Current(ref mut stream) => return stream.offset_flush(),
                ConsumerState::Terminated => {
                    warn!("offset_flush called but stream is terminated");
                    return Box::pin(async { Ok(()) });
                }
                _ => {
                    let notify = self.notify.clone();
                    run_block_on(async move {
                        notify.notified().await;
                    });
                }
            }
        }
    }
```

sehz: For WASM, can use  `feature` config to customize trait per target.  `run_block_on` make it difficult to port to other async runtime.

fraidev: This works for `offset_flush` that is async already.
I propose a new PR to make `offset_commit` async too. But this would be a breaking change for users.

sehz: it's bump up already,  so it's ok.   Let's create  PR to convert `offset` api to async as pre-requisite.

crates/fluvio/src/consumer/retry.rs
```diff
@@ -0,0 +1,356 @@
...
+        loop {
+            match self.state {
+                ConsumerState::Current(ref mut stream) => return stream.offset_commit(),
+                ConsumerState::Terminated => {
+                    warn!("offset_commit called but stream is terminated");
+                    return Box::pin(async { Ok(()) });
+                }
+                _ => {
+                    let notify = self.notify.clone();
+                    fluvio_future::task::run_block_on(async move {
```
sehz, line 178: let's try to remove `run_block_on`

PR #4395, opened 2025-02-17
chore: prevent batch write when storage error occurs

sehz:
Currently SPU allow to write to replica even if there was error occurred.   This could leads to undefined states in the storage.  This PR mark replica with flag to prevent that from happening.

This however doesn't cover all cases which need to be address in the future.
- if rollover fails, revert any changes including new segments.

Also, making logging more clear and change some logging to info as they are useful

### DIFF
--- a/crates/fluvio-storage/src/replica.rs
+++ b/crates/fluvio-storage/src/replica.rs
@@ -62,7 +62,7 @@ impl ReplicaStorage for FileReplica {
             .build()
             .map_err(|err| StorageError::Other(format!("failed to build cleaner config: {err}")))?;
 
-        Self::create_or_load(
+        Self::create_or_load_inner(
             replica.topic.clone(),
             replica.partition,
             0,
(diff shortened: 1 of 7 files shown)


crates/fluvio-storage/src/replica.rs
```diff
@@ -399,26 +405,73 @@ impl FileReplica {
...
+        }
+
         self.size
             .store_active(self.active_segment.occupied_memory());
         Ok(())
     }
+
+    /// perform roll over which involves creating new segment and
+    #[instrument(skip(self))]
+    async fn rollver(&mut self) -> Result<()> {
```
fraidev, line 459: Is it a typo? rollover instead rollver?

crates/fluvio-storage/src/replica.rs
```diff
@@ -399,26 +405,73 @@ impl FileReplica {
...
+                return Err(err);
+            }
+        }
+
         self.size
             .store_active(self.active_segment.occupied_memory());
         Ok(())
     }
+
+    /// perform roll over which involves creating new segment and
```
fraidev, line 457: I think we don't need this "and"

fraidev (approved):
LGTM!

PR #4405, opened 2025-02-20
fix: commit and flush retry on other states

fraidev:
I got some offset consumer tests failing sometimes, it's related to offset_commit and offset_flush have not been called on no idle or terminated states. I fixed it keeping state in the struct to be able to call it every moment.

Also, I changed the signature of `ConsumerStream` methods for wasm, making it to write async code for both.

I ran tests many times and now seems ok!

### DIFF
--- a/crates/fluvio/src/consumer/retry.rs
+++ b/crates/fluvio/src/consumer/retry.rs
@@ -61,7 +61,7 @@ type BoxConsumerFuture = Pin<
 
 #[derive(Clone)]
 pub struct ConsumerRetryInner {
-    fluvio_client: Arc<Fluvio>,
+    fluvio_config: FluvioConfig,
     next_offset_to_read: Option<i64>,
     consumer_config: ConsumerConfigExt,
 }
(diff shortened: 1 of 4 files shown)


crates/fluvio/src/consumer/retry.rs
```diff
@@ -90,70 +94,79 @@ impl ConsumerWithRetry {
...
+    }
 }
 
 impl Stream for ConsumerWithRetry {
     type Item = Result<ConsumerRecord, ErrorCode>;
 
-    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
-        // We use a loop so that after a state transition we immediately poll again.
-        let this = self.get_mut();
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
```
sehz, line 116: Can you check if it is possible to implement using `https://crates.io/crates/async-stream` instead of implementing custom polling?

fraidev: I think it's possible, but I need to rewrite all implementation of `ConsumerStream` it needs Stream from `futures-core`

fraidev: I propose a new PR for this technical debt

fraidev: here: https://github.com/infinyon/fluvio/issues/4406

fraidev:
> Also, we should add/update integration test for offset api

I added a commit with unit test

sehz (approved):
Nice!

Issue #4412, opened 2025-02-25
[Bug]: smdk load error w/ missing SmartModule.toml


digikata:
**What happened**
`smdk load` when the SmartModule.toml was unintentionally missing, emits a misleading error error:

```
$ smdk load
Loading package at: /path
Error: No such file or directory (os error 2)

```
**Expected behavior**
The specific missing file name should be emitted with the error.

**Describe the setup**
- fluvio 0.15.2 and rust 1.85.0

**How to reproduce it (as minimally and precisely as possible)**
1. smdk generate  `<sm-name>`
2. cd `<sm-name>`
3. smdk build
4. rm Smartmodule.toml  # repro only
5. smdk load

digikata:
fixed behavior:
```
$ smdk load
Loading package at: ~/test/sm-test
Error: reading smartmodule metadata file ~/sm-test/SmartModule.toml, No such file or directory (os error 2)
```

PR #4413, opened 2025-02-25
fix: smdk error for missing Smartmodule.toml #4412

digikata:
- **fix: consistent SMARTMODULE_TOML const**
- **fix: smdk error for missing Smartmodule.toml #4412**


sehz (approved):
LGTM

PR #4431, opened 2025-03-04
Export ReplyHubref{} and HUB_REF_ENDPOINT

crajcan:
Export the endpoint responsible for getting the hub url, and the struct used to deserialize the response.

This is so the UIs can use it, notably here: https://github.com/infinyon/infinyon-cloud-ui/pull/1348


digikata (approved):
LGTM!

LeoBorai (approved):

Issue #4421, opened 2025-02-28
improve recovery capabilities on fluvio producer


morenol:
Currently. in producer, If `producer.flush` fails due to a network error, consecutive calls to `producer.send` fail early too due to the previous error.  Forcing us to call `clear_errors` to recover the producer.

Probably we need to reconsider if we want to force the client to call `clear_errors` to recover from these errors

fraidev:
Solved by https://github.com/infinyon/fluvio/pull/4441

PR #4441, opened 2025-03-10
fix: reconnect producer when spu is down

fraidev:
Should solve: https://github.com/infinyon/fluvio/issues/4421

### DIFF
--- a/crates/fluvio/src/producer/partition_producer.rs
+++ b/crates/fluvio/src/producer/partition_producer.rs
@@ -1,10 +1,12 @@
 use std::sync::Arc;
-use std::time::Duration;
 
 use adaptive_backoff::prelude::{
     Backoff, BackoffBuilder, ExponentialBackoff, ExponentialBackoffBuilder,
 };
 use async_lock::RwLock;
+use fluvio_types::defaults::{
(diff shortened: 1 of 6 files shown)


crates/fluvio/src/producer/partition_producer.rs
```diff
@@ -24,6 +28,10 @@ use super::{
 use super::accumulator::{BatchEvents, BatchesDeque};
 use super::event::EventHandler;
 
+pub const BACKOFF_MIN_DURATION: Duration = Duration::from_secs(1);
+pub const BACKOFF_MAX_DURATION: Duration = Duration::from_secs(30);
+pub const BACKOFF_FACTOR: f64 = 1.1;
+
```
digikata, line 34: Should these use common values to the consumer side of things, and maybe be in `fluvio-types/src/defaults.rs`?

digikata (commented):
Looks good so far!
- minor nit on the constants location
- is there a test that can be written to cover this?

crates/fluvio/src/producer/partition_producer.rs
fraidev: Sure! I moved the default values to defaults.rs

Also we're already have good test for reconnecting producer at crates/fluvio-test/src/tests/reconnection/mod.rs. I just remove the `producer.clear_errors().await;` that is not needed anymore when spu is down.

sehz (approved):
LGTM

Issue #4436, opened 2025-03-09
[Bug]: when fluvio topics are deleted, consumer connect w/ consumer id's can cause panic


digikata:
**What happened**
On a local fluvio cluster:

1. a fluvio topic was deleted
2. consumer_id's with offsets existed for that topic
3. the topic was recreated
4. On rerun of the fluvio consumer w/ consumer id, fluvio panicked comparing offsets
    - the condition was not a returned error, it was a hard panic.

** what should happen **
All consumer ids associated with the topic should be deleted when the topic is deleted

fraidev:
solved by https://github.com/infinyon/fluvio/pull/4448#event-1677253838

fraidev:
Solved by https://github.com/infinyon/fluvio/pull/4448 and https://github.com/infinyon/fluvio/pull/4459

Issue #4308, opened 2025-01-03
[Bug]: Delete the consumer when delete topic


fraidev:
When deleting a topic with consumers, their consumer should also be deleted.

Steps for the bug:
```
▷ flvd consumer list
  CONSUMER       TOPIC                                 PARTITION  OFFSET  LAST SEEN
  test-consumer  0689fe79-75d8-4346-9ab8-63a70de714e7  0          1       8s

▷ flvd topic list                                                                                                                                       $! add_offset_management_integration
  NAME                                  TYPE      PARTITIONS  REPLICAS  RETENTION TIME  COMPRESSION  DEDUPLICATION  STATUS                   REASON
  foobar                                computed  1           1         7days           any          none           resolution::provisioned
  0689fe79-75d8-4346-9ab8-63a70de714e7  computed  1           1         7days           any          none           resolution::provisioned

▷ flvd topic delete 0689fe79-75d8-4346-9ab8-63a70de714e7
topic "0689fe79-75d8-4346-9ab8-63a70de714e7" deleted

▷ flvd topic list
  NAME                                  TYPE      PARTITIONS  REPLICAS  RETENTION TIME  COMPRESSION  DEDUPLICATION  STATUS                   REASON
  foobar                                computed  1           1         7days           any          none           resolution::provisioned

~/workspace/infinyon/fluvio-client-python
▷ flvd consumer list
  CONSUMER       TOPIC                                 PARTITION  OFFSET  LAST SEEN
  test-consumer  0689fe79-75d8-4346-9ab8-63a70de714e7  0          1       42s
```

fraidev:
solved by https://github.com/infinyon/fluvio/pull/4448#event-1677253838

PR #4448, opened 2025-03-12
fix: delete consumer after topic delete

fraidev:
Should solve: https://github.com/infinyon/fluvio/issues/4436 and https://github.com/infinyon/fluvio/issues/4308

### DIFF
--- a/crates/fluvio-spu/src/core/global_context.rs
+++ b/crates/fluvio-spu/src/core/global_context.rs
@@ -197,7 +197,7 @@ mod file_replica {
         sc_api::remove::ReplicaRemovedRequest, replica::Replica,
         spu_api::update_replica::UpdateReplicaRequest,
     };
-    use fluvio_protocol::{link::ErrorCode, record::ReplicaKey};
+    use fluvio_protocol::record::ReplicaKey;
     use fluvio_types::{defaults::CONSUMER_STORAGE_TOPIC, PartitionId};
     use tracing::{trace, warn};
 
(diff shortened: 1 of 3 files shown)


crates/fluvio-spu/src/core/global_context.rs
```diff
@@ -478,5 +484,44 @@ mod file_replica {
                 error!("leader switch failed: {}", err);
             }
         }
+
+        /// Delete consumers offset for given replica if it is leader consumer
+        async fn delete_consumers_offset_if_leader_consumer(
+            &self,
```
sehz, line 490: is this necessary? this only works if this SPU is leader is same as consumer but that may not case for multiple SPU

sehz:
This doesn't seem like full solution.  Instead of trying to delete consumer.  Rather validate consumer offset to ensure it's valid

fraidev:
> Rather validate consumer offset to ensure it's valid

When/where?

crates/fluvio-spu/src/core/global_context.rs
sehz: but need to test case where consumer offset topic is not on same SPU

sehz:
> > This doesn't seem like full solution. Instead of trying to delete consumer. Rather validate consumer offset to ensure it's valid
> 
> Do you mean make it a soft delete? Not show on `consumer list` when the topic not exist anymore?

Don't even need to anything when delete happens.   The command `consumer list` will just filter out topic that doesn't exists.

fraidev:
@sehz I changed to filter it on fetch/list. Much simpler.

sehz (approved):

PR #4450, opened 2025-03-16
fix: recalculate replicas on timeout and rejection

fraidev:


sehz:
Let's add integration tests here https://github.com/infinyon/fluvio/blob/master/crates/fluvio-spu/src/replication/test.rs to cover these cases

sehz (approved):
LGTM

PR #4456, opened 2025-03-18
fix: return tls cluster connect error when built w/o tls

### LINKED ISSUE #4445: [Bug]: fluvio profiles with TLS can fail to connect  when configured with default-features = false
**What happened**
What a fluvio client is built with `default-features = false`,
a fluvio connect call, with a profile pointing to a cluster that requires TLS can fail with:
`connect error Socket io server has terminated connection`

when built with fluvio crate `0.26.1` and possibly earlier crates.

**Expected behavior**
IF a client is built with `default-features = false`, `Fluvio::connect()` call with a TLS 
configured cluster profile should return a better error: e.g.  `Profile requires TLS and client was not built with TLS enabled`

**Describe the setup**
A project with fluvio crate with `Cargo.toml` dep configured as
`fluvio = { version = "0.26.1", default-features = false }`

should be configured as:
`fluvio = { version = "0.26.1" }`

digikata:
fixes #4445

### DIFF
--- a/crates/fluvio/src/fluvio.rs
+++ b/crates/fluvio/src/fluvio.rs
@@ -79,7 +79,7 @@ impl Fluvio {
         #[cfg(not(any(feature = "openssl", feature = "rustls")))]
         if crate::config::TlsPolicy::Disabled != config.tls {
             return Err(anyhow::anyhow!(
-                "Error: TLS is not supported in this build, but the cluster config requires tls.\nPlease enable the `openssh`feature."
+                "Error: TLS is not supported in this build, but the cluster config requires tls.\nPlease enable the `openssl` feature."
             ));
         }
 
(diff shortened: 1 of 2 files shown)


crates/fluvio/src/fluvio.rs
```diff
@@ -75,6 +75,14 @@ impl Fluvio {
     /// # }
     /// ```
     pub async fn connect_with_config(config: &FluvioClusterConfig) -> Result<Self> {
+        // if crate tls is not configured and the profile has tls, return an error
+        #[cfg(not(any(feature = "openssl", feature = "rustls")))]
+        if crate::config::TlsPolicy::Disabled != config.tls {
+            return Err(anyhow::anyhow!(
+                "Error: TLS is not supported in this build, but the cluster config requires tls.\nPlease enable the `openssh`feature."
```
morenol, line 82:
```suggestion
                "Error: TLS is not supported in this build, but the cluster config requires tls.\nPlease enable the `openssl` feature."
```

digikata: Thanks @morenol !

morenol (approved):
lgtm, but don't we need a whitespace between `openssl` and feature?

fraidev (approved):

Issue #4455, opened 2025-03-18
[Bug]: Panic in consumer_with_config most likely due to only partial topic deletion


caemor:
**What happened?**

I created the `cloud-command` topic. I posted a few messages in it and deleted the topic using `fluvio topic delete ...`.
I then created the same topic using a mirror connection.
Now when I tried to create a consumer in rust for this topic, it panicked.
The problem is in `OffsetInner::resolve`, where start (72) was greater than last (2), causing a min > max panic. (see log)

I'd guess that there was an issue with deleting the topic, where the pointer to start wasn't deleted correctly. The cli didn't show any partition and topic between the delete and re-create.

**Expected behaviour**

Don't panic by checking that start is a valid value, and completely delete all information about a topic when deleting it, so that start can never have such an invalid value.

**Workaround**

Calling `fluvio cluster delete`, `fluvio cluster start` and then `fluvio home connect` once more, solved the issue 😄 

**Describe the setup**
- Local Fluvio Installation
- Versions: rust: `0.25` - cli: `0.15.1`

**Log output**

Backtrace:
```
2025-03-18T08:51:13.594288Z DEBUG consumer_stream_with_config:inner_stream_batches_with_config{consumer_id=Some("edge-device")}:request_stream{offset=Offset { inner: FromBeginning(0) } consumer_id=Some("edge-device")}:create_serial_socket:create_serial_socket_from_leader{leader_id=5001}:connect_to_leader{leader=5001}: fluvio_socket::multiplexing: spawning dispatcher socket=15
2025-03-18T08:51:13.595546Z DEBUG consumer_stream_with_config:inner_stream_batches_with_config{consumer_id=Some("edge-device")}:request_stream{offset=Offset { inner: FromBeginning(0) } consumer_id=Some("edge-device")}: fluvio::offset: fetching offset for replica: cloud-command-0
2025-03-18T08:51:13.596393Z DEBUG consumer_stream_with_config:inner_stream_batches_with_config{consumer_id=Some("edge-device")}:request_stream{offset=Offset { inner: FromBeginning(0) } consumer_id=Some("edge-device")}:send_and_receive{self=MultiplexerSocket 15}: fluvio_socket::multiplexing: sending request api=1002 correlation_id=1
2025-03-18T08:51:13.600487Z DEBUG dispatcher_loop{self=MultiplexDisp(15)}: fluvio_socket::multiplexing: received frame correlation_id=1 len=48 remaining=48
[...]

PR #4459, opened 2025-03-19
fix: hard delete consumer after topic delete

fraidev:
At https://github.com/infinyon/fluvio/pull/4448 we prefer just filter consumers with non deleted topic on list/fetch. 
But we still some problems, like https://github.com/infinyon/fluvio/issues/4455

This PR is hard deleting the consumer from the topic when topic is deleted.

### DIFF
--- a/crates/fluvio-spu/src/core/global_context.rs
+++ b/crates/fluvio-spu/src/core/global_context.rs
@@ -197,14 +197,12 @@ mod file_replica {
         sc_api::remove::ReplicaRemovedRequest, replica::Replica,
         spu_api::update_replica::UpdateReplicaRequest,
     };
-    use fluvio_protocol::record::ReplicaKey;
-    use fluvio_types::{defaults::CONSUMER_STORAGE_TOPIC, PartitionId};
     use tracing::{trace, warn};
 
     use fluvio_storage::FileReplica;
     use flv_util::actions::Actions;
 
-    use crate::{core::SpecChange, kv::consumer::ConsumerOffsetKey};
+    use crate::core::SpecChange;
 
     use super::*;
 
@@ -371,16 +369,16 @@ mod file_replica {
         }
 
         async fn remove_replica(&self, outputs: &mut Vec<ReplicaChange>, replica: Replica) {
-            if let Err(err) = self.delete_consumers_offset(&replica).await {
-                error!("error: {} deleting consumers offset: {}", err, replica);
-            }
-
             if replica.leader == self.local_spu_id() {
                 outputs.push(ReplicaChange::Remove(
-                    self.remove_leader_replica(replica).await,
+                    self.remove_leader_replica(replica.clone()).await,
                 ));
             } else {
-                self.remove_follower_replica(replica).await;
+                self.remove_follower_replica(replica.clone()).await;
+            }
+
+            if let Err(err) = self.delete_consumers_offset(&replica).await {
+                error!("error: {} deleting consumers offset: {}", err, replica);
             }
         }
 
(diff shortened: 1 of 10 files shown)


crates/fluvio-spu/src/core/global_context.rs
```diff
@@ -478,5 +481,41 @@ mod file_replica {
                 error!("leader switch failed: {}", err);
             }
         }
+
+        /// Delete consumers offset for given replica if it is leader consumer
+        async fn delete_consumers_offset(&self, replica: &Replica) -> anyhow::Result<()> {
+            let consumers_replica_id =
```
sehz, line 487: this should be const since this is same for all invocation

crates/fluvio-spu/src/core/global_context.rs
```diff
@@ -478,5 +481,41 @@ mod file_replica {
...
+                ReplicaKey::new(CONSUMER_STORAGE_TOPIC, <PartitionId as Default>::default());
+            let Some(ref replica_consumer) = self.leaders_state().get(&consumers_replica_id).await
```
sehz, line 489: let's move this as separate api something like `is_consumer_leader` or something like that

crates/fluvio-spu/src/core/global_context.rs
fraidev: move to `default.rs` as tuple to not import fluvio-protocol there.

so we can now use it as `CONSUMER_REPLICA_KEY.into()`

crates/fluvio-spu/src/core/global_context.rs
```diff
@@ -478,5 +481,41 @@ mod file_replica {
...
+                debug!("no consumer replica found");
+                return Ok(());
+            };
+
+            let consumer_storage = self
+                .consumer_offset()
+                .get_or_insert(replica_consumer, self.follower_notifier())
+                .await?;
+
+            let consumer_offset_keys = consumer_storage
```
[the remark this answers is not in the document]
fraidev: moved to SharableConsumerOffsetStorage and added unit tests

crates/fluvio-spu/src/core/global_context.rs
fraidev: moved to ReplicaLeadersState

crates/fluvio-spu/src/kv/consumer.rs
```diff
@@ -169,6 +169,29 @@ impl SharableConsumerOffsetStorage {
         self.0.write().await.delete(key).await
     }
 
+    pub async fn delete_by_replica_key(&self, replica: &ReplicaKey) -> Result<()> {
```
sehz, line 172: this doesn't need to be public.
add tracing

sehz:
Looks good. couple of nits

crates/fluvio-spu/src/core/global_context.rs
```diff
@@ -382,7 +368,21 @@ mod file_replica {
             outputs
         }
 
-        /// reemove leader replica
+        async fn remove_replica(&self, outputs: &mut Vec<ReplicaChange>, replica: Replica) {
+            if let Err(err) = self.delete_consumers_offset(&replica).await {
```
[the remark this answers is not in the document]
fraidev: sure, done!

crates/fluvio-spu/src/kv/consumer.rs
fraidev: sure, done! it's pub(crate) now

sehz (approved):
LGTM

Issue #4467, opened 2025-03-24
Re-export PartitionId.


ruseinov:
Currently one has to do this:
```
pub type PartitionId = u32;
```

It would be great if `PartitionId` was re-exported from `fluvio` package: `fluvio::PartitionId`. 
That can be achieved by modifying it's `use` statement to `pub use`.

ajhunyady:
@ruseinov can you please send a pull request?

ruseinov:
> [@ruseinov](https://github.com/ruseinov) can you please send a pull request?

Last time I checked I was not able to open a PR from my fork to your repo.

ruseinov:
It seems to work now, strange.

PR #4481, opened 2025-04-02
fix: re-export PartitionId from fluvio package.

ruseinov:
fixes https://github.com/infinyon/fluvio/issues/4467


sehz (approved):
Thanks for contributing.  LGTM

PR #4491, opened 2025-04-06
chore: pos release fluvio 0.17.1

fraidev:


VERSION
```diff
@@ -1 +1 @@
-0.17.1
\ No newline at end of file
+0.17.1-dev-1
```
digikata, line 1: `0.17.2-dev-1` ?

fraidev: Indeed! fixed.

digikata (approved):
LGTM!

PR #4494, opened 2025-04-07
chore: move boxed stream into main consumer api

sehz:
BoxedConsumerStream is handy.  This moves into core part of consumer stream to make it easier to use it

### DIFF
--- a/crates/fluvio/src/consumer/mod.rs
+++ b/crates/fluvio/src/consumer/mod.rs
@@ -61,7 +61,7 @@ pub type BoxConsumerStream =
 pub type BoxConsumerStream =
     Pin<Box<dyn ConsumerStream<Item = Result<ConsumerRecord, ErrorCode>> + Send + 'static>>;
 
-/// Type alias for the future returned by our retry logic.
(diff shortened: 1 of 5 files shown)


crates/fluvio/src/consumer/mod.rs
```diff
@@ -50,6 +53,39 @@ pub use fluvio_smartmodule::dataplane::smartmodule::SmartModuleExtraParams;
...
 
+/// Type alias for the consumer record stream.
+#[cfg(target_arch = "wasm32")]
+pub type BoxConsumerStream =
+    Pin<Box<dyn ConsumerStream<Item = Result<ConsumerRecord, ErrorCode>> + 'static>>;
+#[cfg(not(target_arch = "wasm32"))]
+pub type BoxConsumerStream =
+    Pin<Box<dyn ConsumerStream<Item = Result<ConsumerRecord, ErrorCode>> + Send + 'static>>;
+
+/// Type alias for the future returned by our retry logic.
```
fraidev, line 64: I think we don't need this comment to "retry logic" here

fraidev (approved):

PR #4499, opened 2025-04-08
chore: store file replica error

fraidev:


nacardin (commented):
This would help in discovering initial error from latest log entry, but we need to solve the issue of spamming error logs in a tight loop.

crates/fluvio-storage/src/replica.rs
```diff
@@ -40,7 +40,7 @@ pub struct FileReplica {
     commit_checkpoint: CheckPoint,
     cleaner: Arc<Cleaner>,
     size: Arc<ReplicaSize>,
-    append_failure: bool, // if this is true, last append failed, should not append again
+    error_msg: Option<String>,
```
sehz, line 43: Error message can be recovered from log, so there is no need hold on this. Beside which error?

sehz: Same comment as @nacardin,  in the SPU, if there are certain number of errors happen,  entire replica should be disabled and return offline status

fraidev:
Fixed it in https://github.com/infinyon/fluvio/pull/4521 and https://github.com/infinyon/fluvio/pull/4527

Issue #4325, opened 2025-01-14
offset management:  settings for offset_start(offset::Beginning()) do not correctly retreive the consumer offset


digikata:
offset_start(Offset::end()) picks the end w/o retrieving the consumer id offset
offset_start(Offset::from_end(1) picks the end - 1 w/o retrieving the consumer id offset

Needs additional test cases

ajhunyady:
I think that’s the only one that works. We should ensure that all other offset flavors work, end(), from_end(), from_beggining(), etc.

fraidev:
fixed by https://github.com/infinyon/fluvio/pull/4500

PR #4500, opened 2025-04-09
fix: consumer offset end

fraidev:
Should solve https://github.com/infinyon/fluvio/issues/4325


crates/fluvio/src/offset.rs
```diff
@@ -21,11 +21,19 @@ impl OffsetInner {
...
+                let resolved = if let Some(consumer_offset) = offsets.consumer_offset {
+                    consumer_offset + offset
+                } else {
+                    offsets.start_offset + offset
+                };
                 resolved.clamp(offsets.start_offset, offsets.last_stable_offset)
             }
             Self::FromEnd(offset) => {
-                let resolved = offsets.last_stable_offset - offset;
+                let resolved = if let Some(consumer_offset) = offsets.consumer_offset {
```
fraidev, line 32: the fix is here

crates/fluvio-spu-schema/src/server/fetch_offset.rs
```diff
@@ -113,6 +113,10 @@ pub struct FetchOffsetPartitionResponse {
 
     /// Last readable offset
     pub last_stable_offset: i64,
+
+    /// Consumer offset
```
sehz, line 117: can you add more description on what is requested?

sehz: Shouldn't this be list of consumer id (string)?

crates/fluvio-spu/src/services/public/offset_request.rs
```diff
@@ -55,7 +55,7 @@ pub async fn handle_offset_request(
                     match fetch_consumer_offset(&ctx, topic, *partition, consumer_id).await {
                         Ok(Some(consumer_offset)) => {
                             debug!(consumer_id, consumer_offset, "consumer offset");
-                            partition_response.start_offset = consumer_offset + 1;
+                            partition_response.consumer_offset = Some(consumer_offset + 1);
                         }
```
sehz, line 59: should have integration test

crates/fluvio-spu-schema/src/server/fetch_offset.rs
fraidev: I changed to `Last committed consumer offset`, what do you think?

crates/fluvio-spu/src/services/public/offset_request.rs
fraidev: I added some tests at `crates/fluvio-test/src/tests/consumer_offsets/` that should cover it

fraidev: Or do you mean another kind of integration test?

crates/fluvio/src/offset.rs
sehz: let' add/update unit test

crates/fluvio/src/offset.rs
```diff
@@ -345,6 +353,7 @@ mod tests {
             partition_index: 0,
             start_offset: 0,
             last_stable_offset: 10,
+            consumer_offset: None,
```
sehz, line 356: use defaults

crates/fluvio/src/offset.rs
```diff
@@ -359,6 +368,7 @@ mod tests {
             partition_index: 0,
             start_offset: 5,
             last_stable_offset: 10,
+            consumer_offset: None,
```
sehz, line 371: use default

fraidev: fixed

crates/fluvio/src/offset.rs
fraidev: fixed

crates/fluvio-spu-schema/src/server/fetch_offset.rs
```diff
@@ -113,6 +113,10 @@ pub struct FetchOffsetPartitionResponse {
 
     /// Last readable offset
     pub last_stable_offset: i64,
+
+    /// Last committed consumer offset
+    #[fluvio(min_version = 24)]
+    pub consumer_offset: Option<i64>,
```
sehz, line 119: Hmm.  There is already `FetchConsumerOffsetsRequest` ? Why is needed here?

crates/fluvio-spu/src/services/public/offset_request.rs
sehz: That's E2E tests, good job on there.

The integration test should go here: https://github.com/infinyon/fluvio/tree/master/crates/fluvio-spu/src/services/public/tests.

crates/fluvio/src/offset.rs
fraidev: sure, added!

crates/fluvio-spu-schema/src/server/fetch_offset.rs
fraidev: This is only requested when the stream is created, we can also change to request `FetchConsumerOffsetsRequest` there, but it will need 2 requested.

As we were handling consumer offset directly at `start_offset` of `FetchOffsetPartitionResponse`, I kept with it.

sehz: Let's use `FetchConsumerOffsetsRequest`. otherwise we have to maintain at two different places.  Making 2 requests are fine as this is only at beginning.

fraidev: Seems that `FetchConsumerOffsetsRequest` return all consumers and doesn't allow passing.
Do you still think that is better use it? If yes, should we add a parameter to filter the replica/consumer_id that we want in the request or in memory?

fraidev: I think that is ok use `FetchOffsetsRequest` as it has consumer_id as a parameter already.
Also is filtered by replica and consumer_id.

image

sehz: Oh. Then need another field in the response? There is no such thing as consumer offset

sehz: Either this is raw partition offset or applied to consumer only.  So there is no need to another field

fraidev: Before we're just replacing the `start_offset` by the `consumer_offset` when is fetched with a `consumer_id`. But this only worked for `beginning` mode.

I added an optional field called `consumer_offset` to have it instead `start_offset`. 
Then when we resolve the offset using the `consumer_offset` right.

fraidev: > Either this is raw partition offset or applied to consumer only. So there is no need to another field

The problem is, we don't pass the offset mode in the request to the server, so we don't know how to resolve it there.

fraidev: We solve it in the client, so I think it makes sense to return the consumer offset state separately from the start/end of the topic.

sehz: Problem is that semantics is unclear.   Maybe then logic should be into the SPU consumer processing not at client.

I would expect that when I get offset, it is valid.  SPU should make sure it's always valid

fraidev: Ok, I can change back to just changing the `start_offset`, but sending in the fetch request the offset (begin, end, ...), and resolve it on SPU.
What do you think?

sehz: It already has following info:
```
  /// First readable offset.
    pub start_offset: i64,

    /// Last readable offset
    pub last_stable_offset: i64,
```

fraidev: I changed to send two requests on stream creation. 
One for topic offset and another one for the consumer_offset.
We're not handling it via forwarding now.

crates/fluvio-spu-schema/src/server/fetch_offset.rs
```diff
@@ -24,9 +24,6 @@ use super::SpuServerApiKey;
 pub struct FetchOffsetsRequest {
     /// Each topic in the request.
     pub topics: Vec<FetchOffsetTopic>,
-
-    #[fluvio(min_version = 23)]
-    pub consumer_id: Option<String>,
```
digikata, line 29: We should consider the compatibility effects of this change on older fluvio clients.  Can they send with this field and still interoperate?  Can the field be deprecated w/o deleting it?

crates/fluvio-spu/src/services/public/offset_request.rs
fraidev: I added integration test

crates/fluvio-spu-schema/src/server/fetch_offset.rs
fraidev: Make sense, I added it back with a deprecated attribute to keep working with older clients.

crates/fluvio-sc/src/controllers/topics/controller.rs
```diff
@@ -196,9 +196,11 @@ where
 
         let mut interval_secs = INTERVAL_STEP;
 
+        sleep(Duration::from_secs(2)).await;
```
sehz, line 199: Is this necessary?  there is already wait below

crates/fluvio-sc/src/controllers/topics/controller.rs
```diff
@@ -196,9 +196,11 @@ where
...
+
         loop {
-            sleep(Duration::from_secs(interval_secs)).await;
             self.ensure_offsets_topic_exists().await;
+            sleep(Duration::from_secs(interval_secs)).await;
```
sehz, line 203: let's add info indicating it is sleeping

crates/fluvio-spu-schema/src/server/consumer_offset.rs
```diff
@@ -100,3 +102,60 @@ impl ConsumerOffset {
         }
     }
 }
+
+#[derive(Decoder, Encoder, Default, Debug)]
+pub struct GetConsumerOffsetRequest {
+    pub replica_id: ReplicaKey,
```
sehz, line 108: to follow other request conventions, this should be list of request.
`replica_id` should be optional

crates/fluvio-spu/src/services/public/tests/consumer_offset.rs
```diff
@@ -0,0 +1,87 @@
...
+        GetConsumerOffsetRequest::new(target_replica.into(), consumer_id.to_owned());
+
+    let response = client_socket
+        .send_and_receive(RequestMessage::new_request(get_consumer_offset_request))
+        .await
+        .expect("send");
+
+    assert_eq!(response.consumer.expect("consumer").offset, 5);
+    assert_eq!(response.error_code, ErrorCode::None);
+
```
sehz, line 84: nic

crates/fluvio-spu/src/services/public/tests/mod.rs
```diff
@@ -10,19 +10,23 @@ use fluvio_controlplane::spu_api::update_smartmodule::SmartModule;
...
 };
+use fluvio_socket::AsyncResponse;
+use fluvio_spu_schema::server::stream_fetch::StreamFetchRequest;
 use fluvio_protocol::{
     fixture::BatchProducer,
     record::{RecordData, Record, RecordSet, Batch, RawRecords},
     ByteBuf,
 };
 use fluvio_storage::{FileReplica, ReplicaStorage};
+use futures_util::StreamExt;
```
sehz, line 21: nit.  move this in third party libraries

fraidev: the order? fixed!

crates/fluvio-sc/src/controllers/topics/controller.rs
fraidev: I added a debug

crates/fluvio-sc/src/controllers/topics/controller.rs
fraidev: no, I added to test something and forget it, I removed it

crates/fluvio-spu-schema/src/server/consumer_offset.rs
fraidev: We have a `FetchConsumerOffsetsRequest` already that does it, maybe merge them? 
also make `replica_id` and `consumer_id` optional

sehz: sure if that make sense

fraidev: sure, done!

crates/fluvio-sc/src/controllers/topics/controller.rs
```diff
@@ -194,9 +194,10 @@ where
         const INTERVAL_STEP: u64 = 15;
         const MAX_INTERVAL: u64 = 120;
 
-        let mut interval_secs = INTERVAL_STEP;
+        let mut interval_secs = 1;
 
         loop {
+            debug!("sleeping for {} seconds", interval_secs);
```
sehz, line 200:
```suggestion
            debug!(interval_secs,"sleeping for");
```

sehz (approved):
LGTM.  Good job

crates/fluvio-sc/src/controllers/topics/controller.rs
fraidev: fixed!

Issue #4324, opened 2025-01-14
example auto management code does not work without odd pauses


digikata:
offset_flush(Duration::from_miliis(0)) should be the default and flush the offset every stream.next()

If optimizing users can call offset_flush with non-zero flush timeouts.

This needs test cases and inline and fluvio-doc documentation

fraidev:
Solved by https://github.com/infinyon/fluvio/pull/4504

PR #4504, opened 2025-04-11
fix: auto management odd pauses in consumer offset

fraidev:
Should solve https://github.com/infinyon/fluvio/issues/4505 and https://github.com/infinyon/fluvio/issues/4324

This PR:
- Adding background checker for offset management when `Auto` is the strategy, before were only checking if should flush or not when receive a new record.
- Improve drop of consumer stream
- Clean up `ConsumerRetryStream`

### DIFF
--- a/crates/fluvio-test/src/tests/consumer_offsets/auto/flush.rs
+++ b/crates/fluvio-test/src/tests/consumer_offsets/auto/flush.rs
@@ -43,11 +43,13 @@ pub async fn test_strategy_auto_periodic_flush(
     for _ in 0..10 {
         ensure_read(&mut stream, &mut counts).await?;
     }
-    sleep(Duration::from_secs(2)).await; // yeild  to drive auto flush flow
+    sleep(Duration::from_secs(3)).await; // yeild  to drive auto flush flow
 
-    let consumer = find_consumer(client, &consumer_id, 0).await?;
-    ensure!(consumer.is_some());
-    ensure!(consumer.unwrap().offset > 0i64);
+    for (partition, offset) in counts.iter().enumerate().take(partitions) {
+        let consumer = find_consumer(client, &consumer_id, partition).await?;
+        ensure!(consumer.is_some());
+        ensure!(consumer.unwrap().offset == *offset);
+    }
 
     drop(stream); //we keep the stream alive to prevent flush on drop occuring
 
--- a/crates/fluvio/src/consumer/mod.rs
+++ b/crates/fluvio/src/consumer/mod.rs
@@ -605,7 +605,8 @@ where
         &self,
         config: ConsumerConfigExt,
     ) -> Result<SinglePartitionConsumerStream<impl Stream<Item = Result<Record, ErrorCode>>>> {
-        let (offset, config, consumer_id, strategy, flush_period) = config.into_parts();
+        let (offset, config, consumer_id, strategy, flush_period, flusher_check_period) =
+            config.into_parts();
         let (stream, start_offset, stream_to_server) = self
             .inner_stream_batches_with_config(offset, config, consumer_id)
             .await?;
@@ -630,6 +631,7 @@ where
             flattened,
             strategy,
             flush_period,
+            flusher_check_period,
             stream_to_server,
         ))
     }
--- a/crates/fluvio/src/consumer/mod.rs
+++ b/crates/fluvio/src/consumer/mod.rs
@@ -73,30 +73,18 @@ pub type BoxConsumerStream =
 pub type BoxConsumerStream =
     Pin<Box<dyn ConsumerStream<Item = Result<ConsumerRecord, ErrorCode>> + Send + 'static>>;
 
+type ShararedConsumerStream = Arc<Mutex<BoxConsumerStream>>;
+
+type ConsumerFutureOutput = (
+    ShararedConsumerStream,
+    Option<Result<(ConsumerRecord, Option<i64>), ErrorCode>>,
+);
+
 /// Type alias to access consume stream as a future.
 #[cfg(target_arch = "wasm32")]
-type BoxConsumerFuture = Pin<
-    Box<
-        dyn Future<
-                Output = (
-                    Arc<Mutex<BoxConsumerStream>>,
-                    Option<Result<(ConsumerRecord, Option<i64>), ErrorCode>>,
-                ),
-            > + 'static,
-    >,
->;
+type BoxConsumerFuture = Pin<Box<dyn Future<Output = ConsumerFutureOutput> + 'static>>;
 #[cfg(not(target_arch = "wasm32"))]
-type BoxConsumerFuture = Pin<
-    Box<
-        dyn Future<
-                Output = (
-                    Arc<Mutex<BoxConsumerStream>>,
-                    Option<Result<(ConsumerRecord, Option<i64>), ErrorCode>>,
-                ),
-            > + Send
-            + 'static,
-    >,
->;
+type BoxConsumerFuture = Pin<Box<dyn Future<Output = ConsumerFutureOutput> + Send + 'static>>;
 
 /// An interface for consuming events from a particular partition
 ///
--- a/crates/fluvio/src/consumer/retry.rs
+++ b/crates/fluvio/src/consumer/retry.rs
@@ -1,12 +1,12 @@
 use std::pin::Pin;
 use std::sync::Arc;
 use std::task::{Context, Poll};
+use std::time::Duration;
 
 use anyhow::Result;
 use adaptive_backoff::prelude::{
     Backoff, BackoffBuilder, ExponentialBackoff, ExponentialBackoffBuilder,
 };
-use async_channel::{Receiver, Sender};
 use async_lock::Mutex;
 use async_trait::async_trait;
 use fluvio_socket::ClientConfig;
@@ -67,7 +67,6 @@ pub struct ConsumerRetryInner {
     consumer_config: ConsumerConfigExt,
     client_config: Arc<ClientConfig>,
     strategy: Arc<dyn ReconnectStrategy>,
-    cmd_channel: (Sender<TaskCommand>, Receiver<TaskCommand>),
     backoff: ExponentialBackoff,
 }
 impl ConsumerRetryInner {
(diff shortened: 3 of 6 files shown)


crates/fluvio-test/src/tests/consumer_offsets/auto/flush.rs
```diff
@@ -45,9 +45,12 @@ pub async fn test_strategy_auto_periodic_flush(
...
     sleep(Duration::from_secs(2)).await; // yeild  to drive auto flush flow
 
-    let consumer = find_consumer(client, &consumer_id, 0).await?;
-    ensure!(consumer.is_some());
-    ensure!(consumer.unwrap().offset > 0i64);
+    for partition in 0..partitions {
+        let offset = (counts[partition]) as i64;
+        let consumer = find_consumer(client, &consumer_id, partition).await?;
+        ensure!(consumer.is_some());
+        ensure!(consumer.unwrap().offset == offset);
```
fraidev, line 52: I tested this without these modifications, and it was falling because it was not including the last read

crates/fluvio/src/consumer/stream.rs
```diff
@@ -175,28 +213,65 @@ impl<T: Stream<Item = Result<Record, ErrorCode>> + Unpin> Stream
 }
 
 impl OffsetManagement {
+    async fn start_background_flusher(&self) {
+        match self {
```
sehz, line 217: shouldn't this be part of flush? why it is part of retry?

sehz: could add this as part of flush configuration

fraidev: Part of retry? This is in the initialization of `SinglePartitionConsumerStream` that is used on all ConsumerStreams

fraidev: do you mean a flag to enable it?

sehz: trying to understand reasons for start background  worker here.  Shouldn't this be part of flush handler?

crates/fluvio/src/consumer/mod.rs
```diff
@@ -73,30 +73,18 @@ pub type BoxConsumerStream =
 pub type BoxConsumerStream =
     Pin<Box<dyn ConsumerStream<Item = Result<ConsumerRecord, ErrorCode>> + Send + 'static>>;
 
+type ShararedConsumerStream = Arc<Mutex<BoxConsumerStream>>;
```
sehz, line 76: what's reason for creating shared consumer stream? This seems to be expensive

crates/fluvio/src/consumer/retry.rs
```diff
@@ -47,18 +101,27 @@ enum ConsumerRetryState {
...
+    Commit,
+    Flush,
+}
+
+impl TaskCommand {
+    async fn apply(self, stream: &mut BoxConsumerStream) {
+        match self {
+            TaskCommand::Commit => {
+                if let Err(e) = stream.offset_commit().await {
+                    warn!(target: SPAN_RETRY, "Commit failed: {}", e);
```
sehz, line 115: should not ignore this error.  this may go into infinite loop.  if commit is failed, entire thing should be retried.

crates/fluvio/src/consumer/retry.rs
```diff
@@ -47,18 +101,27 @@ enum ConsumerRetryState {
...
+                }
+            }
+            TaskCommand::Flush => {
+                if let Err(e) = stream.offset_flush().await {
```
sehz, line 119: same comment.  error must not be ignored

crates/fluvio/src/consumer/retry.rs
```diff
@@ -307,52 +292,113 @@ impl ConsumerRetryStream {
...
+            }
+        }
+    }
+
+    /// Handles the consumer records stream.
+    async fn handle_consumer_records(
+        inner: &mut ConsumerRetryInner,
+        stream: ShararedConsumerStream,
+    ) -> Option<ConsumerFutureOutput> {
+        let mut stream_lock = stream.lock().await;
```
sehz, line 319: locking is expensive.  what's performance drop here?

crates/fluvio/src/consumer/retry.rs
```diff
@@ -307,52 +292,113 @@ impl ConsumerRetryStream {
...
+                    warn!(target: SPAN_RETRY, "Could not connect to stream on {}: {}", inner.consumer_config.topic, e);
+
+                    match inner.consumer_config.retry_mode {
+                        RetryMode::TryUntil(max) if attempts >= max => {
+                            return Err(ErrorCode::MaxRetryReached);
+                        }
+                        RetryMode::Disabled => {
+                            return Err(ErrorCode::Other(format!("{}", e)));
                         }
+                        _ => {} // Continue retrying.
```
sehz, line 401: ? this should be explicit

crates/fluvio/src/consumer/retry.rs
sehz, line 1: unit test should be updated and add new test cases

sehz:
Questioning whether it is good idea to flush in the client.   Instead, should explore do this in the SPU.  It's much easier to detect produce connection is ended and generate flush or auth flush.

Alternatively this can be modeled similar like Kakfa auto commit where commits are flushed periodically.

If not, simplest implementation would be just flush on user exists CLI.

crates/fluvio/src/consumer/retry.rs
fraidev: Before we were using lock already, I changed to Optional with `take` to borrow, but this is not too safe, and also it makes really hard to test.

fraidev:
> Questioning whether it is good idea to flush in the client. Instead, should explore do this in the SPU. It's much easier to detect produce connection is ended and generate flush or auth flush.
> 

We don't handle flush offset at SPU, we will need to change many things to do it. I don't think it's worth, to be honest.

> Alternatively this can be modeled similar like Kakfa auto commit where commits are flushed periodically.
> 

But this background check is doing it, no? It's checking only for auto mode, if it's the right time to flush.
By default, we use 2 seconds. Kafka uses 5.

> If not, simplest implementation would be just flush on user exists CLI.

We have it too. But we can't be 100% sure of that will happen in a forced exit. 
Also, there is the case that user still with the consumer on, but do not receive anything for a large amount of time. The last record will not be flushed without the background flusher.

crates/fluvio/src/consumer/retry.rs
fraidev: I added an explicit `continue`

crates/fluvio/src/consumer/retry.rs
fraidev: fixed

crates/fluvio/src/consumer/retry.rs
fraidev: fixed

crates/fluvio/src/consumer/retry.rs
```diff
@@ -307,52 +292,113 @@ impl ConsumerRetryStream {
...
+        }
+    }
+
+    /// Handles the consumer records stream.
+    async fn handle_consumer_records(
+        inner: &mut ConsumerRetryInner,
+        stream: ShararedConsumerStream,
+    ) -> Option<ConsumerFutureOutput> {
+        let mut stream_lock = stream.lock().await;
+        loop {
```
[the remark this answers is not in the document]
fraidev: fixed

crates/fluvio/src/consumer/retry.rs
fraidev: I added more unit tests

crates/fluvio/src/consumer/retry.rs
```diff
@@ -175,33 +197,31 @@ impl Stream for ConsumerRetryStream {
...
-            loop {
-                match self.state {
-                    ConsumerRetryState::Idle | ConsumerRetryState::Terminated => {
-                        return self.stream.lock().await.offset_commit().await;
-                    }
-                    _ => {
-                        notify.notified().await;
-                    }
-                }
+            if let Some(mut stream) = self.stream.try_lock() {
```
sehz, line 201: Seems like you could put timer here which would avoid need to put another layer on top.
So you could do:

```
        let timer = fluvio::sleep(timer).
         loop {
              select! {
              
                   etc

            }
         
         
         }
```

fraidev: A loop here would not work well in the case where stream is locked waiting for a record that never arrives.

But I agree that this code smells, thinking in somehow to improve it. 

I removed the expected for now.

fraidev: maybe if we have a timer also in `handle_consumer_records`, to not allow lock it forever. So we will not need this layer

fraidev: I removed this layer

crates/fluvio/src/consumer/retry.rs
```diff
@@ -307,51 +245,114 @@ impl ConsumerRetryStream {
...
+    }
+
+    /// Keep trying until a new stream is created.
+    async fn handle_reconnection_loop(
+        inner: &mut ConsumerRetryInner,
+        mut backoff: ExponentialBackoff,
+        mut attempts: u32,
+        new_config: ConsumerConfigExt,
+    ) -> Result<ShararedConsumerStream, ErrorCode> {
+        info!(target: SPAN_RETRY, "Reconnecting to stream with offset: {:?}", new_config.offset_start);
```
sehz, line 331: nit: use structure format

crates/fluvio/src/consumer/retry.rs
```diff
@@ -307,51 +245,114 @@ impl ConsumerRetryStream {
...
+        loop {
+            match inner
+                .strategy
+                .reconnect(inner, new_config.clone(), backoff.clone())
+                .await
+            {
+                Ok(new_stream) => {
+                    info!(target: SPAN_RETRY, "Created new consumer stream with offset: {:?}", new_config.offset_start);
```
sehz, line 339: same

crates/fluvio/src/consumer/retry.rs
fraidev: fixed

crates/fluvio/src/consumer/retry.rs
fraidev: fixed

crates/fluvio/src/consumer/stream.rs
```diff
@@ -175,28 +211,58 @@ impl<T: Stream<Item = Result<Record, ErrorCode>> + Unpin> Stream
...
+            OffsetManagement::Auto { auto_flusher, .. } => loop {
+                let check_period = sleep(std::time::Duration::from_millis(100));
```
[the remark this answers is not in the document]
fraidev: fixed

sehz (approved):
LGTM

Issue #4501, opened 2025-04-09
Error log storm when SPU is out of disk


nacardin:
The follow log entry is emitted millions of time per hour by the SPU when it runs out of disk space.
```
Error writing to replica: Other(
    "last append failed, disabling append",
)
```

This overwhelms the logging infrastructure/alerting and makes it hard to identify the root cause.

If this error is recoverable, I think we should implement exponential backoff. If not the SPU should terminate.

sehz:
Two things to address:
- SPU should have send back to client that error happens, instead of swallowing 
- If too many error happens, SPU should consider itself as offline.  That way, other SPU will pick up traffic

PR #4521, opened 2025-05-01
fix: out of space problems

fraidev:
Now on producing, it will show this error:

```
Producer error: Producer received an error code: Storage is full
```

Also change the SPU and Partition status.

Partition status as ~Offiline~ `OutOfStorage`:
```
▷ flvd partition list
  ID    NAME             STATUS        TYPE      RACK  PUBLIC          PRIVATE
  5001  custom-spu-5001  OutOfStorage  "custom"  -     localhost:9010  localhost:9011
```
                                   
Partition status as ~OutOfStorage~ Online:                                                                                                                                                                         
```
▷ flvd cluster spu list
  ID    NAME             STATUS          TYPE      RACK  PUBLIC          PRIVATE        
  5001  custom-spu-5001  Online  "custom"  -     localhost:9010  localhost:9011 
```

Should fix: https://github.com/infinyon/fluvio/issues/4501


sehz:
This should be flipped.  SPU Status indicate overall SPU's status.   That shouldn't not be effected by out of storage which is topic/partition issue.  In the future,  SPU could have multiple volumes so only some of the volumes could be effected.

crates/fluvio-controlplane-metadata/src/spu/status.rs
```diff
@@ -34,6 +34,7 @@ impl SpuStatus {
             SpuStatusResolution::Online => "online",
             SpuStatusResolution::Offline => "offline",
             SpuStatusResolution::Init => "Init",
+            SpuStatusResolution::OutOfStorage => "out of storage",
```
sehz, line 37: See overall comment.  No change here since this is for SPU health

crates/fluvio-protocol/src/link/error_code.rs
```diff
@@ -57,6 +57,12 @@ pub enum ErrorCode {
     #[fluvio(tag = 71)]
     #[error("Offset {offset} is evicted. The next available is {next_available}")]
     OffsetEvicted { offset: i64, next_available: i64 },
+    #[fluvio(tag = 72)]
+    #[error("Storage is full")]
```
sehz, line 61: Should be something like PartitionFull/PartitionShortCircuited

crates/fluvio-controlplane-metadata/src/spu/status.rs
fraidev: btw, should we create a new status for partition with the same name (OutOfStorage) or just use offline there?

sehz: Yes, add `OufOfStorage` for partition status

crates/fluvio-protocol/src/link/error_code.rs
fraidev: fixed

crates/fluvio-sc-schema/src/objects/mod.rs
```diff
@@ -15,7 +15,7 @@ pub use list::*;
 pub use watch::*;
 pub use metadata::*;
 
-pub(crate) const COMMON_VERSION: i16 = 18; // from now, we use a single version for all objects
+pub(crate) const COMMON_VERSION: i16 = 19; // from now, we use a single version for all objects
```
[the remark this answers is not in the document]
fraidev: fixed

crates/fluvio-controlplane-metadata/src/spu/status.rs
fraidev: fixed

crates/fluvio-controlplane-metadata/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "fluvio-controlplane-metadata"
 edition = "2021"
-version = "0.30.4"
+version = "0.30.5"
```
sehz, line 4: bump to `0.36.0` as well

sehz: and fluvio as well

crates/fluvio-spu/src/control_plane/dispatcher.rs
```diff
@@ -136,6 +139,7 @@ impl ScDispatcher<FileReplica> {
 
                 _ = status_timer.next() =>  {
                     self.send_lrs_status_back_to_sc(&mut sink).await?;
```
sehz, line 141: Hmm.  this is sending too many individual requests.  This should be batched so only a a single status back to SC

sehz (approved):

crates/fluvio-spu/src/control_plane/dispatcher.rs
fraidev: I filtered with a hashset at least also applied a little of DRY (Don't Repeat Yourself)

PR #4474, opened 2025-03-31
fix: replication stuck when syncing large batches

fraidev:
Should fix replica sync problem: https://github.com/infinyon/fluvio/issues/4472


I also notice more perfomance with replicas now:

before:
```
 fluvio benchmark producer -p 3 -r 3 -k --num-records 50000                                                                                                                                                                                                                                                              $ fix_replication_large_2 
Benchmark started

latencies: 19.7µs min, 7.4ms avg, 33.1ms max, 7.2ms p0.50, 15.1ms p0.95, 17.3ms p0.99
50000 total records sent, 71942 records/sec: (369.1 MB/sec), total time: 695.9ms

**Per Record E2E Latency**

|Variable| p0.00 |p0.50|p0.95 |p0.99 |p1.00 |
|--------|-------|-----|------|------|------|
|Latency |19.7µs |7.2ms|15.1ms|17.3ms|33.1ms|

**Throughput (Total Produced Bytes / Time)**

|     Variable      |   Speed    |
|-------------------|------------|
|Produced Throughput|369.1 MB/sec|

Benchmark completed
```


now: 

```
▷ flvdr benchmark producer -p 3 -r 3 -k --num-records 50000                                                                                                                                                                                                                                                               $ fix_replication_large_2 
Benchmark started

latencies: 69.2µs min, 5.9ms avg, 23.2ms max, 5.6ms p0.50, 12.6ms p0.95, 14.6ms p0.99
50000 total records sent, 99800 records/sec: (512.0 MB/sec), total time: 501.5ms

**Per Record E2E Latency**

|Variable| p0.00 |p0.50|p0.95 |p0.99 |p1.00 |
|--------|-------|-----|------|------|------|
|Latency |69.2µs |5.6ms|12.6ms|14.6ms|23.2ms|

**Throughput (Total Produced Bytes / Time)**

|     Variable      |   Speed    |
|-------------------|------------|
|Produced Throughput|512.0 MB/sec|

Benchmark completed
```

PR #4522, opened 2025-05-02
fix: consume large and incomplete batches

fraidev:
Similar to https://github.com/infinyon/fluvio/pull/4474, but for consume.

Solves https://github.com/infinyon/fluvio/issues/4523

In others places like `RecordSet` we have `Batch<RawRecords>` and we ignore errors of `UnexpectedEof`, because we can receive uncompleted records limited by max_bytes.

For consuming, we handle records with `ConsumerRecord` that is created by converting `Batch<RawRecords>` to `Batch<MemoryRecords>`. So we need to ignore these errors here too.

### DIFF
--- a/crates/fluvio-protocol/src/record/batch.rs
+++ b/crates/fluvio-protocol/src/record/batch.rs
@@ -1,8 +1,10 @@
 use std::io::Error;
+use std::io::ErrorKind;
 use std::mem::size_of;
 use std::fmt::Debug;
 use bytes::Bytes;
 use fluvio_types::PartitionId;
+use tracing::debug;
 use tracing::trace;
 
 use fluvio_types::Timestamp;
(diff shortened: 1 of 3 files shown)


crates/fluvio-protocol/src/record/batch.rs
```diff
@@ -917,4 +928,22 @@ mod test {
...
+        // when
+        let mut batch_raw_records: Batch<RawRecords> = Batch::try_from(batch).unwrap();
+        let mut other = bytes::BytesMut::from(batch_raw_records.records.0.clone());
+        other.put_slice(&[0, 1, 2, 3]); // add random bytes to be handle as uncompleated
+        batch_raw_records.records.0 = other.freeze();
+        let batch_result = batch_raw_records.memory_records();
+
+        // then
+        let batch = batch_result.unwrap();
+        assert_eq!(batch.len(), 3);
```
sehz, line 947: check offset and record value.
also add test case for compressed

fraidev: sure, I added record value check. 

> also since this covers incomplete batches, would need cover that scenario as well.
Actually, this test is doing it `other.put_slice(&[0, 1, 2, 3]);` is adding an incomplete record at the end

sehz (approved):
LGTM

Issue #4524, opened 2025-05-08
[Bug]: Read should still work even if storage limit is reached.


fraidev:
Working on https://github.com/infinyon/fluvio/pull/4521, I noticed that I was not able to consume after hit a out of storage error.

Seems something related to replica cursor been changed wrongly on out of storage.

Error with zero copy:

```
2025-05-07T20:26:39.036230Z ERROR fluvio_future::zero_copy: error sendfile: ENOTSUP: Operation not supported
2025-05-07T20:26:39.036481Z ERROR fluvio_spu::services::public::stream_fetch: error starting stream fetch handler: Io {
    source: Custom {
        kind: Other,
        error: "zero copy failed: Nix error: ENOTSUP: Operation not supported",
    },
    msg: "zero copy failed: Nix error: ENOTSUP: Operation not supported",
}
```

without zero copy:
```
2025-05-07T20:42:06.842657Z ERROR fluvio_spu::services::public::stream_fetch: error starting stream fetch handler: Io {
    source: Custom {
        kind: Other,
        error: "pread failed: ESPIPE: Illegal seek",
    },
    msg: "pread failed: ESPIPE: Illegal seek",
}
```

fraidev:
Fixed by https://github.com/infinyon/fluvio/pull/4527

PR #4527, opened 2025-05-09
fix: prevent double free on storage error

fraidev:
Should solve: https://github.com/infinyon/fluvio/issues/4524

We're not calling `std::mem::forget` on write_all error cases. So we're having double free problems. 
Now consumer works even on out of storage state.


sehz (approved):
Good catch.  We need to wrap this with more safe abstraction later.

PR #4569, opened 2025-06-20
chore: use rustls instead openssl

fraidev:
Needs: 
- [x] https://github.com/infinyon/flv-tls-proxy/pull/38

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -125,7 +125,9 @@ mime = "0.3"
 mockall = { version = "0.13.1", default-features = false }
 nix = { version = "0.29.0", default-features = false }
 once_cell = "1.7.2"
+openssl = { version = "0.10", default-features = false }
 parking_lot = { version = "0.12.3", default-features = false }
+lib-cargo-crate = "0.2.1"
 pin-project = "1.1.0"
 pin-utils = "0.1.0"
 portpicker = "0.1.1"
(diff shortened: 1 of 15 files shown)


Cargo.toml
```diff
@@ -125,7 +125,9 @@ mime = "0.3"
 mockall = { version = "0.13.1", default-features = false }
 nix = { version = "0.29.0", default-features = false }
 once_cell = "1.7.2"
+openssl = { version = "0.10", default-features = false }
```
sehz, line 128: reason why this is needed anymore?

crates/cdk/Cargo.toml
```diff
@@ -30,6 +30,7 @@ sysinfo = { workspace = true, default-features = false }
 tempfile = { workspace = true }
 toml = { workspace = true, features = ["parse", "display", "preserve_order"] }
 tracing = { workspace = true }
+openssl = { workspace = true, features = ["vendored"] } # cargo-generate requires openssl
```
sehz, line 33: shouldn't this be using fluvio future's openssl dep?

Cargo.toml
fraidev: cdk and smdk uses cargo-generate that needs it vendored

sehz (approved):

crates/cdk/Cargo.toml
fraidev: cdk needs openssl only for `cargo-generate`, we can also enable the feature at fluvio future.

PR #4574, opened 2025-06-27
chore: ensure that crypto provider is set

fraidev:
Rustls is tricky to handle when an application is using more than one crypto provider.
When more than one provider is enabled, we have these errors: 

```
no process-level CryptoProvider available -- call CryptoProvider::install_default() before this point
```

Easily to have more than one provider enable if using sub library from rustls ecosystem that only support one of them.

~This PR is adding a new feature for Fluvio crate called `rustls-ring` to use ring provider instead of `aws-ls-rs`. That way we'll have sure that one is settled at least and control it by feature flags.~

This PR is adding two new feature for Fluvio crate called `ring` and `aws-ls-rs`. `aws-ls-rs` is the default.
That way we'll have sure that one is settled at least and control it by feature flags.

### DIFF
--- a/crates/fluvio/Cargo.toml
+++ b/crates/fluvio/Cargo.toml
@@ -15,15 +15,17 @@ name = "fluvio"
 path = "src/lib.rs"
 
 [features]
-default = ["rustls", "compress"]
+default = ["rustls", "compress", "aws-lc-rs"]
 admin = ["fluvio-sc-schema/use_serde"]
 smartengine = ["fluvio-smartengine"]
 openssl = ["fluvio-future/openssl_tls"]
-rustls = ["fluvio-future/rust_tls", "dep:rustls", "rustls?/aws-lc-rs"]
-rustls-ring = ["fluvio-future/rust_tls", "dep:rustls", "rustls?/ring"]
+rustls = ["fluvio-future/rust_tls", "dep:rustls"]
 compress = ["fluvio-compression/compress", "fluvio-protocol/compress"]
 nightly = []
 unstable = []
+# Crypto providers for rustls (mutually exclusive)
+aws-lc-rs = ["rustls?/aws-lc-rs", "rustls?/prefer-post-quantum"]
+ring = ["rustls?/ring"]
 
 [dependencies]
 adaptive_backoff = { workspace = true }
@@ -54,7 +56,11 @@ fluvio-future = { workspace = true, features = [
     "task",
     "retry",
 ] }
-rustls = { workspace = true, optional = true, default-features = false }
+rustls = { workspace = true, optional = true, default-features = false, features = [
+    "std",
+    "logging",
+    "tls12",
+] }
 fluvio-types = { workspace = true, features = ["events"] }
 fluvio-sc-schema = { workspace = true }
 fluvio-socket = { workspace = true }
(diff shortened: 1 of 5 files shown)


digikata:
@nacardin should probably review this too, as `aws-ls-rcs` was integrated by him in a few other places.  Is there are reason to use ring over `aws-ls-rcs`?

fraidev:
> This PR seems to mix many things. can you separate it out? Also Rustls related changes should be made in fluvio-future

yeah, let's fix clippy rules in that PR first: https://github.com/infinyon/fluvio/pull/4575

fraidev:
> @nacardin should probably review this too, as `aws-ls-rcs` was integrated by him in a few other places. Is there are reason to use ring over `aws-ls-rcs`?

Sure! Actually, we're using `aws-ls-rs` as default, like `rustls`. But, as now we're forcing with `default_provider().install_default()`, we need feature flags to control it.

Also make it easier to test, for example, in Infinyon Cloud I got a weird bug related to the size of client hello of handshake (already fixed) that only happens with `aws-ls-rs` but works well with `ring`.

nacardin (commented):
Try using https://docs.rs/rustls/latest/rustls/crypto/struct.CryptoProvider.html#method.get_default to avoid panic

fraidev:
> Try checking for existing default using https://docs.rs/rustls/latest/rustls/crypto/struct.CryptoProvider.html#method.get_default before installing to avoid panic.

Sure, done!

crates/fluvio/Cargo.toml
```diff
@@ -15,14 +15,17 @@ name = "fluvio"
...
 admin = ["fluvio-sc-schema/use_serde"]
 smartengine = ["fluvio-smartengine"]
 openssl = ["fluvio-future/openssl_tls"]
-rustls = ["fluvio-future/rust_tls"]
+rustls = ["fluvio-future/rust_tls", "dep:rustls"]
 compress = ["fluvio-compression/compress", "fluvio-protocol/compress"]
 nightly = []
 unstable = []
+# Crypto providers for rustls (mutually exclusive)
+aws-lc-rs = ["rustls?/aws-lc-rs", "rustls?/prefer-post-quantum"]
```
sehz, line 27: can we get use consistent feature name?  it should be something like `rustls-aws`,  `rusts-ring`.

crates/fluvio/src/config/tls.rs
```diff
@@ -241,13 +241,29 @@ cfg_if::cfg_if! {
...
         }
-    }  else if #[cfg(feature = "rustls")] {
+    } else if #[cfg(any(feature = "rustls"))] {
 
         impl TryFrom<TlsPolicy> for DomainConnector {
             type Error = anyhow::Error;
 
             fn try_from(config: TlsPolicy) -> Result<Self, Self::Error> {
 
+                if rustls::crypto::CryptoProvider::get_default().is_none() {
```
sehz, line 251: Since fluvio is library,  not sure this make sense.  If other binary decided to do change how this work, this will override and could cause conflict.  

I think it's better this to be done outside fluvio and move to `fluvio future` and make it available as API and part of initialization

crates/fluvio/Cargo.toml
fraidev: Sure, I'll create a new PR for that.

PR #4626, opened 2025-12-29
fix(CRD): remove duplicate source field in topic schema

### LINKED ISSUE #4610: [Bug]: Cannot apply the sys chart server-side
We use argocd, which just runs `helm template` and then applies the generated manifest. When I tried applying the manifest server-side, I get this error:
```
one or more objects failed to apply, reason: failed to create typed patch object (/topics.fluvio.infinyon.com; apiextensions.k8s.io/v1, Kind=CustomResourceDefinition): .spec.versions[0].schema.openAPIV3Schema.properties.spec.properties.replicas.properties.mirror.properties.home.properties.partitions.source: field not declared in schema
```

Here is how chatGPT is suggesting we fix it:
https://chatgpt.com/share/68fa46e7-9e48-8009-bdc0-ca5c44ab867d

--

applying server-side also removes all of the `nullable: false` lines from the crd, which makes the chart always appear out of sync.

syedazeez337:
The topic CRD had a malformed schema with the  field defined twice - once correctly inside partitions.items and once incorrectly as a sibling to partitions. This caused server-side apply (used by ArgoCD) to fail with 'field not declared in schema' error.

Fixes #4610

### DIFF
--- a/.github/workflows/cd_dev.yaml
+++ b/.github/workflows/cd_dev.yaml
@@ -85,7 +85,7 @@ jobs:
         if: ${{ matrix.cluster_type == 'k8' }}
         run: ./k8-util/cluster/reset-k3d.sh
       - name: Setup Fluvio
-        uses: infinyon/fluvio/.github/actions/setup-fluvio@master
+        uses: fluvio-community/fluvio/.github/actions/setup-fluvio@master
         with:
           version: latest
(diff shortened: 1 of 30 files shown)


morenol (approved):
LGTM, thanks! Replicated the issue locally

PR #4632, opened 2026-01-05
chore: remove hub related subcommands in cli

morenol:


digikata (approved):
LGTM, we can think about how to readd somesort of hub in the future

digikata (approved):

PR #4633, opened 2026-01-06
feat(connectors): honor meta.producer.max-request-size

xpload32004:
In production we hit a hard failure mode where source connectors crash-loop when they attempt to
produce a single record larger than Fluvio’s default producer request limit (~1MiB). Operators can
already configure meta.producer in connector YAML, but max-request-size was not being parsed/applied by
the connector framework, meaning the setting had no effect and the connector would still exit on larger
payloads.
This PR adds max-request-size (and max_request_size alias) to the connector config schema and wires it
into the producer configuration builder, so the producer is created with the requested max request
size. This allows connectors to handle larger individual records safely when the environment requires
it (e.g. large log/event payloads).


digikata:
Thanks for the PR, LGTM too

